PR #5, opened 2019-07-17
Model descriptions yx

yongqx:


docs/source/models/cause.rst
```diff
@@ -4,66 +4,108 @@
...
-
-Excess mortality
-++++++++++++++++
-
-Disability weight
-+++++++++++++++++
+.. note::
+
+   Currently there are GBD shared function tools that return you draw-level values;
+   alternatively, you can also pull data by using ``vivarium_inputs`` tools.
```
halfhorst: we should probably make this an explicit link to our tutorial over there

docs/source/models/cause.rst
```diff
@@ -4,66 +4,108 @@
...
++------------------+-------------------------------------------------------+
+| **Tool**         | **Requirements**                                      |
++------------------+-------------------------------------------------------+
+| get_draw("como") | only accepts one of the following types:              |
+|                  | cause_id, rei_id, or sequela_id                       |
++------------------+-------------------------------------------------------+
+| get_draw("epi")  | with rare exceptions, Prevalence, Incidence, EMR, and |
+|                  | WCMR values are non-final and should not be used      |  
++------------------+-------------------------------------------------------+
+| get_raw_data()   | modelable entity, measure, and lcoation               |
```
halfhorst: typo in `location` :D

docs/source/models/cause.rst
halfhorst: https://vivarium-inputs.readthedocs.io/en/latest/tutorials/pulling_data.html

docs/source/models/cause.rst
yongqx: Thanks, James introduced me the reStructuredText Directives that includes a lot of contents such as how to make a table using "list-table". I will definitely make those documents easily editable as i learning through rst markdowns.

docs/source/models/cause.rst
yongqx: Totally agree, the ultimate goal is to fill the modeling part in details. For example, in the cause modeling we will explicitly explain what's the cause in a way of understanding by GBD, and also what microsim cares about. We will eventually give the handy example how we initialize a simple cause model. Add some workflow images additional to contents in paragraph.

PR #8, opened 2019-09-06
Document the documentation process

aflaxman:
heavy-meta

PR #10, opened 2019-10-10
Feature/doc process

collijk:
To replace https://github.com/ihmeuw/vivarium_research/pull/8

Typically, workflow-wise, we'd handle something like this with additional PRs, but Abie submitted his from a fork (https://guides.github.com/activities/forking/).  **Don't submit PRs from forks**.  It's a totally fine work style, just not the one we use here. Instead, follow the contributing guide: https://vivarium-research.readthedocs.io/en/latest/contributing.html#contributing

Otherwise, what's in this PR:

- Expanded out Abie's take on the process
- Provided a silly example.  
- Moved the documentation process docs to their own rst file and referenced them in the main modeling process index.  The modeling process index will eventually contain links to all of our standard modeling process stuff that is currently located on the hub.

### DIFF
--- a/docs/source/modeling_process/documentation_sessions.rst
+++ b/docs/source/modeling_process/documentation_sessions.rst
@@ -1,0 +1,92 @@
+.. _process_documentation:
+
+======================
+Documentation Sessions
+======================
+
+The Simulation Science team uses weekly documentation sessions to organize
+and produce high-quality documentation of our modeling process.  These sessions
+offer an opportunity for team members to share ideas and modeling techniques
+and to formalize those techniques that we've found especially successful.
+
+
+.. contents:
+   :local:
+   :depth: 1
+
+
+The process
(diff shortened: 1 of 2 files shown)


docs/source/modeling_process/documentation_sessions.rst
```diff
@@ -0,0 +1,92 @@
...
+   `hosted documentation <https://vivarium-research.readthedocs.io/en/latest/>`_.
+
+
+A silly but hopefully helpful example
+-------------------------------------
+
+1. **Topic Kickoff** - James gives and overview of building cause models.
+2. **Sub-topic Assignment** - Collectively, the team identifies the sub-topics
+   and we assign them to research team members:
+   - Cause Modeling Overview: Abie
```
aflaxman: add a line here to typeset better

PR #12, opened 2019-10-10
Feature/cause model validation

yongqx:
I have added the notes for TB/HIV model validation & verification. Will add more details later. Sorry for getting those commits together and i should separate `cause_model_validation.rst` out.


collijk:
This PR has way too much stuff in it.  Branch from master and PR only the cause validation stuff.

PR #13, opened 2019-10-11
resumbit cause-model-validation.rst to make small PR

yongqx:


docs/source/models/gbd2017/causes/tb_hiv_model_vv.rst
```diff
@@ -0,0 +1,87 @@
...
+External
+	- Model results should be checked against local TB epidemiology
+		- e.g. rate of decline in burden of disease should be compared
+		  with historical evidence.
+	- Compare our results (e.g. ICERs) to similar models or empirical
+	  assessments where possible
+Internal
+	- Calibrate simulation baseline against GBD 2017 results
+		- Sim outputs mean should perfectly match GBD results
+		- Sim outputs uncertainty should wider than GBD results
```
aflaxman: should *be* wider than

docs/source/models/gbd2017/causes/tb_hiv_model_vv.rst
```diff
@@ -0,0 +1,87 @@
...
+			- Stochastic + individual heterogeneity
```
aflaxman: bullet is unclear, but I think you mean "The uncertainty intervals of sim outputs should be wider than GBD results, because they include parameter uncertainty *and* stochastic uncertainty".  (Heterogeneity should not increase the uncertainty interval, I think.)

docs/source/models/gbd2017/causes/tb_hiv_model_vv.rst
```diff
@@ -0,0 +1,87 @@
...
+logic
+-----
+- Parent cause is the sum of child causes
+	- Fatal: Deaths (CSMR, Excess MR), YLLs
+	- Non-fatal: Cases, YLDs, Prevalence, Incidence
+	- DALYs = YLLs + YLDs
+- By location-/age-/sex-
+- Start from count space
+- We expect that active TB cases < LTBI cases when LTBI prevalence
+  as denominator
```
aflaxman: This confuses me.  Do you mean the incidence *rate*, since you say denominator? or do you mean the count, since only a fraction of LTBI progresses to active TB?

docs/source/models/gbd2017/causes/tb_hiv_model_vv.rst
```diff
@@ -0,0 +1,87 @@
...
+
+Cause hierarchy
+---------------
+.. image:: cause_hierarchy.png
```
aflaxman: Add LTBI on third row (?)

docs/source/models/gbd2017/causes/tb_hiv_model_vv.rst
```diff
@@ -0,0 +1,87 @@
...
+	- Constrained verification
+		- e.g. compare (disease counts / person time) to incidence in GBD
+	- Unconstrained verification
+		- CSMR
+		- YLLs
+		- YLDs
+
+Minimal model verification
+--------------------------
+.. image:: minimal_model_verification.png
```
aflaxman: I don't understand this figure. What do the arrows mean?  What does the figure mean at a high level?

collijk:
I'm concerned that this document is only about TB.  The goal was to write down general v&v principals for cause models.  e.g. Given a cause model and a definition of it's data sources, how do we go about validating the data is correct?  That the model is correct?  What can we use for external validation data?  What shorthand epi formulas can we use to check that the data is internally consistent?  What's the difference between checking incidence measures as output vs. checking prevalence or CSMR?  Is one or the other better?  More useful?  Why?


This seems like a good start to the V&V we want to do for TB, but it should be part of a document about the TB model, along with the concept diagram and tables of input data.

collijk:
The generic V&V should be a part of the document on building cause models at docs/source/models/cause.rst

docs/source/models/gbd2017/causes/tb_hiv_model_vv.rst
yongqx: I will rephrase it. Here i mean the simulation output on event count, the total LTBI cases should be greater than active TB cases (active TB cases / LTBI cases < 1).

yongqx:
I agree that this is a start of TB validation criteria and going to write up a general validation and verification principle for cause model to the directory James suggested.

docs/source/models/gbd2017/causes/tb_hiv_model_vv.rst
yongqx: Thanks for clarifying it!

collijk:
@yongqx Let's talk about this PR during office hours, please.

PR #14, opened 2019-10-23
Added data source documentation excluding mortality/disability

alibow:
I am saving the mortality and disability data sources for Yaqi to detail

### DIFF
--- a/docs/source/models/cause.rst
+++ b/docs/source/models/cause.rst
@@ -102,14 +102,14 @@ Birth Prevalence
 Birth prevalence is defined as the **proportion of live births in a given population that possess a given 
 condition or trait at birth.**
  
-	For example, the birth prevalence for cleft lip in the United States in 2006 was 10.63 per 10,000 live 
-	births, or 0.163%.
+	For example, the birth prevalence for cleft lip in the United States in 2006 was 10.6 per 10,000 live 
+	births, or 0.106%.
 
 Birth prevalence data can be used to **initialize neonatal cause model states** and represent the **probability that a 
 simulant who is born during the simulation will be born into a given neonatal cause model state.** 
 
 	For example, the probability that a simulant born during a simulation of cleft lip in the United States 
-	in 2006 is 0.00163, or 0.163%.
+	in 2006 is 0.00106, or 0.106%.
 
 Incidence Rates
 +++++++++++++++
--- a/docs/source/models/cause.rst
+++ b/docs/source/models/cause.rst
@@ -265,7 +265,7 @@ Depending on the specific cause model at hand, the prefered data source may vary
 and instantaneous incidence (or hazard) rates. The table below discusses some considerations that may influence 
 which data source is preferable. In general...
 
-**Annual incidence rates are preferable when:**
+**Annual rates are preferable when:**
 
 - The assumption of uniform and constant distribution of new cases is **valid**
 
@@ -278,7 +278,7 @@ which data source is preferable. In general...
 - The assumption of uniform and constant distribution of new cases is **invalid**, but the assumption will not 
 impact model results in a meaningful way
 
-**Instantaneous incidence (hazard) rates are preferable when:**
+**Instantaneous (hazard) rates are preferable when:**
 
 - There is not a uniform or constant distribution of new cases over an annual timeframe
 
--- a/docs/source/models/cause.rst
+++ b/docs/source/models/cause.rst
@@ -121,11 +121,12 @@ simulant who is born during the simulation will be born into a given neonatal ca
 Incidence
 +++++++++
 
-Incidence rates are defined as the **number of new cases of a condition that occur per person-year of the at-risk 
-population (individuals without condition).** 
+Incidence rates are defined as the **number of new cases of a condition that occur per person-year of the 
+at-risk population (individuals without condition).** Specifically, the at-risk population can be represented as 
+`1 - condition prevalence`.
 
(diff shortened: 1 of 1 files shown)


docs/source/models/cause.rst
```diff
@@ -41,29 +41,251 @@ Common data sources for cause models
...
+	For example, the probability that a simulant in a model of obesity in the United States beginning in 
+	2016 will begin the simulation as obese is 0.4 or 40%.
+
+Birth Prevalence
+++++++++++++++++
+
+Birth prevalence is defined as the **proportion of live births in a given population that possess a given 
+condition or trait at birth.**
+ 
+	For example, the birth prevalence for cleft lip in the United States in 2006 was 10.63 per 10,000 live
```
KjellSwedin: Typo here:
In [2]: (10.63 / 10000) * 100                                                                                           
Out[2]: 0.10630000000000002
0.163% -> 0.106%

alibow: Thanks, good catch! It's fixed now.

docs/source/models/cause.rst
```diff
@@ -41,29 +41,251 @@ Common data sources for cause models
...
+|Weights       |                                |                    |
++--------------+--------------------------------+--------------------+
+|Cause-Specific|INSERT DEFINITION (Yaqi?)       |INSERT USES (Yaqi?) |
+|Mortality     |                                |                    |
++--------------+--------------------------------+--------------------+
+|Excess        |INSERT DEFINITION (Yaqi?)       |INSERT USES (Yaqi?) |
+|Mortality     |                                |                    |
++--------------+--------------------------------+--------------------+
 
 Prevalence
```
aflaxman: It is possible to add hyperlinks between the entries and headings, but I don't remember how to do it.

docs/source/models/cause.rst
```diff
@@ -41,29 +41,251 @@ Common data sources for cause models
...
+	1 year = 100,000 person-years), we would expect 2.8 of these individuals to develop MS within this timeframe. 
+
+	Alternatively, if we followed 50,000 individuals without MS for 2 years each (50,000 people * 2 years = 100,000 
+	person-years), we also would expect 2.8 of these individuals to develop MS within this timeframe.
+
+Incidence can be used to **estimate cause model transition rates** and can represent the **probability that a simulant 
+will transition from a susceptible state to an infected state within a given timestep.** 
+
+	For example, with a timestep of one year, the probability that a simulant will transition from a 
+	susceptible (without MS) cause model state to an infected (with MS) cause model state is 2.8*10^(-5).
```
aflaxman: I think that this is an approximation of the precise probability, and the approximation is less accurate when the product is large.

docs/source/models/cause.rst
```diff
@@ -41,29 +41,251 @@ Common data sources for cause models
...
+
+	This suggests that if we followed 10,000 individuals with type II diabetes for one year each, we would 
+	expect to see 2.4 individuals recover from type II diabetes.
+
+Remission rates can be used to estimate cause model transition rates and represent the probability that a 
+simulant in an infected (with condition) state will transition to a non-infected (without condition) state. 
+
+	For example, with a time step of one year, the probability that a simulant in the infected state in a 
+	model of type II diabetes in the United States will transition to a susceptible or recovered state 
+	within a timestep is 2.4*10^(-4), or 0.024%.
```
aflaxman: Same comment as above in the incidence rate section about my confusion---is 0.024% precise or is it an approximation of `1 - exp(2.4e-4)`.  Not a big deal in this case, because they are really similar values.  But it could be important for getting the concept across clearly.

docs/source/models/cause.rst
```diff
@@ -41,29 +41,251 @@ Common data sources for cause models
...
+
+	Consider a cause model in which an individual can trasition from a susceptible state to a mild disease 
+	state OR from a susceptible state to a severe disease state.
+
+	In this case, the incidence rate for overall disease (mild and severe) does not help us estimate the 
+	transition rates from susceptible to mild disease or to severe disease. In these cases, incidence rates 
+	specific to mild and severe disease are needed to inform the specific transitions present in the cause model.
+
+Lastly, see the section on hazard rates in non-standard data sources below to determine when hazard rates may be 
+preferrable to annual incidence rates as a data source for cause model transition rates.
```
aflaxman: Perhaps my confusion about the formula for incidence/remission probabilities below is related to the distinction between hazards and annual rates alluded to here.  If so, maybe there is a way to introduce and contrast these concepts up front to help me and other readers understand sooner.

docs/source/models/cause.rst
alibow: Done!

docs/source/models/cause.rst
alibow: I made some changes that I hope make this more clear.

I was hoping that this example represented the "ideal" scenario in which the denominator actually represented the at-risk population, so the probability would not be an approximation. 

Then, I expanded the examples when the incidence rate would be an approximation of the transition rate below. 

Let me know if you think this still needs additional clarification after my updates. Thanks!

aflaxman: Important clarifications, thanks.  There is still something else that I am confused about, which is a distinction between "instantaneous rates" and "annual rates".  But let's merge to keep things moving, and remember that there is still something I am confused about here.

collijk:
@alibow Please branch from the main repository and submit this pr from there.  See https://vivarium-research.readthedocs.io/en/latest/contributing.html#contributing for instructions, or come chat during office hours if you need help.

PR #17, opened 2019-10-28
add cause model impact doc

yaqi7:


alibow:
I think we need to consider how this portion of the documentation should fit in with the existing cause.rst file. Do we want a separate page for cause model impact or should it be worked into the overall cause model documentation page?

docs/source/cause_model_impact.rst
```diff
@@ -0,0 +1,53 @@
+===============================
+Vivarium Research Documentation
+===============================
+
+Mode impact 
+-------------
+
+1. **What is model impact**
+ Mode impact is about how causes affect mortality/disability
```
alibow: Mode --> model 
Two typos here

docs/source/cause_model_impact.rst
```diff
@@ -0,0 +1,53 @@
...
+ - Use GBD data to estimate mortality rate (three cases)
+
+   case 1: no cause models
+    :math:`\lambda=ACMR`
+   case 2: one cause model
+    :math:`\lambda = ACMR - CSMR + EMR(disease\_status)`
+   case 3: multiple causes model 
+    :math:`\lambda_{i} = ACMR - \sum_{diseases}CSMR_{c} + \sum_{diseases}EMR_{c}(disease\_status_{c})`
+   what did the person die of
+    :math:`P_{c} = \frac{EMR_{c}}{CDMR + \sum_{diseases}EMR_{c}(disease\_status_{c})}`
```
alibow: I think we should to define the acronyms ACMR, CSMR, and EMR

docs/source/cause_model_impact.rst
```diff
@@ -0,0 +1,53 @@
...
+
+3. **Disability** 
+
+ - Years lived with disability (YLD): This includes conditions such as influenza, which may last for only a few days, or epilepsy, which can last a lifetime. It is measured by taking the prevalence of the condition multiplied by the disability weight for that condition. 
+
+ - Disability weights represent the severity of health loss  associated with a health state. Disability weights reflect the severity of different conditions.
+
+       :math:`Disability\_weight_{c}`: proportion of health lost due to condition c
+        
+       :math:`(1 - Disability\_weight_{c})`: proportion of life lived in spite of condition c
```
aflaxman: awkward phrasing, consider alternative such as "proportion of health remaining when living with condition c"

docs/source/cause_model_impact.rst
```diff
@@ -0,0 +1,53 @@
...
+       
+       :math:`\prod_{c\in causes}(1 - Disability\_weight_{c})`: proportion of life lived despite all causes
```
aflaxman: suggest similar rephrasing

docs/source/cause_model_impact.rst
yaqi7: fixed, thanks

docs/source/cause_model_impact.rst
yaqi7: Do you think it's okay as following:
:math:`Disability\_weight_{c}`: proportion of health lost when living with condition c
:math:`(1 - Disability\_weight_{c})`: proportion of health remaining when living with condition c

yaqi7: Do you think it's okay as following:
:math:Disability\_weight_{c}: proportion of health lost due to condition c
:math:(1 - Disability\_weight_{c}): proportion of health remaining when living with condition c

docs/source/cause_model_impact.rst
yaqi7: Is it okay that I add following annotation afterwards:
All cause mortality rate (ACMR)
Cause specific mortality rate (CSMR)
Excess mortality rate (EMR)

alibow: Yeah, I think that's perfect. Thanks :)

yaqi7:
> I think we need to consider how this portion of the documentation should fit in with the existing cause.rst file. Do we want a separate page for cause model impact or should it be worked into the overall cause model documentation page?

Good point, I think it's better to fit this part in existing cause.rst file. How about adding it to 'the structure of a cause model' session? Model impact should be a component of cause model.

docs/source/models/cause.rst
```diff
@@ -16,6 +16,62 @@ What is a cause?
 The structure of a cause model
 ------------------------------
 
+Model impact
```
aflaxman: Let's call this "Cause model impact" to make to completely clear that it is about the impact of a cause model on the demographic model.

yaqi7: changed, thanks

docs/source/models/cause.rst
```diff
@@ -16,6 +16,62 @@ What is a cause?
 The structure of a cause model
 ------------------------------
 
+Cause model impact 
+-------------
+
+1. **What is model impact**
+ Model impact is about how causes affect mortality/disability
+2. **Mortality**
```
collijk: @yaqi7 The content here is good because it operationally describes how the data gets used in the simulation, but we also want to write some information about what the actual quantities that get used are.  

In particular, you want to think about what the cause specific mortality rate is and what the excess mortality rate is.  I think it's probably useful to have a section on case fatality rates as well.  The format ali has is useful to frame this.  E.g.

The cause specific mortality rate is a measure of ....

For example the CSMR due to X is Y.  In a population of 10K people over 1 year, this means....

In the simulation we use the CSMR to ...

In the simulation the CSMR modifies and indivuals mortality rate by...

collijk:
@yaqi7 Let's figure out how to work this into the format Ali has laid out on monday

PR #18, opened 2019-11-01
Feature/gbd cause detail

ctroeger:
I have completed an attempt to describe GBD Cause of Death including input data, modeling process, and where/how to pull results.

I would especially like feedback on where this document should live and the appropriate amount of useful information that I have included.


yongqx:
~/vivarium_research/docs/source/models/gbd_cause_details.rst, It could be under same repository as cause.rst

docs/source/models/gbd2017/causes/causes.rst
```diff
@@ -0,0 +1,95 @@
...
+Causes in the GBD are organized in a hierarchy where at every level of that hierarchy
+causes of death or disease are mutually exclusive and collectively exhaustive. 
+
+- All causes
+
+  - Injuries, communicable diseases, non-communicable diseases
+
+    - ... down to the most detailed level
+
+Some causes are only associated with non-fatal outcomes, such as lower back pain.
```
alibow: I think this line was repeated and should be deleted :)

docs/source/models/gbd2017/causes/causes.rst
```diff
@@ -0,0 +1,95 @@
...
+
+Measures for causes include:  
+Deaths, incidence, prevalence, YLDs, YLLs, and DALYs
+
```
alibow: Should we also include metrics here? As in counts/rates/percent?

docs/source/models/gbd2017/causes/causes.rst
```diff
@@ -0,0 +1,95 @@
...
+	
+Modeling causes of death
+------------------------
+
+.. image:: fatal_flowchart.png
+   :width: 600
+
+**Modeling approach** 
+
+Most causes use a centrally-maintained Bayesian ensemble model (CODEm)
```
alibow: "use" seems a little vague here to me. Maybe something like "estimates for most causes are computed using a centrally-maintained...." (if that is factually correct)

docs/source/models/gbd2017/causes/causes.rst
ctroeger: Thanks!

docs/source/models/gbd2017/causes/causes.rst
ctroeger: I added that suggestion!

docs/source/models/gbd2017/causes/causes.rst
ctroeger: I changed the text to "Most causes are estimated using ..."

docs/source/models/gbd2017/causes/causes.rst
```diff
@@ -0,0 +1,96 @@
...
+.. image:: fatal_flowchart.png
+   :width: 600
+
+**Modeling approach** 
+
+Most causes are estimated using a centrally-maintained Bayesian ensemble modeling tool (CODEm)
+
+- Bayesian: estimates based on probabilities
+- Ensemble: combines many unique submodels into a final model
+
```
aflaxman: Paper out today that provides some details on CODEm in a style intended to be understandable:  https://doi.org/10.1093/ije/dyz223

PR #19, opened 2019-11-05
Data Sources

alibow:
This pull request has been resubmitted directly from the master branch instead of from a fork

### DIFF
--- a/docs/source/models/cause.rst
+++ b/docs/source/models/cause.rst
@@ -171,29 +171,294 @@ Common data sources for cause models
 
 .. todo::
 
-   Format as table with measure, measure definition, data sources and
-   their uses.
+   Update mortality-related data sources within existing format.
+
+Once a cause model structure is specified, data is needed to inform its states and transitions. For our purposes, cause models generally have the following data needs:
+
+1. The probability that a simulant will start the simulation in a given state within the cause model
+2. The probability that a simulant will transition to a new state within the cause model in a given timestep
+3. The disability weight for each state in the cause model
+4. The probability that a simulant in a given cause model state will die in a given timestep
+
+There are several common data sources that can be used for these needs, which are outlined in the table below 
+and discussed in more detail afterward. `Non-standard data sources`_ are discussed later on this page.
+
++----------------------------+--------------------------------+--------------------+
+|Measure                     |Definition                      |Uses                |
++============================+================================+====================+
+|Prevalence_                 |Proportion of population        |Initialize cause    |
+|                            |with a given condition          |model states        |
++----------------------------+--------------------------------+--------------------+
+|`Birth Prevalence`_         |Proportion of all live births   |Initialize neonatal |
+|                            |born with a given condition     |cause model states  |
++----------------------------+--------------------------------+--------------------+
+|Incidence_                  |Number of new cases of a given  |Estimate transition |
+|                            |condition per person-year       |rates               |
++----------------------------+--------------------------------+--------------------+
+|Remission_                  |Number of recovered cases from a|Estimate transition |
+|                            |given condition per person-year |rates               |
++----------------------------+--------------------------------+--------------------+
+|`Disability Weights`_       |Proportion of full health not   |Measure disability  |
+|                            |experienced due to disability   |attributed to       |
+|                            |associated with condition       |cause model states  |
++----------------------------+--------------------------------+--------------------+
+|`Cause-Specific Mortality`_ |Number of deaths due to a given |
+|                            |cause per person-year in the    |
+|                            |total population                |
++----------------------------+--------------------------------+--------------------+
+|`Excess Mortality`_         |Number of deaths due to a given |
+|                            |cause per person-year in the    |
+|                            |population with the condition   |
+
++----------------------------+--------------------------------+--------------------+
+
+.. _Prevalence:
+
+Prevalence
++++++++++
+
+Prevalence is defined as the **proportion of a given population that possesses a given condition or trait** at a 
+given timepoint.
+
+  For example, the prevalence of obesity in the United States was approximately 40% in 2016.
+
+When a time *frame* (such as 2016, i.e. 1/1/16 to 12/31/16) instead of a *timepoint* (such as 1/1/16) is 
+reported, it is commonly assumed that the reported prevalence represents the prevalence of the *midpoint* of 
+that timeframe (7/1/16 is the midpoint for the time frame of all of 2016). However, this may not always be the 
+case and it should be considered when relevant.
+
+Prevlance data can be used to **initialize cause model states** and represents the **probability that a simulant 
+will begin the simulation in a given state.**
+
+  For example, the probability that a simulant in a model of obesity in the United States beginning in 
+  2016 will begin the simulation as obese is 0.4 or 40%.
+
+.. _`Birth Prevalence`:
+
+Birth Prevalence
+++++++++++++++++
+
+Birth prevalence is defined as the **proportion of live births in a given population that possess a given 
+condition or trait at birth.**
+ 
+  For example, the birth prevalence for cleft lip in the United States in 2006 was 10.6 per 10,000 live 
+  births, or 0.106%.
+
+Birth prevalence data can be used to **initialize neonatal cause model states** and represent the **probability that a 
+simulant who is born during the simulation will be born into a given neonatal cause model state.** 
+
+  For example, the probability that a simulant born during a simulation of cleft lip in the United States 
+  in 2006 is 0.00106, or 0.106%.
+
+.. _Incidence:
 
 Incidence
 +++++++++
 
-Birth prevalence
-++++++++++++++++
+Incidence rates are defined as the **number of new cases of a condition that occur per person-year of the 
+at-risk population (individuals without condition).** Specifically, the at-risk population can be represented as 
+`1 - condition prevalence`.
+
+  For example, the incidence of multiple sclerosis (MS) in the United States is 2.8 per 100,000 
+  person-years of the at-risk population. 
+
+  This suggests that if we followed 100,000 individuals without MS for 1 year each (100,000 people * 
+  1 year = 100,000 person-years), we would expect 2.8 of these individuals to develop MS within this timeframe. 
+
+  Alternatively, if we followed 50,000 individuals without MS for 2 years each (50,000 people * 2 years = 100,000 
+  person-years), we also would expect 2.8 of these individuals to develop MS within this timeframe.
+
+Incidence can be used to **estimate cause model transition rates** and can represent the **probability that a simulant 
+will transition from a susceptible state to an infected state within a given timestep.** 
+
+  For example, with a timestep of one year and using incidence as the transition rate data source, the 
+  probability that a simulant will transition from a susceptible (without MS) cause model state to an 
+  infected (with MS) cause model state is 2.8*10^(-5).
+
+.. _above:
+
+**A Few Considerations for Incidence Data Sources:**
+
+As mentioned above, the denominator for incidence is person-years of the *at-risk* population, or the population 
+*without* condition (``1 - condition prevalence``). However, in certain scenarios, this may not always be the 
+case. 
+
+  In situations when the general population is represented in the denominator rather than the at risk population...
+
+    
+    If the prevalence of a condition is *small*, ``1 - prevalence`` ~ ``1``. In these cases, incidence 
+    calculated as the number of new cases per person-years in the *entire* population will be 
+    *approximately* equal to the number of new cases per person-years in the *at-risk* population. 
+    Therefore, the approximation will be fairly accurate and likely not have a large impact on the 
+    model transition rates.
+
+    If the prevalence of a condition is *large*, ``(1 - prevalence)`` < ``1``. In these cases, the 
+    approximation will be more inaccurate and may bias the model transition rates. 
+
+  Therefore, it is important to understand how incidence data sources used for cause models are measured 
+  and whether the population in the denominator represents the at risk population or the general 
+  population. If the population in the denominator represents the general population, the impact on the 
+  model and potential solutions to limit bias should be considered.
+
+    A potential solution may be to represent the transition rate with the following:
+
+    ``incidence rate`` * ``population size`` / ``(1 - prevalence)``
+
+Further, it is important to consider that cause models are *state*-specific and not necessarily 
+*disease*-specific. What does this mean?
+
+  Consider a cause model in which an individual can trasition from a susceptible state to a mild disease 
+  state OR from a susceptible state to a severe disease state.
+
+  In this case, the incidence rate for overall disease (mild and severe) does not help us estimate the 
+  transition rates from susceptible to mild disease or to severe disease. In these cases, incidence rates 
+  specific to mild and severe disease are needed to inform the specific transitions present in the cause model.
+
+Lastly, see the section on `hazard rates`_ in non-standard data sources below to determine when hazard rates may be 
+preferrable to annual incidence rates as a data source for cause model transition rates.
+
+.. _Remission:
 
 Remission
-+++++++++
-
-Prevalence
 ++++++++++
 
-Cause-specific mortality
+Remission rates are defined as the **number of newly recovered cases from a condition that occur per person-year 
+of the population with the condition.**
+
+  For example, the remission rate of type II diabetes in the United States is 2.4 per 10,000 person-years.
+
+  This suggests that if we followed 10,000 individuals with type II diabetes for one year each, we would 
+  expect to see 2.4 individuals recover from type II diabetes.
+
+Remission rates can be used to estimate cause model transition rates and represent the probability that a 
+simulant in an infected (with condition) state will transition to a non-infected (without condition) state. 
+
+  For example, with a time step of one year, the probability that a simulant in the infected state in a 
+  model of type II diabetes in the United States will transition to a susceptible or recovered state 
+  within a timestep is 2.4*10^(-4), or 0.024%.
+
+.. NOTE::
+
+  The considerations discussed in the incidence section above apply to remission rates as well. See above_ 
+  for details.
+
+.. _`Disability Weights`:
+
+Disability Weights
+++++++++++++++++++
+
+(TO-DO)
+
+.. _`Cause-Specific Mortality`:
+
+Cause-Specific Mortality
 ++++++++++++++++++++++++
 
-Excess mortality
+(TO-DO)
+
+.. _`Excess Mortality`:
+
+Excess Mortality
 ++++++++++++++++
 
-Disability weight
-+++++++++++++++++
+(TO-DO)
 
-Non-standard data sources for cause models
+.. _`Non-standard data sources`:
+
+Non-Standard Data Sources for Cause Models
 ------------------------------------------
+
+Duration
+++++++++
+
+In certain situations, there may be restrictions on the amount of time a simulant may occupy a given cause model 
+state. In these cases, it is important to specify the duration that simulants may remain in the state of interest.
+
+For example, in conditions that have acute and chronic phases, it may be necessary to specify the length 
+of time an individual occupies the acute phase before transitioning into the chronic phase.
+
+  E.g. In a cause model of ischemic heart disease, a simulant may transition from susceptible to a 
+  myocardial infarction state, where they remain for 28 days, before they transition to a ischemic 
+  heart disease state.
+
+Restrictions
+++++++++++++
+
+In addition to time-related restrictions discussed above, certain situations may require additional restrictions 
+to be placed on cause model states. Examples of possible restrictions include:
+
+- Age range restrictions (e.g. only simulants under 5 years old may enter this state) 
+- Sex restrictions (e.g. only female simulants may enter this state) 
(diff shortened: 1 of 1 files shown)


alibow:
I am unsure about how we want to handle the disability weight/mortality data sources. Should I expand upon these data sources in the same manner as I have done for the other included data sources, or do we want to merge the information from Yaqi's cause model impact pull request here?

yaqi7:
> I am unsure about how we want to handle the disability weight/mortality data sources. Should I expand upon these data sources in the same manner as I have done for the other included data sources, or do we want to merge the information from Yaqi's cause model impact pull request here?

@alibow The cause model impact is mainly about how does cause affect disability weight/mortality. It doesn't contain a lot description of disability weight/mortality data. I think it would be more systematic if these data sources could be in the same manner with other included data sources in data source section. If there is any overlap, I'd like to work on that too.

alibow:
> @alibow The cause model impact is mainly about how does cause affect disability weight/mortality. It doesn't contain a lot description of disability weight/mortality data. I think it would be more systematic if these data sources could be in the same manner with other included data sources in data source section. If there is any overlap, I'd like to work on that too.

@yaqi7 Ok! In this pull request, there are sections for disability weights, cause-specific mortality, and excess mortality that I did not complete. Feel free to clone this branch and add whatever information you want to these sections. Let me know if you would like me to work on any of it or if you have questions about how I structured things. Thanks :)

docs/source/models/cause.rst
```diff
@@ -144,7 +146,8 @@ What does a model document look like?
 * Validation criteria
 * Assumptions about the model
 
-[[to be updated based on experience from LTBI cause model document, and generalization thereof]]
+[[to be updated based on experience from LTBI cause model document,
+and generalization thereof]]
```
collijk: Everything above here is just line break fixes.

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+   within the cause model.
+#. The probability that a simulant will transition to a new state within the
+   cause model in a given time-step.
+#. The disability weight for each state in the cause model
+#. The probability that a simulant in a given cause model state will die in a
+   given time-step.
+
+There are several common data sources that can be used for these needs, which
+are outlined in the table below and discussed in more detail afterward.
+`Non-standard data sources`_ are discussed later on this page.
```
collijk: I don't think it's necessarily worthwhile to split standard and non-standard data sources into two sections.

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+and transitions. For our purposes, cause models generally have the following
+data needs:
+
+#. The probability that a simulant will start the simulation in a given state
+   within the cause model.
+#. The probability that a simulant will transition to a new state within the
+   cause model in a given time-step.
+#. The disability weight for each state in the cause model
+#. The probability that a simulant in a given cause model state will die in a
+   given time-step.
```
collijk: The biggest point this section needs to make is that our models use (approximately) instantaneous, individual-based probabilities to make decisions, but all of our data sources are period- or point-based population estimates.  That's the context for every subsection below.  

I think it's worth framing that we, in general, take population level estimates and use them as our best-guess predictors for individual-level estimates.  We then add heterogeneity to those guesses by adding more complexity to the model.  Your point 4 here has made this last step implicit.  We want to know, generally, what is the probability that a simulant will die in a given time step.  The data we actually need  for our model is how does being in a particular state in a cause model change the probability of death.

collijk: Which we inform by starting with a population level estimate (ACMR), deleting out the population level effect of the cause (CSMR) and adding back in our best guess at the additional mortality incurred by having the condition described by particular states in the cause model (EMR).

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+.. list-table:: Data Definitions
+   :widths: 20 30 30
+   :header-rows: 1
+
+   * - Measure
+     - Definition
+     - Uses
+   * - `Prevalence`_
+     - Proportion of population with a given condition.
+     - Initialize cause model states.
```
collijk: I think the Uses column of this table should describe what the quantity is used as rather than what it is used for.

E.g. Prevalence is the proportion of the population with a condition and it is used as the probability that an individual enters the simulation with the condition.

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
-Incidence
-+++++++++
+Prevalence
+++++++++++
+
+Prevalence is defined as the **proportion of a given population that possesses
+a given condition or trait** at a given time-point.
 
-Birth prevalence
+  For example, the prevalence of obesity in the United States was
```
collijk: Use a cause as an example here.  When we talk about obesity and other risk factors, we call it exposure.

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+     - Definition
+     - Uses
+   * - `Prevalence`_
+     - Proportion of population with a given condition.
+     - Initialize cause model states.
+   * - `Birth Prevalence`_
+     - Proportion of all live births born with a given condition.
+     - Initialize neonatal cause model states.
+   * - `Incidence`_
+     - Number of new cases of a given condition per person-year.
```
collijk: per person year of at risk

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+     - Estimate transition rates.
+   * - `Remission`_
+     - Number of recovered cases from a given condition per person-year.
```
collijk: per person year of with condition

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+
+- Age range restrictions (e.g. only simulants under 5 years old may
+  enter this state)
+- Sex restrictions (e.g. only female simulants may enter this state)
+- Restrictions related to states in other cause models (e.g. only simulants
+  who are susceptible to condition X may enter this state)
+- Etc.
+
+Hazard Rates
+++++++++++++
```
collijk: Ali to move to PR about survival analysis

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
 ++++++++++++++++
 
-Remission
+Birth prevalence is defined as the **proportion of live births in a given
+population that possess a given condition or trait at birth.**
+
+  For example, the birth prevalence for cleft lip in the United States in 2006
+  was 10.6 per 10,000 live births, or 0.106%.
+
+Birth prevalence data can be used to **initialize neonatal cause model
```
collijk: Prevalence is used to initialize states for the starting pop, birth prevalence for sims who enter by being born in the simulation.

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+  cleft lip in the United States in 2006 is 0.00106, or 0.106%.
+
+Incidence
 +++++++++
 
-Prevalence
+Incidence rates are defined as the **number of new cases of a condition that
+occur per person-year of the at-risk population (individuals without
+condition).** Specifically, the at-risk population can be represented as
+`1 - condition prevalence`.
```
collijk: Doing a little slight of hand with the definition of prevalence here.

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+
+  For example, the incidence of multiple sclerosis (MS) in the United States
```
collijk: Only true for immortal pops.  Maybe three examples
1. in a pop who never dies
2. in a realistic pop
3. thinking about age specific when you have a realistic pop and different size age bins (e.g. 7-28 day olds).

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+
+  This suggests that if we followed 100,000 individuals without MS for 1
+  year each (100,000 people * 1 year = 100,000 person-years), we would expect
+  2.8 of these individuals to develop MS within this timeframe.
+
+  Alternatively, if we followed 50,000 individuals without MS for 2 years each
+  (50,000 people * 2 years = 100,000 person-years), we also would expect 2.8
+  of these individuals to develop MS within this timeframe.
+
+Incidence can be used to **estimate cause model transition rates** and can
```
collijk: Be consistent about incidence vs hazard vs transition etc.

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+represent the **probability that a simulant will transition from a susceptible
+state to an infected state within a given timestep.**
```
collijk: Show how, then ref survival.rst (make a stub file for now).

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+
+  For example, with a timestep of one year and using incidence as the
+  transition rate data source, the probability that a simulant will transition
+  from a susceptible (without MS) cause model state to an infected (with MS)
+  cause model state is 2.8*10^(-5).
```
collijk: Only true for small rates where x ~= 1 - exp(-x)

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+As mentioned above, the denominator for incidence is person-years of the
+*at-risk* population, or the population *without* condition (``1 - condition
+prevalence``). However, in certain scenarios, this may not always be the
+case.
+
+  In situations when the general population is represented in the denominator
+  rather than the at risk population...
+
+
+    If the prevalence of a condition is *small*, ``1 - prevalence`` ~ ``1``.
```
collijk: This is not necessarily true.  Consider a low-prevalence, unrecoverable, morbidity only condition (e.g. amputee or untreatable blindness).  Because using the whole population in the denominator **biases** our results to be low, you'll see significant divergence in the disease prevalence in sufficiently long simulations.  

The point being, it's an okay approximation if you're just trying to do some data comparisons, but you should take care using biased approximators for simulation as it can have dramatic consequences.

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+    In these cases, incidence calculated as the number of new cases per
+    person-years in the *entire* population will be *approximately* equal to
+    the number of new cases per person-years in the *at-risk* population.
+    Therefore, the approximation will be fairly accurate and likely not have a
+    large impact on the model transition rates.
+
+    If the prevalence of a condition is *large*, ``(1 - prevalence)`` < ``1``.
+    In these cases, the approximation will be more inaccurate and may bias the
+    model transition rates.
```
collijk: Not may, definitely will.

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+
+  Therefore, it is important to understand how incidence data sources used for
+  cause models are measured and whether the population in the denominator
+  represents the at risk population or the general population. If the
+  population in the denominator represents the general population, the impact
+  on the model and potential solutions to limit bias should be considered.
```
collijk: I think this whole paragraph applies broadly to all the different kinds of data types.

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+
+    A potential solution may be to represent the transition rate with the
+    following:
+
+    ``incidence rate`` * ``population size`` / ``(1 - prevalence)``
```
collijk: Just incidence rate / (1 - prevalence).  The incidence rate is already in per person year terms.

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+
+Further, it is important to consider that cause models are *state*-specific
+and not necessarily *disease*-specific. What does this mean?
```
collijk: Try to use cause wherever you might be inclined to say disease.

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
 ++++++++++
 
-Cause-specific mortality
+Remission rates are defined as the **number of newly recovered cases from a
+condition that occur per person-year of the population with the condition.**
+
+  For example, the remission rate of type II diabetes in the United States is
+  2.4 per 10,000 person-years.
+
+  This suggests that if we followed 10,000 individuals with type II diabetes
```
collijk: Same thing as incidence.  Start with an immortal pop (this example), then talk about a real pop in which people die.

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+  for one year each, we would expect to see 2.4 individuals recover from type
+  II diabetes.
+
+Remission rates can be used to estimate cause model transition rates and
+represent the probability that a simulant in an infected (with condition)
+state will transition to a non-infected (without condition) state.
+
+  For example, with a time step of one year, the probability that a simulant
```
collijk: Maybe actually use an infectious disease as an example.  The remission rate for diarrheal diseases is like 70.  And that kind of extremely high rate is important to talk about.

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
 
-Disability weight
-+++++++++++++++++
+Non-Standard Data Sources
+-------------------------
+
+Duration
+++++++++
+
+In certain situations, there may be restrictions on the amount of time a
```
collijk: Add a todo here about describing the relationship that duration and transition rates can play when there are multiple ways out of a state.  The LTBI model will be a good example.  We'll talk this through later, just the todo for now.

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+phase before transitioning into the chronic phase.
+
+  E.g. In a cause model of ischemic heart disease, a simulant may transition
+  from susceptible to a myocardial infarction state, where they remain for
+  28 days, before they transition to a ischemic heart disease state.
+
+Restrictions
+++++++++++++
+
+In addition to time-related restrictions discussed above, certain situations
```
collijk: These are in every model

docs/source/models/cause.rst
NathanielBlairStahn: I think these calculations actually compute the **cumulative incidence** (according to Wikipedia, this is the number of new cases during a period divided by the number of at risk subjects at the beginning). As a function of time, this is the same as the cumulative distribution function for the random variable representing the time someone gets the condition. 

The incidence we want here (the one the model needs) is the **incidence rate**. The relationship is
(cumulative incidence) = 1 - exp[-(incidence rate) * duration],
assuming a constant rate over the time period.

So I think maybe the issue is not necessarily about simpler vs more complicated scenarios, but rather just being clear about which thing we're talking about.

docs/source/models/cause.rst
NathanielBlairStahn: I read through an introduction to survival analysis, and if I am understanding things correctly, all three of these are different terms for exactly the same thing, just in different contexts or coming from different disciplines:

incidence rate (epidemiology) = hazard rate (survival analysis) = transition rate (Markov processes) = force of mortality (demographics) = failure rate (reliability engineering)

Namely, all of these are the negative of the logarithmic derivative of the survival function S(t) = complementary cumulative distribution function for the random variable T representing the time until the relevant event happens:

S(t) = P(T >t)
lambda(t) = - d/dt (log S(t)) = -S'(t)/S(t)

The question then becomes how to estimate lambda(t) from data.

Does that seem accurate?

And yes, we should be consistent with terminology. I just wanted to make sure I understand what we're talking about.

docs/source/models/cause.rst
NathanielBlairStahn: What exactly is the sleight of hand again?

docs/source/models/cause.rst
NathanielBlairStahn: @collijk What do you mean by "used as" vs. "used for"? Do you mean we should be more specific about how we are using prevalence to initialize model states?

docs/source/models/cause.rst
NathanielBlairStahn: Or for small time steps (e.g. maybe a day, but not a year). Again, the difference is between the instantaneous rate vs. the cumulative effect over time.

docs/source/models/cause.rst
collijk: I mean we use, e.g., the prevalence *as* the probability that an individual has a condition when they enter which is used *for* initialization of the cause model states.

docs/source/models/cause.rst
collijk: Above we define prevalence as a point estimate and here we are implicitly using it as an annual average.

docs/source/models/cause.rst
collijk: You're accurate, but we're trying to bridge GBD terminology (incidence as essentially cumulative incidence over the year in an age group) with our transition rates.

docs/source/models/cause.rst
collijk: They do compute the cumulative incidence, but that's what GBD calls incidence.  We use it as essentially our best guess of the incidence hazard.

docs/source/models/cause.rst
```diff
@@ -171,29 +174,321 @@ Common data sources for cause models
...
+
+Prevalence is defined as the **proportion of a given population that possesses
+a given condition or trait** at a given time-point.
 
-Birth prevalence
+  For example, the prevalence of obesity in the United States was
+  approximately 40% in 2016.
+
+When a *time-frame* (such as 2016, i.e. 1/1/16 to 12/31/16) instead of a
+*time-point* (such as 1/1/16) is reported, it is commonly assumed that the
```
NathanielBlairStahn: This is vague. When is this assumption being made, and by whom?  And why?

docs/source/models/cause.rst
alibow: I changed to:  Specifically, the at-risk population at a given time can be 
represented as `1 - condition prevalence at that time`

alibow: (I changed it locally... remote updates coming soon!)

docs/source/models/cause.rst
alibow: Really? 

I was thinking we want (new cases)/(at risk person-year)

If we have (new cases)/(overall person-year), can we then multiply by (overall population)/(at-risk population) to get what we (new cases)/(at risk person-year)

Thinking about this again it seems like it may not be this simple....

Do we know if GBD always uses the at-risk population as the denominator for incidence estimates?

docs/source/models/cause.rst
```diff
@@ -262,16 +310,32 @@ the simulation will be born into a given neonatal cause model state.**
...
+
+  However, the above example does not consider the important impact of mortality within a population. 
+
+      So, let's imagine that we intend to follow 10 individuals for one year. Now let's imagine that two of the individuals die six months into our observation window. The other eight individuals remain alive and we follow each of them for the full year period. In this case, even though we intended to observe 10 person-years, we only observed 9 person-years! This is because the two individuals who died contributed only 0.5 person-years each to our observaton period.
+
+      Now, with this in mind, let's revist our previous example. We intend to follow a cohort of 100,000 individuals for one year each. However, several of these individuals die within that year and we end up observing a total of 90,000 person-years. Therefore, we would expect 2.52 individuals in this cohort to develop MS within this timeframe (2.8 cases / 100,000 py * 90,000 py).
```
alibow: I'm still working on these... please disregard for now!

docs/source/models/cause.rst
alibow: I updated this in my most recent commit

docs/source/models/cause.rst
alibow: done

docs/source/models/cause.rst
NathanielBlairStahn: Does the GBD always use cumulative incidence? For example, the Measles section in the YLD appendix specifically refers to the incidence rate:
> We modelled measles incidence in count space with a mixed-effects linear regression of case notifications on routine measles vaccination rates (first- and second-dose measles-containing vaccines) and coverage of supplementary immunisation activities (SIAs). More precisely, log-transformed **incidence rates** were regressed on the log of the proportion unvaccinated with first- and second-dose measles-containing vaccine, and additional SIA coverage lagged by one, two, three, four, and five years, with super-region, region, and country-level random effects:
Yij =beta_0+beta_1 MCV1ij +beta_2 MCV2ij +beta_a3 SIAaij +uj+eij,

docs/source/models/cause.rst
alibow: I added some clarification here

docs/source/models/cause.rst
alibow: updated this to just say that it will be biased, especially when prevalence is large

docs/source/models/cause.rst
alibow: I added a discussion of this before the table. Let me know if you think it could still use more detail!

PR #20, opened 2019-11-05
Common cause models

alibow:
Moved this pull request directly to master branch instead of from a fork. Added in images.


alibow:
Dropping neonatal cause models definitely sounds good to me. In my mind, it's helpful to have the basic SI, SIS, SIR models defined, but I understand why they might not be helpful to include here because they are more simplistic than our actual cause models. I can add in a disclaimer and drop neonatal models in a new commit or just close this PR based on if people think it is useful or not. I am leaning towards keeping it in with a disclaimer, but let me know!

collijk:
We might be able to use @aflaxman's model of diarrheal disease as our template for SI models and @NathanielBlairStahn's measles for SIR models.  We don't currently have anyone working on something that would be a good example of SI models.

docs/source/models/cause.rst
```diff
@@ -146,26 +146,107 @@ What does a model document look like?
...
++---------+--------------------------------+---------------------------------------------------------------------+
+|SIS_     |Susceptible-Infected-Susceptible|Simulants can recover from infected (with condition) state and can   |
+|         |                                |become infected again after recovery                                 |
+|         |                                |                                                                     |
++---------+--------------------------------+---------------------------------------------------------------------+
+|SIR_     |Susceptible-Infected-Recovered  |Simulants can recover from infected (with condition) state and cannot| 
+|         |                                |become infected again after recovery                                 |
+|         |                                |                                                                     |
++---------+--------------------------------+---------------------------------------------------------------------+
+|Neonatal_|With and without condition      |Simulants are born either with or without condition                  |
```
NathanielBlairStahn: I'm currently working on these models and thinking about how to label the model states. What do people think about "F" for "**Free** of Condition" and "C" for "With **Condition**"?

docs/source/models/cause.rst
```diff
@@ -146,26 +146,107 @@ What does a model document look like?
...
+.. image:: Neonatal_no_recover.png
+
+An example of a condition appropriate for a neonatal cause model with potential recovery from the condition 
+state is cleft palate (which may be treated later in life).
+
+.. image:: Neonatal_recover.png
+
+.. _`Non-Standard Cause Models`:
+
+Non-Standard Cause Models
```
NathanielBlairStahn: Perhaps "Non-Standard" is not the right terminology, because there isn't really a "Standard," but I'm not sure what a better term would be...

NathanielBlairStahn:
> I'm not totally sure this section is necessary. This document should be filled with references to actual cause model documents we use in production (e.g. a diarrhea model, an IHD model, the LTBI model, etc.)
> 
> If we do put in this section, it should come with a big warning that every cause model we've investigated in detail has been more complex than this, but that these serve as useful starting points in the conception of a cause model.

I think we should keep the section to provide context for looking at the specific diseases, and include links to diseases that are examples of each of the patterns. And yes, maybe a couple disclaimer sentences at the beginning explaining that the examples here are simple building blocks, and most of our models will be more complicated.

> Also, we should drop the neonatal models. There are like a half-dozen neonatal conditions which should be referenced directly because they don't form a pattern. They all have birth prevalence. Some also have remission. Some also have incidence. Not all are fatal. At least one of them has an issue where the fatal and nonfatal sides of GBD disagree about what the definition of the cause is. That's too much heterogeneity to talk about as a pattern.

Perhaps we can mention neonatal and congenital cause models in the section currently labeled "Non-Standard Cause Models" (but I think should be labeled something else). I kind of like having some of the diagrams here as examples because they're so simple, but I'm open to leaving them out if people think it's not worth including them.

docs/source/models/cause.rst
alibow: I changed the headings to "Basic Cause Model Structures" and "Other Cause Model Structures" in the most recent commit

docs/source/models/cause.rst
alibow: I think this could work! I think I remember someone mentioning that dismod has its own naming convention... maybe we could use that for consistency. Does anyone happen to know the abbreviations used in dismod?

NathanielBlairStahn: Abie's book on DisMod-MR uses S for "susceptible" and C for "with condition." But I don't like "susceptible" for the neonatal/congenital disorders because you're never actually susceptible - you're either born with the condition or not.

NathanielBlairStahn:
> Dropping neonatal cause models definitely sounds good to me. In my mind, it's helpful to have the basic SI, SIS, SIR models defined, but I understand why they might not be helpful to include here because they are more simplistic than our actual cause models. I can add in a disclaimer and drop neonatal models in a new commit or just close this PR based on if people think it is useful or not. I am leaning towards keeping it in with a disclaimer, but let me know!

@alibow Yes, I also think we should drop the section about neonatal models. Above, I suggested mentioning them in the "Other Cause Model Structures" section instead of having their own section at the same level as the basic cause models.  I would perhaps group "congenital disorders" (like neural tube defects) in with the neonatal models, as there are some similarities.  And on second thought, it's probably not worth including the diagrams here, just links to the actual cause models.

alibow:
> > Dropping neonatal cause models definitely sounds good to me. In my mind, it's helpful to have the basic SI, SIS, SIR models defined, but I understand why they might not be helpful to include here because they are more simplistic than our actual cause models. I can add in a disclaimer and drop neonatal models in a new commit or just close this PR based on if people think it is useful or not. I am leaning towards keeping it in with a disclaimer, but let me know!
> 
> @alibow Yes, I also think we should drop the section about neonatal models. Above, I suggested mentioning them in the "Other Cause Model Structures" section instead of having their own section at the same level as the basic cause models. I would perhaps group "congenital disorders" (like neural tube defects) in with the neonatal models, as there are some similarities. And on second thought, it's probably not worth including the diagrams here, just links to the actual cause models.

Thanks! I updated this in the last commit.

docs/source/models/cause.rst
```diff
@@ -146,25 +146,101 @@ What does a model document look like?
...
+transition to a recovered state where they will remain for the remainder
+of the simulation. Notably, the cause model allows individuals to become 
+infected only once in a simulation.
+
+This cause model structure is appropriate for conditions for which individuals 
+can only have a single case, but do not stay in the with condition state 
+forever.
+
+An example of a condition potentially appropriate for an SIR cause model 
+structure is measles.
```
NathanielBlairStahn:
```suggestion
structure is :ref:`measles <2017_cause_measles>`.
```

docs/source/models/cause.rst
```diff
@@ -146,25 +146,101 @@ What does a model document look like?
...
+In this cause model structure, simulants in the susceptible state can 
+transition to the infected state and simulants in the infected state can 
+transition to the susceptible state. Notably, this cause model allows for
+simulants to enter the infected state more than once in a simulation. 
+
+This cause model structure is appropriate for conditions for which individuals 
+can have multiple cases over their lifetimes.
+
+Examples of conditions potentially appropriate for an SIS cause model 
+structure include diarrheal diseases.
```
NathanielBlairStahn:
```suggestion
structure include :ref:`diarrheal diseases <2017_cause_diarrhea>`.
```

Might as well link to the causes we have documents for.

PR #21, opened 2019-11-07
tuberculosis cause model doc

yongqx:
The tables in data sources section only visible for html format because i created modifiable csv file instead of hardcode it.


collijk:
Looks excellent so far!  Can you add a section that discusses
1. The process (at a high level) of dismoding the ltbi incidence data.
2. Some discussion about the tb remission (we suspect there is a difference based on hiv status, whatever discussions you had, and our decision to not disaggregate based on expected time/effort to do so)

docs/source/models/gbd2017/causes/tuberculosis/tuberculosis.rst
```diff
@@ -0,0 +1,135 @@
+.. _tuberculosis_model:
+
+=====================
+Modeling Tuberculosis
+=====================
+
+TB Overview
+-----------
+Tuberculosis (TB) is a disease caused by the bacteria Mycobacterium
```
collijk: Are there relevant GBD sources we can cite here as well?  I think it's a good idea to link out to any info the GBD has published about their TB/Hiv estimates.

docs/source/models/gbd2017/causes/tuberculosis/tuberculosis.rst
```diff
@@ -0,0 +1,135 @@
...
+--------------
+Active TB disease developed within 2 years called fast progression, Individuals
+who do not have rapid progression are classified as having slow-progressing latent
+tuberculosis infection. With latent infection, individuals experience no adverse
+health effects and will not transmit Mycobacterium tuberculosis, but they face
+an ongoing risk of developing active tuberculosis through reactivation.
+
+What is TB/HIV model?
+---------------------
+It's a complex joint model that includes both tuberculosis and HIV/AIDS in the model.
```
collijk: We should include as part of the justification for the joint model the fact that this is what GBD does as well (as reflected in the hierarchy diagram below.

docs/source/models/gbd2017/causes/tuberculosis/tuberculosis.rst
```diff
@@ -0,0 +1,135 @@
...
+We model TB and HIV together since GBD 2017 TB model only involves HIV negative TB portion
+and we intend to model the impact of HIV on TB as well.  
+
+GBD hierarchy
+-------------
+.. image:: cause_hierarchy.svg
+
+- {Anemia} = [no_anemia, mild_anemia, moderate_anemia, severe_anemia]
```
collijk: Include some discussion of this diagram.  In particular, that the tb cases in different sections of the cause hierarchy are mutually exclusive.

docs/source/models/gbd2017/causes/tuberculosis/tuberculosis.rst
```diff
@@ -0,0 +1,135 @@
...
+.. csv-table:: *State Measures*
+   :file: state_inputs.csv
+   :header-rows: 1
+
+.. csv-table:: *Transitions*
+   :file: transitions.csv
+   :header-rows: 1
+
+The most recent model inputs documentation lives at
+`J:/Project/simulation_science/latent_tuberculosis_infection/report/model_inputs_researcher_ltbi.docx`
```
collijk: Should drop this.  *This document* should contain the mos recent model inputs documentation.

docs/source/models/gbd2017/causes/tuberculosis/tuberculosis.rst
```diff
@@ -0,0 +1,135 @@
...
+Data sources
+------------
+.. csv-table:: *State Measures*
+   :file: state_inputs.csv
+   :header-rows: 1
+
+.. csv-table:: *Transitions*
+   :file: transitions.csv
+   :header-rows: 1
+
```
collijk: Include assumptions about the model here.  recovery rates are the same is an important one.

docs/source/models/gbd2017/causes/tuberculosis/tuberculosis.rst
```diff
@@ -0,0 +1,135 @@
...
+The most recent model inputs documentation lives at
+`J:/Project/simulation_science/latent_tuberculosis_infection/report/model_inputs_researcher_ltbi.docx`
+
+========================================
+TB/HIV model validation and verification
+========================================
```
collijk: Your section headers are not going to render correctly because they're at the wrong levels.  Try building the docs locally and see how they look.

docs/source/models/gbd2017/causes/tuberculosis/tuberculosis.rst
```diff
@@ -0,0 +1,135 @@
...
+
+Obejective
+----------
+External
```
collijk: A lot of this section still seems unclear to me.  One way you could phrase this is a set of hypotheses and then how you plan to check them.

For example:

Hypothesis: The sum of the prevalences of all model states should equal the GBD prevalence of HIV (id 298) plus the prevalence of TB (id 297).
Check 1 (Data): Once the model input data is produced and put in the artifact, produce a graph of the sum of the input data prevalences and compare it to the GBD data not in the model.
Check 2 (Sim initialization): Initialize a simulation using the model input data and count people with condition to make sure they match with GBD data not in the model
Check 3 (Historical calibration): Run a simulation from 2005 to 2010 and count people at the end of the sim with condition to make sure they match with GBD data not in the model.
etc.

docs/source/models/gbd2017/causes/tuberculosis/tuberculosis.rst
```diff
@@ -0,0 +1,164 @@
...
+no evidence of clinically manifest active TB. (WHO)
+
+What is Active TB?
+------------------
+Active tuberculosis, also known as reactivation TB. It occurs when individual's
+immune system becomes weakened and is no longer able to contain the latent bacteria;
+then the TB bacteria become "active", overwhelm the immune system and cause a person
+to become ill. (UpToDate) In GBD, Active TB consist of 6 causes: Drug-susceptible
+TB, Multidrug-resistant TB, Extensively drug-resistant TB, HIV/AIDS - Drug-susceptible
+TB, HIV/AIDS - Multidrug-resistant TB, and HIV/AIDS - Extensively drug-resistant TB.
```
NathanielBlairStahn: Can we say something about why GBD models TB and HIV together? Even just a single sentence explaining why this is important would be good.

docs/source/models/gbd2017/causes/tuberculosis/tuberculosis.rst
```diff
@@ -0,0 +1,164 @@
...
+.. image:: cause_hierarchy.svg
+
+- {Anemia} = [no_anemia, mild_anemia, moderate_anemia, severe_anemia]
+- {HIV/AIDS} = [early_hiv, symptomatic_hiv, hiv_aids_with_art, aids]
+The causes in the GBD at any level of hierarchy are mutually exclusive
+and collectively exhaustive. 
+
+Cause model diagram
+------------------------
+.. image:: joint_model_diagram.svg
```
NathanielBlairStahn: I think the cause model diagram should have a key mapping the label in each state to what it stands for. I see that the data sources table below has definitions for each state, which is good, but I would also like to have a quick reference that tells me that, e.g. "LTBI, S_HIV" stands for "Latent Tuberculusis Infection, Susceptible to HIV" (is that accurate?).

docs/source/models/gbd2017/causes/tuberculosis/tuberculosis.rst
NathanielBlairStahn: Also please include some text explaining what is in the tables. For example, what does the "Initialization" column tell me? What do all your variable names represent (e.g. what is prev_954 or csmr_300)? What exactly does the formula "SUM(prev_s * dw_s) / SUM(prev)" in the disability_weight column mean? Is that calculation literally exactly the same for 5 of the states? I could try to guess the answers to these questions, but I shouldn't have to.

docs/source/models/gbd2017/causes/tuberculosis/tuberculosis.rst
```diff
@@ -0,0 +1,164 @@
...
+health effects and will not transmit Mycobacterium tuberculosis, but they face
+an ongoing risk of developing active tuberculosis through reactivation.
+
+What is TB/HIV model?
+---------------------
+It's a complex joint model that includes both tuberculosis and HIV/AIDS in the model.
+We model TB and HIV together since GBD has also studied the impact of HIV on TB as well.
+Note that we have excluded the active TB drug resistance for our model and GBD predicted
+the proportion of new TB cases with MDR-TB or XDR-TB by ST-GPR.
+(GBD appendix: 03_Hmwe_TB_Nonfatal2017_resub)
```
NathanielBlairStahn: I'm assuming that MDR-TB stands for "Multidrug-resistant TB" and XDR-TB stands for "Extensively drug-resistant TB," but you should explicitly say this somewhere so that the reader doesn't have to figure it out.

Why did you exclude active TB drug resistance from the model? Why is it important to mention that GBD predicted MDR-TB and XDR-TB cases with ST-GPR?

docs/source/models/gbd2017/causes/tuberculosis/tuberculosis.rst
NathanielBlairStahn: Also, I'm guessing that "extensively drug resistant" is worse than "multidrug resistant" -- is that correct? It might be nice to have a brief description of why there is a distinction between these two causes.

yongqx:
Thanks for all the comments! i will wrap it up and update ASAP.

NathanielBlairStahn:
@yongqx You should rename your `tuberculosis.rst` file to `index.rst` - then it will match our new directory structure and show up in the table of contents when you build with Sphinx.

PR #23, opened 2019-11-12
data sources intro/summary into separate PR for review

alibow:
organization/intro summary for data sources section... content to be reviewed in separate PR


collijk:
In person review was conducted with Nathaniel and Yongquan.  We decided to split Ali's bigger PR into pieces since there are so many important questions about each part.

PR #24, opened 2019-11-12
Prevalence data sources

alibow:


docs/source/models/cause.rst
```diff
@@ -260,9 +260,56 @@ Cause Model Initialization
...
+
+  For example, the prevalence of diabetes mellitus in the United States was
+  approximately 6.5% in 2017.
+
+When a *time-frame* (such as 2016, i.e. 1/1/16 to 12/31/16) instead of a
+*time-point* (such as 1/1/16) is reported, it is commonly assumed that the
+reported prevalence represents the prevalence of the *midpoint* of
+that time-frame (7/1/16 is the midpoint for the time frame of all of 2016).
+However, this may not always be the case and it should be considered when
+relevant.
```
NathanielBlairStahn: I still want some clarity about this point. Is it we as Vivarium modelers who are assuming this about our input data, or is it commonly assumed in other contexts?

aflaxman: My understanding is that this assumption is shared between we vivarium modelers and our colleagues on GBD who make these estimates.

NathanielBlairStahn: Cool, could we just be more specific then? Maybe "commonly assumed at IHME"? Or "commonly assumed with GBD data"? Or "commonly assumed when using GBD data in a Vivarium model"?

alibow: In our discussion with @collijk, he said that vivarium assumes the prevalence to be constant across the entire timeframe it represents. 

I'm thinking of changing this section to say "When prevalence is reported for a timeframe instead of a timepoint, Vivarium assumes the prevalence to be constant across the entire timeframe." And maybe an additional point such as "for seasonal conditions, additional considerations may be necessary."

I'm thinking about leaving out the midpoint value entirely because I'm not sure it's relevant to our uses after all. 

Do you all agree?

NathanielBlairStahn: Sounds good!

collijk: Nathaniel's direction is the right way to go.  We can make the model implementation in vivarium work in whatever way you think is best.  The framing is about what you as researchers commonly assume when interpreting GBD data.

PR #25, opened 2019-11-14
Feature/start measles

NathanielBlairStahn:
So far I just have the general disease description for measles. How does it look? Too much information? Not enough? Just right? Comments welcome!

### DIFF
--- a/docs/source/models/gbd2017/causes/measles.rst
+++ b/docs/source/models/gbd2017/causes/measles.rst
@@ -4,10 +4,10 @@
 Measles
 ==================
 
-Disease Information
+Disease Description
 ===================
 
-Measles is a highly contagious, serious disease caused by a virus.
+Measles_ is a highly contagious, serious disease caused by the `measles virus`_.
 Before the introduction of measles vaccine in 1963 and widespread vaccination,
 major epidemics occurred approximately every 2–3 years and measles caused an
 estimated 2.6 million deaths each year.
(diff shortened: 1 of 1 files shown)


NathanielBlairStahn:
Oops, I voided the previous approvals by pushing a new commit to fix some minor punctuation issues. Oh well, more people should review it anyway.

@collijk For future reference, how should we do the line lengths and headers correctly the first time so you don't have to make an extra commit?

docs/source/models/gbd2017/causes/measles.rst
```diff
@@ -34,7 +34,7 @@ measles caused an estimated 2.6 million deaths each year. Despite the
 availability of a vaccine, approximately 110,000 people died from measles in
 2017, mostly children under the age of 5 years. However, due to accelerated
 immunization activities, global measles deaths have decreased 80% during the
-period 2000–2017, from an estimated 545 000 to  110 000, and measles
+period 2000–2017, from an estimated 545,000 to  110,000, and measles
 vaccination prevented an estimated  21.1 million deaths during
 2000–2017 [WHO]_.
```
ctroeger: I think that you could add information about the global estimates for vaccine coverage or efficacy if you wanted. I anticipate that the vaccine, and potentially how GBD estimates vaccine coverage, might be important in your model.

collijk:
> Oops, I voided the previous approvals by pushing a new commit to fix some minor punctuation issues. Oh well, more people should review it anyway.
> 
> @collijk For future reference, how should we do the line lengths and headers correctly the first time so you don't have to make an extra commit?

I wrap lines at 80 characters as long as it doesn't do something dumb like break an html link.  Try not to break lines early.  I can probably find a thing to automate the wrapping at some point.  

Heading level guide with example: http://docutils.sourceforge.net/docs/user/rst/quickstart.html#sections

docs/source/models/gbd2017/causes/measles.rst
NathanielBlairStahn: Good points. When I do the GBD modeling section, I'll mention how they used estimates of vaccine coverage, and I can add some data to the description section if it seems relevant.

PR #27, opened 2019-11-15
Add concept model diagrams for Large Scale Food Fortification (vivarium_conic_lsff)

NathanielBlairStahn:
I'm going to try writing the LSFF concept model document(s) in the research documentation repo. Here's an initial commit.


collijk:
You can view the diagrams by clicking the "Display the rich diff" button at the top of the file.

yongqx:
I agree, it's good to add coverage cap "lack of food fortification" between intervention component and modifiable outcomes. An indicator of using coverage gap framework!

NathanielBlairStahn:
@collijk @yongqx Hmm, ok. I'm not that familiar with the coverage gap framework; I've just seen it mentioned in other concept model documents, but not fully explained. I have a few questions:

1. What exactly is the "coverage gap framework"?

>From our perspective, the "intervention" is increasing the coverage of fortification, not the fortifcation itself.

2. What is the difference between "the fortification itself" and "increasing the coverage of fortification"?

3. Why is it important to separate the intervention (or, rather, NOT the intervention, but the fortification?)  from the lack of coverage?

4. For the minimal LSFF models I am working on now, the intervention is essentially, "Assume the coverage of fortification increases linearly as a function of time from X% to Y% over 5 years," so I am failing to see the distinction between the fortification component and the coverage component. Does the difference become more important if we want to make a more complex model of the intervention?  Am I missing something?

PR #29, opened 2019-11-15
Feature/gbd cause detail

ctroeger:
I added an outline for the description of non-fatal modeling data, tools, and outputs for GBD. Anything that you think I am missing?

(Small pull request per recommendations).


docs/source/models/gbd2017/causes/causes.rst
```diff
@@ -94,3 +94,58 @@ specific reason, draws from CoDCorrect should be used to estimate cause-specific
...
+
+The documentation for GBD causes (write-ups) are typically separate Word documents by cause that become the Appendices
+for the study publication. If you have access to the shared drives at IHME, you can find the documentation on the
+I Drive: I:/RTs_and_Projects/GBD/Publications/Capstone Lancet Papers 2018/Cause Write-Ups/COD/Resubmission/4_final/
+
+The Appendix for the GBD 2017 Cause of Death manuscript is also available and Open Access at the Lancet website 
+`GBD COD Capstone
+<https://www.thelancet.com/journals/lancet/article/PIIS0140-6736(18)32203-7/fulltext#seccestitle540>`_
+
+
```
alibow: Maybe we could add a quick note about the type of info we might find here? Otherwise this all looks good to me!

docs/source/models/gbd2017/causes/causes.rst
```diff
@@ -6,7 +6,7 @@ Cause Models in GBD
 
 What is a cause?
 ----------------
-A cause in the GBD is something that is responsible for death, disease, or disability. 
+A cause in the GBD is something that is responsible for death, disease, injury or disability.
```
NathanielBlairStahn: This seems slightly confusing to me, in terms of being clear with our definitions. The way I think of it is that a cause is something that is responsible for death or disability, and the two categories of causes are diseases and injuries.

In other words, the cause **is** the disease or injury, and these are the things responsible for death and or disability. Does that match other people's understanding?

PR #30, opened 2019-11-15
Data sources/transitions

alibow:
In the spirit of small PRs, I am asking for review on my description of incidence as a data source, although there is more to come for the incidence description (see to-do notes in file) as well as more to come for the transition section in general.

Overall, I would like to know if the text is clear, if the examples are helpful, and if it is accurate within the GBD/Vivarium context. Any other thoughts and suggestions are welcome as well :)

### DIFF
--- a/docs/source/models/cause.rst
+++ b/docs/source/models/cause.rst
@@ -297,17 +297,16 @@ Generally, incidence is a measure of new cases of a given condition that occur
 in a given timeframe and population. The count value of new cases of the 
 condition of interest will always be the numerator of incidence measures. The 
 denominator of incidence measures is somewhat more complex and is critical to 
-ensuring an accurate data source to inform cause model transition rates.
+ensuring an accurate data source to inform cause model transition rates. 
 
-Incidence rates that are estimated by the Global Burden of Disease study are 
-measures of *cumulative incidence.* Cumulative incidence is defined using the 
-following concepts: 
+**Cumulative incidence** is a specific type of incidence measure that can be 
+used to represent cause model transition probabilities. Cumulative incidence 
+is defined using the following concepts: 
 
 .. todo: 
 
 	insert formatted equation for: new cases of condition / person-time of at 
   risk population
-
 
 **Person-time:** person-time is a measure of the number of individuals 
 multiplied by the amount of time they individually occupy the population of 
@@ -332,14 +331,6 @@ following events occur:
 	- Individuals with the condition recover from the condition and re-enter the 
   at-risk population as susceptible (in the case of conditions with remission)
 
-Cumulative incidence contrasts with other measures of incidence that may have 
-different denominators. For instance, incidence rate may refer to new cases of 
-a given condition that occur in a specified timeframe (e.g. 10 cases per year) 
-or may refer to new cases per person-time of the general population rather 
-than the at-risk population. Incidence measures other than cumulative with 
-person-time of the at-risk population as the denominator are not appropriate 
-for incidence data sources for cause model transition probabilities.
-
 Because the denominator for cumulative incidence is person-time in the at-risk 
 population, it can represent the probability of a new case of the condition 
 occuring in an individual without the condition in a given time frame. 
@@ -347,11 +338,12 @@ Therefore, it can be used to represent the probability that a simulant will
 transition from a susceptible to infected cause model state in a given 
 timestep.
 
-	For instance, as reported by the GBD, the global incidence of injuries in 
-  2017 was approximately 6,800 cases per 100,000 person-years, or 0.068 cases 
-  per person-year.
+	For instance, consider an example in which the global cumulative incidence 
+  of injuries in 2017 was 6,800 cases per 100,000 person-years, or 0.068 cases
+  per person-year. In this example, 6,800 new injuries occurred among 100,000 
+  person-years of observation among the non-injured population.
 
-	Consider a cause model with a susceptible (not injured) state and an 
+	Now, consider a cause model with a susceptible (not injured) state and an 
   infected (injured) state with a simulation timestep of 1 year. In this case, 
   the probability that a simulant will transition from the susceptible to 
   infected state within a single timestep (i.e. the transition probability) 
@@ -363,6 +355,25 @@ timestep.
   timestep. Therefore, if the timestep of the cause model considered above 
   were six months instead of one year, the transition probability would be 
   0.034 (0.034 cases per 0.5 person-years). 
+
+Cumulative incidence contrasts with other measures of incidence that may have 
+different denominators. For instance, incidence rate may refer to new cases of 
+a given condition that occur in a specified timeframe (e.g. 10 cases per year) 
+or may refer to new cases per person-time of the general population rather 
+than the at-risk population. Incidence measures other than cumulative with 
+person-time of the at-risk population as the denominator are not appropriate 
+for incidence data sources for cause model transition probabilities.
+
+.. warning:
+
+  GBD does not estimate cumulative incidence, but rather a true incidence rate 
+  of new cases per person-time of the total population. Incidence measures 
+  retrieved directly from GBD results will therefore not be accurate 
+  representations of transition probabilities. 
+
+  Vivarium automatically calculates cumulative incidence rates from GBD 
+  estimates of true incidence rates for use as cause model transition rate 
+  data source.
 
 There are several key assumptions and limitations of this approach, disscussed 
 below.
--- a/docs/source/models/cause.rst
+++ b/docs/source/models/cause.rst
@@ -278,45 +278,50 @@ Cause Model Transitions
 .. todo:
 
 	- Add progression transitions, deterministic transitions, and severity splits to summary table above? (should this be the case? or should these only be discusse din the transition section? I am thinking the latter)
-	- Add/enhance blurb to beginning of cause model transition section about how we use probabilies to inform cause model transitions (to come in next PR)
-	- Detail incidence, remisison, and duration-based transition sections (to come in next PR)
-	- Detail progression transitions, deterministic transitions, and severity splits (to come later)
+	- Enhance blurb to beginning of cause model transition section about how we use probabilies to inform cause model transitions (to come in next commits)
+	- Detail incidence, remisison, and duration-based transition sections (to come in next commits)
+	- Detail progression transitions, deterministic transitions, and severity splits (to come in future PRs)
 
 Vivarium uses probabilities to make decisions about how and when simulants move between cause model states. 
 
 Incidence Rates
 ^^^^^^^^^^^^^^^
 
-Generally, incidence is a measure of new cases of a given condition that occur in a given time frame. The count of new cases of the condition of interest will always be the numerator of incidence measures. The denominator of incidence measures is somewhat more complex and is critical to ensuring an accurate data source to inform cause model transition rates.
+Generally, incidence is a measure of new cases of a given condition that occur in a given timeframe and population. The count value of new cases of the condition of interest will always be the numerator of incidence measures. The denominator of incidence measures is somewhat more complex and is critical to ensuring an accurate data source to inform cause model transition rates.
 
 Incidence rates that are estimated by the Global Burden of Disease study are measures of *cumulative incidence.* Cumulative incidence is a defined as the following: 
 
-	- to-do: insert formatted equation for...
+.. todo: 
 
-		- new cases of condition / person-time of at risk population
+	insert formatted equation for: new cases of condition / person-time of at risk population
+
 
 **Person-time:** person-time is a measure of the number of individuals multiplied by the amount of time they individually occupy the population of interest.
 
 	For example, if one individual is in our population of interest for two years, they contribute two person-years. If another individual is in our population of interest for 6 months, they contribute 0.5 person-years. Together, these individuals contribute a total of 2.5 person-years.
 
-**At-risk population:** the at-risk population is defined as the population that *does not* have the condition of interest; in other words, the susceptible population that is at risk of developing the condition. Notably, this population will change overtime with the following scenarios:
+**At-risk population:** the at-risk population is defined as the population that *does not* have the condition of interest; in other words, the susceptible population that is at risk of developing the condition. Notably, the number of individuals in this population will change overtime as the following events occur:
 
 	- Members of the at-risk population develop of the condition and are no longer susceptible
 	- Members of the at-risk population die and are no longer susceptible 
 	- Individuals are born or age into the at-risk population and become susceptible
 	- Individuals with the condition recover from the condition and re-enter the at-risk population as susceptible (in the case of conditions with remission)
 
-Because the denominator for cumulative incidence is person-time in the at-risk population, it can be used to represent the probability of a new case of a condition occuring in an individual without the condition in a given time frame. Therefore, it can be used to represent the probability that a simulant will transition from a susceptible to infected cause model state in a given timestep.
+Cumulative incidence contrasts with other measures of incidence that may have different denominators. For instance, incidence rate may refer to new cases of a given condition that occur in a specified timeframe (e.g. 10 cases per year) or may refer to new cases per person-time of the general population rather than the at-risk population. Incidence measures other than cumulative with person-time of the at-risk population as the denominator are not appropriate for incidence data sources for cause model transition probabilities.
+
+Because the denominator for cumulative incidence is person-time in the at-risk population, it can represent the probability of a new case of the condition occuring in an individual without the condition in a given time frame. Therefore, it can be used to represent the probability that a simulant will transition from a susceptible to infected cause model state in a given timestep.
 
 	For instance, as reported by the GBD, the global incidence of injuries in 2017 was approximately 6,800 cases per 100,000 person-years, or 0.068 cases per person-year.
 
 	Consider a cause model with a susceptible (not injured) state and an infected (injured) state with a simulation timestep of 1 year. In this case, the probability that a simulant will transition from the susceptible to infected state within a single timestep (i.e. the transition probability) would be represented as 0.068.
 
-	Notably, in order to represent the transition probability for a single simulant within a single timestep, the cumulative incidence value needs to be scaled so that the person-time denominator is equal to the simulation timestep. Therefore, if the timestep of the cause model considered above were six months instead of one year, the transition probability would be 0.34 (0.34 casese per 0.5 person-years). 
+	Notably, in order to represent the transition probability for a single simulant within a single timestep, the cumulative incidence value needs to be scaled so that the person-time denominator is equal to the simulation timestep. Therefore, if the timestep of the cause model considered above were six months instead of one year, the transition probability would be 0.34 (0.34 cases per 0.5 person-years). 
 
 There are several key assumptions and limitations of this approach, disscussed below.
 
-	to-do: add assumptions and limitations
+.. todo:
+
+	Add description of key assumptions and limitations
 
 Remission Rates
 ^^^^^^^^^^^^^^^
--- a/docs/source/models/cause.rst
+++ b/docs/source/models/cause.rst
@@ -275,14 +275,30 @@ Birth Prevalence
 Cause Model Transitions
 +++++++++++++++++++++++
 
-Incidence
-^^^^^^^^^
+.. todo:
 
-Remission
-^^^^^^^^^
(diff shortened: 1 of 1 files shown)


docs/source/models/cause.rst
```diff
@@ -275,14 +275,68 @@ Birth Prevalence
...
+**Person-time:** person-time is a measure of the number of individuals multiplied by the amount of time they individually occupy the population of interest.
+
+	For example, if one individual is in our population of interest for two years, they contribute two person-years. If another individual is in our population of interest for 6 months, they contribute 0.5 person-years. Together, these individuals contribute a total of 2.5 person-years.
+
+**At-risk population:** the at-risk population is defined as the population that *does not* have the condition of interest; in other words, the susceptible population that is at risk of developing the condition. Notably, the number of individuals in this population will change overtime as the following events occur:
+
+	- Members of the at-risk population develop of the condition and are no longer susceptible
+	- Members of the at-risk population die and are no longer susceptible 
+	- Individuals are born or age into the at-risk population and become susceptible
+	- Individuals with the condition recover from the condition and re-enter the at-risk population as susceptible (in the case of conditions with remission)
```
alibow: Do we think this is helpful? Or distracting/unnecessary?

docs/source/models/cause.rst
```diff
@@ -275,14 +275,68 @@ Birth Prevalence
...
+
+	For instance, as reported by the GBD, the global incidence of injuries in 2017 was approximately 6,800 cases per 100,000 person-years, or 0.068 cases per person-year.
+
+	Consider a cause model with a susceptible (not injured) state and an infected (injured) state with a simulation timestep of 1 year. In this case, the probability that a simulant will transition from the susceptible to infected state within a single timestep (i.e. the transition probability) would be represented as 0.068.
+
+	Notably, in order to represent the transition probability for a single simulant within a single timestep, the cumulative incidence value needs to be scaled so that the person-time denominator is equal to the simulation timestep. Therefore, if the timestep of the cause model considered above were six months instead of one year, the transition probability would be 0.34 (0.34 cases per 0.5 person-years).
```
aflaxman: 0.034 not 0.34, IIUC

docs/source/models/cause.rst
NathanielBlairStahn: I think it's helpful!

docs/source/models/cause.rst
```diff
@@ -275,14 +275,68 @@ Birth Prevalence
...
+	- Detail progression transitions, deterministic transitions, and severity splits (to come in future PRs)
+
+Vivarium uses probabilities to make decisions about how and when simulants move between cause model states. 
+
+Incidence Rates
+^^^^^^^^^^^^^^^
+
+Generally, incidence is a measure of new cases of a given condition that occur in a given timeframe and population. The count value of new cases of the condition of interest will always be the numerator of incidence measures. The denominator of incidence measures is somewhat more complex and is critical to ensuring an accurate data source to inform cause model transition rates.
+
+Incidence rates that are estimated by the Global Burden of Disease study are measures of *cumulative incidence.* Cumulative incidence is a defined as the following:
```
NathanielBlairStahn:
```suggestion
Incidence rates that are estimated by the Global Burden of Disease study are measures of *cumulative incidence.* Cumulative incidence is a defined using the following concepts: 
```

docs/source/models/cause.rst
```diff
@@ -275,14 +275,68 @@ Birth Prevalence
...
+	insert formatted equation for: new cases of condition / person-time of at risk population
+
+
+**Person-time:** person-time is a measure of the number of individuals multiplied by the amount of time they individually occupy the population of interest.
+
+	For example, if one individual is in our population of interest for two years, they contribute two person-years. If another individual is in our population of interest for 6 months, they contribute 0.5 person-years. Together, these individuals contribute a total of 2.5 person-years.
+
+**At-risk population:** the at-risk population is defined as the population that *does not* have the condition of interest; in other words, the susceptible population that is at risk of developing the condition. Notably, the number of individuals in this population will change overtime as the following events occur:
+
+	- Members of the at-risk population develop of the condition and are no longer susceptible
```
NathanielBlairStahn:
```suggestion
	- Members of the at-risk population develop the condition and are no longer susceptible
```

docs/source/models/cause.rst
NathanielBlairStahn: I think this is helpful

docs/source/models/cause.rst
```diff
@@ -275,14 +275,68 @@ Birth Prevalence
...
+
+Cumulative incidence contrasts with other measures of incidence that may have different denominators. For instance, incidence rate may refer to new cases of a given condition that occur in a specified timeframe (e.g. 10 cases per year) or may refer to new cases per person-time of the general population rather than the at-risk population. Incidence measures other than cumulative with person-time of the at-risk population as the denominator are not appropriate for incidence data sources for cause model transition probabilities.
+
+Because the denominator for cumulative incidence is person-time in the at-risk population, it can represent the probability of a new case of the condition occuring in an individual without the condition in a given time frame. Therefore, it can be used to represent the probability that a simulant will transition from a susceptible to infected cause model state in a given timestep.
+
+	For instance, as reported by the GBD, the global incidence of injuries in 2017 was approximately 6,800 cases per 100,000 person-years, or 0.068 cases per person-year.
```
NathanielBlairStahn: Do the GBD incidence measures mean person-years for the total population as opposed to the at-risk population? It would be good if we clarified this somewhere.

NathanielBlairStahn: If so, then in the example of injuries below, we're approximating at-risk person-years by total person-years, which should be valid because the injury rate is low. Is that correct?

collijk:
Hold on merging til our working session monday.  We need to make some formatting fixes that I'll show you how to do.

docs/source/models/cause.rst
alibow: Yes, thank you! I fixed this in a new commit

docs/source/models/cause.rst
```diff
@@ -275,14 +275,68 @@ Birth Prevalence
...
+	- Detail progression transitions, deterministic transitions, and severity splits (to come in future PRs)
+
+Vivarium uses probabilities to make decisions about how and when simulants move between cause model states. 
+
+Incidence Rates
+^^^^^^^^^^^^^^^
+
+Generally, incidence is a measure of new cases of a given condition that occur in a given timeframe and population. The count value of new cases of the condition of interest will always be the numerator of incidence measures. The denominator of incidence measures is somewhat more complex and is critical to ensuring an accurate data source to inform cause model transition rates.
+
+Incidence rates that are estimated by the Global Burden of Disease study are measures of *cumulative incidence.* Cumulative incidence is a defined using the following concepts:
```
NathanielBlairStahn:
```suggestion
Incidence rates that are estimated by the Global Burden of Disease study are measures of *cumulative incidence.* Cumulative incidence is defined using the following concepts: 
```

docs/source/models/cause.rst
```diff
@@ -275,14 +281,125 @@ Birth Prevalence
...
+  were six months instead of one year, the transition probability would be 
+  0.034 (0.034 cases per 0.5 person-years). 
+
+Cumulative incidence contrasts with other measures of incidence that may have 
+different denominators. For instance, incidence rate may refer to new cases of 
+a given condition that occur in a specified timeframe (e.g. 10 cases per year) 
+or may refer to new cases per person-time of the general population rather 
+than the at-risk population. Incidence measures other than cumulative with 
+person-time of the at-risk population as the denominator are not appropriate 
+for incidence data sources for cause model transition probabilities.
```
NathanielBlairStahn: Again, change cumulative incidence to (average) incidence rate.

Ideally, I would like some context about when these different measures of rates are appropriate.

For example, why does GBD use total population person-time rather than at-risk person-time when reporting incidence?

docs/source/models/cause.rst
```diff
@@ -275,14 +281,125 @@ Birth Prevalence
...
+Incidence Rates
+^^^^^^^^^^^^^^^
+
+Generally, incidence is a measure of new cases of a given condition that occur 
+in a given timeframe and population. The count value of new cases of the 
+condition of interest will always be the numerator of incidence measures. The 
+denominator of incidence measures is somewhat more complex and is critical to 
+ensuring an accurate data source to inform cause model transition rates. 
+
+**Cumulative incidence** is a specific type of incidence measure that can be
```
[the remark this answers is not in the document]
NathanielBlairStahn: Aha! I figured out the relationship between the incidence rate and the hazard rate function, using the observation that the person-time of the at-risk population is the initial population times the area under the survival function curve, and the fact that the hazard function h(t) = -S'(t) / S(t).

Namely, I think the standard definition of incidence rate is (incident cases) / (person-time of at-risk population), and the following are equivalent characterizations:

incidence rate over an interval [t_1, t_2]
= (incident cases in [t_1 ,t_2]) / (person-time of at-risk population during [t_1, t_2])
= (decrease in S(t) from t_1 to t_2) / (area under S(t) from t_1 to t_2)
= - (S(t_2) - S(t_1)) / ((average fraction of population at risk during [t_1, t_2]) * (t_2 - t_1))
= (average rate of decrease of at-risk population) / (average at-risk population during [t_1, t_2])
= (integral of h(t)*S(t) from t_1 to t_2) / (integral of S(t) from t_1 to t_2)
= population-weighted average of h(t) on the interval [t_1, t_2]

In particular, the incidence rate is **not** equal to what I called lambda_t above (the unweighted average of h(t) on [t_1, t_2]), unless the hazard rate is constant.

I think these two different averages should have the following interpretations (but I haven't thought about it carefully):

- The unweighted average hazard rate should be the constant hazard rate that results in the same decrease in S(t) over the interval [t_1, t_2] if you started with equal at-risk populations.

- The population-weighted hazard rate (i.e. the incidence rate) should be the constant hazard rate that results in the same decrease in S(t) over the interval [t_1, t_2] if the two at-risk populations had the same person-time during the interval.

Anyway, not all of this should go on the page about incidence rates, but some of it should go on the page about survival analysis.

docs/source/models/cause.rst
alibow: This is addressed in most recent commit - thanks for the feedback!

docs/source/models/cause.rst
```diff
@@ -275,14 +275,116 @@ Birth Prevalence
...
+**Person-time:** person-time is a measure of the number of individuals 
+multiplied by the amount of time they individually occupy the population of 
+interest.
+
+	For example, if one individual is in our population of interest for two 
+  years, they contribute two person-years. If another individual is in our 
+  population of interest for 6 months, they contribute 0.5 person-years. 
+  Together, these individuals contribute a total of 2.5 person-years.
+
+**At-risk population:** the at-risk population is defined as the population
```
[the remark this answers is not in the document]
alibow: Done in most recent commit!

docs/source/models/cause.rst
```diff
@@ -275,14 +281,125 @@ Birth Prevalence
...
+denominator of incidence measures is somewhat more complex and is critical to 
+ensuring an accurate data source to inform cause model transition rates. 
+
+**Cumulative incidence** is a specific type of incidence measure that can be 
+used to represent cause model transition probabilities. Cumulative incidence 
+is defined using the following concepts: 
+
+.. math::
+
+  CumulativeIncidence = \frac{n_{incident cases}}{persontime_{atrisk}}
```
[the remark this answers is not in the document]
alibow: I updated the formula formatting in the most recent commit - thanks!

docs/source/models/cause.rst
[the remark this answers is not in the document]
alibow: In the most recent commits, I changed "cumulative incidence" to what we discussed as "total population incidence rate" and "susceptible population incidence rate". So far, I have focused on defining what these are and how they may be used in our model.

I am planning to get into how the susceptible population incidence rate and hazard rate relate in the limitation section (currently just a to-do still on this page) and in the linked survival analysis page.

Let me know if you think more of the information from these comments belongs in the existing text of this PR instead of the to-do limitation and survival analysis sections and I can work on adding it in. Otherwise, this will be helpful moving forward onto those sections, so thank you!

docs/source/models/cause.rst
alibow: Interesting... I am not exactly sure, but my guess is that incidence/total_pop is a more relevant measure for policy discussions than incidence/susceptible_pop. Incidence/total_pop allows for a quick measure of "what percent of my country's population will develop the disease this year". With incidence/susceptible_pop, a quick calculation (including condition prevalence) would be necessary to figure out that number.

Do you think this would be worth discussing in this section?

docs/source/models/cause.rst
```diff
@@ -275,14 +273,155 @@ Birth Prevalence
...
+    For example, if one individual is occupies the population of interest for 
+    two years, they contribute two person-years. If another individual is in 
+    our population of interest for 6 months, they contribute 0.5 person-years.
+    Together, these two individuals contribute a total of 2.5 person-years.
+
+  **Susceptible or At-Risk Population:** the susceptible population, also 
+  referred to as the at-risk population, is defined as the population that *
+  does not* have the condition of interest; in other words, the susceptible 
+  population that is at risk of developing the condition. Notably, the number 
+  of individuals in this population will change overtime as the following
```
NathanielBlairStahn:
```suggestion
  of individuals in this population will change over time as the following 
```

docs/source/models/cause.rst
```diff
@@ -275,14 +273,155 @@ Birth Prevalence
...
+**Susceptible/At-Risk Population Incidence Rate** as discussed here is also 
+referred to as incidence density rate, person-time incidence rate, and in some 
+cases may simply be referred to as the incidence rate. It is defined as:
+
+.. math::
+
+  \frac{n_\text{incident cases}}{\text{person-time}_\text{susceptible population}}
+
+Because the denominator for the susceptible population incidence rate is 
+person-time in the at-risk population, it can represent the probability of a
```
NathanielBlairStahn: I don't like the word "represent" here because the rate has dimensions of 1/time and is not the same thing as a probability, which is dimensionless (the rate is like a "probability density"). Rather, we can compute the probability if we know the rate (by integrating and then exponentiating). Here's a suggestion for different wording:

```suggestion
person-time in the at-risk population, this incidence rate can be used to compute the probability of a 
```

docs/source/models/cause.rst
```diff
@@ -275,14 +273,155 @@ Birth Prevalence
...
+cases may simply be referred to as the incidence rate. It is defined as:
+
+.. math::
+
+  \frac{n_\text{incident cases}}{\text{person-time}_\text{susceptible population}}
+
+Because the denominator for the susceptible population incidence rate is 
+person-time in the at-risk population, it can represent the probability of a 
+new case of the condition occuring in an individual without the condition in a 
+given time frame. Therefore, it can be used to represent the probability that
```
NathanielBlairStahn: Again, the rate is distinct from the probability:

```suggestion
given time frame. Therefore, it can be used to compute the probability that 
```

docs/source/models/cause.rst
```diff
@@ -275,14 +273,155 @@ Birth Prevalence
...
+
+Because the denominator of this measure is not specific to a particular cause 
+model state, it is **not** an appropriate data source for cause model 
+transition rates between states. 
+
+.. note:: 
+
+  GBD estimates of total population incidence rate require transformation 
+  prior to use as a cause model transition probability data source (see below 
+  for more detail).
```
NathanielBlairStahn: This sounds a little too absolute to me. Do we **always** transform the data prior to use, or are there cases (e.g. if we don't have good prevalence data but we know the prevalence is small), where we would just use the total population incidence rate in the model as a close-enough approximation? @collijk , any input on this?

Also, saying that the GBD estimates of total population incidence rates are "not an appropriate data source" makes it sound like we can't use them and we'll have to get data somewhere else. But that's not true - we do use them, we just have to combine this estimate with a different estimate to get the number we want.  So perhaps changing the wording of that sentence would help.

collijk: For all the models we use now, we always perform the transformation.  That is up to y'all though.

docs/source/models/cause.rst
NathanielBlairStahn: Or perhaps, "after scaling by time, the incidence rate can represent the probability...". I think we just have to be careful about glossing over the rate/probability distinction because the rates can be bigger than 1, whereas probabilities cannot. The rate can only be interpreted as a probability after multiplying it by a sufficiently small time step. For sufficiently small time steps, rate*time becomes a good estimate of the probability, but how small of a time step you need to get a good approximation depends on how large the rate is.

docs/source/models/cause.rst
```diff
@@ -275,14 +273,155 @@ Birth Prevalence
...
+  the probability that a simulant will transition from the susceptible to 
+  infected state within a single timestep (i.e. the transition probability) 
+  would be represented as 0.068.
+
+  Notably, in order to represent the transition probability for a single 
+  simulant within a single timestep, the cumulative incidence value needs to 
+  be scaled so that the person-time denominator is equal to the simulation 
+  timestep. Therefore, if the timestep of the cause model considered above 
+  were six months instead of one year, the transition probability would be 
+  0.034 (0.034 cases per 0.5 person-years).
```
NathanielBlairStahn: As per my comments above, I think it would be good to clarify that these are approximations of the probability, which (I think) are valid because the rate is small enough.

I'm not sure the best way to reword this. Perhaps the current version is good enough for now, and we can think about how to edit it to make it more precise in a future iteration.

docs/source/models/cause.rst
```diff
@@ -275,14 +273,155 @@ Birth Prevalence
...
+transition rates between states. 
+
+.. note:: 
+
+  GBD estimates of total population incidence rate require transformation 
+  prior to use as a cause model transition probability data source (see below 
+  for more detail).
+
+**Susceptible/At-Risk Population Incidence Rate** as discussed here is also 
+referred to as incidence density rate, person-time incidence rate, and in some
```
collijk: Both this and the total population incidence rate are technically incidence density rates.  That term just means you have person-time in the denominator.

NathanielBlairStahn: I think *incidence rate* and *incidence density rate* (or *incidence density*) are now essentially treated as synonyms. See e.g.:

http://www.sjsu.edu/faculty/gerstman/hs161/pre-inc-key-points.PDF

https://en.wikipedia.org/wiki/Incidence_(epidemiology)

https://www.investopedia.com/terms/i/incidence-rate.asp

I like the sound of "incidence rate" better than the other two options for general use, but it is good that we note the alternative terminology. In fact, when we say "... and in some cases may simply be referred to as the incidence rate," it may be more accurate to say that *usually*, the term "incidence rate" means the measure that we are calling the "susceptible population incidence rate."

Also note that in the last item of the first link above, they say that at-risk person time can be approximated by *(average population) x (time interval)*. It looks like they mean "average total population," which means that using this measure in the denominator would actually compute what we are calling the "total population incidence rate" that the GBD uses. So perhaps this is a common way to estimate the susceptible incidence rate when total population is the only data available?

docs/source/models/cause.rst
alibow: I will add some clarity to the language about how incidence as retrieved from GBD final results is not an appropriate data source **before** transformation to the at-risk population. 

Also, for context: as discussed with @collijk, our transformation is simply un-doing a transformation that GBD made to get incidence in the total population from an estimate of incidence in the susceptible population.

docs/source/models/cause.rst
```diff
@@ -275,14 +273,155 @@ Birth Prevalence
...
+
+.. note::
+
+  Because GBD estimates total population incidence rates, Vivarium 
+  automatically transforms GBD results into susceptible population incidence 
+  rates that can be used as an appropriate data source for cause model 
+  transition probabilities. 
+
+  This transformation from total population incidence rate to an approximation 
+  of the susceptible populatoin incidence rate is performed with the following
```
beatrixkh: population (sp)

docs/source/models/cause.rst
NathanielBlairStahn: Hmm, yes, I think it would be worth putting in a sentence or two explaining what GBD does and why, especially given this comment from below:

> Also, for context: as discussed with @collijk, our transformation is simply un-doing a transformation that GBD made to get incidence in the total population from an estimate of incidence in the susceptible population.

It looks like the todo mentions this, so perhaps we can add something in a future version.

PR #31, opened 2019-11-15
Incorporate some feedback from documentation session

NathanielBlairStahn:
I added a "Model assumptions and limitations" section, and I added some detail to some of the todo's. I also added some hyperlinks and made a few other small edits.

### DIFF
--- a/docs/source/models/gbd2017/causes/measles.rst
+++ b/docs/source/models/gbd2017/causes/measles.rst
@@ -13,31 +13,30 @@ the virus, and last 7-10 days. Symptoms include fever, cough, runny nose,
 conjunctivitis, characteristic white spots inside the cheek (called Koplik's
 spots), and a red, flat, blotchy skin rash that develops on average 14 days
 after exposure to the virus (range, 7-21 days) and lasts 5-6 days. Recovery from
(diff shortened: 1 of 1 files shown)


docs/source/models/gbd2017/causes/measles.rst
```diff
@@ -8,60 +8,105 @@ Disease Description
...
+   country's vaccine coverage went into the estimation of measles incidence
+   rates, which are then multiplied by an average disease duration of 10 days to
+   compute prevalence.
+
+   Describe enough of the data sources and modeling process to verify that even
+   though  measles can lead to diarrhea or other causes that we include in our
+   Vivarium models, we won't be double counting mortality and morbidity from
+   these causes. For example, a death caused by diarrheal dehydration due to
+   measles should be counted in the GBD as a death due to measles, not as a
+   death due to diarrheal diseases.
```
collijk: Check on the measles sequelae as well. I'm not sure what the non-fatal outcomes look like.  We've tended not to pay much attention to the non-fatal side of esp. under 5 diseases, but it sounds like some of the things here can persist well after someone recovers from measles and maybe that's important to think about.

NathanielBlairStahn: Sounds good, thanks for the suggestion.

PR #32, opened 2019-11-16
Attempt to get table of contents to work

NathanielBlairStahn:
The ..toctree:: directive is not finding my files in the `vivarium_conic_lsff` subfolder. I don't yet understand Sphinx well enough to fix it. Any suggestions?

### DIFF
--- a/docs/source/models/gbd2017/concept_models/index.rst
+++ b/docs/source/models/gbd2017/concept_models/index.rst
@@ -8,4 +8,5 @@ Simulation Concept Models
    :maxdepth: 2
    :glob:
 
+   */*
    *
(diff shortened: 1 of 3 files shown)


NathanielBlairStahn:
Now the table of contents should work. I had to add `*/*` instead of just `*` because the single wildcard wasn't catching subdirectories.

docs/source/models/gbd2017/concept_models/index.rst
```diff
@@ -0,0 +1,12 @@
...
+
+=========================
+Simulation Concept Models
+=========================
+
+.. toctree::
+   :maxdepth: 2
+   :glob:
+
+   */*
```
collijk: Use */index here.  You'll pick up a bunch of files you don't want with the raw glob.

NathanielBlairStahn: I incorporated this change in the "develop" branch where I'm writing the documentation: https://github.com/ihmeuw/vivarium_research/tree/develop/concept_model/lsff

PR #42, opened 2019-11-22
Add LSFF concept model diagram, cause & risk names

NathanielBlairStahn:
I'd love feedback on:
1. Titles of diagram columns (particularly my use of the word "covariates" - is it appropriate in this situation?)
2. Do the arrows in the diagram's title row look good, or are they distracting/unnecessary?
3. Is the diagram clear? Any other comments on the diagram?
4. How does the organization of the "Model Components" section look?


collijk:
Don't use covariates, as that's already a gbd thing that means something else. I don't like the arrows at the top.

NathanielBlairStahn:
> Don't use covariates, as that's already a gbd thing that means something else. I don't like the arrows at the top.

Any suggestions for what else to call the column with population coverage? How about "Population Features"?

collijk:
Why not just treatment coverage?

The problem with something generic like population features is that you could also use it to describe the risk and disease models.  And where would we put something like, e.g., household income which would just be a feature we track without a direct causal effect.

collijk:
And I don't like the arrows at the top because the flow of effect is usually, but not always, left to right.  Example: for a PAF of 1 cause like diabetes we may trust the cause model information more than the 
risk factor information for fasting plasma glucose.  We might then use the cause model to determine fpg expsoure and then the fpg exposure to determine the incidence of other diseases.

NathanielBlairStahn:
@collijk Does this look better? Good enough to approve?

docs/source/models/gbd2017/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -23,44 +23,58 @@ Scope of Modeling
...
+:ref:`Diarrheal Diseases <2017_cause_diarrhea>`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Folic Acid Fortification
-++++++++++++++++++++++++
+:ref:`Neural Tube Defects <2017_cause_neural_tube_defects>`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. image:: folic_acid_diagram.svg
+Lower Respiratory Infections
```
collijk: Can you go ahead and make stubs for all these other cause docs?

docs/source/models/gbd2017/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -23,44 +23,58 @@ Scope of Modeling
...
 ++++++++++++++++++++
 
-:ref:`Measles <2017_cause_measles>`
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Low Birth Weight and Short Gestation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-:ref:`Diarrheal Diseases <2017_cause_diarrhea>`
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Risk-Cause Relationships
```
collijk: Let's call these risk-outcome relationships.  Things we know exist but have not modeled yet: risk-risk relationships where exposure to one risk increases your exposure to another.

docs/source/models/gbd2017/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Yes, I'll create stubs for the missing ones. Lu created the LRI document, so that should be merged into master now.

docs/source/models/gbd2017/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Ok, I'll change this.

docs/source/models/gbd2017/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: @collijk Three questions:

1. Do we want to keep adding cause model documents in files directly in the gbd2017/causes folder, or would it be better to create a subfolder for each cause that includes the main page (renamed to index.rst to allow easy globbing?) as well as the image files, etc.?

2. Would it be better to push the requested changes to this pull request, then merge, or merge first, then create a new pull request adding stubs for missing causes and renaming the "Risk-Cause Relationships" section to "Risk-Outcome Relationships"?

3. Where should I put the stubs for the PAF-of-1 causes?  In the cause folder, or in a separate folder for "risk-attributable causes"?

PR #43, opened 2019-11-23
add cause model diagram for LRI.

XiaoluQianUW:


NathanielBlairStahn:
> Most people use remission rate to refer to the transition from I to S, and incidence rate to refer to S to I. I would like you to flip the labels on your arrows in your figure.

Good catch, I hadn't noticed they were reversed.

PR #46, opened 2019-11-26
clarify that gbd point prevalence represents year midpoint

alibow:
This PR is to update the existing description of prevalence as a data source for cause model initialization to clarify that GBD prevalence estimates are estimates of point prevalence at the midpoint of a given year and to discuss assumptions about how we use this point prevalence estimate across full simulation year(s). 

Let me know if you have any feedback! Thanks

### DIFF
--- a/docs/source/models/cause.rst
+++ b/docs/source/models/cause.rst
@@ -289,20 +289,11 @@ scenarios:
 	simulant is *born* into a simulation. See the below section on birth 
 	prevalence for how cause model states are initialized in this scenario.
 
-While GBD results are estimates of *point* prevalence at the year midpoint, 
-cause model initialization occurs continuously throughtout a full simulation 
-year. The following are two possible assumptions on how the GBD point 
-prevalence estimate might apply across the full simulation year that it 
-represents.
-
-- Point prevalence estimate is constant across the full year. This is likely 
-  appropriate in cases where prevalence is relatively constant over time. 
-
-- Prevalence estimates across several years follow a smooth line between 
-  annual midpoint point prevalence estimates. This approach would require 
-  fitting a curve across existing point prevalence estimates and may be 
-  appropriate when prevalence of a given condition is steeply increasing or 
-  decreasing over time.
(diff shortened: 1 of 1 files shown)


docs/source/models/cause.rst
```diff
@@ -298,9 +285,24 @@ scenarios:
...
 
-Prevalence is **not** used to initialize cause model states when a simulant 
-is *born* into a simulation. See the below section on birth prevalence for 
-how cause model states are initialized in this scenario.
+	Prevalence is **not** used to initialize cause model states when a 
+	simulant is *born* into a simulation. See the below section on birth 
+	prevalence for how cause model states are initialized in this scenario.
+
+While GBD results are estimates of *point* prevalence at the year midpoint, 
+cause model initialization occurs continuously throughtout a full simulation
```
collijk: Cause model initialization does indeed occur every time step, but prevalence is only used at simulation start.  Afterwards birth prevalence is used.  This might be different if we included something like migration.

docs/source/models/cause.rst
```diff
@@ -298,9 +285,24 @@ scenarios:
...
+While GBD results are estimates of *point* prevalence at the year midpoint, 
+cause model initialization occurs continuously throughtout a full simulation 
+year. The following are two possible assumptions on how the GBD point 
+prevalence estimate might apply across the full simulation year that it 
+represents.
+
+- Point prevalence estimate is constant across the full year. This is likely 
+  appropriate in cases where prevalence is relatively constant over time. 
+
+- Prevalence estimates across several years follow a smooth line between
```
collijk: Eh, maybe.  There are both age and time axes to consider here when we're talking about what we should fit a line too.  I'll draw a picture at coworking.

docs/source/models/cause.rst
alibow: Updated in last commit!

docs/source/models/cause.rst
alibow: Updated in last commit!

PR #47, opened 2019-11-27
Add stubs for causes in LSFF model

NathanielBlairStahn:
I added stubs for the neonatal causes and the PAF-of-1 vitamin A and iron deficiency causes, and I fixed a copy/paste typo in the "Risk Outcome Relationship Models" stub. I started a new directory structure where each cause goes in its own subfolder to keep the related files together, and I modified the appropriate index files accordingly.

### DIFF
--- a/docs/source/models/gbd2017/causes/index.rst
+++ b/docs/source/models/gbd2017/causes/index.rst
@@ -9,3 +9,4 @@ Cause Models
    :glob:
 
    *
+   */index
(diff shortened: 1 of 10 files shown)


NathanielBlairStahn:
> Looks good. If you don't mind doing some extra bookkeeping, submit a second PR that moves the current cause models to their own directories.

I'll create a pull request to reorganize the files that no one has worked on yet, and I'll coordinate with @XiaoluQianUW , @kiranklc , and @aflaxman for putting the files for LRI, measles, and diarrheal diseases in their own subfolders.

PR #48, opened 2019-12-01
Update GBD Hierarchy diagram and Move measles files into a folder.

kiranklc:


docs/source/models/gbd2017/causes/measles/index.rst
```diff
@@ -83,6 +83,31 @@ Modeling Measles in GBD 2017
...
++----------------------------------+------------+-----------+
+| Communicable, maternal, neonatal | 295        | 1         |
+| and nutritional diseases         |            |           | 
++----------------------------------+------------+-----------+
+| Other infectious diseases        | 961        | 0         |
++----------------------------------+------------+-----------+
+| Measles                          | 341        | 3         |
++----------------------------------+------------+-----------+
+
+- {Measles} = [Moderate Measles, Severe Measles]
```
aflaxman: Let's make it clearer that this is about the sequelae of measles, which we use only for calculating disability weights

docs/source/models/gbd2017/causes/measles/index.rst
```diff
@@ -83,6 +83,31 @@ Modeling Measles in GBD 2017
...
+
+**Hierarchy Diagram:**
+
+.. image:: measles_gbd_hierarchy.svg
+   :alt: Measles GBD hierarchy diagram
+
```
aflaxman: I don't think we need a hierarchy diagram for a simple model. For the HIV/TB model it was very helpful, so if others like to see it for simple models, too, please let me know here.

docs/source/models/gbd2017/causes/measles/index.rst
```diff
@@ -83,6 +83,31 @@ Modeling Measles in GBD 2017
...
++----------------------------------+------------+-----------+
+| **Cause**                        | **GBD ID** | **Level** |
++----------------------------------+------------+-----------+
+| All causes                       | 294        | 0         |
++----------------------------------+------------+-----------+
+| Communicable, maternal, neonatal | 295        | 1         |
+| and nutritional diseases         |            |           | 
++----------------------------------+------------+-----------+
+| Other infectious diseases        | 961        | 0         |
```
collijk: Think this should be level 2

docs/source/models/gbd2017/causes/measles/index.rst
```diff
@@ -83,6 +83,31 @@ Modeling Measles in GBD 2017
 
    Make sure to check on measles sequelae as well. Our models so far have not paid much attention to the nonfatal side, but it looks like some of the complications can persist well after someone recovers from measles, so maybe that's important to think about.
 
+GBD Hierarchy
+-------------
+
+**Hierarchy table:**
+
++----------------------------------+------------+-----------+
```
collijk: For simple tables like this, I highly suggest using list-table http://docutils.sourceforge.net/docs/ref/rst/directives.html#list-table

docs/source/models/gbd2017/causes/measles/index.rst
alibow: I think that it is worth including to avoid any surprises in GBD (just thinking about how anemia due to maternal hemorrhage is under a separate cause from maternal hemorrhage... could be worth including to be sure).

collijk:
Don't need both the table and the hierarchy.  I'd say just always put in the hierarchy image.  Use a different box type for sequelae though.

docs/source/models/gbd2017/causes/measles/index.rst
NathanielBlairStahn: I think it's nice to see the hierarchy diagram, particularly for people who aren't yet that familiar with the GBD.  And it could be useful if we want to add complexity to the model later. And for catching surprises like Ali said.

However, for simple models, it might be easier/faster to start with a table, then later put in the effort to replace it with a pretty diagram.

I think the diagram should label the different cause levels (0, 1, 2, 3, 4), and perhaps also indicate which level is the most detailed for the cause, and as James suggested below, use a different box type to visually distinguish the sequelae.

@kiranklc, I like how you show the prevalences of the sequelae in the diagram.

docs/source/models/gbd2017/causes/measles/index.rst
NathanielBlairStahn: @kiranklc Somewhere I would also quote the YLD appendix where they say they assume 50% of measles cases are moderate and 50% are severe. I don't remember seeing any data backing that up (though if we could find some, that would be good) - we should clearly indicate that this is an assumption of our model, based on what they did in the GBD.

kiranklc: @NathanielBlairStahn Sure. I will add that below the hierarchy diagram.

docs/source/models/gbd2017/causes/measles/index.rst
kiranklc: Thanks James. I will update it.

docs/source/models/gbd2017/causes/measles/index.rst
kiranklc: I will modify it the table to list.

kiranklc:
I have removed the hierarchy table and modified hierarchy diagram as per our discussion. Please review.

PR #51, opened 2019-12-06
Survival analysis page

alibow:
Hello! This is the start of the survival analysis page. For now, there are only headers and to-do notes in a separate survival analysis document. Please let me know if you think any sections are missing, unnecessary, out of order, etc. Also, if there is a better place to put the document please let me know. Thanks!

### DIFF
--- a/docs/source/models/survival_analysis.rst
+++ b/docs/source/models/survival_analysis.rst
@@ -1,0 +1,102 @@
+=================
+Survival Analysis
+=================
+
+.. todo::
+
+	Insert summary/introduction that explains the purpose of this section, how 
+	it relates to cause modeling, and what will be covered
+
+.. contents::
+	:local:
+
+----
+++++
+^^^^
+
+Terms and Definitions
+---------------------
+
+.. todo::
+
+	Add definitions and descriptions to terms
+
+.. list-table:: 
+   :widths: 20 20
+   :header-rows: 1
+
+   * - Term
+     - Definition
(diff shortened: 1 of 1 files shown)


docs/source/models/survival_analysis.rst
```diff
@@ -0,0 +1,117 @@
...
+
+	Link to cause model transition rates data source page
+
+	Briefly discuss how incidence is used as transition probability and 
+	assumption that it is constant over the timeframe it represents
+
+	Discuss how this assumption is valid for constant hazard rates, with 
+	examples
+
+	Discuss how this assumption is less valid for the following scenarios,
```
NathanielBlairStahn: This comment doesn't apply to this pull request specifically, but it's something to think about as you're writing this section:

There's some subtlety in determining what is a "valid assumption," because what is valid depends on the context in which the assumption is being used.

For example, just because the hazard function is not constant does not mean it is inappropriate to replace it with its average (i.e. the incidence rate) in a model - it depends on what we want the model to do.

If all we need the model to do is compute the incident cases in a specified time period, then replacing the nonconstant hazard function with its average over that period will still give the right answer.  On the other hand, if we need to model variability in incident cases over the time period, then we need more localized data about the hazard function than the overall average.  So the validity of an assumption depends on the goals of the model as well as the underlying process and data.

alibow: Good point! Thank you

PR #54, opened 2019-12-11
add sequelae descriptions and age restrictions for neural tube defects

NathanielBlairStahn:


docs/source/models/gbd2017/causes/neural_tube_defects/index.rst
```diff
@@ -114,6 +140,16 @@ Data Description
...
 
+
+Restrictions
+++++++++++++
+
+yll_age_group_id_start=2,
+yll_age_group_id_end=17,
+
+yld_age_group_id_start=2,
+yld_age_group_id_end=235,
```
aflaxman: Let's say in words what these age groups mean. IIUC, `2` is "early neonatal".  Does `235` mean "birth prevalence"?  And what is `17`?

docs/source/models/gbd2017/causes/neural_tube_defects/index.rst
```diff
@@ -68,6 +68,32 @@ separately and then fit to a total model of all neural tube defects.
...
+
+Sequelae and health states associated with Neural Tube Defects
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+The Neural Tube Defects cause has 85 sequelae, which can be summarized as follows:
+
+- All infants with anencephaly are assigned the health state of severe motor and
+  cognitive impairment.
+- Cases of spina bifida and encephalocele are split into
+  every combination of mild, moderate and severe motor impairment, all severities
```
aflaxman: how many ID severities are there?  I proposing including here as "all X severities of ID."

docs/source/models/gbd2017/causes/neural_tube_defects/index.rst
NathanielBlairStahn: I need to look these up.  I moved this into an explicit to-do in the next pull request (https://github.com/ihmeuw/vivarium_research/pull/55).  At the doc session today we discussed how to format the restrictions table, including listing actual ages rather than ids, so I'll incorporate that when I fill in the Restrictions section.

docs/source/models/gbd2017/causes/neural_tube_defects/index.rst
NathanielBlairStahn: It's probably the same as for motor impairment (mild, moderate, severe), unless there's also an 'asymptomatic' category.  I'll check and incorporate this in the next pull request.

PR #58, opened 2019-12-13
Add data description tables for measles model

kiranklc:


docs/source/models/gbd2017/causes/measles/index.rst
```diff
@@ -123,9 +128,128 @@ outbreaks due to lack of vaccination in small communities.
...
+   * - population
+     - Demography
+     - Mid-year population for |br| given country
+     - 
+   * - incidence_rate_c341
+     - COMO
+     - Incidence rate for measles
+     - 
+   * - remission_rate_c341
+     - Calculation
```
aflaxman: Apologies if you've already discussed this, but I thought the calculation would appear in this table directly, not just the word "calculation"

kiranklc: May be I misunderstood during the discussion, I have included calculation as value in the Transitions table. I will move that here.

docs/source/models/gbd2017/causes/measles/index.rst
```diff
@@ -123,9 +128,128 @@ outbreaks due to lack of vaccination in small communities.
...
+   
+   * - Transition
+     - Source 
+     - Sink 
+     - Value
+     - Notes
+   * - i
+     - S
+     - I
+     - incidence_rate_c341
```
NathanielBlairStahn: As per @beatrixkh 's comment in https://github.com/ihmeuw/vivarium_research/pull/60, we should change this to `incidence_rate_c341 / (1 - prevalence_c341)` (or `incidence_c341 / (1 - prevalence_c341)` if we're leaving "rate" implicit in our variable names...), since everything in this table should be numbers we can directly pull from GBD.

docs/source/models/gbd2017/causes/measles/index.rst
```diff
@@ -123,9 +128,128 @@ outbreaks due to lack of vaccination in small communities.
...
+     - Incidence rate in |br| total population
```
NathanielBlairStahn: Correspondingly, the note should say that we are taking the "total population incidence rate" and dividing by 1-prevalence to get the incidence rate among the susceptible population.

docs/source/models/gbd2017/causes/measles/index.rst
NathanielBlairStahn: We just discussed putting YLD Appendix for the source, and adding a note in the Notes column that we are using a constant remission rate based on the average duration of 10 days that they assumed.

NathanielBlairStahn:
> I think we need to come up with a clearer way to indicate a duration-based transition rather than a rate. I'll think about it some more and see if I can come up with any specific suggestions.

Actually, we are going with a standard rate-based transition, not duration based, so we can postpone figuring out duration-based transitions until later. There are still a few minor fixes to make though - see other comments.

docs/source/models/gbd2017/causes/measles/index.rst
```diff
@@ -123,9 +128,128 @@ outbreaks due to lack of vaccination in small communities.
...
+   * - r
+     - I
+     - R
+     - remission_rate_c341 |br| :math:`= \frac{\text{1}}{\text{10 person-days * 365 person-days}}` |br| :math:`= 36.5`
```
NathanielBlairStahn: This math is not quite right - should be:
rate = (1 / average duration) = (1 / 10 days) * (365 days / year) = 36.5 / year

docs/source/models/gbd2017/causes/measles/index.rst
```diff
@@ -123,9 +128,128 @@ outbreaks due to lack of vaccination in small communities.
 Data Description
 ----------------
 
-.. todo::
+.. list-table:: Definitions
+   :widths: 25 50
+   :header-rows: 1
+
+   * - State
+     - Definition
```
NathanielBlairStahn: @kiranklc @beatrixkh @XiaoluQianUW On Friday we discussed adding a "State Name" column to this table. Do we still agree that that's a good idea?

NathanielBlairStahn: It might seem a bit redundant for simple SIR or SIS models, but I think it becomes more important to come up with short descriptive names in models with less standard state definitions, like neural tube defects, neonatal hemolytic disease, or the tuberculosis model.

kiranklc: I agree that's a good idea. I will add the column to the definitions table.

kiranklc:
Updated the tables as per the comments. Please review the changes.

NathanielBlairStahn:
Looks good! Thanks, @kiranklc

docs/source/models/gbd2017/causes/measles/index.rst
```diff
@@ -123,9 +128,132 @@ outbreaks due to lack of vaccination in small communities.
...
+     - Notes
+   * - i
+     - S
+     - I
+     - :math:`\frac{\text{incidence_rate_c341}}{\text{1 - prevalence_c341}}`
+     - Incidence rate in total population |br| is divided by 1-prevalence_c341 |br| to get incidence rate among the |br| susceptible population.
+   * - r
+     - I
+     - R
+     - remission_rate_c341 |br| :math:`= \frac{\text{1} \times \text{365 person-days}}{\text{10 person-days} \times \text{1 year}}` |br| :math:`= \frac{\text{36.5}}{\text{year}}`
```
NathanielBlairStahn: You don't need the `\text{1} \times` in the numerator

docs/source/models/gbd2017/causes/measles/index.rst
```diff
@@ -123,9 +128,132 @@ outbreaks due to lack of vaccination in small communities.
...
+     - :math:`\frac{\text{deaths_c341}}{\text{population} \times \text{prevalence_c341}}`
+     - 
+   * - I
+     - disability weights
+     - disability_weight_s117 :math:`\times` |br| prevalence_s117+ |br| disability_weight_s118 :math:`\times` |br| prevalence_s118
+     - GBD assumes 50% of measles |br| cases as severe and other 50% |br| as moderate |br| [GBD-2017-YLD-Capstone-Appendix-1]_.
+   * - R
+     - prevalence
+     - 0
+     - Clearly room for improvement. |br| This is done to simplify the model |br| as the focus is on LSFF but not |br| on measles.
```
NathanielBlairStahn: Might be good to explicitly say that we don't have data for the number of recovered people, but the incidence rate we compute includes recovered as well as susceptible.

docs/source/models/gbd2017/causes/measles/index.rst
```diff
@@ -123,9 +128,132 @@ outbreaks due to lack of vaccination in small communities.
...
+   * - Transition
+     - Source 
+     - Sink 
+     - Value
+     - Notes
+   * - i
+     - S
+     - I
+     - :math:`\frac{\text{incidence_rate_c341}}{\text{1 - prevalence_c341}}`
+     - Incidence rate in total population |br| is divided by 1-prevalence_c341 |br| to get incidence rate among the |br| susceptible population.
```
NathanielBlairStahn: Actually, for this particular model, this incidence rate will be the incidence among the recovered as well as susceptible population, so we should note that here.

docs/source/models/gbd2017/causes/measles/index.rst
NathanielBlairStahn: Oops, forgot about the recovered state - dividing by 1-prevalence will give incidence among everyone who doesn't currently have measles, including those who are already recovered.

kiranklc:
Removed line breaks and done minor updates as per the comments.

PR #57, opened 2019-12-13
Created outline of IHD cause model

SantoniS:
This PR is for the outline. I'll then follow-up with PR requests for specific sections of this cause model.

PR #59, opened 2019-12-13
Ischemic stroke model

SantoniS:
This PR is for the outline. I'll then follow-up with PR requests for specific sections of this cause model.


SantoniS:
> This PR has your IHD model in it, too. But as long ar you merge the IHD PR #57 first, that should go away.

Thanks for your approval, Abie! I saw that the IHD model was in it so I must have made an error when I made this PR. I'll figure out why and make sure it doesn't happen in future PRs.

docs/source/models/gbd2017/causes/ischemic_stroke/ischemic_stroke.rst
```diff
@@ -0,0 +1,68 @@
...
+Strokes in GBD 2017
+++++++++++++++++++++++++++++++++++++++
+
+
+Modeling Strategy for Ischemic Stroke
+++++++++++++++++++++++++++++++++++++++++++++
+
+Cause Model Diagram
+--------------------------
+.. image:: ischemic_stroke_states.png
```
beatrixkh: Consider standardizing this to look more like some of the others--circles instead of rectangles, and labelled with the shorthand state name, to be referenced in the data input tables.

beatrixkh: Oops that's not actually accurate---it looks like Nathaniel's neural tube defects diagram is between the format you have and the one I described. Maybe we can confirm during coworking that we don't need to standardize!

docs/source/models/gbd2017/causes/ischemic_stroke/ischemic_stroke.rst
```diff
@@ -0,0 +1,68 @@
...
+  :width: 150
+
+Transitions Diagram
+--------------------------
+.. image:: ischemic_stroke_transitions.png
+  :width: 200
+
+Model Assumptions and Limitations
++++++++++++++++++
```
beatrixkh: I _think_ we're treating this one as it's own section (---- as opposed to +++), though I don't know if this has yet been discussed/standardized!

docs/source/models/gbd2017/causes/ischemic_stroke/ischemic_stroke.rst
NathanielBlairStahn: My old neural tube defects diagram had lots of words in the state bubbles, but in the new one, each state is labeled with a single letter (new diagram was merged into master on Friday).

docs/source/models/gbd2017/causes/ischemic_stroke/ischemic_stroke.rst
NathanielBlairStahn: That's where I've been putting it - as its own section, so it should be ------. I'm open to discussion if we want to change the organization, but that's what will match the current documents.

SantoniS: I updated this to be its own section (-----). Thanks, y'all!

PR #60, opened 2019-12-16
added data tables for LRI

XiaoluQianUW:


docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
```diff
@@ -46,10 +49,104 @@ Data Description
...
+     - Measure
+     - Value
+     - Notes
    * - S
-     - Susceptible but does not currently have LRI
+     - prevalence
+     - 1-prevalence_c322
+     -
+   * - S
+     - excess mortality |br| rate
```
beatrixkh: these |br| comments are showing up strangely in the compiled version. maybe remove?

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
```diff
@@ -46,10 +49,104 @@ Data Description
...
+     - prevalence
+     - prevalence_c322
+     -
+   * - I
+     - excess mortality |br| rate
+     - :math:`\frac{\text{deaths_c322}}{\text{population * prevalence_c322}}`
+     -
+   * - I
+     - disability weights
+     - disability_weight_s670* |br| prevalence_s670+ |br| disability_weight_s669* |br| prevalence_s669
```
beatrixkh: Maybe:

:math:`(\text{disability_weight_s670}\times\text{prevalence_s670})(\text{disability_weight_s669}\times\text{prevalence_s669})`

or without the parentheses--i think anything that gets rid of the |br|s works, unless perhaps my locally compiled version isn't what it will look like once published, and those are line breaks or something?

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
```diff
@@ -46,10 +49,104 @@ Data Description
...
+
+   * - Transition
+     - Source
+     - Sink
+     - Value
+     - Notes
+   * - i
+     - S
+     - I
+     - incidence_rate_c322
```
beatrixkh: I think we've been just labeling this as "incidence_c322", but we can revisit this, and if we're in agreement on that?

I was also writing this as incidence_c322/(1-prevalence_c322), to convert it to incidence within the susceptible population. I wasn't sure from @alibow 's documentation on incidence if the software engineers do this conversion themselves, or if we should do it?

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
```diff
@@ -46,10 +49,104 @@ Data Description
...
+     - DisMod-MR
+     - Remission rate for LRI
+     -
+   * - disability_weight_{sid}
+     - GBD YLD Appendix
+     - Disability weights associated |br| with each sequelae
+     -
+   * - prevalence_{sid}
+     - COMO
+     - Prevalence of each sequelae
```
beatrixkh: We decided to standardize to all lowercase for sources other than YLD: so como, codcorrect, demography---but then YLD appendix.

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
```diff
@@ -46,10 +49,104 @@ Data Description
...
+     -
+   * - incidence_rate_c322
+     - COMO
+     - Incidence rate for LRI
+     -
+   * - remission_rate_c322
+     - DisMod-MR
+     - Remission rate for LRI
+     -
+   * - disability_weight_{sid}
```
beatrixkh: I adjusted to Nathaniel's on this one---with disability_weight_s{sid} to make it absolutely clear that the id is a sequela_id (also prevalence_s{sid})

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
XiaoluQianUW: S

o






r




r


y

XiaoluQianUW: Sorry for the confusion, I fixed the |br| in my second commit. It was not right before

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
XiaoluQianUW: Thanks! I will fix this.

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
NathanielBlairStahn: I think we should explicitly put in the conversion of dividing by 1-prevalence, since the numbers we can directly pull from GBD are the "total population incidence rate" and the prevalence. I think Vivarium Inputs is written to do this transformation automatically, but we need to explicitly say whether this is what we want or if we want something different.

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
```diff
@@ -46,10 +49,104 @@ Data Description
 
    * - State
      - Definition
```
NathanielBlairStahn: On Friday we discussed adding a "State Name" column to this table. Do people still agree on that?

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
NathanielBlairStahn: > I think we've been just labeling this as "incidence_c322", but we can revisit this, and if we're in agreement on that?

What do you think @beatrixkh ? I think James recommended leaving out "rate" since this is the standard assumption, but I lean towards being more explicit and calling it incidence_rate.

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
collijk: Shouldn't State and State name be synonymous.  We don't have to use single letters in the diagram.

NathanielBlairStahn: I'm open to discussion about it, but I think it is useful to have both an abbreviation/label and a short name. Unless we want to disallow abbreviations and *only* use short names in the diagrams. And the Definition field is a place to further explain the precise meaning of the state in case it's not perfectly clear from the name (so maybe sometimes redundant, but sometimes not).

collijk:
think you can pull the break statements now ( run `git pull origin master` to bring in @kiranklc's changes).

PR #61, opened 2019-12-16
Filled out the section for DisMod in the non-fatal description.

ctroeger:
I added a text description for DisMod in the non-fatal section of the page.

I would like feedback on its completeness and level of detail. I also would be curious to hear ideas about how to make it more readable. I am not sure writing out long paragraphs is the best way to communicate the information.

### DIFF
--- a/docs/source/models/gbd2017/causes/causes.rst
+++ b/docs/source/models/gbd2017/causes/causes.rst
@@ -23,11 +23,11 @@ causes of death or disease are mutually exclusive and collectively exhaustive.
 
     - ... down to the most detailed level
 
-Measures for causes include:  
-Deaths, incidence, prevalence, YLDs, YLLs, and DALYs
+*Measures* are a quantity of disease burden. Measures for causes include:  
+deaths, incidence, prevalence, remission, excess mortality, YLDs, YLLs, and DALYs
 
-And metrics for causes include:  
-Counts, rates, and fractions
+*Metrics* are units that describe disease burden. Metrics for causes include:  
+counts, rates, and fractions
 
 Modeling causes of death
 ------------------------
@@ -108,13 +108,41 @@ The Appendix for the GBD 2017 Cause of Death manuscript is also available and Op
 Modeling non-fatal outcomes (Outline)
 ---------------------------
 
-- Figure of process
 .. image:: gbd_nonfatal_flowchart_sim_team.jpg
    :width: 600
 
-**Modeling approach** 
+**DisMod** 
 
-- Modeling tools like DisMod, ST-GPR
+Disease Model Meta Regression 2.1 (DisMod, sometimes DisMod MR) is a statistical modeling tool developed for the Global
+Burden of Disease study to estimate non-fatal disease burden. It is the most frequently used tool for non-fatal modeling
+in the GBD and models that are run outside of it are frequently called *custom models*. 
+
+DisMod has a user interface. The 
+`External Version
+<https://vizhub.healthdata.org/epi/>`_
+shows published and final models for GBD 2017.
+The 
+`Internal Version
+<https://internal.ihme.washington.edu/epi/>`_
+is where new and ongoing modeling occurs and is for GBD 2019 (but shows GBD 2017 best models). 
+
+There are three main components to DisMod. The first is that it
+is a meta-regression statistical model. This means that it uses point estimates with uncertainty around those estimates and covariates
+to predict disease burden for every location/year/age/sex. The second is that it is a compartmental model of disease. This
+means that DisMod is solving differential equations while fitting meta-regression estimates to enforce consistency between
+disparate measures of disease like incidence, prevalence, remission, and excess mortality. The third is that DisMod is age-integrating.
+Input data in DisMod must have an age or age range associated with those data. DisMod can account for ranges in age in the input
+data to help it produce continuous estimates of disease burden across all ages.
+
+Input data for DisMod can be any of the measures of disease that are estimated within it. These include prevalence, incidence,
+remission, excess mortality, and cause-specific mortality. Input data will be described in more detail in the following section.
+
+Results from DisMod are internally consistent *within* that model. Incidence, prevalence, remission, and excess mortality are
+linked in the DisMod estimation process and so the results from a model will include all these measures of disease. However, results
+from DisMod will *not* be consistent with final GBD estimates because of processes like COMO that rescale prevalence and sometimes
+incidence of non-fatal disease models. 
+
+**Other Modeling tools- ST-GPR**
 
 - Connection between fatal/non-fatal models
 
--- a/docs/source/models/gbd2017/causes/causes.rst
+++ b/docs/source/models/gbd2017/causes/causes.rst
@@ -27,7 +27,7 @@ causes of death or disease are mutually exclusive and collectively exhaustive.
 deaths, incidence, prevalence, remission, excess mortality, YLDs, YLLs, and DALYs
 
 *Metrics* are units that describe disease burden. Metrics for causes include:  
-counts, rates, and fractions
+counts, rates, and percentages
 
 Modeling causes of death
 ------------------------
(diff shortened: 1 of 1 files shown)


docs/source/models/gbd2017/causes/causes.rst
```diff
@@ -108,13 +108,41 @@ The Appendix for the GBD 2017 Cause of Death manuscript is also available and Op
...
+The 
+`Internal Version
+<https://internal.ihme.washington.edu/epi/>`_
+is where new and ongoing modeling occurs and is for GBD 2019 (but shows GBD 2017 best models). 
+
+There are three main components to DisMod. The first is that it
+is a meta-regression statistical model. This means that it uses point estimates with uncertainty around those estimates and covariates
+to predict disease burden for every location/year/age/sex. The second is that it is a compartmental model of disease. This
+means that DisMod is solving differential equations while fitting meta-regression estimates to enforce consistency between
+disparate measures of disease like incidence, prevalence, remission, and excess mortality. The third is that DisMod is age-integrating.
```
alibow: I think that this section could be numbered or bullet-ed to help keep it straight for the reader. Maybe bolded components that follow with a non-bold description of the component?

docs/source/models/gbd2017/causes/causes.rst
```diff
@@ -108,13 +108,41 @@ The Appendix for the GBD 2017 Cause of Death manuscript is also available and Op
...
+Input data in DisMod must have an age or age range associated with those data. DisMod can account for ranges in age in the input
+data to help it produce continuous estimates of disease burden across all ages.
```
alibow: Maybe we should bold or underline "continuous" estimates across all ages in order to emphasize this point because my understanding that this is not the case for final gbd results which are categorized by age group (assuming my understanding is accurate)... I'm only brainstorming here, but maybe this could be helpful for us to use one day for under 5 models or something that have some variation within age bins

docs/source/models/gbd2017/causes/causes.rst
```diff
@@ -23,11 +23,11 @@ causes of death or disease are mutually exclusive and collectively exhaustive.
 
     - ... down to the most detailed level
```
beatrixkh: Would be nice to represent this as a tree, such that you could see branches diverging

docs/source/models/gbd2017/causes/causes.rst
```diff
@@ -108,13 +108,41 @@ The Appendix for the GBD 2017 Cause of Death manuscript is also available and Op
 Modeling non-fatal outcomes (Outline)
 ---------------------------
 
-- Figure of process
 .. image:: gbd_nonfatal_flowchart_sim_team.jpg
```
beatrixkh: Currently, it notes under "Final results in GBD-Compare" that one can pull deaths, incidence, prev, YLDs, YLLs, DALYs. Might be nice to have another arrow pointing into that final green rectangle (from 'fatal modeling pipeline' or something similar), to show that half of those aren't a product of everything else visualized here--or in some other way differentiate those

The boxes "epi database" and "modeling and computational tools" might be clearer as labels for those sections if they don't have their own bubbles? But that's definitely up to personal preference!

docs/source/models/gbd2017/causes/causes.rst
```diff
@@ -108,13 +108,41 @@ The Appendix for the GBD 2017 Cause of Death manuscript is also available and Op
...
+DisMod has a user interface. The 
+`External Version
+<https://vizhub.healthdata.org/epi/>`_
+shows published and final models for GBD 2017.
+The 
+`Internal Version
+<https://internal.ihme.washington.edu/epi/>`_
+is where new and ongoing modeling occurs and is for GBD 2019 (but shows GBD 2017 best models). 
+
+There are three main components to DisMod. The first is that it
```
beatrixkh: I think it's worth specifying that (if I understand correctly) DisMod estimates each year that it estimates for separately--so it ends up creating excellent age patterns but sometimes rather poor time trends. (Though perhaps the second half of that sentence is unnecessary).

docs/source/models/gbd2017/causes/causes.rst
```diff
@@ -108,13 +108,41 @@ The Appendix for the GBD 2017 Cause of Death manuscript is also available and Op
...
+in the GBD and models that are run outside of it are frequently called *custom models*. 
+
+DisMod has a user interface. The 
+`External Version
+<https://vizhub.healthdata.org/epi/>`_
+shows published and final models for GBD 2017.
+The 
+`Internal Version
+<https://internal.ihme.washington.edu/epi/>`_
+is where new and ongoing modeling occurs and is for GBD 2019 (but shows GBD 2017 best models).
```
beatrixkh: I think this will be more digestible once the reader has learned, conceptually, what dismod is/does

docs/source/models/gbd2017/causes/causes.rst
```diff
@@ -108,13 +108,41 @@ The Appendix for the GBD 2017 Cause of Death manuscript is also available and Op
...
+Input data in DisMod must have an age or age range associated with those data. DisMod can account for ranges in age in the input
+data to help it produce continuous estimates of disease burden across all ages.
+
+Input data for DisMod can be any of the measures of disease that are estimated within it. These include prevalence, incidence,
+remission, excess mortality, and cause-specific mortality. Input data will be described in more detail in the following section.
+
+Results from DisMod are internally consistent *within* that model. Incidence, prevalence, remission, and excess mortality are
+linked in the DisMod estimation process and so the results from a model will include all these measures of disease. However, results
+from DisMod will *not* be consistent with final GBD estimates because of processes like COMO that rescale prevalence and sometimes
+incidence of non-fatal disease models.
```
beatrixkh: maybe add something like, "because of post-processes like COMO that rescale prevalence and sometimes incidence of different non-fatal models in order to achieve consistency across disease models."?

docs/source/models/gbd2017/causes/causes.rst
```diff
@@ -23,11 +23,11 @@ causes of death or disease are mutually exclusive and collectively exhaustive.
...
 
-Measures for causes include:  
-Deaths, incidence, prevalence, YLDs, YLLs, and DALYs
+*Measures* are a quantity of disease burden. Measures for causes include:  
+deaths, incidence, prevalence, remission, excess mortality, YLDs, YLLs, and DALYs
 
-And metrics for causes include:  
-Counts, rates, and fractions
+*Metrics* are units that describe disease burden. Metrics for causes include:  
+counts, rates, and fractions
```
beatrixkh: You later have a bullet with percentages rather than fractions---either seems fine, but I think it's helpful to choose one

docs/source/models/gbd2017/causes/causes.rst
beatrixkh: I would also consider adding (i don't know if you or if james or someone else mocked up the outline?) 'epi database' to the list of things that you're going to define

docs/source/models/gbd2017/causes/causes.rst
```diff
@@ -108,13 +108,41 @@ The Appendix for the GBD 2017 Cause of Death manuscript is also available and Op
...
+`External Version
+<https://vizhub.healthdata.org/epi/>`_
+shows published and final models for GBD 2017.
+The 
+`Internal Version
+<https://internal.ihme.washington.edu/epi/>`_
+is where new and ongoing modeling occurs and is for GBD 2019 (but shows GBD 2017 best models). 
+
+There are three main components to DisMod. The first is that it
+is a meta-regression statistical model. This means that it uses point estimates with uncertainty around those estimates and covariates
```
beatrixkh: Regarding covariates: I'm not sure when we'll switch from GBD 2017 to 2019--because I'm visualizing covariates as a dismod sub-section for 2017, and a crosswalk sub-section for 2019---but it seems relevant to mention the heterogeneity of most epi data in these models, and the role of covariates (/types of covariates) in making that data usable. Though I could also see an argument for keeping this section simple.

docs/source/models/gbd2017/causes/causes.rst
NathanielBlairStahn: I agree bullets or a numbered list would be good. You could also make it less wordy and make the 3rd point parallel in structure to the first two, e.g.

There are three main components to DisMod:

1. A meta-regression statistical model. This means...
2. A compartmental model of disease. This means...
3. Age integration. This means...

You could also try a definition list (rather than a numbered list) to implement Ali's suggestion of a bolded component followed by a description.

docs/source/models/gbd2017/causes/causes.rst
NathanielBlairStahn: Yeah, I'm still a little fuzzy about what exactly is meant by "age-integrating," why this is important, and why this aspect of DisMod is described at the same level as the other two components.

docs/source/models/gbd2017/causes/causes.rst
NathanielBlairStahn: Yes, I think this is definitely important to put in somewhere -- including, and perhaps especially, the second half of the sentence because that gives me an idea of some of the limitations of DisMod!

docs/source/models/gbd2017/causes/causes.rst
aflaxman: 👍

docs/source/models/gbd2017/causes/causes.rst
aflaxman: Age-integrating is a term I made up, to refer to how DisMod-MR can use data on very different age groups together.  It literally does an integral of the age-specific rate to figure out what the level should be for age a0 to a1. This is necessary when different studies report data for different age groups.  It is perhaps not too important for our simulation science work, but you can see how I define and defend this approach in Chapter 5 of _An Integrative Metaregression Framework for Descriptive Epidemiology_:
image

PR #62, opened 2019-12-16
Updated Ischemic stroke cause model states and transitions diagrams.

SantoniS:
@collijk reviewed the 'Cause model' and 'Transitions' diagrams and requested updates. This PR is for those updates of simplifying the diagrams.


SantoniS:
I updated the diagrams to be SVG file formats.

NathanielBlairStahn:
@SantoniS @collijk Is it necessary to have two separate diagrams, or can we just have one cause model diagram with both states and transitions (i.e. the one labeled transition diagram here)?

collijk:
Yeah, we just want your "transitions" diagram

SantoniS:
> @SantoniS @collijk Is it necessary to have two separate diagrams, or can we just have one cause model diagram with both states and transitions (i.e. the one labeled transition diagram here)?

I made this update so 'Cause Model Diagram' includes both states and transitions. Thanks!

PR #64, opened 2019-12-17
Cause model diagram and data tables for NN Enceph

NathanielBlairStahn:


docs/source/models/gbd2017/causes/neonatal_encephalopathy/index.rst
```diff
@@ -20,4 +20,176 @@ occur when an infant is deprived of oxygen during delivery or sustains physical
...
+.. image:: neonatal_encephalopathy_cause_model_diagram.svg
+
+There is no transition between the states. **Thus, incidence and remission
+rates are zero.**
+
+Data Description
+----------------
+
+State and Transition Data Tables
+++++++++++++++++++++++++++++++++
```
beatrixkh: Is there going to be anything additional under the first heading, besides the second heading? If not, maybe choose one? I only have 'data description' in mine right now, but the second one is more accurate.

NathanielBlairStahn: Ahh, at first I had the "Restrictions" in this section as well, so there were two subsections.  Hmm, if there won't be anything else in the Data Descriptions section, perhaps we should change the name?  Or maybe there are other things that should go in this section?  Not sure what the best way is - I think I'll leave it as it is for now, but we should discuss further and come back to this point.  I'll add a todo in that section.

PR #67, opened 2019-12-17
Add custom css file to wrap tables automatically

kiranklc:


NathanielBlairStahn:
Looks good, but I am slightly concerned about whether the tables will always render properly on different devices.

For example, if I make my (Chrome) browser window really narrow, it cuts off the table, and there's no scroll bar to see the right-hand side that gets cut off. Unless I make the window even narrower, in which case it reverts to the original style with scroll bars and no wrapping.

Are we fine with this behavior? Has anyone tried it on their phone to see if it still looks ok?

collijk:
This is better than what was there before.  If we notice problems in rendering, we'll make adjustments.

PR #69, opened 2019-12-18
Add assumptions and restrictions table for measles model

kiranklc:


docs/source/models/gbd2017/causes/measles/index.rst
```diff
@@ -117,9 +117,52 @@ vaccination campaign.
...
 outbreaks due to lack of vaccination in small communities.
 
+Some of the **assumptions** made in this model are:
+
+1. There is no data available for population in recovered state in GBD. 
+So, we considered all the population who do not have measles as susceptible  This includes both susceptible and recovered population.
+To compensate this, the incidence rate among susceptible and recovered population is used for transition.
+
+2. There is no data avaialable for remission rate in GBD. So a constant remission rate is 
+calculated from duration assumption of 10 days [GBD-2017-YLD-Capstone-Appendix-1]_.
```
NathanielBlairStahn: Let's say we calculate a constant rate from the assumption of an *average* 10 day duration.

docs/source/models/gbd2017/causes/measles/index.rst
```diff
@@ -117,9 +117,52 @@ vaccination campaign.
 2. This model uses country-level data, and cannot be used to model local measles
 outbreaks due to lack of vaccination in small communities.
 
+Some of the **assumptions** made in this model are:
+
+1. There is no data available for population in recovered state in GBD. 
+So, we considered all the population who do not have measles as susceptible  This includes both susceptible and recovered population.
+To compensate this, the incidence rate among susceptible and recovered population is used for transition.
```
NathanielBlairStahn: This is good for now, but we should do a bit more work to try to estimate the effect of this assumption. For example:

- There are 2 ways people can be in the "recovered" state - either they get measles and then recover, or they get vaccinated and move directly into the "recovered" state without ever having the disease. We should look into measles vaccination rates in the countries we're interested in (Nigeria, India, Ethiopia) and compare this to the number of people who actually get measles. If the number of vaccinated people is much higher than the number who get the disease, then our assumption will have a smaller effect, because the few people who enter the recovered state in our model will be be a small proportion of the total number of people in the recovered state, and the GBD incidence rate is already accounting for people who are "recovered" by vaccination.
- We should also look at the case fatality rate / excess mortality rate for measles, as this will also have an impact on the effect of this assumption, as well as on our assumption of a constant remission rate.
- There are ways we could try to estimate the people who are in the recovered state, but it is probably not worth the effort and added complexity for this model, particularly because we are not explicitly modeling vaccinations.
- For our assumption of a constant remission rate (below), we should think about what the actual hazard function for remission should look like (we should be able to get some idea about this from the disease description), and estimate how replacing it with a constant rate will affect our results.

@kiranklc Perhaps you can add versions of these specific comments to the to-do box.

kiranklc:
Updated as per the comments

docs/source/models/gbd2017/causes/measles/index.rst
```diff
@@ -117,9 +117,68 @@ vaccination campaign.
 2. This model uses country-level data, and cannot be used to model local measles
 outbreaks due to lack of vaccination in small communities.
 
+Some of the **assumptions** made in this model are:
```
beatrixkh: You might also add some of the assumptions that Natalie told us about---I don't remember all the specifics, but I believe in their nonfatal model they assumed that 95% of unvaccinated populations, globally, would be incident cases of measles? And I know this is in your states data table, but it also seems worth stressing that they assume a 50/50 split of moderate/severe measles in their estimates

kiranklc: @beatrixkh Thanks for your comments. But I think 95% attack rate assumption will be too confusing as this documentation is intended for all kinds of audience. We have already cited YLD Appendix and it is mentioned there. But I am open to discussion and if everyone think its important, I will add that.

kiranklc: I will include 50/50 split assumption. Thanks for mentioning that.

PR #70, opened 2019-12-18
Added remission section to cause page

beatrixkh:
In addition to flagging errors/suggestions re:the content already here, I have two types of questions:

1. To what degree should this (or any given section on this page) be able to stand alone?
- Some of what I've written is redundant with the DisMod description--do we want to keep it in?
- I left out some definitions (such as person-time) that are defined directly above, in the incidence section. 
2. Are there sections to add?
- I went into much less detail than Ali's incidence section, because these are so similar. However, I can add further examples and considerations analogous to her section.


docs/source/models/cause.rst
```diff
@@ -558,6 +558,53 @@ below.
...
+**Definition**
+
+Remission is a measure of cases that recover from a with-condition state, given 
+a specified population and time period. Just as with incidence, the numerator is 
+given by the count of recovered (or *remitted*) cases, and the denominator is 
+the cumulative person-time [link to person-time definition in incidence 
+section] during which cases are able to go into remission.
+
+  For example, consider diarrhea cases in the Philippines in 2017. Say there
+  were 1 million prevalent cases of diarrhea over the course of 2017, and each
```
NathanielBlairStahn: Are you referring to the period prevalence here as opposed to a point prevalence?  If so, we should probably clarify that because we usually mean point prevalence.

docs/source/models/cause.rst
```diff
@@ -558,6 +558,53 @@ below.
...
+refer to rates within an infected, or with-condition population. {Is this true? 
+This is based on limited google searching!}
+
+**Remission within GBD**
+
+Most nonfatal models in GBD are run using DisMod [link to DisMod page]. DisMod 
+estimates compartmental models of disease, which includes an estimate of 
+remission in terms of **{remitted cases in the with-condition population} per 
+{person-year}**. Thus DisMod's estimates of various measures (prevalence, 
+incidence, remission, excess mortality rate, etc.) are internally consistent for
```
NathanielBlairStahn: Are there more measures DisMod estimates that go into the "etc.", or are these the only four measures?

docs/source/models/cause.rst
```diff
@@ -558,6 +558,53 @@ below.
...
+  rate depends on the denominator we choose; some examples include:
+
+.. math::
+	
+	\frac{\text{1/5 case}}{\text{1 person-days}}=\frac{365/5\text{ cases}}{\text{1
+   person-year}}=73\text{ cases/person-year}
+ 
+Note that, unlike incidence, there is not a strong precedent for referring to 
+remission rates within the entire population; rather, remission rates typically 
+refer to rates within an infected, or with-condition population. {Is this true?
```
NathanielBlairStahn: My understanding was that this is also usually true of incidence rates, and using total population in the denominator is a quirk of GBD (or is used an approximation when more detailed data isn't available). But I could be wrong.

docs/source/models/cause.rst
beatrixkh: Yes, I am. I could also have said incidence; do you think that would be clearer @NathanielBlairStahn ?

docs/source/models/cause.rst
beatrixkh: DisMod also estimates CSMR, RR, standardized mortality ratio, with-condition mortality ratio, and other-cause mortality rate. Some of those are implicit in the other measures. My inclination would be to keep this with the 'etc.' and potentially push those off to the DisMod section? But happy to include if you/others think it would be helpful!

alibow: I agree with Beatrix about pushing off to the dismod section here

docs/source/models/cause.rst
alibow: Agreed!

docs/source/models/cause.rst
alibow: This is an interesting point! 

In GBD, incidence of diarrheal diseases >>> (point) prevalence of diarrheal diseases because of the high remission.

I'm not sure what the best approach to discussing this in this section will be. FYI Nathaniel, we're in coworking now and will discuss.

collijk:
Beatrix to add more examples and clarify the distinction between using period and point prevalence

docs/source/models/cause.rst
```diff
@@ -558,6 +558,113 @@ below.
...
+  and add the incident cases over the course of the year (approximately 150 
+  million cases). We emphasize that this would also not tell us the prevalence 
+  of TB among this cohort on January 1, 2018, because we also need to account 
+  for mortality in both the susceptible and infected groups over the course of 
+  the year, in addition to remission. This is thus a less helpful measure in 
+  this example.
+
+So, to which population does the remission rate apply? For the purposes of our 
+model, the *remission rate* will be the probability that a given prevalent case 
+who does *not* die during a given timestep remitts.
```
beatrixkh: @collijk i made this up---i'm don't actually know how you handle this or deal with mortality vs. remission. Can we either talk next coworking or office hours, or could you correct here?

docs/source/models/cause.rst
```diff
@@ -558,6 +558,113 @@ below.
...
+  Note that in this final example, as the duration of infection is on the same 
+  order of magnitude as the timestep, *period prevalence* is not approximated by 
+  either incidence or point prevalence. To find period prevalence, we would start 
+  with the point prevalence on January 1, 2017 (approximately 30 million cases), 
+  and add the incident cases over the course of the year (approximately 150 
+  million cases). We emphasize that this would also not tell us the prevalence 
+  of TB among this cohort on January 1, 2018, because we also need to account 
+  for mortality in both the susceptible and infected groups over the course of 
+  the year, in addition to remission. This is thus a less helpful measure in 
+  this example.
```
beatrixkh: @alibow @NathanielBlairStahn i also made this up. Is this how this would work?

As discussed with @alibow , i wanted to include examples of causes with remission rates that were much smaller than, much larger than, and on the same order as the timesteps being taken (here, one year)

I think that that works more to illustrate the difference between point and period prevalence than remission. I think it _does_ relate to remission insofar as which population to apply the remission rate to. (see section below).

is this correct? if so, should we keep it in, or is this getting to convoluted?

we could alternately leave this section to just contain examples of duration of condition being simply greater or less than the timestep.

docs/source/models/cause.rst
```diff
@@ -558,6 +558,113 @@ below.
...
+  For example, consider diarrhea cases in the Philippines in 2017. Say there
+  were 53,362,490 prevalent cases of diarrhea over the course of 2017, and each
+  case remitted after an average of 5 days. Note that here we are using *period
+  prevalence*; however, GBD (and thus we) typically mean *point prevalence* when 
+  we refer to prevalence. Then, the way we then define this remission rate 
+  depends on the denominator we choose. Some examples include:
+
+  .. math::
+
+    \frac{\text{1/5 case}}{\text{1 person-days}}=\frac{365/5\text{ cases}}{\text{1 person-year}}=73\text{ cases/person-year}
```
NathanielBlairStahn: 2 comments:

1. It might be clearer to say (1 case) / (5 person-days) rather than putting the fraction 1/5 in the numerator.
2. Since the paragraph starts with the period prevalence of diarrheal diseases, I was expecting that number to show up in the calculation, but this calculation is independent of that number - it just uses the average duration. It would be less confusing if you didn't mention the period prevalence at all (or introduce it later if you need it for something else).

docs/source/models/cause.rst
```diff
@@ -558,6 +558,113 @@ below.
...
+
+  We also note that in the context of a one-year timestep, as diarrheal diseases
+  have such a high remission rate relative to one year, period prevalence is
+  effectively the same as incidence. We can see how this would not be the case
```
NathanielBlairStahn: This is not immediately obvious to me. Perhaps it will become obvious if I spend a minute or two thinking about it, but it's probably better to add a brief explanation of why period prevalence is close to incidence in this case so I don't have to spend that minute or two...

docs/source/models/cause.rst
beatrixkh: EDIT: this was too convoluted, simplified/removed this section

docs/source/models/cause.rst
beatrixkh: EDIT: i removed this section/simplified the whole thing.

docs/source/models/cause.rst
beatrixkh: 1. Sounds good

2. Yes! I think I backed myself into a corner with point/period prevalence, when it isn't entirely relevant here. I've simplified the whole section down.

NathanielBlairStahn: I like the simplified version much better!

docs/source/models/cause.rst
```diff
@@ -558,6 +558,85 @@ below.
...
+  diabetes (both type I and type II) in Moldova amongst males in 2017, and of 
+  those 142,794 cases, 509 remitted in 2017. This gives us the following rate:
+
+  .. math::
+
+    \frac{\text{509 cases}}{\text{142,794 person-years}} =
+    \frac{\text{0.0036 cases}}{\text{1 person-year}}
+
+  It is important to note that this is a remission rate for diabetes at all
+  ages. In the context of the diarrheal diseases example, this makes it seem as
```
NathanielBlairStahn: Why is it important to note that this is a remission rate for diabetes at all ages? (And, for example, why was it not important to note this for diarrhea? Presumably because remission rates vary significantly across ages for diabetes, but not for diarrhea?) It's just a bit unclear how this sentence fits in with the rest of the discussion, since then you specifically discuss type I vs. type II, but don't mention ages again.

In general, though I think this new version is much clearer.

docs/source/models/cause.rst
```diff
@@ -558,6 +558,85 @@ below.
...
+  It is important to note that this is a remission rate for diabetes at all
+  ages. In the context of the diarrheal diseases example, this makes it seem as 
+  if diabetes cases remit, on average, after :math:`\frac{1}{0.0036}\simeq 279` 
+  years--which clearly does a poor job of capturing the behavior of diabetes. 
+  This sort of description was only appropriate for diarrhea, as there is a 
+  uniform remission rate across all cases. With diabetes, remission is not 
+  guaranteed for type II, and doesn't occur for type I, so remission rate does 
+  *not* tell us the average duration that an individual will experience diabetes.
+
+For both of the above rates, we think of :math:`r_1=0.2` as the probability that
```
NathanielBlairStahn: I was a bit confused about where r_1=0.2 came from (since the number 0.2 doesn't appear anywhere above), until I realized it was the remission rate for diarrhea measured in cases/person-day.

I think this gets at a technical mathematical point where we tend to be a bit sloppy in daily usage. Namely, the rates themselves don't represent probabilities, because they have physical dimensions of 1/time, whereas a probability needs to be dimensionless. Hence my confusion, because why can we use the daily remission rate as a probability but not the yearly rate (which clearly doesn't make sense because it's bigger than 1)?

Rather, to get a probability, we always have to multiply the rate by some unit of time (in this case 1 day for diarrhea, and 1 year for diabetes). Then we get a truly dimensionless number, z=(rate x time), and the probability of remission is given by 1-exp(-z). If z is **small enough**, then 1-exp(-z) is well-approximated by z, so we can use z=(rate x time) itself as an approximation for the probability.

A lot of this should eventually be explained on the survival analysis page, but I think whenever we want to convert rates to probabilities, we should always explicitly multiply by the appropriate time unit, and explicitly say that we can use this as an approximation to the true probability since (rate x time) is small enough.

docs/source/models/cause.rst
```diff
@@ -558,6 +558,85 @@ below.
...
+a rate within the infected or with-condition population. This is true both in
+general, and in the context of GBD--unlike with incidence, which GBD calculates 
+within the entire population, as discussed above.
+
+**Remission within GBD**
+
+Most nonfatal models in GBD are run using DisMod (:ref:`2017_cause_models`). DisMod 
+estimates compartmental models of disease, which includes an estimate of 
+remission in terms of **{remitted cases in the with-condition population} per 
+{person-year}**. Thus DisMod's estimates of various measures (prevalence,
```
NathanielBlairStahn: I think it's more important to stress that the **person-years** are measured for the with-condition population. Remitted cases by definition have to come from the with-condition population, whereas person-years can be measured for different populations.

docs/source/models/cause.rst
```diff
@@ -558,6 +558,85 @@ below.
 Remission Rates
 ^^^^^^^^^^^^^^^
 
+**Definition**
```
NathanielBlairStahn: Why not use another level of .rst subheadings instead of merely bolding the different subsection headings?

docs/source/models/cause.rst
NathanielBlairStahn: Also, I think this definition (i.e. clarifying that remission uses person-time in the with-condition population) might make more sense in the previous paragraph where you first mention this.

docs/source/models/cause.rst
```diff
@@ -558,6 +558,85 @@ below.
...
+GBD runs standardizing processes on estimates of prevalence, incidence, and 
+estimated mortality rate. Note then that these final estimates are no longer 
+consistent with the DisMod estimates. However, as remission is not needed to 
+calculate YLDs, the latest-stage estimate of remission produced by GBD comes 
+from DisMod models.
+
+**Implementing remission rates in cause models**
+
+For a given simulation with timesteps of length {x time-units}, we convert 
+remission rates to the form of cases remitted per person-{x time-units}. This
```
NathanielBlairStahn: I think the curly braces and variable name x here are distracting. Maybe try to write this in plain English without explicit math-y or code-y bits.

docs/source/models/cause.rst
```diff
@@ -558,6 +558,85 @@ below.
...
+    \frac{\text{0.0036 cases}}{\text{1 person-year}}
+
+  It is important to note that this is a remission rate for diabetes at all
+  ages. In the context of the diarrheal diseases example, this makes it seem as 
+  if diabetes cases remit, on average, after :math:`\frac{1}{0.0036}\simeq 279` 
+  years--which clearly does a poor job of capturing the behavior of diabetes. 
+  This sort of description was only appropriate for diarrhea, as there is a 
+  uniform remission rate across all cases. With diabetes, remission is not 
+  guaranteed for type II, and doesn't occur for type I, so remission rate does 
+  *not* tell us the average duration that an individual will experience diabetes.
```
NathanielBlairStahn: Hmm, this makes me curious exactly what assumptions are necessary for different sorts of calculations. Not something to address here, but perhaps something to delve into a bit on the survival analysis page...

docs/source/models/cause.rst
beatrixkh: Ohh. Ok interesting; thank you for explaining that. I was a bit confused about how that worked but had been glazing over it. Updating.

docs/source/models/cause.rst
beatrixkh: updating

docs/source/models/cause.rst
beatrixkh: For whatever reason when I was trying this the first time, they were all showing up the same regardless of what symbol I was using. Updating now, though.

beatrixkh:
thanks for the close read @NathanielBlairStahn ! I liked your comments/suggestions and have updated accordingly.

docs/source/models/cause.rst
```diff
@@ -558,6 +558,95 @@ below.
...
+calculate YLDs, the latest-stage estimate of remission produced by GBD comes 
+from DisMod models.
+
+Implementing remission rates in cause models
+""""""""""""""""""""""""""""""""""""""""""""
+For a given simulation with timesteps of length *time unit* for some arbitrary 
+time unit, we convert remission rates to the form of cases remitted per 
+person-time-unit. This rate can then be used to compute the probability of a 
+simulant transitioning from an infected or with-condition state to a susceptible 
+or free-of-condition state in a given timestep.
```
NathanielBlairStahn: I think saying "arbitrary time unit" is misleading, for the reason I mentioned earlier - (rate) x (time-step) is only an accurate estimate of the probability if the time-step is small compared to the rate. I think we should make that explicit here.

Another way to think of this is that the time-step of a simulation is not arbitrary, but needs to be chosen small enough relative to all the rates involved so that we get reasonable numbers when calculating probabilities of events occurring within one time-step.

docs/source/models/cause.rst
```diff
@@ -558,6 +558,95 @@ below.
...
+  average duration that an individual will experience diabetes.
+
+  How do we apply this to our simulants? Say we randomly selected 10 people with 
+  diarrhea in the Philippines on a random day in 2017. In the next day, they 
+  would accumulate 10 person days. Our rate tells us that in the next day, the 
+  expected value for cases remitted is given by:
+
+  .. math::
+
+    \frac{\text{1 case of diarrhea}}{\text{5 person-days}}\times\text{10 person-days} = \text{2 cases of diarrhea}.
```
alibow: I would clarify in the equation portion that we are talking about 2 _remitted_ cases of diarrhea

docs/source/models/cause.rst
```diff
@@ -558,6 +558,95 @@ below.
...
+  .. math::
+    \frac{\text{number of remitted cases}}{\text{person-time in the with-condition 
+    population}}
+
+  For example, consider diarrhea cases in the Philippines in 2017. Say that that 
+  year, every such case remitted after an average of 5 days:
+
+  .. math::
+
+    \frac{\text{1 case}}{\text{5 person-days}}=\frac{365/5\text{ cases}}{\text{1 person-year}}=73\text{ cases/person-year}
```
alibow: add clarification for how the days --> years calculation is happening

docs/source/models/cause.rst
```diff
@@ -558,6 +558,95 @@ below.
...
+
+    \frac{\text{1 case of diarrhea}}{\text{5 person-days}}\times\text{10 person-days} = \text{2 cases of diarrhea}.
+
+  Similarly, we can take the rate of remission of diabetes, and for a randomly 
+  selected case of diabetes in Moldova in 2017, consider if they will remit some 
+  time in the next year. The expected value for cases remitted is then given by: 
+
+.. math::
+
+  \frac{\text{0.0036 cases of diabetes}}{\text{1 person-year}}\times\text{1 person-year} = \text{0.0036 cases of diabetes}.
```
alibow: again, clarify _remitted_ cases of diabetes in formula

beatrixkh:
- is the continuous indent appropriate? the examples stretch on for most of the section
- is this clear? would it help to see a visualization (from epiviz) of a remission rate over time for diabetes vs diarrhea?
- what to do about person-time definition? (can i link to the incidence section?)

beatrixkh:
notes:

fix "that that" wording
add hyperlink to 'person time' definition

beatrixkh:
make clear cases are remitted cases
make conversion with (365pdays/1py) clear

docs/source/models/cause.rst
```diff
@@ -558,6 +558,95 @@ below.
...
+a specified population and time period. Just as with incidence, the numerator is 
+given by the count of recovered (or *remitted*) cases, and the denominator is 
+the cumulative person-time [link to person-time definition in incidence 
+section] during which cases are able to go into remission:
+
+  .. math::
+    \frac{\text{number of remitted cases}}{\text{person-time in the with-condition 
+    population}}
+
+  For example, consider diarrhea cases in the Philippines in 2017. Say that that
```
NathanielBlairStahn: Perhaps reword to avoid "that that"?

docs/source/models/cause.rst
```diff
@@ -558,6 +558,95 @@ below.
 Remission Rates
 ^^^^^^^^^^^^^^^
 
+Definition
+""""""""""
+Remission is a measure of cases that recover from a with-condition state, given 
+a specified population and time period. Just as with incidence, the numerator is 
+given by the count of recovered (or *remitted*) cases, and the denominator is 
+the cumulative person-time [link to person-time definition in incidence 
+section] during which cases are able to go into remission:
```
NathanielBlairStahn: @beatrixkh Here are instructions for referencing an arbitrary location in a document: http://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-arbitrary-locations

docs/source/models/cause.rst
beatrixkh: thank you; updating

docs/source/models/cause.rst
beatrixkh: thank you!

docs/source/models/cause.rst
beatrixkh: I'm going to mention this in the "Implementing remission rates in cause models" subsection i have at the bottom @NathanielBlairStahn !

docs/source/models/cause.rst
```diff
@@ -558,6 +560,97 @@ below.
...
+from DisMod models.
+
+Implementing remission rates in cause models
+""""""""""""""""""""""""""""""""""""""""""""
+For a given simulation with timesteps of length *time unit* and a given time 
+unit, we convert remission rates to the form of cases remitted per 
+person-time-unit. If the rate is small with respect to the timestep (that is, if
+the rate is less than 1 per the time step), it can be used to compute the 
+probability of a simulant transitioning from an infected or with-condition state 
+to a susceptible or free-of-condition state in a given timestep.
```
beatrixkh: @NathanielBlairStahn do you think this is sufficiently clear?

NathanielBlairStahn: @beatrixkh Yeah, looks good!

PR #71, opened 2019-12-18
Added a disease summary description.

SantoniS:
I added the summary disease description at the top. Should there be more information around prevalence, incidence, age groups?

### DIFF
--- a/docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
+++ b/docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
@@ -7,16 +7,13 @@ Ischemic Heart Disease
 Disease Description
 -------------------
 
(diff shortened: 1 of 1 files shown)


NathanielBlairStahn:
> I added the summary disease description at the top. Should there be more information around prevalence, incidence, age groups?

Yeah, it seems like more information wouldn't hurt. If you don't want to add it now, put it in the todo.

beatrixkh:
I agree more wouldn't hurt, but so far looks good! You might also look into (/add to the todo) what risks, impairments, and sequelae IHD is linked to within GBD---both to help 'fuel' what to look more into outside of GBD, and to know if GBD is missing some major component.

SantoniS:
I just added more of a description to include symptoms, and information about the ranking of this cause from GBD.

PR #72, opened 2019-12-19
Added ICD codes for cause description of neonatal other

beatrixkh:


SantoniS:
I really liked the way you displayed the ICD codes table for the cause description. I'd like to incorporate the same style into the IHD and Ischemic stroke cause models!

beatrixkh:
Waiting on response from CoD PO before confirming this list is definitive.

NathanielBlairStahn:
I think we should do something to more clearly visually highlight the differences between the fatal and nonfatal inclusions in the last two columns of the table. A couple ideas:

- Leave cells blank if they're the same (which is most of them), and somehow only mark the rows that are different
- Maybe use True/False instead of Incl./Excl.
- Maybe use different colors (can we easily define a role in Sphinx to do this?)

I'm not sure what the best/easiest strategy is.

beatrixkh:
I tried just labelling "excl" for the relevant rows, and leaving everything else blank.

With our current formatting, if i extend the titles to make them more descriptive, they don't wrap, and the table becomes unmanageably wide. We can change that behavior by editing the CSS file, I believe.If we want to color any of the boxes, similarly, we'll need to edit the CSS file.  Do we think this is clear enough as is? Happy to go through and look that up if we decide it's worth it. @NathanielBlairStahn

beatrixkh:
Cool. I'm still waiting on Eve (CoD PO) to confirm that these are the correct ICD codes.

PR #74, opened 2019-12-20
Added restriction table and updated the disease description for LRI

XiaoluQianUW:


docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
```diff
@@ -17,12 +17,29 @@ syncytial virus (*RCV*). LRI can be caused by multiple pathogens and the pathoge
...
+.. todo::
+
+   Describe more about deaths and complications due to LRI.
+   Talk about current vaccination against influenza and pneumonia.
+   https://apps.who.int/iris/bitstream/handle/10665/241904/WER8714_129-144.PDFp
+
 Modeling LRI in GBD 2017
 ------------------------
 
 The GBD 2017 defined the time to recovery for LRI as an average of 10 days(*5-15 days*),
```
NathanielBlairStahn:
```suggestion
The GBD 2017 defined the time to recovery for LRI as an average of 10 days (*5-15 days*),
```
Missing a space. You're also missing a space after a comma in the first paragraph: "(Hib),influenza" should be "(Hib), influenza"

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
```diff
@@ -39,13 +56,21 @@ Cause Model Diagram
...
 Model Assumptions and Limitations
 ---------------------------------
+This model is designed to be used for estimating DALYs due to LRI that are averted from a
+country-level intervention(e.g. food fortification or supplementation given to a percentage of the population)
+that can reduce IRI incidence as a downstream effect.
+
 There is substantial additional effort in GBD to divide LRI
-burden into the aetiologies of LRI, but we have not included
+burden into the aetiologies of LRI, but for the use of this model, we do not included
 aetiologies in this simple model.
```
NathanielBlairStahn: With the new wording, "included" should change to "include" to be grammatical. Also, it now sounds redundant because it says both "for the use of this model" and "in this simple model," so maybe re-word to only mention the model once.

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
```diff
@@ -17,12 +17,29 @@ syncytial virus (*RCV*). LRI can be caused by multiple pathogens and the pathoge
 Pneumococcal pneumonia is the largest cause of LRI
 mortality. [Wikipedia]_, [GBD-2017-YLD-Capstone-Appendix-1]_
 
+Transmission of LRI may occur via several pathways, including direct physical contact,
+fomites, direct droplet spread, and suspended small particles. Intermingling of
+large numbers of people can facilitate transmission of respiratory pathogens. [CDC]_
+
+Malnutrition was identified as a leading risk factor for lower respiratory infection
```
NathanielBlairStahn: Maybe preface this with "In GBD 2016, malnutrition was identified..." to make it clear what research this is referring to.

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
```diff
@@ -39,13 +56,21 @@ Cause Model Diagram
...
+This model is designed to be used for estimating DALYs due to LRI that are averted from a
+country-level intervention(e.g. food fortification or supplementation given to a percentage of the population)
+that can reduce IRI incidence as a downstream effect.
+
 There is substantial additional effort in GBD to divide LRI
-burden into the aetiologies of LRI, but we have not included
+burden into the aetiologies of LRI, but for the use of this model, we do not included
 aetiologies in this simple model.
 
 There are three sequelae associated with LRI. We do not consider Guillain-Barré syndrome
```
NathanielBlairStahn: @XiaoluQianUW I just noticed that here and in the State Data table, it says we are not including sequela s671 (Guillain-Barre due to LRI). I believe that we *do* want to include this sequela (and that is what the Vivarium model currently does, correct, @collijk ?).

The model limitation is that we are not tracking the long-term effects of this sequela (which can include paralysis, for example), but are only counting its YLDs *while the simulant is experiencing a bout of LRI*. The point is that since the prevalence of GBS is so low, there would probably not be much benefit in attempting to capture the long-term YLDs in addition to the short term YLDs.

@beatrixkh Does this match your understanding for diarrheal diseases (which also has GBS as a sequela)?

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
```diff
@@ -39,13 +56,21 @@ Cause Model Diagram
 
 Model Assumptions and Limitations
 ---------------------------------
+This model is designed to be used for estimating DALYs due to LRI that are averted from a
+country-level intervention(e.g. food fortification or supplementation given to a percentage of the population)
+that can reduce IRI incidence as a downstream effect.
```
aflaxman: Typo: "IRI" should be "LRI" right?

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
```diff
@@ -12,17 +12,33 @@ and bronchiolitis, is a major global killer of both children and adults. Symptom
 include shortness of breath, weakness, fever, coughing and fatigue. It is important to check for a fever. Symptoms can last about 7 days and the infection is contagious
 to others shortly before and while experiencing symptoms. It is mainly
 caused by four microorganisms - Streptococcus pneumoniae (*pneumococcal
```
beatrixkh: Here microorganisms = pathogens, yeah? Maybe just call them pathogens to be consistent?

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
```diff
@@ -12,17 +12,33 @@ and bronchiolitis, is a major global killer of both children and adults. Symptom
...
-pneumonia*), Haemophilus influenzae type B (*Hib*),influenza, and respiratory
+pneumonia*), Haemophilus influenzae type B (*Hib*), influenza, and respiratory
 syncytial virus (*RCV*). LRI can be caused by multiple pathogens and the pathogens may co-infect.
```
beatrixkh: Is there a difference between being caused by multiple pathogens, and pathogens co-infecting?

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
```diff
@@ -12,17 +12,33 @@ and bronchiolitis, is a major global killer of both children and adults. Symptom
...
+
+In GBD 2016, malnutrition was identified as a leading risk factor for lower respiratory infection
+mortality among children younger than 5 years and, together with air pollution (both household and ambient)
+and increased antibiotic use, was identified as a focus for targeted intervention measures. [Lancet]_
+
+.. todo::
+
+   Describe more about deaths and complications due to LRI.
+   Talk about current vaccination against influenza and pneumonia.
+   https://apps.who.int/iris/bitstream/handle/10665/241904/WER8714_129-144.PDFp
```
beatrixkh: Might also be nice to specify which organs comprise "lower respiratory tract". (Wiki says:
"
The lower respiratory tract or lower airway is derived from the developing foregut and consists of the trachea, bronchi (primary, secondary and tertiary), bronchioles (including terminal and respiratory), and lungs (including alveoli).[6] It also sometimes includes the larynx.
"
)

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
beatrixkh: For diarrhea, both GBD and we are including GB as one of the sequelae (the other three being mild, moderate, and severe diarrhea). This is because this etiology involves one of the pathogens that can cause diarrhea also causing GB, so if one could avert that diarrhea case, one would also have averted the GB case, if I understand correctly.

GB can be life threatening--but can also be nonfatal, and yes, contribute to significant long-term nonfatal burden, so yes, stopping the model after 5 years does fail to capture those long-term YLDs averted.

I believe the mechanism for LRI--> GB is similar, so yes, it would make sense to include it in this model!

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
```diff
@@ -92,7 +117,7 @@ Data Description
      -
    * - I
      - disability weights
-     - disability_weight_s670 :math:`\times` prevalence_s670+ disability_weight_s669 :math:`\times` prevalence_s669
+     - disability_weight_s670 :math:`\times` prevalence_s670+ disability_weight_s669 :math:`\times` prevalence_s669 + disability_weight_s671 :math:`\times` prevalence_s671
      - GBD assumes 15% of LRI cases as severe and 85% as moderate [GBD-2017-YLD-Capstone-Appendix-1]_.
```
beatrixkh: this currently links to the measles page

beatrixkh: In response to @NathanielBlairStahn 's comment--from the YLD appendix, it sounds like they did a systematic review to determine that LRI cases split into 85/15 moderate/severe. If GB is one of the sequelae you get when you pull GB sequelae, then I believe that that would've been redistributed to GB during como, and after the initial severity split calculations, to result in those three sequelae (moderate LRI, severe LRI, GB).

beatrixkh: We can verify by pulling the prevalence of those sequelae!

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
NathanielBlairStahn: My understanding when I read this was that "can be caused by multiple pathogens" means that any one of the listed pathogens can cause LRI, and that "pathogens may co-infect" means that someone who has LRI may be infected by more than one of the pathogens simultaneously. Does that sound accurate, @XiaoluQianUW ?

Perhaps it could be reworded to be clearer? Any suggestions, @beatrixkh ?

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
NathanielBlairStahn: > this currently links to the measles page

This is a systematic problem we need to figure out how to fix: Sphinx is written so that citations have to be unique across the entire documentation, not just within a page. So in order to get citations to work correctly, each writer will have to be responsible for coming up with unique citation abbreviations for their references, which seems a bit burdensome, though we could try to come up with a standard naming convention to help facilitate this.

Other options would be to see if we can reconfigure Sphinx (@collijk said he would look into this), or to come up with a totally different system for doing citations.  Any suggestions would be welcome.

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
NathanielBlairStahn: Thanks for the clarification, @beatrixkh ! I'm still a bit confused about whether the three sequelae are mutually exclusive, though in your other comment you indicate this has something to do with como...

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
NathanielBlairStahn: > In response to @NathanielBlairStahn 's comment--from the YLD appendix, it sounds like they did a systematic review to determine that LRI cases split into 85/15 moderate/severe. If GB is one of the sequelae you get when you pull GB sequelae, then I believe that that would've been redistributed to GB during como, and after the initial severity split calculations, to result in those three sequelae (moderate LRI, severe LRI, GB).

So, assuming the 3 sequelae are mutually exclusive and their prevalences add up to 100% of LRI, does that mean essentially that people with the GB sequela can be assumed to also have an 85/15 moderate/severe split for the severity of LRI they experience?

Is it possible to pull GB sequelae directly, since it's an impairment not a cause (i.e. as opposed to searching for all the causes that list GB as one of the sequelae)? I guess I'm confused about how impairments work and how COMO works. What does it mean that "that would've been redistributed to GB during como"?

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
XiaoluQianUW: I have talked with Beatrix and I believe the three sequelae are not mutually exclusive. I was able to pull the prevalence for GBS from GBD output. It is a very small number though but still exists.

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
XiaoluQianUW: Thanks for the comment. I will make sure it is consistent.

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
XiaoluQianUW: I might have not use the best wording for this, but that sound accurate. I will try to reword this to make it clearer.

docs/source/models/gbd2017/causes/lower_respiratory_infections/index.rst
XiaoluQianUW: Thanks for the suggestion! I will include it in this section.

PR #75, opened 2019-12-20
Added disease description: hemolytic disease/neonatal jaundice

beatrixkh:


docs/source/models/gbd2017/causes/neonatal_jaundice/index.rst
```diff
@@ -7,24 +7,44 @@ Haemolytic disease and other neonatal jaundice
 Disease Description
 -------------------
 
-Case definition in GBD 2017
-+++++++++++++++++++++++++++
+`Hemolytic disease`_ and `neonatal jaundice`_ are two separate etiologies by which
```
NathanielBlairStahn: Is jaundice an etiology, or is it a symptom (i.e. yellowish skin that is a visible sign of hyperbilirubinemia) for which hemolytic disease is one etiology (and, e.g. G6PD is another) that is a precursor to developing extreme hyperbilirubinemia and kernicterus?

docs/source/models/gbd2017/causes/neonatal_jaundice/index.rst
```diff
@@ -7,24 +7,44 @@ Haemolytic disease and other neonatal jaundice
...
+resulting in brain damage: bilirubin-induced neurologic dysfunction (BIND).
+
+The most common etiologies of hyperbilirubinemia include incompatibilty between
+the mother and infant's blood types, lack of breastfeeding, recieving breastmilk 
+if the mother's antigens are incompatible with the infant (called Breastmilk
+jaundice; this would develop a few days after birth), and poor liver function in
+the infant.
+
+**The etiologies modeled by GBD** include EHB from `Rhesus (Rh) disease`_, preterm
+birth, `glucose-6-phosphate dehydrogenase deficiency`_ (G6PD), and other causes.
```
NathanielBlairStahn: It looks like G6PD is a genetic condition which is not captured in the etiologies you describe above, and Rh disease is a subset of hemolytic disease, if I'm understanding correctly. We should probably also explain how preterm birth leads to jaundice, and we should probably ask the GBD modelers what they mean by "other causes."

docs/source/models/gbd2017/causes/neonatal_jaundice/index.rst
beatrixkh: Oops yes--that is correct. When GBD uses the label "haemolytic disease and other neonatal jaundice", they mean "a collection of etiologies which lead to extreme hyperbilirubinemia", and they define collection to include {Rh disease, preterm birth, G6PD, and other causes}.

The 'other' makes it sound like they mean all etiologies of EHB, or at least all such etiologies that fed into the data they were able to find.

Updating accordingly.

docs/source/models/gbd2017/causes/neonatal_jaundice/index.rst
beatrixkh: I've updated to describe all etiologies modeled by GBD, and i've put in a todo my guess (from reading the appendix) what "other causes" is. I've messaged helena to confirm/deny that guess. @NathanielBlairStahn

beatrixkh:
I've bumped the thread with helena asking about the definition of 'all causes'.

Let's discuss in coworking, regarding cases like this when we're waiting on info from someone, if we'd prefer to merge this PR with the open todo, or to leave the PR open until receiving said info. Alternately, we (I) could refrain from opening PRs until we have all relevant info.

PR #76, opened 2019-12-20
Added document outline and cause model diagram for neonatal sepsis.

kiranklc:


docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
```diff
@@ -13,8 +13,92 @@ Case definition in GBD 2017
...
+
+
+Cause Model Diagram
+-------------------
+
+.. image:: neonatal_sepsis_cause_model.svg
+   :alt: SI cause model diagram for neonatal sepsis and other infections
+
+
+Model Assumptions and Limitations
```
NathanielBlairStahn: In the neural tube defects and NN Enceph, I moved this section to come after the Data Description section, in order to have the data tables be adjacent to the cause model diagram. What do other people think about the section ordering?

docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
```diff
@@ -13,8 +13,92 @@ Case definition in GBD 2017
...
 .. todo::
-   Describe cause model
+   Add more information and references. In particular, find data about differences in global and 
+   gbd definitions, global prevalence and relation to folic acid during pregnancy.
+
+Modeling Measles in GBD 2017
+----------------------------
+
+It is worth noting that the ICD 10 and ICD 9 codes used in non-fatal and fatal GBD models are different. 
+The ICD 10 codes used in non fatal model are P36-P36.9, P38-P39.9. K29.0-K29.00, K29.1-K29.20, P77-P78.1
```
NathanielBlairStahn:
```suggestion
The ICD 10 codes used in non fatal model are P36-P36.9, P38-P39.9. The codes K29.0-K29.00, K29.1-K29.20, P77-P78.1 
```
This was a bit hard to read with all the codes adjacent to each other. Editing this line and the next for clarity.

docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
```diff
@@ -13,8 +13,92 @@ Case definition in GBD 2017
...
-   Describe cause model
+   Add more information and references. In particular, find data about differences in global and 
+   gbd definitions, global prevalence and relation to folic acid during pregnancy.
+
+Modeling Measles in GBD 2017
+----------------------------
+
+It is worth noting that the ICD 10 and ICD 9 codes used in non-fatal and fatal GBD models are different. 
+The ICD 10 codes used in non fatal model are P36-P36.9, P38-P39.9. K29.0-K29.00, K29.1-K29.20, P77-P78.1 
+codes are also used in hospital/claims analyses of non fatal model The ICD 9 codes used are 771, 771.4-771.89.
```
NathanielBlairStahn:
```suggestion
are also used in hospital/claims analyses of non fatal model. The ICD 9 codes used are 771, 771.4-771.89. 
```
Minor edits for readability.

docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
```diff
@@ -13,8 +13,92 @@ Case definition in GBD 2017
...
+Modeling Measles in GBD 2017
+----------------------------
+
+It is worth noting that the ICD 10 and ICD 9 codes used in non-fatal and fatal GBD models are different. 
+The ICD 10 codes used in non fatal model are P36-P36.9, P38-P39.9. K29.0-K29.00, K29.1-K29.20, P77-P78.1 
+codes are also used in hospital/claims analyses of non fatal model The ICD 9 codes used are 771, 771.4-771.89. 
+For hospital/claims analyses, 535.0-535.00, 535.3-535.30, 777.5-777.7 are also used in addition to the above mentioned ICD 9 codes.
+While the fatal model uses only P36-P36.9, P38-P39.9 ICD 10 codes and 771.4-771.9 ICD 9 codes 
+[GBD-2017-YLD-Capstone-Appendix-1-Neonatal-sepsis-and-other-infections]_, [GBD-2017-CoD-Appendix-1-Neonatal-sepsis-and-other-infections]_.
+
```
NathanielBlairStahn: I think I am still confused about which codes were used where...

docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
beatrixkh: I kind of like it before the data description, because i like the idea of going from most high-level to most nitty-gritty as you go through the document. In that schema, in my mind, it would go:
model diagram,
then limitations,
then data description

(as kiran has).

If we wanted the diagram near the tables that reference it (which i also like), what if we did model diagram, then data descriptions, then limitations? to maintain some of the newspaper-style high level ---> detailed ordering?

docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
beatrixkh: Is K29.0-K29.00 a range? or ==K29.0?

docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
beatrixkh: Is 771==771.0? This is a question I also had for neonatal other. I've emailed the CoD team's PO to ask about that notation---though if that way of describing ranges is common to ICD codes and someone on our team is familiar with it, that would be helpful to know!

docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
beatrixkh: Kiran and I had started to talk about this---but we should talk in coworking about whether these are important to list.

docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
kiranklc: Here, I followed the template we used for measles model.

docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
kiranklc: I am not sure. I am assuming its a range. I checked the ICD codes in the range. They seem relevant but couldn't find all of the codes in the range.

docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
kiranklc: Would it look better if I add a table like in YLD appendix but two rows one for fatal and the other for non-fatal?

docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
kiranklc: K29.0 and K29.00 are different. K29.00 is "Acute gastritis without bleeding". K29.0 is "Acute gastritis"

docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
kiranklc: I have added a table for ICD codes. Hope it looks better now.

docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
```diff
@@ -13,8 +13,111 @@ Case definition in GBD 2017
...
-pneumonia, however, is not included -- it is captured in our modelling of
+pneumonia, however, is not included -- it is captured in the modelling of
 pneumonia as a separate entity.
 
 .. todo::
-   Describe cause model
+   Add more information and references. In particular, find data about differences in global and 
+   gbd definitions, global prevalence and relation to folic acid during pregnancy.
+
+Modeling Measles in GBD 2017
```
NathanielBlairStahn: Should be NN Sepsis instead of Measles. Or, if we follow James's suggestion for the cause model template, to avoid copy/past errors like this, we could omit the cause name entirely and just call this section something like _GBD 2017 Modeling Strategy_.

PR #77, opened 2019-12-23
Iron deficiency PAF of 1 cause doc

alibow:
Hi, all. So far this PR contains a general overview of iron deficiency and how it is modeled in GBD 2017. Details of the modeling process and how we plan to operationalize the cause will come in subsequent PRs!

Feel free to provide feedback about the specifics of the iron deficiency description as well as the general outline for risk-attributable cause documentation (and how we might want to approach it differently than non-risk-attributable causes)


docs/source/models/gbd2017/risk_attributable_causes/iron_deficiency/index.rst
```diff
@@ -4,5 +4,152 @@
...
+The sequela associated with the dietary iron deficiency cause in GBD 2017 
+include mild iron deficiency anemia, moderate iron deficiency anemia, and 
+severe iron deficiency anemia. The severity of iron deficiency anemia is 
+determined by the WHO age- and sex- specific hemoglobin concentrations, as 
+described in the table below.
+
+.. list-table:: Hemoglobin Thresholds (g/L)
+   :widths: 15, 15, 15, 15
+   :header-rows: 1
+
```
alibow: I am having trouble getting this table to render. If you spot an issue in my code please let me know!

SantoniS: Is it possible that your 'commas used in :widths: ',' are creating the issue?

SantoniS: In various list table examples online, it shows :widths: as never using commas, but I am not sure if that is mjust a trivial difference and the issue is something else but it's worth a try.

NathanielBlairStahn: The commas should be fine, according to the specification in https://docutils.sourceforge.io/docs/ref/rst/directives.html#list-table

@alibow , the problem is your indentation: The bullets are not lined up with the line above your list (i.e. ":header-rows: 1" - the asterisks should line up with the first colon in this line), so the compiler doesn't recognize the list as being part of the list-table directive.

NathanielBlairStahn: Looking at the whitespace characters in Atom, it looks like you have hard tab characters (\t) instead of spaces for your list, so the spacing might look right depending on your text editor settings, but it won't be recognized as correct `.rst`. You should delete the hard tabs in front of the list, and replace them with the correct number of spaces (3, it appears).

It looks like you have some hard tabs in other places as well (namely your todo boxes), which could potentially cause problems at some point, so it might be good to get rid of those as well.

NathanielBlairStahn: If you're using SublimeText, presumably there should be a setting that automatically inserts soft tabs (some number of spaces) instead of hard tabs when you hit the tab key, and shortcut keys to automatically increase or decrease the indentation level with the correct number of soft tabs.

docs/source/models/gbd2017/risk_attributable_causes/iron_deficiency/index.rst
```diff
@@ -4,5 +4,152 @@
...
+ chronic blood loss, altered red blood cell morphology, inadequate nutritional 
+ intake, and others.
+
+**Iron deficiency anemia** is a type of anemia that is due to insufficient 
+iron levels, which lead to a deficiency of hemoglobin in the blood. Notably, 
+iron deficiency anemia can occur when dietary intake of iron is insufficient, 
+although it may occur in other situations as well, such as when iron is lost 
+through bleeding (ex: menstrual disorders, hookworm disease, etc.). Iron 
+deficiency anemia is the most common cause of anemia globally in most 
+populations.
```
beatrixkh: This sounds a bit redundant---by 'populations' do you mean something along the lines of most age-sex (-race?) groups?

docs/source/models/gbd2017/risk_attributable_causes/iron_deficiency/index.rst
```diff
@@ -4,5 +4,152 @@
...
+fraction (PAF) of 1 cause with the iron deficiency risk factor. This means 
+that 100% of the dietary iron deficiency cases are attributable to the iron 
+deficiency risk factor. Notably, there are additional causes other than 
+dietary iron deficiency that are attributable to the iron deficiency risk 
+factors (ex: maternal disorders).
+
+Additionally, there is an anemia *impairment* modeled in GBD 2017 that 
+represents **all** forms of anemia that are attributable to several causes, 
+including causes such as hemoglobinopathies and hemolytic anemias that are not 
+causally attributable to iron deficiency.
```
beatrixkh: I think this is really clear!

docs/source/models/gbd2017/risk_attributable_causes/iron_deficiency/index.rst
```diff
@@ -4,5 +4,152 @@
...
+YLD-only cause, meaning that it contributes to morbidity, but not mortality.
+
+Modeling Strategy for the Dietary Iron Deficiency Cause
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The dietary iron deficiency cause in GBD 2017 is not modeled directly. Rather, 
+the dietary iron deficiency cause is estimated as the remaining cases of 
+anemia after all of the other anemic cases in the overall anemia envelope were 
+causally attributed to their respective attributable causes (ex: 
+hemoglobinopathies and hemolytic anemias).
```
beatrixkh: Even though these are risk-attributable, I think it is the case that the cause model is fully responsible for the YLDs, and then they are retroactively attributed to the risk exposure. (though it might be slightly more complicated than that). Thus it is easier for me to process/understand, i think, if this section comes before the risk factor section.

I think it would also be helpful to have a high-level understanding of how the parent envelope is modeled, and maybe at what stage they subtract out the other child models?

PR #78, opened 2019-12-23
Added IHD modeling in GBD 2017 section.

SantoniS:


docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
```diff
@@ -11,20 +11,58 @@ Ischemic heart disease (IHD) is a non-communicable cardiovascular disease which
...
-
-Modeling Strategy for Ischemic Heart Disease
-++++++++++++++++++++++++++++++++++++++++++++
+GBD 2017 models IHD using Myocardial infarction (MI) sequelae to estimate the prevalence of IHD, due to the challenges of disease detection and varying symptoms across the population.
+
+
+1. Acute myocardial infarction (MI): Definite and possible MI according to the third universal
+definition of myocardial infarction:
+
+a. When there is clinical evidence of myocardial necrosis in a clinical setting consistent with
```
aflaxman: These nested lists are not rendering well in GitHub.  Maybe there is some tweak to them that will look better.

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
```diff
@@ -11,20 +11,58 @@ Ischemic heart disease (IHD) is a non-communicable cardiovascular disease which
...
+b. Detection of a rise and/or fall of cardiac biomarker values and with at least one of the following: 
+  i. symptoms of ischaemia, 
+  ii. new or presumed new ST-segment-T wave changes or new left bundle branch block, 
+  iii. development of pathological Q waves in the 277 ECG, 
+  iv. imaging evidence of new loss of viable myocardium or new regional wall motion abnormality, or 
+  v. identification of an intracoronary thrombus by angiography or autopsy.
+c. Sudden (abrupt) unexplained cardiac death, involving cardiac arrest or no evidence of a noncoronary cause of death
+d. Prevalent MI is considered to last from the onset of the event to 28 days after the event and is divided into an acute phase (0–2 days) and subacute (3–28 days).
+
+2. Chronic IHD
```
aflaxman: Perhaps there is a way to connect these GBD Modeling section on MI and chronic IHD to the Vivarium model we designed.  I know that they are connected, but it would be good to help a less involved reader see this, too.

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
SantoniS: Thank you for the feedback! I'll find a better method and make an update.

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
SantoniS: Thank you for pointing this out, Abie! That is definitely missing and I'll add a connection between the two and request folks to review if it makes sense.

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
```diff
@@ -11,20 +11,58 @@ Ischemic heart disease (IHD) is a non-communicable cardiovascular disease which
...
+a. Angina; clinically diagnosed stable exertional angina pectoris or definite angina pectoris
+according to the Rose Angina Questionnaire, physician diagnosis, or taking nitrate
+medication for the relief of chest pain.
+b. Asymptomatic ischemic heart disease following myocardial infarction; survival to 28 days
+following incident MI. The GBD study does not use estimations.
```
beatrixkh: Can you clarify the 'The GBD study does not use estimations' comment?

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
```diff
@@ -11,20 +11,58 @@ Ischemic heart disease (IHD) is a non-communicable cardiovascular disease which
...
+
+A systematic review for myocardial infarction was not performed for GBD 2017. Updates to systematic
+reviews are performed on an ongoing schedule across all GBD causes; an update for myocardial infarction
+will be performed in the next one to two iterations.
+A systematic review was done for myocardial infarction for GBD 2015. The search strings used were
+extensive; a full list will be provided on request.
+The dates of the search were 1/1/2009 に 2/3/2015. 38,522 studies were returned; 194 were extracted
```
beatrixkh: haha you ended up with a japanese character here

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
```diff
@@ -11,20 +11,58 @@ Ischemic heart disease (IHD) is a non-communicable cardiovascular disease which
...
+(this number includes extractions that were done for STEMI/NSTEMI models and revascularisation models
+that are not currently part of the MI modelling process but may be in the future).
+A systematic review for myocardial infarction was also done for GBD 2013. The extensive search terms for
+that review will be provided on request.
```
beatrixkh: I actually think we might not need this info---and that you could just say what types of data goes into the GBD model.

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
```diff
@@ -11,20 +11,58 @@ Ischemic heart disease (IHD) is a non-communicable cardiovascular disease which
...
+(this number includes extractions that were done for STEMI/NSTEMI models and revascularisation models
+that are not currently part of the MI modelling process but may be in the future).
+A systematic review for myocardial infarction was also done for GBD 2013. The extensive search terms for
+that review will be provided on request. 
+
+Vital registration, verbal autopsy data were used to model ischaemic heart disease in GBD 2017. The model outliered verbal
+autopsy data in countries and subnational locations where high-quality vital registration data were also
+available. The model also outliered non-representative subnational verbal autopsy data points, ICD8 and ICD9
+BTL data points which were inconsistent with the rest of the data and created implausible time trends,
+and data in a number of Indian states identified by experts as poor-quality.
```
beatrixkh: Is this for the fatal, or nonfatal model?

beatrixkh: I think you could list, for the fatal and nonfatal models each, the type of data they used {epi data from a systematic review, vital registration, and verbal autopsy data, (other?)} without the outliering details! Though perhaps it's worth noting (either here or in the limitations section) that GBD didn't trust their data for india for their (nonfatal?) model

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
NathanielBlairStahn: @SantoniS I think the problem might be with your indentation and missing blank lines. See these references for the correct syntax:

https://docutils.sourceforge.io/docs/user/rst/quickref.html#bullet-lists
https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#bullet-lists

Basically, the two rules you need to follow for lists (including nested lists) are:

1. A blank line is required before the first item in the list and after the last item in the list, *including sublists*. For example, your list numbered i. - v. is not a proper sublist because it is not preceded or followed by a blank line.

2. The items in a sublist must be aligned with the *text* in the parent list. For example, the list labeled a. - d. is not a sublist, because the labels a. b. c. d. are not indented 3 spaces to align with the word "Acute" in the parent list's item 1.

SantoniS: Thank you for the information! I'll make these updates shortly.

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
SantoniS: My text cut off. Fixed it! Thanks for pointing this out.

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
SantoniS: Haha, I removed that! When I pasted some information into this file, it looks like the encoding broke.

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
SantoniS: It makes sense to remove it. I'll rephrase to specify what types of data goes into the model. Thanks!

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
SantoniS: Thanks for the feedback! I've updated this section. Whenever you have a chance, can you please take a look if my update works? Thank you!

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
```diff
@@ -53,6 +81,22 @@ Restrictions
 
    Describe more assumptions and limitations of the model.
 
+Cause Model Diagram
+--------------------
+.. image:: ischemic_heart_disease_transitions.svg
+  :width: 150
```
NathanielBlairStahn: This is pretty small and hard to see. If you just leave out the width, I think the image will automatically scale to a reasonable size.

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
```diff
@@ -11,37 +11,65 @@ Ischemic heart disease (IHD) is a non-communicable cardiovascular disease which
 
 Acute coronary events may cause symptoms such as angina, cold sweats, dizziness, nausuea, neckpain, shortness of breath, sleep disturbances, or weakness.
```
beatrixkh: Above you mention that IHD is a leading cause of YLLs. Is it also a significant cause of YLDs?

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
SantoniS: Thank you for the input! Will do that before merging with master.

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
```diff
@@ -11,37 +11,65 @@ Ischemic heart disease (IHD) is a non-communicable cardiovascular disease which
...
 
-Modeling Strategy for Ischemic Heart Disease
-++++++++++++++++++++++++++++++++++++++++++++
+1. Acute myocardial infarction (MI): Definite and possible MI according to the third universal definition of myocardial infarction:
 
-Cause Model Diagram
---------------------
-.. image:: ischemic_heart_disease_transitions.svg
-  :width: 150
+When there is clinical evidence of myocardial necrosis in a clinical setting consistent with
```
beatrixkh: Not super important, but it might be nice to indent (or otherwise visually distinguish) everything that is a subset of "1.", so that it's clear what falls under that definition

SantoniS: Great point. Just updated that!

PR #79, opened 2019-12-24
ltbi_overview

yongqx:
In this PR, i included three sections: 1) TB definitions, 2) GBD 2017 modeling methodologies, 3) Cause model description.

### DIFF
--- a/docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
+++ b/docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
@@ -4,6 +4,119 @@
 Latent Tuberculosis
 ===================
 
-.. todo:
+TB Overview
+-----------
 
-   all of this
+TB definitions
+++++++++++++++
+Tuberculosis (TB) is a disease caused by the bacteria Mycobacterium
+tuberculosis. It most commonly affects the lungs, although it can affect
+other parts of the body. TB can be fatal if not recognized and treated.
+It also can spread from person to person to infect others. Anyone who 
+inhales the airborne droplets that contains TB bacteria is called a "contact".
+A contact can be someone who living with the person who has infectious TB,
+like a family member, friend, or coworker. [UpToDate]_
+
+What is latent TB?
+++++++++++++++++++
+Latent tuberculosis infection (LTBI) is defined as a state of persistent
+immune response to stimulation by Mycobacterium tuberculosis antigens with
+no evidence of clinically manifest active TB. [WHO]_
+
+What is active TB?
+++++++++++++++++++
+Active tuberculosis, also known as reactivation TB. It occurs when individual's
+immune system becomes weakened and is no longer able to contain the latent
+bacteria; then the TB bacteria become "active", overwhelm the immune system and
+cause a person to become ill. [UpToDate]_ In GBD, active TB consist of 6 child
+causes: Drug-susceptible TB, Multidrug-resistant TB, Extensively drug-resistant
+TB, HIV/AIDS - Drug-susceptible TB, HIV/AIDS - Multidrug-resistant TB, and
+HIV/AIDS - Extensively drug-resistant TB.
+
+TB progression
+++++++++++++++
+Active TB disease developed within 2 years called fast progression, Individuals
+who do not have rapid progression are classified as having slow-progressing latent
+tuberculosis infection. With latent infection, individuals experience no adverse
+health effects and will not transmit Mycobacterium tuberculosis, but they face
+an ongoing risk of developing active tuberculosis through reactivation.
+
+GBD 2017 Modeling Strategy
+++++++++++++++++++++++++++
+GBD 2017 TB model made a refinement in the mortality-to-incidence (MI) ratio
+regression approach upon 2016 model. The SDI was replaced by Healthcare Access
+and Quality index (HAQI) in the MI ratio regression. First, GBD estimated
+risk-weighted prevalence of LTBI by demographic characteristics using data from
+population-based tuberculin surveys and cohort studies reporting the risk of
+developing active TB disease as a function of induration size. Next, they
+divided the inputs on prevalence, incidence, and cause-specific mortality rate
+(CSMR) by the risk-weighted LTBI prevalence in order to model TB among at risk
+populatoin. GBD ran a mixed effects regression (with region random effects)
+using logit transformed MI ratios from locations with a 4-star or above rating
+(countries with poor surveillance systems are classified as less than 4 stars)
+on causes of death with HAQI as a covariate anchoring the lower bound of the
+HAQI scale with a data point reported in the Bangalore study to predict age-sex
+specific MI ratios for all locations and years. Then the age-sex-specific
+incidence were estimated based on those predicted MI ratios. GBD used DisMod-MR
+2.1 (Bayesian meta-regression tool) to generate consistent trends in all
+parameters, and multiplied the outputs by the risk-weighted prevalence of LTBI
+to get population-level estimates of incidence and prevalence. GBD disaggregated
+the all-form TB outputs from DisMod-MR 2.1 into MDR-TB and XDR-TB by HIV status.
+To do so, they estimated the proportions of TB cases with MDR-TB for all
+locations and years, and then estimated the proportions of MDR-TB among
+HIV-negative and HIV-positive individuals based on the risk of MDR-TB associated
+with HIV infection from a meta-analysis. To split MDR-TB into MDR-TB with and
+without extensive drug resistance, GBD pooled the notification and survey data
+on the proportion of MDR-TB cases with extensive drug resistance by
+super-region, and applied these proportions to MDR-TB cases among HIV-negative
+and HIV-positive individuals, respectively. [GBD-2017-YLD-Capstone-Appendix-1]_
+
+Summary model description
++++++++++++++++++++++++++
+It's a complex joint model that includes both tuberculosis and HIV/AIDS. We
+model TB and HIV together because HIV has impacts on both progression rate of
+LTBI to active TB and active TB excess mortality rate. Note that we have
+excluded the active TB drug resistance and GBD predicted the proportion of new
+TB cases with MDR-TB or XDR-TB by ST-GPR. In our model, each simulant occupies
+a state at each discrete time step. It is convenient to distinguish two
+sub-state types: a healthy (susceptible) state and a disease (with condition)
+state. Many simulants began the simulation in a susceptible state with, for
+example, no TB and remain in the susceptible state until they contract LTBI,
+progress to active TB, or die due to TB or other causes. Each state is mutually
+exclusive (e.g. a simulant can be susceptible or be a prevalent case of TB, but
+not both). However, a simulant may have TB and still be susceptible to HIV or
+vice versa. Prevalence is the measure that describes the proportion of individuals
+with disease in the total population during the given simulation time step. The
+prevalence of a disease state determines what proportion of the population will
+occupy that disease state during each time step. Each disease state has transition
+attributes defining the simulant transitions into and out of the state. We
+calculate transition probabilities from rates estimated in the GBD study and
+external data sources as required. In addition, it is worth noting that this
+model for CSU contracted project is not a dynamic transmission model. We will
+not model infection passing between specific individuals in the simulation,
+although the model is individual-based. In other words, individuals in the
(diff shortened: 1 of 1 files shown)


docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,119 @@
...
 
-   all of this
+TB definitions
+++++++++++++++
+Tuberculosis (TB) is a disease caused by the bacteria Mycobacterium
+tuberculosis. It most commonly affects the lungs, although it can affect
+other parts of the body. TB can be fatal if not recognized and treated.
+It also can spread from person to person to infect others. Anyone who 
+inhales the airborne droplets that contains TB bacteria is called a "contact".
+A contact can be someone who living with the person who has infectious TB,
```
beatrixkh: is infectious TB==active TB? I think it's worth using consistent language there

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,119 @@
...
+immune system becomes weakened and is no longer able to contain the latent
+bacteria; then the TB bacteria become "active", overwhelm the immune system and
+cause a person to become ill. [UpToDate]_ In GBD, active TB consist of 6 child
+causes: Drug-susceptible TB, Multidrug-resistant TB, Extensively drug-resistant
+TB, HIV/AIDS - Drug-susceptible TB, HIV/AIDS - Multidrug-resistant TB, and
+HIV/AIDS - Extensively drug-resistant TB.
+
+TB progression
+++++++++++++++
+Active TB disease developed within 2 years called fast progression, Individuals
```
beatrixkh: At risk of being pedantic, i would start just by saying that TB progression is specifically referring to the progression from latent to active TB. (as opposed to TB-negative to latent to active, or regular TB to drug resistant TB), before differentiating between rapid vs slow-progressing!

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,119 @@
...
+who do not have rapid progression are classified as having slow-progressing latent
+tuberculosis infection. With latent infection, individuals experience no adverse
+health effects and will not transmit Mycobacterium tuberculosis, but they face
+an ongoing risk of developing active tuberculosis through reactivation.
+
+GBD 2017 Modeling Strategy
+++++++++++++++++++++++++++
+GBD 2017 TB model made a refinement in the mortality-to-incidence (MI) ratio
+regression approach upon 2016 model. The SDI was replaced by Healthcare Access
+and Quality index (HAQI) in the MI ratio regression. First, GBD estimated
```
beatrixkh: I think because your audience isn't necessarily familiar with GBD's 2016 model, you could either omit this, or perhaps move it later, rather than opening with it

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,119 @@
...
+risk-weighted prevalence of LTBI by demographic characteristics using data from
+population-based tuberculin surveys and cohort studies reporting the risk of
+developing active TB disease as a function of induration size. Next, they
```
beatrixkh: prior to this point, i think it would be helpful to explain TSTs and what induration size is!

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,119 @@
...
+tuberculosis infection. With latent infection, individuals experience no adverse
+health effects and will not transmit Mycobacterium tuberculosis, but they face
+an ongoing risk of developing active tuberculosis through reactivation.
+
+GBD 2017 Modeling Strategy
+++++++++++++++++++++++++++
+GBD 2017 TB model made a refinement in the mortality-to-incidence (MI) ratio
+regression approach upon 2016 model. The SDI was replaced by Healthcare Access
+and Quality index (HAQI) in the MI ratio regression. First, GBD estimated
+risk-weighted prevalence of LTBI by demographic characteristics using data from
```
beatrixkh: what does risk-weighted mean here? are the risks a set collection of risks defined by GBD, or is a case weighted based on its corresponding induration size?

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,119 @@
...
+
+What is active TB?
+++++++++++++++++++
+Active tuberculosis, also known as reactivation TB. It occurs when individual's
+immune system becomes weakened and is no longer able to contain the latent
+bacteria; then the TB bacteria become "active", overwhelm the immune system and
+cause a person to become ill. [UpToDate]_ In GBD, active TB consist of 6 child
+causes: Drug-susceptible TB, Multidrug-resistant TB, Extensively drug-resistant
+TB, HIV/AIDS - Drug-susceptible TB, HIV/AIDS - Multidrug-resistant TB, and
+HIV/AIDS - Extensively drug-resistant TB.
```
beatrixkh: I think it would be helpful to add the abbreviations here, so it's clear what you mean when you use them later. e.g., "Multidrug-resistant TB (MDR-TB)"

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,119 @@
...
+HIV-negative and HIV-positive individuals based on the risk of MDR-TB associated
+with HIV infection from a meta-analysis. To split MDR-TB into MDR-TB with and
+without extensive drug resistance, GBD pooled the notification and survey data
+on the proportion of MDR-TB cases with extensive drug resistance by
+super-region, and applied these proportions to MDR-TB cases among HIV-negative
+and HIV-positive individuals, respectively. [GBD-2017-YLD-Capstone-Appendix-1]_
+
+Summary model description
++++++++++++++++++++++++++
+It's a complex joint model that includes both tuberculosis and HIV/AIDS. We
```
beatrixkh: Maybe swap "It's" for "Our model is", to be clear you're no longer referring to the GBD model?

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,119 @@
...
+model TB and HIV together because HIV has impacts on both progression rate of
+LTBI to active TB and active TB excess mortality rate. Note that we have
+excluded the active TB drug resistance and GBD predicted the proportion of new
+TB cases with MDR-TB or XDR-TB by ST-GPR. In our model, each simulant occupies
```
beatrixkh: Can you clarify if you excluded MDR-TB and XDR-TB cases entirely, or if you've lumped them in with your TB-prevalent cases?

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,119 @@
...
+sub-state types: a healthy (susceptible) state and a disease (with condition)
+state. Many simulants began the simulation in a susceptible state with, for
+example, no TB and remain in the susceptible state until they contract LTBI,
+progress to active TB, or die due to TB or other causes. Each state is mutually
+exclusive (e.g. a simulant can be susceptible or be a prevalent case of TB, but
+not both). However, a simulant may have TB and still be susceptible to HIV or
+vice versa. Prevalence is the measure that describes the proportion of individuals
+with disease in the total population during the given simulation time step. The
+prevalence of a disease state determines what proportion of the population will
+occupy that disease state during each time step. Each disease state has transition
```
beatrixkh: These two sentences seem redundant. They also could be misinterpreted to imply that you re-input prevalence at every timestep, rather than initializing a prevalence of HIV, TB-HIV, and TB, and then letting the transition rates determine prevalence at each timestep thereafter.

I would consider just dropping this, because the cause section has a detailed description of prevalence.

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,119 @@
...
+attributes defining the simulant transitions into and out of the state. We
```
beatrixkh: This part also seems like a general attribute of vivarium models, and perhaps doesn't need to be specified here (though if there are nuances to the transition rates in this model, that makes sense to me to include/clarfiy!)

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,119 @@
...
+calculate transition probabilities from rates estimated in the GBD study and
+external data sources as required. In addition, it is worth noting that this
```
beatrixkh: By this are you referring to your estimation of ltbi incidence? that process seems worth mentioning here!

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,119 @@
...
+model for CSU contracted project is not a dynamic transmission model. We will
+not model infection passing between specific individuals in the simulation,
+although the model is individual-based. In other words, individuals in the
+simulation will, for example, move from uninfected to infected states, but the
+model will not specify that individual X infected individual Y. As described
+above, the model will be calibrated in aggregated level, the outputs from
+simulation baseline scenario should match the population level rates estimated
+in the GBD.
```
beatrixkh: This section could potentially end up in your 'assumptions/restrictions' section---maybe we could discuss in coworking and standardize, as i think we're now using both formats on different cause model pages.

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
yongqx: Good catch! It’s important to explain the relationship between TST and induration size because the distribution is dependent on sub-population selection.

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
yongqx: Yes, external sources are needed to calculate transition data like LTBI incidence or active tb remission disaggregated by HIV status.

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
yongqx: When I say exclude MDR-TB and XDR-TB, it actually means we only split active TB by HIV status, not drug resistance.

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
alibow: It might be good to specifically state that you're talking about the cause model here too!

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
aflaxman: I think the infectious disease docs make a distinction between active TB that "pulmonary TB" (affecting the lungs) and therefore is infectious, and "extrapulmonary TB", which is active but unlikely to infect others.  However, we do not use this level of detail in our sim.  I support changing this phrase to "active TB" or striking infectious and writing just "TB"

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,119 @@
...
+immune system becomes weakened and is no longer able to contain the latent
+bacteria; then the TB bacteria become "active", overwhelm the immune system and
+cause a person to become ill. [UpToDate]_ In GBD, active TB consists of 6 child
+causes: Drug-susceptible TB, Multidrug-resistant TB, Extensively drug-resistant
+TB, HIV/AIDS - Drug-susceptible TB, HIV/AIDS - Multidrug-resistant TB, and
+HIV/AIDS - Extensively drug-resistant TB.
+
+TB progression
+++++++++++++++
+Active TB disease developed within 2 years is called fast progression. Individuals
```
aflaxman: "fast progression" on this line is the same as "rapid progression" on next, right?  If so, let's call it "rapid progression" everywhere.

yongqx:
Thanks for all your comments, I should use consistent phrase and it’s much clear for all level audience.

PR #80, opened 2019-12-24
ltbi cause hierarchy diagram and table

yongqx:


beatrixkh:
Though if you leave the table in, perhaps consider adding a reference to the sequelae into the table!

yongqx:
Yes, I am thinking about add a full list of sequelae_id for each cause.

Get Outlook for iOS<https://aka.ms/o0ukef>

PR #81, opened 2019-12-24
ltbi cause model diagram and relevant data sources

yongqx:


docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,274 @@
...
+   * - csmr
+     - cause-specific mortality rate
+   * - emr
+     - excess mortality rate
+   * - dw
+     - disability weight
+   * - {measure}_{gbd_id_type}{gbd_id}
+     - e.g. prev_c954 = prevalence of LTBI
+
+.. list-table:: **State Measures**
```
beatrixkh: oh boy 😂 ; it looks like your models were a test case we should've been using when selecting formatting. I think your equations are too wide, and now this table is compiling off the side of the page. I think you could either change your notation system, or we should reformat the CSS @kiranklc 

The other thing i would try is just defining less-wide columns than you currently have, and see to what degree it helps. It seems like (someone correct me if they know better) even when i've defined them quite small, they've expanded to accommodate the text I write.

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,274 @@
...
 
-.. todo:
+Cause Model Diagram
+-------------------
+.. image:: joint_model_diagram.svg
 
-   all of this
+Data Description
+----------------
+.. list-table:: **Abbreviations**
```
beatrixkh: We came up with a standardized format for these tables (see measles, diarrheal diseases, neonatal enceph for examples). So we have a 'state' 'state name' and also 'definition' column for this particular table.

I think it would also make sense to have a row for every bubble in your diagram--so one for S_TB, I_HIV and another for S_TB, S_HIV, etc.

yongqx:
The easier way to see a neat table is to make html, we can discuss the format in coworking next time!

Get Outlook for iOS<https://aka.ms/o0ukef>

PR #82, opened 2019-12-24
ltbi cause model V&V criteria

yongqx:
To-do: make the V&V criteria into a pipeline!

### DIFF
--- a/docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
+++ b/docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
@@ -4,6 +4,111 @@
 Latent Tuberculosis
 ===================
 
-.. todo:
+Validation and Verification Criteria
+------------------------------------
 
-   all of this
+Obejective
+++++++++++
+External
+ - Model results should be checked against local TB epidemiology
+    - e.g. rate of decline in burden of disease should be compared
+      with historical evidence.
+ - Compare our results (e.g. ICERs) to similar models or empirical
+   assessments where possible
+Internal
+ - Test hypotheses without develop a full simulation. 
+ - Calibrate simulation baseline against GBD 2017 results
+    - Sim outputs mean should perfectly match GBD results.
+    - Sim outputs uncertainty should be wider than GBD results,
+      because both stochastic and parametric uncertainty are included.
+
+V&V strategy
+++++++++++++
+Model Validation
+ - Check the logical structure and input data for concept model,
+   make sure
+    - the theories and assumptions underlying the conceptual model are correct.
+    - the data to build, evaluate, and test model are correct.
+Model Verification
+ - Check the translation from concept model to script based on test data,
+   make sure
+    - the computer programming and implementation of the concept model is correct.
+    - the output of the model can be calibrated against GBD results.
+
+logic
++++++
+- Parent cause is the sum of child causes
+    - Fatal: Deaths (CSMR, Excess MR), YLLs
+    - Non-fatal: Cases, YLDs, Prevalence, Incidence
+    - DALYs = YLLs + YLDs
+- By location-/age-/sex-
+- Start from count space
+- We expect that total active TB cases less than LTBI cases for sim output
+  of event count
+
+How GBD post-processing results
++++++++++++++++++++++++++++++++
+GBD starts from All-form TB results
+    1. Find proportion of HIV+ cases among all TB cases
+    2. Disaggregate into HIV+ TB and HIV- TB
+    3. Find proportion of drug-resistant cases among HIV+ TB cases
+       and HIV- TB cases
+    4. Disaggregate into:
+        - drug-susceptible TB, multidrug-resistant TB, and extensively
+          drug-resistant TB
+        - HIV+ drug-susceptible TB, HIV+ multidrug-resistant TB,
+          and HIV+ extensively drug-resistant TB
+
+Formula
++++++++
+For certain location-/age-/sex-
+    - Deaths due to all causes equal to sum of:
+        - Deaths due to all-form TB (aggregate all child active TB causes)
+        - Deaths due to HIV resulting in other diseases
+        - Deaths due to other causes
+Apply the formula to other measures (e.g. DALYs)
+
+Steps of model verification
++++++++++++++++++++++++++++
+1. Set hypothesis
+    - The sum of the prevalences of all model states should equal
+      to the GBD TB prevalence plus HIV prevalence. (Pre_297 + Prev_298
+      = Sum(Prev_s))
+    - The sum of the cause-specific mortality of all model states
+      should equal to the GBD TB CSMR plus HIV CSMR. (CSMR_297 + 
+      CSMR_298 = Sum(Prev_s * ExcessMR_s))
+    - The prevalence weighted sum of the disability weight of all model states
+      should equal to the GBD TB YLDs plus HIV YLDs. (ylds_297 + ylds_298 
+      = Sum(Prev_s * dw_s))
+2. Check for proposed hypothesis (e.g. prevalence for the whole model)
+    - **Data:** Once the model input data is produced and put in the artifact,
+      produce a graph of the sum of the input data prevalences and compare
+      it to the GBD data not in the model.
+    - **Sim initialization:** Initialize a simulation using the model input data
+      and count the disease event to make sure it matches with GBD data 
+      not in the model.
+    - **Historical calibration:** Run a simulation from 2012 to 2017 and count
+      the disease event at the end of the sim to make sure it matches with
+      GBD data not in the model.
+    - **Baseline verification:** Run a simulation from 2020 to 2025 and count
+      the disease event at the end of the sim to make sure the baseline
+      model outcomes match with GBD 2017 results.
+
+Measure types in verification
++++++++++++++++++++++++++++++
+Constrained verification
+ - Compare (disease preson time / total person time) to prevalence in GBD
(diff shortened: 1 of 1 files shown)


docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,111 @@
 Latent Tuberculosis
 ===================
 
-.. todo:
+Validation and Verification Criteria
+------------------------------------
 
-   all of this
+Obejective
```
beatrixkh: Objective (sp)

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,111 @@
...
+++++++++++
+External
+ - Model results should be checked against local TB epidemiology
+    - e.g. rate of decline in burden of disease should be compared
+      with historical evidence.
+ - Compare our results (e.g. ICERs) to similar models or empirical
+   assessments where possible
+Internal
+ - Test hypotheses without develop a full simulation.
```
beatrixkh: by this, do you mean a paired-down sim? or a back-of-the-envelope calculation?

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,111 @@
...
+Model Validation
+ - Check the logical structure and input data for concept model,
+   make sure
+    - the theories and assumptions underlying the conceptual model are correct.
+    - the data to build, evaluate, and test model are correct.
+Model Verification
+ - Check the translation from concept model to script based on test data,
+   make sure
+    - the computer programming and implementation of the concept model is correct.
+    - the output of the model can be calibrated against GBD results.
```
beatrixkh: By this do you mean that the outputs match GBD, or something more complicated? (such as for the intervention sim?). If the former, consider simplifying the language. if the latter (like, if you expect the intervention sim to have lower burden than the baseline?), consider elaborating? (Though obviously I'd welcome pushback on that)

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,111 @@
...
+logic
++++++
+- Parent cause is the sum of child causes
+    - Fatal: Deaths (CSMR, Excess MR), YLLs
+    - Non-fatal: Cases, YLDs, Prevalence, Incidence
+    - DALYs = YLLs + YLDs
+- By location-/age-/sex-
+- Start from count space
+- We expect that total active TB cases less than LTBI cases for sim output
+  of event count
```
beatrixkh: I think clarify this---were these also meant to be sub-bullets of the above?

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,111 @@
...
+
+How GBD post-processing results
```
beatrixkh: I'm a little unclear on the purpose of this section; is this explaining some sort of discrepancy you guys had in your baseline sim vs. gbd? or just to provide context for what you're validating against?

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,111 @@
...
+    2. Disaggregate into HIV+ TB and HIV- TB
+    3. Find proportion of drug-resistant cases among HIV+ TB cases
+       and HIV- TB cases
+    4. Disaggregate into:
+        - drug-susceptible TB, multidrug-resistant TB, and extensively
+          drug-resistant TB
+        - HIV+ drug-susceptible TB, HIV+ multidrug-resistant TB,
+          and HIV+ extensively drug-resistant TB
+
+Formula
```
beatrixkh: This seems redundant with your 'logic' section

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,111 @@
...
+      it to the GBD data not in the model.
+    - **Sim initialization:** Initialize a simulation using the model input data
+      and count the disease event to make sure it matches with GBD data 
+      not in the model.
+    - **Historical calibration:** Run a simulation from 2012 to 2017 and count
+      the disease event at the end of the sim to make sure it matches with
+      GBD data not in the model.
+    - **Baseline verification:** Run a simulation from 2020 to 2025 and count
+      the disease event at the end of the sim to make sure the baseline
+      model outcomes match with GBD 2017 results.
```
beatrixkh: Thank you for giving us an example 😂 . this is helpful to see.

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,111 @@
...
+
+Measure types in verification
++++++++++++++++++++++++++++++
+Constrained verification
+ - Compare (disease preson time / total person time) to prevalence in GBD
```
beatrixkh: person (sp)

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,111 @@
...
+ - Compare (disease counts / person time) to incidence in GBD
+Unconstrained verification
+ - CSMR
```
beatrixkh: Maybe this should be obvious to me--but why are these unconstrained? Does that mean unconstrained in your model?

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
yongqx: The bullet points in external part are things we can check based on non-GBD evidences. Back envelope calculation is the step before post-processing raw sim outputs that tells whether your results fit into a reasonable magnitude.

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
yongqx: These are some tips when we validating against GBD results, I’ll rewrite this section into a readable paragraph!

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,111 @@
...
 
-.. todo:
+Validation and Verification Criteria
+------------------------------------
 
-   all of this
+Obejective
+++++++++++
+External
+ - Model results should be checked against local TB epidemiology
```
alibow: I would take out "TB" here so that it can be applied as a guide for all cause models

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,111 @@
...
+    - Sim outputs uncertainty should be wider than GBD results,
+      because both stochastic and parametric uncertainty are included.
+
+V&V strategy
+++++++++++++
+Model Validation
+ - Check the logical structure and input data for concept model,
+   make sure
+    - the theories and assumptions underlying the conceptual model are correct.
+    - the data to build, evaluate, and test model are correct.
```
alibow: Also that the model is sufficiently complex to capture the necessary dynamics to produce a valid result

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,111 @@
...
+
+logic
++++++
+- Parent cause is the sum of child causes
+    - Fatal: Deaths (CSMR, Excess MR), YLLs
+    - Non-fatal: Cases, YLDs, Prevalence, Incidence
+    - DALYs = YLLs + YLDs
+- By location-/age-/sex-
+- Start from count space
+- We expect that total active TB cases less than LTBI cases for sim output
```
alibow: Again, clarify that TB-specific stuff is an example or use general language

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
```diff
@@ -4,6 +4,111 @@
...
+Formula
++++++++
+For certain location-/age-/sex-
+    - Deaths due to all causes equal to sum of:
+        - Deaths due to all-form TB (aggregate all child active TB causes)
+        - Deaths due to HIV resulting in other diseases
+        - Deaths due to other causes
+Apply the formula to other measures (e.g. DALYs)
+
+Steps of model verification
```
alibow: I think it would be useful to clarify what is done automatically in vivarium post-processing by the software engineers and what should be done by the research folks. I'm not sure what the answer to this is, but I think it would be good to include here.

docs/source/models/gbd2017/causes/latent_tuberculosis/index.rst
alibow: Oh, sorry - I was wrong about this!! I though that this was a PR for the template, not for LTBI. This looks good in that case :)

beatrixkh:
From documentation session:

We would like to clarify what the purpose of this section is within each cause model writeup. There will be similar V&V across all causes, so do we want to separate that out and write a generic V&V page? Do we want to have explicit calculations in this section? And at what point in the modeling process are we theoretically writing this? Is it before anything has been run? This will change the purpose of this section and how it needs to be laid out.

Would like further discussion with @collijk

PR #86, opened 2019-12-27
neonatal jaundice: added restrictions table

beatrixkh:


docs/source/models/gbd2017/causes/neonatal_jaundice/index.rst
```diff
@@ -156,4 +156,35 @@ Data Descriptions
...
+	  - age_group_id = 2; [0-0.01917808 years)
+	* - YLL age group end
+	  - Post neonatal
+	  - age_group_id = 4; [0.07671233-1 years)
+	* - YLD age group start
+	  - Early neonatal
+	  - age_group_id = 2; [0-0.01917808 years)
+	* - YLD age group end
+	  - 90-94
+	  - age_group_id = 32; [90-94 years)
```
NathanielBlairStahn: GBD mapping says this should be age group 235

beatrixkh: I pulled this by checking get_cause_metadata(cause_set_id = 3, gbd_round_id = 5). That table lists yld_age_end as 95 for neonatal jaundice. cause_set_id 3 is "GBD reporting"; I wonder if that's how I'm getting a different number? I need to ask for help with getting gbd_mapping set up so I can cross reference!

beatrixkh: I've clarified my confusion with gbd_mapping, and have also asked central comp about the notation yld_age_end=95 that i'm getting from get_cause_metadata

beatrixkh: image
It looks like I ought to switch to gbd_mapping

PR #87, opened 2019-12-27
diarrheal diseases: added restrictions table

beatrixkh:


docs/source/models/gbd2017/causes/neonatal_jaundice/index.rst
```diff
@@ -156,4 +156,35 @@ Data Descriptions
 	  - YLD appendix
 	  - Disability weight of sequela with id {id}
 	  -
-	
\ No newline at end of file
+
+.. list-table:: Restrictions
```
collijk: Shouldn't be in this PR.  If you're working on totally separate documents, remember to make all branches from master.

docs/source/models/gbd2017/causes/diarrheal_diseases/index.rst
```diff
@@ -222,6 +222,37 @@ Data Description
...
+	  - age_group_id = 2; [0-0.01917808 years)
+	* - YLL age group end
+	  - 90-94
+	  - age_group_id = 32; [90-94 years)
+	* - YLD age group start
+	  - Early neonatal
+	  - age_group_id = 2; [0-0.01917808 years)
+	* - YLD age group end
+	  - 90-94
+	  - age_group_id = 32; [90-94 years)
```
NathanielBlairStahn: GBD mapping says the age group id should be 235, and so should the one for yll age group end.

docs/source/models/gbd2017/causes/neonatal_jaundice/index.rst
beatrixkh: oops; will make sure to do so in the future. thanks!

docs/source/models/gbd2017/causes/diarrheal_diseases/index.rst
beatrixkh: thanks for the catch!

PR #88, opened 2019-12-27
Neonatal other: added restrictions section

beatrixkh:


docs/source/models/gbd2017/causes/neonatal_other/index.rst
```diff
@@ -36,3 +36,38 @@ neonatal disorders.
...
+	  - age_group_id = 2; [0-0.01917808 years)
+	* - YLL age group end
+	  - Post neonatal
+	  - age_group_id = 4; [0.07671233-1 years)
+	* - YLD age group start
+	  - Early neonatal
+	  - age_group_id = 2; [0-0.01917808 years)
+	* - YLD age group end
+	  - 90-94
+	  - age_group_id = 32; [90-94 years)
```
NathanielBlairStahn: Should be [90-95 years)

docs/source/models/gbd2017/causes/neonatal_other/index.rst
```diff
@@ -36,3 +36,38 @@ neonatal disorders.
...
+	  -
+	* - YLL only
+	  - False
+	  -
+	* - YLD only
+	  - False
+	  -
+	* - YLL age group start
+	  - Early neonatal
+	  - age_group_id = 2; [0-0.01917808 years)
```
NathanielBlairStahn: If we're going to put in age intervals, it might be nicer to use days as units for the early age groups to make it more readable. E.g. [0, 7 days) for early neonatal, and [28 days, 1 year) for post neonatal.

docs/source/models/gbd2017/causes/neonatal_other/index.rst
```diff
@@ -36,3 +36,38 @@ neonatal disorders.
 
 .. todo::
    Describe cause model
+
+Restrictions
+------------
```
NathanielBlairStahn: We have yet to finalize a document structure, but I've been putting Restrictions as a subsection within Model Assumptions and Limitations, which I've been putting after the Data Description section. I'm open to discussion though.

Since you don't have headings for the other sections yet, it doesn't matter too much at this stage, but eventually we should standardize. I'll create a pull request for the cause model template soon so we can work out details there.

docs/source/models/gbd2017/causes/neonatal_other/index.rst
XiaoluQianUW: I thought it is 95 plus.

NathanielBlairStahn: Good catch, @XiaoluQianUW , looks like it should be age group id 235, not 32

docs/source/models/gbd2017/causes/neonatal_other/index.rst
collijk: Also, it shouldn't be intervals.  There is an implicit bit in the restrictions that the `age_group_start` is the left hand side of the bin represented by the id and the age_group_end is the right hand side of it's respective bin.

docs/source/models/gbd2017/causes/neonatal_other/index.rst
collijk: 90-94 is right.  It's a demographer's notation thing. It doesn't really mix with the math open/closed interval notation.

NathanielBlairStahn: @collijk There are two issues here:
1. GBD mapping says the age_group_id for YLD age group end for other neonatal disorders should be 235, not 32.
2. If the correct age group id _were_ 32, the table already includes the correct age group name, "90-94" in the "Value" column. The interval appears in the "Notes" column, presumably as a clarification to the human reader that this age group actually includes all ages up to but not including 95.

docs/source/models/gbd2017/causes/neonatal_other/index.rst
NathanielBlairStahn: I think @beatrixkh included the intervals so that the unfamiliar reader will know what the "Early neonatal" age group means. These are the intervals that Vivarium inputs returns - what's wrong with including them in the Notes column, @collijk ?

beatrixkh: Yes; @NathanielBlairStahn is correct, I included the intervals in the notes column as a definition of those age_groups. I can keep or delete---we can also discuss in coworking @collijk

docs/source/models/gbd2017/causes/neonatal_other/index.rst
beatrixkh: Thanks for creating the additional page! I will update this page with those if we decide on a format before closing this PR.

beatrixkh:
From documentation session:

We would like to clarify what the purpose of this section is within each cause model writeup. There will be similar V&V across all causes, so do we want to separate that out and write a generic V&V page? Do we want to have explicit calculations in this section? And at what point in the modeling process are we theoretically writing this? Is it before anything has been run? This will change the purpose of this section and how it needs to be laid out.

Would like further discussion with @collijk

NathanielBlairStahn:
> From documentation session:
> 
> We would like to clarify what the purpose of this section is within each cause model writeup. There will be similar V&V across all causes, so do we want to separate that out and write a generic V&V page? Do we want to have explicit calculations in this section? And at what point in the modeling process are we theoretically writing this? Is it before anything has been run? This will change the purpose of this section and how it needs to be laid out.
> 
> Would like further discussion with @collijk

@beatrixkh Did you mean to put this comment in PR https://github.com/ihmeuw/vivarium_research/pull/82 ?

beatrixkh:
@NathanielBlairStahn oops; yes. I just commented it to 82. thank you

PR #89, opened 2019-12-28
Add data tables to neonatal sepsis model

kiranklc:


docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
```diff
@@ -18,3 +18,137 @@ pneumonia as a separate entity.
...
+     - disabilty weights
+     - 0
+     -
+   * - I
+     - prevalence
+     - prevalence_c383
+     - 
+   * - I
+     - birth prevalence
+     - 1-birth_prevalence_c383
```
aflaxman: Typo?  I was expecting `birth_prevalence_c383` here, not one minus it

docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
```diff
@@ -18,3 +18,137 @@ pneumonia as a separate entity.
...
+   * - Transition
+     - Source 
+     - Sink 
+     - Value
+     - Notes
+   * - i
+     - S
+     - I
+     - :math:`\frac{\text{incidence_rate_m1594}}{\text{1 - prevalence_c383}}`
+     - Incidence rate is estimated in dismod but does not exists in como.
```
aflaxman: If incidence rate comes from dismod, I think we don't need to divide through by prevalence.

docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
kiranklc: Sorry, its a typo.. I copied from the previous row and forgot to change it. I will modify it.

docs/source/models/gbd2017/causes/neonatal_sepsis/index.rst
kiranklc: I will change it.

PR #91, opened 2019-12-28
Cause Model Template: outline and getting started instructions

NathanielBlairStahn:
Please review the section titles and the order of the sections, and post here if you want to suggest changes. I've put a few specific questions/concerns in special "Todo for template development" boxes within the .rst document itself.

PR #93, opened 2020-01-02
Cause Model Template: Change section names and structure; add GBD Restrictions table

NathanielBlairStahn:
Here's a new version that incorporates some suggestions from the previous PR https://github.com/ihmeuw/vivarium_research/pull/91 :

1. Remove "Disease Description" section, and just put the description at the top level of the document.
2. Remove the cause name from the GBD Modeling section title.
3. Move the restrictions table into the GBD Modeling section.
4. Remove the section "Model Assumptions and Limitations," and replace it with a section called "Vivarium Modeling Strategy," with two subsections, "Scope," and "Assumptions and Limitations." This new section comes right before the "Cause Model Diagram" section.

I have copied relevant comments from several pull requests into the template todo boxes, so we can see all the discussion relevant to section structure in one place.

Post comments in this PR if you have opinions about the evolution of the section structure.

### DIFF
--- a/docs/source/models/gbd2017/causes/cause_model_template.rst
+++ b/docs/source/models/gbd2017/causes/cause_model_template.rst
@@ -4,6 +4,8 @@
 Cause Model Template
 ====================
 
+.. _Pull Request 64: https://github.com/ihmeuw/vivarium_research/pull/64
+.. _Pull Request 76: https://github.com/ihmeuw/vivarium_research/pull/76
 .. _Pull Request 91: https://github.com/ihmeuw/vivarium_research/pull/91
 
 .. important::
@@ -56,9 +58,6 @@ Cause Model Template
       `directive <https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#directives>`_
       from :file:`index.rst`.
 
-Disease Description
--------------------
-
 .. todo::
 
    Add a general clinical overview of the cause.
@@ -72,14 +71,6 @@ Disease Description
    template to instruct the cause modeler how to fill out the cause model
    document.
 
-   - Confirm the name of this section. Will "Disease Description" be inaccurate
-     or confusing if we are modeling a cause that is not technically a disease?
-     I was aiming for something that matched everyday usage better than
-     "Cause Description."
-
-     - In `Pull Request 91`_, @James suggested dropping the subsection
-       altogether and putting the description at the top level.
-
    - In the above todo box, add more details about what we're looking for in
      cause descriptions, such as:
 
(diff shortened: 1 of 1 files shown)


NathanielBlairStahn:
I just remembered that one reason I had put in the "Disease Description" section in the first place was to allow subsections within the disease description, such as @yongqx does for TB ( see https://github.com/ihmeuw/vivarium_research/pull/79 ), without forcing these subsections to be at the same level as the other top-level sections in the document.

Do people think it is important to have the flexibility to add subsections within the disease description?

docs/source/models/gbd2017/causes/cause_model_template.rst
```diff
@@ -91,125 +82,193 @@ Disease Description
...
+    something like "State and Transition Data Tables" be more descriptive?
 
-        I think `Data Description` is fine.
+    In `Pull Request 91`_, @James said:
+
+      I think `Data Description` is fine.
 
-    - Is there anything else that belongs in a "Data Description" section
-      besides the data tables?
+  - Should this section have subsections (such as "State and Transition Data
```
alibow: I like state and transition data tables, but either is fine in my opinion

docs/source/models/gbd2017/causes/cause_model_template.rst
```diff
@@ -91,125 +82,193 @@ Disease Description
...
+    In `Pull Request 91`_, @Yongquan said:
 
-        one data-relevant section needed for more complicated models is
-        non-standard data sources where external calculation occurs.
+      one data-relevant section needed for more complicated models is
+      non-standard data sources where external calculation occurs.
 
-    - Add instructions in other sections, including:
+    Does "non-standard data sources" warrant a subsection, even if we don't need
+    it for all cause models?
```
alibow: I think we can handle non-standard data sources on an ad hoc basis and probably don't need a subsection for it given my understanding that it is not necessary for most causes? If I'm wrong in this assumption and it is in fact necessary for most causes, then I think we can include a subsection.

docs/source/models/gbd2017/causes/cause_model_template.rst
```diff
@@ -91,125 +82,193 @@ Disease Description
...
 
 .. admonition:: Todo for template development
 
-  - Are the sections in a good order? In `Pull Request 91`_, @Lu said:
+  Decide on section names and overall structure.
 
-      The template looks good to me. I was putting the model assumptions and
-      limitations section right after the cause model diagram. But I think
-      this order makes more sense.
+  **Question:** Are the sections in a good order?
```
alibow: I like the section order as stands in this PR

docs/source/models/gbd2017/causes/cause_model_template.rst
```diff
@@ -91,125 +82,193 @@ Disease Description
...
-+++++
+    I am good with this structure. But, we have to make changes to the
+    causes that are finished. Also, for restrictions I like it under
+    Assumptions and Limitations section. We can add subsections if there are
+    different types of restrictions.
 
-Validation Criteria
--------------------
+  **Question:** Do we have examples of restrictions we would apply that are
+  different from GBD restrictions?
```
alibow: My understanding is that the restrictions that go in a cause model diagram are GBD-specific only (ex: male only, yld only). Restrictions for a given cause model in Vivarium would go in the _concept_ model document (ex: track under five mortality and morbidity only). I'm not sure about examples of any other type of restrictions.

docs/source/models/gbd2017/causes/cause_model_template.rst
```diff
@@ -4,6 +4,8 @@
 Cause Model Template
 ====================
 
+.. _Pull Request 64: https://github.com/ihmeuw/vivarium_research/pull/64
```
collijk: Why are you holding on to links to PRs?

docs/source/models/gbd2017/causes/cause_model_template.rst
```diff
@@ -91,125 +82,193 @@ Disease Description
...
+.. todo::
+
+  Describe the clinical and mathematical assumptions made for this cause model,
+  and the limitations these assumptions impose on the applicability of the
+  model.
+
+.. admonition:: Todo for template development
+
+  What else goes in the `Vivarium Modeling Strategy` section? Should there be
+  any other subsections besides `Scope` and `Assumptions and Limitations`?
```
collijk: The cause model diagram, data description, and validation criteria are perhaps better suited as subsections of  `Vivarium modeling strategy`

docs/source/models/gbd2017/causes/cause_model_template.rst
collijk: you can give the tables titles.  Don't need the extra subsections

docs/source/models/gbd2017/causes/cause_model_template.rst
collijk: The non-standard data sources should be handled in table 4 (the data definitions).  The point of the table layouts we came up with was to be flexible enough to handle all known use cases.

docs/source/models/gbd2017/causes/cause_model_template.rst
```diff
@@ -91,125 +82,193 @@ Disease Description
...
-Scope
-+++++
+    I am good with this structure. But, we have to make changes to the
+    causes that are finished. Also, for restrictions I like it under
+    Assumptions and Limitations section. We can add subsections if there are
+    different types of restrictions.
 
-Validation Criteria
--------------------
+  **Question:** Do we have examples of restrictions we would apply that are
```
collijk: A cause example would be something where the mortality restrictions are wider than the morbidity restrictions, meaning we have deaths due to a cause that GBD says there is no prevalence for.

docs/source/models/gbd2017/causes/cause_model_template.rst
NathanielBlairStahn: To refer back to people's comments so we can see what decisions were made and why

NathanielBlairStahn:
> Do people think it is important to have the flexibility to add subsections within the disease description?

No one has answered this question - this is important for the overall structure because it determines whether we should put the "disease description" back in its own section or leave it as it is now, at the top level.

If we follow @collijk 's framing above, it seems like there should be 3 top-level sections: 1) what do experts think, 2) what does GBD think, 3) what does Vivarium think. Section 1 would be the "disease description" (or some other title). Everything in the document would go into one of these 3 sections or some subsection thereof.

What do people think of that overall structure?

@collijk @alibow @kiranklc @beatrixkh @XiaoluQianUW @yongqx @aflaxman @SantoniS @Ninicorn

collijk:
> > Do people think it is important to have the flexibility to add subsections within the disease description?
> 
> No one has answered this question - this is important for the overall structure because it determines whether we should put the "disease description" back in its own section or leave it as it is now, at the top level.
> 
> If we follow @collijk 's framing above, it seems like there should be 3 top-level sections: 1) what do experts think, 2) what does GBD think, 3) what does Vivarium think. Section 1 would be the "disease description" (or some other title). Everything in the document would go into one of these 3 sections or some subsection thereof.
> 
> What do people think of that overall structure?
> 
> @collijk @alibow @kiranklc @beatrixkh @XiaoluQianUW @yongqx @aflaxman @SantoniS @Ninicorn

My argument about dropping the first heading is mostly an aesthetic one.  I was thinking of the expert view on the cause as, essentially, an introduction or overview to the document.  If the expert view is complex enough that you think it should have subsections, then we should put back the heading we dropped and consider what a paragraph for the overview should be.

collijk:
My preferred layout for docs is 

```
=====
Title
=====

Summary paragraph describing what's in the document.

.. contents::
   :local:
   :depth: 1

Section 1
---------

Section overview

Subsection 1.1
++++++++++++++

Subsection contents.

Etc.
```

docs/source/models/gbd2017/causes/cause_model_template.rst
```diff
@@ -56,9 +58,6 @@ Cause Model Template
       `directive <https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#directives>`_
       from :file:`index.rst`.
 
-Disease Description
--------------------
-
 .. todo::
```
beatrixkh: I think it would be helpful here to add a heading above this todo--perhaps on the same level as "GBD 2017 framework" and "Vivarium framework", as i think @collijk suggested at some point.

docs/source/models/gbd2017/causes/cause_model_template.rst
beatrixkh: I'm also happy with how these are ordered. I think the one drawback is that the general disease description is going to be somewhat tailored towards what the GBD and vivarium models are concerned with, but that doesn't necessarily become evident until you've read the "restrictions" and "scope" sections below.

But I think the benefit of keeping this section first --in that it is quite helpful in orienting the reader-- outweighs the above concern.

docs/source/models/gbd2017/causes/cause_model_template.rst
beatrixkh: Oh oops I was viewing this inline, and missed those review comments above--in reply to your aesthetic concern @NathanielBlairStahn , even if expert opinion isn't as immediately relevant to the construction of our model as the gbd/vivarium sections, it's conceptually clearer to me to have that orient-me-to-the-disease/injury under a top-level heading that is distinct from the other two. I think that heading could be either of the flavor 'disease overview' _or_ 'introduction'

docs/source/models/gbd2017/causes/cause_model_template.rst
```diff
@@ -4,6 +4,8 @@
 Cause Model Template
```
beatrixkh: My preference for this title would be to actually write "Full Name of Cause in GBD 2017" here instead of "Cause Model Template", in order to make it clear that this .rst is indeed a template, rather than a document about what template to use.

docs/source/models/gbd2017/causes/cause_model_template.rst
NathanielBlairStahn: I meant, should there be more than one subsection, one of which would be the tables. If the tables are the only thing in the "data description" section, then I agree with Ali (and Beatrix in PR 64) that it would be more descriptive to change the name of the whole section to "State and Transition Data Tables" since that clarifies what data we're talking about.

docs/source/models/gbd2017/causes/cause_model_template.rst
NathanielBlairStahn: The main problem with that approach is that when you build the documentation in Sphinx, the page will show up in the table of contents as "Full Name of Cause in GBD 2017," which would be confusing.

I guess one solution would be to have both a title and a subtitle for the template, e.g.
```
-----------------------
Cause Model Template
-----------------------

(Description of what's in this document --
delete this and everything above once you start writing)

============================
Full Name of Cause in GBD 2017
============================
```

I think that will work because it looks like Sphinx doesn't require the heading symbols to be consistent between documents. Therefore, "Cause Model Template" will render as the title until it is deleted, at which point the "Full Name of Cause in GBD 2017" subtitle will become the document's title (I guess I should test this to make sure).

What do you think of that idea, @beatrixkh ?

NathanielBlairStahn:
> My preferred layout for docs is
> 
> ```
> =====
> Title
> =====
> 
> Summary paragraph describing what's in the document.
> 
> .. contents::
>    :local:
>    :depth: 1
> 
> Section 1
> ---------
> 
> Section overview
> 
> Subsection 1.1
> ++++++++++++++
> 
> Subsection contents.
> 
> Etc.
> ```

Aesthetically, I like that layout also (that's how I organized my PhD thesis). It requires a bit more work, in that the writer needs to spend time writing an overview for each section, but I think it is helpful for orienting the reader.

docs/source/models/gbd2017/causes/cause_model_template.rst
NathanielBlairStahn: @beatrixkh , what would be the difference between a 'disease overview' and an 'introduction' in your view?

If we put the expert opinion back in its own section, as you suggest, and we follow @collijk 's suggestion for layout, then there would also be a brief paragraph between the document title and the first section, giving an overview of the entire document.

docs/source/models/gbd2017/causes/cause_model_template.rst
beatrixkh: Ahh I see. In that case--then yes, I like this @NathanielBlairStahn !

docs/source/models/gbd2017/causes/cause_model_template.rst
beatrixkh: I'm picturing 'disease overview' containing only the expert opinion (or perhaps other general information specific to the disease/cause, though we haven't done anything like that thus far and i don't have examples), and excluding metadata about the page itself.

In the format you and james are suggesting, if we include a table of contents, i was picturing 'introduction' effectively being a broad enough word to be able to hold both a ToC and the disease overview.

PR #95, opened 2020-01-09
Neonatal encephalopathy: add hierarchy diagram

NathanielBlairStahn:


NathanielBlairStahn:
> This is the structure I was expecting. I did not confirm that the id numbers are correct, though. I think you should double check them and then merge. :)

Double checked!

PR #96, opened 2020-01-09
Neonatal Encephalopathy: add restrictions table

NathanielBlairStahn:
Sequel to PR https://github.com/ihmeuw/vivarium_research/pull/95 (adding a commit on top of that branch).

Just getting the restrictions table in the document for now. I'll rearrange the sections once we finalize the structure in the cause model template.

### DIFF
--- a/docs/source/models/gbd2017/causes/neonatal_encephalopathy/index.rst
+++ b/docs/source/models/gbd2017/causes/neonatal_encephalopathy/index.rst
@@ -181,9 +181,42 @@ Model Assumptions and Limitations
 Restrictions
 ++++++++++++
 
-.. todo::
+The following table describes any restrictions in GBD 2017 on the effects of
+this cause (such as being only fatal or only nonfatal), as well as restrictions
+on the ages and sexes to which the cause applies.
 
-  Describe data restrictions in table format.
+.. list-table:: GBD 2017 Cause Restrictions
+   :widths: 15 15 20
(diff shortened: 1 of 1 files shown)


docs/source/models/gbd2017/causes/neonatal_encephalopathy/index.rst
```diff
@@ -181,9 +181,42 @@ Model Assumptions and Limitations
...
+     - [0, 7 days), age_group_id=2
+   * - YLL age group end
+     - Post Neonatal
+     - [28 days, 1 year), age_group_id=4
+   * - YLD age group start
+     - Early Neonatal
+     - [0, 7 days), age_group_id=2
+   * - YLD age group end
+     - 95 plus
+     - [95, 125 years), age_group_id=235
```
beatrixkh: I wrote this as "95 years +", but can standardize to yours--I think I like this format more

NathanielBlairStahn:
> This looks good--in the future though I think it's preferable to restrict new content to one PR, and deal with the merge conflict! Open to push-back on that.

@beatrixkh I set the "base branch" for this PR as `cause/nn_enceph` rather than `master` so that only the new differences will show up rather than all differences between this branch and master. If the changes were in different files, then I'd definitely use two separate branches from master. But since these were relatively straightforward changes to the same file, I thought this would be easier.

PR #99, opened 2020-01-10
Cause Model Template: add references to YLD & CoD appendices

NathanielBlairStahn:
Switching gears temporarily from section structure, here are some proposed templates for the YLD/CoD appendix references. There are several issues where it would be good to get feedback:

- Does the formatting look good?
- Is it clear what is being referenced? I included a link to the appendix, plus the full citation for the capstone paper. Does this make sense? Is there a better way to cite the appendices?
- I put in question marks (???-???) as a place holder to fill in the correct pages in the appendices. Is this useful, or will it be annoying to fill in the correct page numbers in each document?
- What do people think of the "Tip" directive I put after the citations? Is it helpful? Should it be an "Important" directive instead, like the one at the beginning of the document?
- I proposed a solution to Sphinx's "global citation" problem: Append the cause name to any source that will get cited on multiple pages (see the Tip and Todo directives within the document). I don't particularly like that solution -- does anyone have any better ideas?

### DIFF
--- a/docs/source/models/gbd2017/causes/cause_model_template.rst
+++ b/docs/source/models/gbd2017/causes/cause_model_template.rst
@@ -216,7 +216,7 @@ References
 
 .. [GBD-2017-YLD-Appendix-Cause-Model-Template]
 
-   Pages ???-??? in `Supplementary appendex 1 to the GBD 2017 YLD Capstone <YLD
+   Pages ???-??? in `Supplementary appendix 1 to the GBD 2017 YLD Capstone <YLD
    appendix on ScienceDirect_>`_:
 
      **(GBD 2017 YLD Capstone)** GBD 2017 Disease and Injury Incidence and
--- a/docs/source/models/gbd2017/causes/cause_model_template.rst
+++ b/docs/source/models/gbd2017/causes/cause_model_template.rst
@@ -8,6 +8,7 @@ Cause Model Template
 .. _Pull Request 76: https://github.com/ihmeuw/vivarium_research/pull/76
 .. _Pull Request 91: https://github.com/ihmeuw/vivarium_research/pull/91
 .. _Pull Request 93: https://github.com/ihmeuw/vivarium_research/pull/93
+.. _Pull Request 99: https://github.com/ihmeuw/vivarium_research/pull/99
 
 .. important::
 
@@ -263,7 +264,8 @@ References
 
    Delete this :code:`.. tip::` `directive
    <https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#directives>`_
-   once you fill in the correct page numbers for your cause in the appendices.
+   once you fill in the correct page numbers for your cause in the appendices
+   and rename the references appropriately.
 
 .. admonition:: Todo for template development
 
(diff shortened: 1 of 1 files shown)


docs/source/models/gbd2017/causes/cause_model_template.rst
```diff
@@ -214,6 +214,66 @@ Validation Criteria
...
+   enclosing the full citation name in brackets and adding a trailing
+   underscore, like this: :code:`[Full-Citation-Name]_`.
+
+   Delete this :code:`.. tip::` `directive
+   <https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#directives>`_
+   once you fill in the correct page numbers for your cause in the appendices.
+
+.. admonition:: Todo for template development
+
+  Is there a better solution to the global citation problem than making citation
```
collijk: I think we're working in an edge case of the sphinx builder.  Do the citations incorrectly resolve?  Or do the citations resolve correctly and throw spurious warnings?  We can write adaptations to the sphinx builder to resolve citations in the most local context in either case.  Or we can make a process for naming citations that we'll have to enforce on a case by case basis.

collijk:
> Switching gears temporarily from section structure, here are some proposed templates for the YLD/CoD appendix references. There are several issues where it would be good to get feedback:
> 
> * Does the formatting look good?

I'm assuming it looks good.  Someone else should check.

> * Is it clear what is being referenced? I included a link to the appendix, plus the full citation for the capstone paper. Does this make sense? Is there a better way to cite the appendices?

No opinion here.

> * I put in question marks (???-???) as a place holder to fill in the correct pages in the appendices. Is this useful, or will it be annoying to fill in the correct page numbers in each document?

Use the placeholder.  It may be annoying to fill in the page numbers, but we should do it anyway.

> * What do people think of the "Tip" directive I put after the citations? Is it helpful? Should it be an "Important" directive instead, like the one at the beginning of the document?

No opinion here.  As long as it's clear what people should do, this doesn't seem super important.

> * I proposed a solution to Sphinx's "global citation" problem: Append the cause name to any source that will get cited on multiple pages (see the Tip and Todo directives within the document). I don't particularly like that solution -- does anyone have any better ideas?

See other comment.

docs/source/models/gbd2017/causes/cause_model_template.rst
```diff
@@ -214,6 +214,66 @@ Validation Criteria
 References
 ----------
 
+.. [GBD-2017-YLD-Appendix-Cause-Model-Template]
+
+   Pages ???-??? in `Supplementary appendex 1 to the GBD 2017 YLD Capstone <YLD
```
beatrixkh: Realistically i'm just going to command-F the document, so I would shy away from adding page numbers---but I'm open to it.

beatrixkh:
From documentation session:

Formatting looks good for citations
Yes, include page numbers (actual as opposed to pdf)
Use unique identifiers for now
Yay tip :+1:

docs/source/models/gbd2017/causes/cause_model_template.rst
NathanielBlairStahn: They resolve incorrectly. For example, if both Measles and LRI have a citation called [WHO], then either the citation on the LRI page will end up pointing to the reference on the Measles page, or vice versa.

The decision in the 1/10/20 documentation session was, at least for now, to follow a naming convention that will result in unique names. For example, the measles page should cite [WHO-Measles], and the LRI page should cite [WHO-LRI].

Let's have everyone try that convention and see how it goes, and meanwhile @collijk , it might be worth looking into our options for adapting the sphinx builder.

PR #100, opened 2020-01-10
Move gbd2017 directory to top level

NathanielBlairStahn:
This implements a re-organization that I suggested awhile ago on Slack:

The list of "GBD 2017 Models" is currently buried within the "Model Descriptions" section, whereas I think it deserves to be at the top level of the documentation.

Moreover, the other pages in the "Model Descriptions" section are not actually model descriptions, but something more along the lines of "How to design and document models for Vivarium." Therefore:

- I renamed the "Model Descriptions" section "Model Design." Is that a good title?
- Do people have other suggestions?
- Perhaps the `gbd2017` folder should be renamed `gbd2017_models` or something (since it's no longer a subfolder of `models`), and perhaps the `models` folder should be renamed something else, but I don't know what.

I was a bit concerned that making this change would interfere with other people's work because it moves all the files people are working on.  But I believe Git is smart enough to correctly track the changes in the files after they have been moved. For example:

-  I merged changes from master into this branch, and it updated the files correctly even though they're in a different location in master branch.
- I created a new `test` branch from master, edited a file, then merged this branch into `test`, and the file was updated correctly in the new location.

Now that we're all experts at checking out remote branches after today's documentation session 😉 , everyone should be able to checkout the `move-gbd2017-to-top-level` branch and see what the changes look like. You might have to run `make html O=-Ea` to force Sphinx to rebuild all the files to get the new table of contents to show up correctly:

https://stackoverflow.com/questions/21019505/sphinx-force-rebuild-of-html-including-autodoc
https://github.com/sphinx-doc/sphinx/issues/6323


beatrixkh:
I like this and agree that GBD 2017 Models should be on the top level.

Along similar lines, I would add two more re-arrangements:

1. in the file "vivarium_research/docs/build/html/gbd2017/causes/causes.rst"; could we rename the first heading from "What is a cause" to something like "What does GBD define as a cause"? I would like to distinguish it in some way from the file (and section) "/vivarium_research/docs/build/html/models/cause.html#what-is-a-cause-and-what-is-a-cause-model". Perhaps the latter could also be renamed to something like "Modeling causes in Vivarium"?

2. It would be clearer to me if the cause model writeup pages:
- Diarrheal diseases
- IHD
-Ischemic Stroke
...
- Neural Tube Defects

Were all within a folder, such that they _weren't_ on the same level as the "Cause Models in GBD" and "Cause Model Template" page. (Though the template could also stay on the same level as the specific cause model writeups; I think either arrangement is reasonable). So this would look something like:
-GBD 2017 Models
   - Background: Cause Models in GBD
   - Cause Model Template
   - Cause Model Writeups
         - Diarrheal Diseases
         - IHD
            ....
____________________________
Along similar lines, what do others think of, instead of "Model Design", "Model Design for Vivarium" for something similar? This would be addressing the issue that until one has memorized the file structure, it is ambiguous as to whether anything referring to a model is background info on GBD or our own model designs. Perhaps this is also not an issue, if we are the only audience(?).
____________________________
I'm going to try moving around some files to see what happens before approving.

beatrixkh:
Also, as far as I can tell @NathanielBlairStahn , the "models" folder==the "Model Design" section. Is that correct? If so, I would say that folder should share whatever title we end up giving the section.

beatrixkh:
This is now getting super nit-picky---but this entire file structure might merit reorganization now that we've started to build it out quite a bit more. I think that if you were new to this set of pages, and looking to start documenting and designing a model, you might click on "The Vivarium Modeling Process", "The Vivarium Framework", "Model Design" _or_ "GBD 2017 Models".

I think we could possibly rename some of those pages (such as "The Vivarium Modeling Process" to "The Vivarium Modeling Workflow", and "The Vivarium Framework" to "The Vivarium Software Framework"?), but we might also want to rehouse them.

I'm viewing {The Vivarium Modeling Process, The Vivarium Framework, Useful Vivarium Tutorials} as all being Vivarium mechanics/team workflow oriented; {Model Design, GBD 2017 Models} as research-related (designing vivarium concept models based on GBD models); and {Useful Vivarium Tutorials, Contributing New Documentation, (Glossary??)} as guides for manipulating metadata relating to models (as opposed to guides to designing models). "Useful Vivarium Tutorials" I've purposely included in two possible categories--I think either works. The Glossary could also be off in its own category.

Perhaps we could both sub-folder some of these, and rename some of them? In particular I'd love to clarify the cause/what_is_a_cause sections, mentioned in my last comment.

In any case, this could all go in another PR. Any thoughts?

NathanielBlairStahn:
> Also, as far as I can tell @NathanielBlairStahn , the "models" folder==the "Model Design" section. Is that correct? If so, I would say that folder should share whatever title we end up giving the section.

Yes, this is correct. I just wasn't sure about the name "Model Design," so I didn't rename the folder yet. If "Model Design" is good enough for now, I'll change the name of the folder before I merge into master.

NathanielBlairStahn:
> 1. in the file "vivarium_research/docs/build/html/gbd2017/causes/causes.rst"; could we rename the first heading from "What is a cause" to something like "What does GBD define as a cause"? I would like to distinguish it in some way from the file (and section) "/vivarium_research/docs/build/html/models/cause.html#what-is-a-cause-and-what-is-a-cause-model". Perhaps the latter could also be renamed to something like "Modeling causes in Vivarium"?

I think the file `vivarium_research/docs/build/html/gbd2017/causes/causes.rst` is in the wrong place - it should perhaps go in the "Understanding GBD Data" section, which would require some further restructuring and/or editing.

I also think that these two sections ("What is a cause?" in the above file, and "What is a cause and what is a cause model?" in the file `vivarium_research/docs/build/html/models/cause.html`) are rather redundant since they're both defining what "cause" means in GBD.

But it probably does make sense to include both sections in different places, since one document is focusing on GBD modeling while the other document is focusing on Vivarium modeling. In this case I think the "what is a cause" section in the Vivarium document should have a link to the "what is a cause" section in thee GBD document. And it might make more sense to include the death certificate example in the GBD document vs. the Vivarium document, but I'm not sure. Given that we should have both sections, I think some sort of renaming along the lines of what @beatrixkh suggested makes sense, though if the documents were organized and named well, it might be clear from context what is meant by each section title.

NathanielBlairStahn:
> 2\. It would be clearer to me if the cause model writeup pages:
> 
> * Diarrheal diseases
> * IHD
>   -Ischemic Stroke
>   ...
> * Neural Tube Defects
> 
> Were all within a folder, such that they _weren't_ on the same level as the "Cause Models in GBD" and "Cause Model Template" page. (Though the template could also stay on the same level as the specific cause model writeups; I think either arrangement is reasonable). So this would look something like:
> -GBD 2017 Models
> 
> * Background: Cause Models in GBD
> * Cause Model Template
> * Cause Model Writeups
>   - Diarrheal Diseases
>   - IHD

As per my comment above, I think "Cause Models in GBD" should not be in this folder at all. In that case, I don't think we need subfolders - the template can be in the same folder as the cause model documents.

NathanielBlairStahn:
> Along similar lines, what do others think of, instead of "Model Design", "Model Design for Vivarium" for something similar? This would be addressing the issue that until one has memorized the file structure, it is ambiguous as to whether anything referring to a model is background info on GBD or our own model designs. Perhaps this is also not an issue, if we are the only audience(?).

@beatrixkh How about "Vivarium Model Design"? I agree that the sections should be named less ambiguously. When I was trying to come up with a name for this section, "Vivarium Modeling Process" seemed good, but that was already taken; I think we should rename that section "Vivarium Modeling Workflow" or something.

NathanielBlairStahn:
> This is now getting super nit-picky---but this entire file structure might merit reorganization now that we've started to build it out quite a bit more. I think that if you were new to this set of pages, and looking to start documenting and designing a model, you might click on "The Vivarium Modeling Process", "The Vivarium Framework", "Model Design" _or_ "GBD 2017 Models".

Good point, @beatrixkh , we should rethink the structure of the repo.

> 
> I think we could possibly rename some of those pages (such as "The Vivarium Modeling Process" to "The Vivarium Modeling Workflow", and "The Vivarium Framework" to "The Vivarium Software Framework"?), but we might also want to rehouse them.

I like these two suggestions for name changes. As I mentioned above, I think "The Vivarium Modeling Process" is too ambiguous; "Workflow" is more accurate.

I googled "software framework" and found this:

https://en.wikipedia.org/wiki/Software_framework

@collijk Is that the sense in which you are using the word "framework"? If so, then I think "The Vivarium Software Framework" would be a good title to help orient us researchers who are not familiar with such things, and perhaps it would be worth linking to the definition of "software framework" from the "vivarium framework" page (and maybe put it in the glossary).

NathanielBlairStahn:
> I'm viewing {The Vivarium Modeling Process, The Vivarium Framework, Useful Vivarium Tutorials} as all being Vivarium mechanics/team workflow oriented; {Model Design, GBD 2017 Models} as research-related (designing vivarium concept models based on GBD models); and {Useful Vivarium Tutorials, Contributing New Documentation, (Glossary??)} as guides for manipulating metadata relating to models (as opposed to guides to designing models). "Useful Vivarium Tutorials" I've purposely included in two possible categories--I think either works. The Glossary could also be off in its own category.
> 
> Perhaps we could both sub-folder some of these, and rename some of them? In particular I'd love to clarify the cause/what_is_a_cause sections, mentioned in my last comment.
> 
> In any case, this could all go in another PR. Any thoughts?

@beatrixkh Not sure what I think of this, other than moving the "Cause Models in GBD" document to the GBD data section as I mentioned above, and clarifying the "what is a cause" sections like you suggest. I think trying out your ideas and opening another PR is the way to go, though perhaps I can implement some of the ideas for renaming sections before I merge this PR, if there is enough agreement.

PR #103, opened 2020-01-16
Added cause hierarchy and restrictions table.

SantoniS:
Referenced GBD for cause hierarchy. I am not sure if I should add more information about the sequelae of IHD in the cause hierarchy. What do y'all think? I also added a Restrictions table, by using gbd_mapping causes tool.


docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
```diff
@@ -57,6 +57,47 @@ For GBD 2017 IHD fatal model, the type of model input data included data from vi
...
+     - [15, 20), age_group_id=8
+   * - YLL age group end
+     - 95 Plus
+     - [95, 126 years), age_group_id=235
+   * - YLD age group start
+     - 15 to 19
+     - [15, 20), age_group_id=8
+   * - YLD age group end
+     - 95 plus
+     - [95, 126 years), age_group_id=235
```
NathanielBlairStahn: According to gbd_mapping, this interval should be [95, 125)

NathanielBlairStahn:
@SantoniS The restrictions table looks good, except I think the endpoint for age group 235 should be 125 not 126.

For the sequelae, we have been including them (at least in summary) in the cause hierarchy diagrams. Since you have 11 sequelae, you probably want to list them in groups rather than listing them all individually.  For examples, see:

https://vivarium-research.readthedocs.io/en/latest/models/gbd2017/causes/neonatal_encephalopathy/index.html
https://vivarium-research.readthedocs.io/en/latest/models/gbd2017/causes/neonatal_jaundice/index.html

Also see:
https://vivarium-research.readthedocs.io/en/latest/models/gbd2017/causes/measles/index.html
https://vivarium-research.readthedocs.io/en/latest/models/gbd2017/risk_attributable_causes/iron_deficiency/index.html

A couple other stylistic things we decided for the hierarchy diagrams but have not documented yet:
1) Include "Other" boxes to capture any branches of the tree that exist but that we're not showing explicitly
2) Include cause id's and sequela id's (which you can find using gbd_mapping)
3) Use rounded rectangles for causes and straight rectangles for sequelae, and use rectangles with dashed edges for groups of sequelae

docs/source/models/gbd2017/causes/ischemic_heart_disease/index.rst
SantoniS: I originally had it at [95, 125) and then mistakenly updated the Notes section, getting confused that the notation meant that it was numbers from 95 up to 125 (not including 126). Thanks for correcting that!

SantoniS:
@NathanielBlairStahn Thanks for your feedback! I updated the cause hierarchy and restrictions table, and hopefully are ready to be merged to master.

SantoniS:
Thanks for your feedback @NathanielBlairStahn! I've made the requested updates, for whenever you're available to review.

PR #105, opened 2020-01-21
Added cause hierarchy, restrictions, and data description tables to diabetes mellitus model.

SantoniS:


SantoniS:
I need to include remission and update the cause model and data tables.

NathanielBlairStahn:
Small detail: If this is a PAF-of-1 cause, it should go in the "Risk-Attributable Cause Models" section:

`vivarium_research/docs/source/models/gbd2017/risk_attributable_causes`

SantoniS:
@NathanielBlairStahn  It is a PAF of one cause so I moved it to that folder. Please let me know if I did something incorrectly! Also, I updated the data transitions table to include remission. I also removed 'overall DM' as a cause in the model diagram.

NathanielBlairStahn:
> @NathanielBlairStahn It is a PAF of one cause so I moved it to that folder. Please let me know if I did something incorrectly! Also, I updated the data transitions table to include remission. I also removed 'overall DM' as a cause in the model diagram.

Looks good to me.

PR #106, opened 2020-01-21
LSFF concept model: reformat outline and add time and demographics

NathanielBlairStahn:


docs/source/models/gbd2017/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -94,5 +98,10 @@ Folic Acid Fortification
 Desired Model Outputs
 ---------------------
 
+Stratification
+++++++++++++++
+
```
SantoniS: Can we add stratification by location? * Locations: **Nigeria, India, Ethiopia**
Does the Gates Foundation want any subpopulations included? 

Can we include in the outline the list of observers needed, to be able to output the desired model outputs? Or is that in a different document?

docs/source/models/gbd2017/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -28,50 +28,54 @@ Concept Model Diagram
...
 
-Additional GBD Risks
-++++++++++++++++++++
+* :ref:`Vitamin A Deficiency / Vitamin A Deficiency <2017_cause_vitamin_a_deficiency>`
 
-:ref:`Low Birth Weight and Short Gestation <2017_risk_lbwsg>`
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* :ref:`Dietary Iron Deficiency / Iron Deficiency <2017_cause_iron_deficiency>`
 
 Risk-Outcome Relationships
```
SantoniS: Was wondering if this document is where the 'Concept model diagram' would go? And if so, where? The diagram of the treatment options, the risk factors, the causes, and desired outputs (morbidity, mortality)?

SantoniS:
How can we test if the links work?

SantoniS:
We added some comments and questions, for reference! Feel free to make changes accordingly but overall, this outline looks sufficient.

docs/source/models/gbd2017/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Yep, concept model diagram was added in a previous commit. Here's the current version: https://vivarium-research.readthedocs.io/en/latest/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.html#concept-model-diagram

docs/source/models/gbd2017/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Hmm, I hadn't thought of it that way -- we **do** want it stratified by location as well, but usually I've seen one output.hdf file per location, and then the different column names are where the further stratification by age/sex/year happen. But perhaps it would be better to explicitly say that we're stratifying by location as well?

The list of observers should eventually go here in the concept model document -- I just haven't gotten to that part yet.

NathanielBlairStahn: Oh, and right now we don't have any requests for specific subpopulations, so I'm keeping it as simple as possible unless and until we hear otherwise.

NathanielBlairStahn:
> How can we test if the reference links work?

I'm using Sphinx-specific syntax (not pure .rst) to reference other documents, so to check if they work, you'd have to checkout this remote branch and build it on your local machine to see the html.

PR #107, opened 2020-01-22
Added cause hierarchy, restrictions, and state and transition data tables for CKD.

SantoniS:
Hi! I'm not sure why my commits for diabetes mellitus cause model document are included in this. I also confirmed that I am on the ckd branch. Is there anything I can do to fix this?


NathanielBlairStahn:
> Hi! I'm not sure why my commits for diabetes mellitus cause model document are included in this. I also confirmed that I am on the ckd branch. Is there anything I can do to fix this?

That would happen if, for example, you did `git branch ckd` from your diabetes branch instead of from master.

SantoniS:
> Not sure what I think of the giant lists of sequelae in the cause hierarchy... It's cool to see all of them, but it also seems a bit gratuitous. Do other people have opinions?

For each list of sequelae, they map up to either being included in 'CKD>with condition>Severe' or 'CKD>with condition>moderate'. Is there a better way to organize the sequelae within each CKD sub cause by severity? Let me know your thoughts!

PR #110, opened 2020-01-28
Updated IHD state and transition tables, based on last week review session.

SantoniS:
Updated cause model, state, transition tables. Need to still update validation criteria.


SantoniS:
> All good and i simply want to confirm on two points:
> 
> 1. prevalence of Acute Myocardial Infarction (A) = 0 refers to no population will be initialized at this disease state.
> 2. Transition rate from S to A = P to A = incidence_c493 means the probability of developing IHD from susceptible to IHD and post IHD are the same. Can you confirm it with GBD modeler?

1. Yes, I confirm that is correct. I will also update this as a limitation of IHD vivarium modeling.

2. Yes, we confirmed this with the GBD Modeler and Greg Roth when we modelled this for IHD and are proposing to use this same modeling approach for Zenon.

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
```diff
@@ -109,15 +109,15 @@ The aspects of the disease this cause model is designed to simulate are the stat
 Model Assumptions and Limitations
 +++++++++++++++++++++++++++++++++
 
-Apart from inpatient hospital and inpatient claims data, GBD 2017 did not include any data from sources other than the literature for myocardial infarction. Given this information, the assumption is that MI is the best and only estimator for the IHD model. The limitation of this assumption and approach is the exclusion of non-MI data sources could be underestimating the IHD model. 
+Apart from inpatient hospital and inpatient claims data, GBD 2017 did not include any data from sources other than the literature for myocardial infarction. Given this information, the assumption is that MI is the best and only estimator for the IHD model. The limitation of this assumption and approach is the exclusion of non-MI data sources could be underestimating the IHD model. In the Vivarium Modeling Strategy, the limitation of the IHD model is we do not initial 'Acute MI' state in the first 28 days, and assume that prevalence = 0.
```
beatrixkh: initialize? (vs initial)

SantoniS: Oops! Typo, my bad. Thanks!

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
```diff
@@ -147,22 +150,46 @@ State and Transition Data Tables
...
+     - prevalence_c493
+     -
+   * - A
+     - prevalence
+     - 0
+     -
+   * - S
+     - prevalence
+     - 1-prevalence_493
+     - simulants not prevalent with IHD
```
kiranklc: A typo.. 1 - prevalence_c493 under S state

SantoniS: It's susceptible though. Wouldn't this make sense, then?

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
Ninicorn: The limitation of this assumption and approach is the exclusion of non-MI data sources which could lead to underestimating **the number of IHD cases** in the IHD model.

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
```diff
@@ -211,13 +238,25 @@ State and Transition Data Tables
...
    * - prevalence_s{sid}
-     - como
+     - dismod-mr
      - Prevalence of sequela with id {id}
      - 
    * - disability_weight_s{sid}
      - YLD appendix
      - Disability weight of sequela with id {id}
      - 
+   * - incidence_493
```
kiranklc: If incidence is from dismod, don't we use meid? The id 493 is different from other ids. Which one's correct?

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
```diff
@@ -109,15 +109,15 @@ The aspects of the disease this cause model is designed to simulate are the stat
...
+Apart from inpatient hospital and inpatient claims data, GBD 2017 did not include any data from sources other than the literature for myocardial infarction. Given this information, the assumption is that MI is the best and only estimator for the IHD model. The limitation of this assumption and approach is the exclusion of non-MI data sources could be underestimating the IHD model. In the Vivarium Modeling Strategy, the limitation of the IHD model is we do not initial 'Acute MI' state in the first 28 days, and assume that prevalence = 0.
 
 .. todo::
 
    Describe more assumptions and limitations of the model.
 
 Cause Model Diagram
 --------------------
-.. image:: ischemic_heart_disease_transitions.svg
+.. image:: cause_model_ihd.svg
```
beatrixkh: I don't think we've documented a model of this type before. Is there a epi standard way to denote this? Is SAP the standard notatoin? I did a very, very quick google of types of compartmental disease models, and didn't see this one. do you know @SantoniS @alibow ? Other people with epi knowledge?

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
Ninicorn: In the Vivarium Modeling Strategy, the limitation of the IHD model is **that** we do not initialize **the** 'Acute MI' state in the first 28 days, and **assume prevalence to be** = 0.

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
```diff
@@ -147,22 +150,46 @@ State and Transition Data Tables
...
      - disability weight
-     - :math:`\displaystyle{\sum_{s\in \text{sequelae_c493}}} \scriptstyle{\text{disability_weight}_s \,\times\, \text{prevalence}_s}`
+     - :math:`\displaystyle{\sum_{s=380}^{385}} \scriptstyle{\text{disability_weight}_s \,\times\, \text{prevalence}_s}`
      -
-   * - All
-     - cause-specific mortality rate
-     - :math:`\frac{\text{deaths_c493}}{\text{population}}`
+   * - A
+     - disability weight
+     - :math:`\scriptstyle{\text{disability_weight_378} \,\times\, \text{prevalence_378} + \text{disability_weight_379} \,\times\, \text{prevalence_379}}`
```
beatrixkh: This table is going off the page for me. I think if you write it like this it won't:

:math:`\displaystyle{\sum_{s\in \{378,379\}}} \scriptstyle{\text{disability_weight}_s \,\times\, \text{prevalence}_s}`

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
```diff
@@ -132,11 +132,14 @@ State and Transition Data Tables
...
+     - Susceptible to IHD
    * - A
-     - Acute Myocardial Infarction (MI)
-     - Simulant that experiences acute MI symptoms listed in 'Myocardial infarction (MI) in GBD 2017'
-   * - C
-     - Chronic IHD
+     - **A**\ cute Myocardial Infarction (MI)
+     - Simulant that experiences acute MI symptoms 
+   * - P
+     - **P**\ ost-MI IHD
```
beatrixkh: Is this encapsulating everyone who has had MI, and is not currently experiencing MI, and they are only presumed to have those symptoms for the first 28 days? Or do they just stay in this box for 28 days, and then go back to susceptible?

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
beatrixkh: I believe the source would be como, here!

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
```diff
@@ -198,11 +225,11 @@ State and Transition Data Tables
      - Source
      - Description
      - Notes
-   * - prevalence_c589
-     - como
-     - prevalence of chronic kidney disease
+   * - prevalence_c493
```
beatrixkh: I haven't double checked all of these---but I believe that for prevalence and incidence, the source is como! i was getting tables from get_outputs(cause_id=493). For deaths and emr, the source should be codcorrect---but i might be misunderstanding, if there's something wonky that goes on with them estimating IHD from the MI data?

beatrixkh: i'm seeing deaths from get_outputs:
go_call

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
SantoniS: I'll update with this. Thank you!

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
SantoniS: This was the documented model that we came up with in a meeting with @aflaxman  @collijk  and @davidbernotas. If it should be updated, let me know!

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
SantoniS: Thank you! I'll update the wording.

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
SantoniS: Thank you for identifying that! Is there a standard checklist of databases to check, when decided the sources? I may have been looking at the IHD flowchart in Appendix 1 incorrectly!

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
SantoniS: @kiranklc Based on my reply above, does it make sense or do you still believe there is a typo?

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
davidbernotas: it's not entirely different from an SIR model with a carrier state.  SI model with a carrier state, I suppose...

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
davidbernotas: There is some residual health impact after the 28 days, so we don't put them back into S after an MI - no matter how long it has been.

docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
davidbernotas: prevalence_493 is the population with IHD, so (1-prev_493) is the population without IHD.

PR #113, opened 2020-01-31
Neonatal Disorders: Cause model diagram and data tables

NathanielBlairStahn:
Start of cause model document for new neonatal disorders parent cause model.


docs/source/gbd2017_models/causes/neonatal_disorders/index.rst
```diff
@@ -0,0 +1,265 @@
...
+     -
+   * - Female only
+     - False
+     -
+   * - YLL only
+     - False
+     -
+   * - YLD only
+     - False
+     -
```
alibow: @kiranklc suggested that we could add a note in this table to clarify that this cause is NOT YLL only in GBD, but that we will be treating it as such for our cause model. Might be nice to highlight this here for clarity.

docs/source/gbd2017_models/causes/neonatal_disorders/index.rst
```diff
@@ -0,0 +1,265 @@
...
+averted by an intervention that affects :ref:`Low Birth Weight and Short
+Gestation (LBWSG) <2017_risk_lbwsg>`. The model groups together all of the
+neonatal sub-causes, and is fatal-only (no disability). The rationale for this
+design is as follows:
+
+- All of the neonatal sub-causes (except sepsis -- see `Assumptions and
+  Limitations`_ below) have the same 2-state cause model structure: "with
+  condition" and "free of condition", with no between-state transitions.
+
+- All of the causes affected by the LBWSG risk factor have the same risk ratios
```
alibow: This is a small note and I am lazy about my RR terminology, but maybe we should write "relative risk" here instead of "risk ratio" to make sure that we are being consistent with GBD language.

docs/source/gbd2017_models/causes/neonatal_disorders/index.rst
NathanielBlairStahn: Good point, thanks @alibow and @kiranklc

docs/source/gbd2017_models/causes/neonatal_disorders/index.rst
NathanielBlairStahn: Ok, seems like a good idea to keep it consistent

PR #117, opened 2020-02-10
Risk stubs

beatrixkh:
Adding stub pages for a risk exposure section. Changes include:

- new stub for risk exposure template
- new stub for "modeling risk exposures in gbd"
- moved all risk exposures housed within "vivarium_research\docs\source\gbd2017_models\risks" into their own subfolders.
- changed the risk section menu to only link to pages one level down (so that the top level menu does not include each heading of each page)

The two stubs have very rough outlines per our discussions in documentation sessions.


docs/source/gbd2017_models/risks/risk_exposure_template.rst
```diff
@@ -0,0 +1,53 @@
...
+Vivarium Modeling Strategy
+--------------------------
+
+Scope
++++++
+
+Assumptions and Limitations
++++++++++++++++++++++++++++
+
+Cause Model Diagram
```
SantoniS: Is this a 'Cause Model' diagram or 'Risk Exposure' diagram? Or would it include 'Risk Exposure' & 'Causes' it impacts?

docs/source/gbd2017_models/risks/risk_exposure_template.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Risk Exposure Overview
+----------------------
+
+ - expert opinion
+
+Risk Exposure description
+-------------------------
+
+Risk Exposures Description in GBD
```
SantoniS: I think there needs to be an added header underline before pushed to master, at least according to how it looks in the file on Git.

docs/source/gbd2017_models/risks/risk_exposure_template.rst
```diff
@@ -0,0 +1,53 @@
+.. _2017_risk_exposure_template:
+
+----------------------
+Risk Exposure Template
```
SantoniS: This may need to be '====' instead of '----'. Do y'all agree?

SantoniS:
I think the outline looks good and the stubs make sense!

docs/source/gbd2017_models/risks/risk_exposure_template.rst
```diff
@@ -0,0 +1,53 @@
...
+  Add instructions, following Nathaniel's cause model template instructions. Flesh out all headings.
+
+
+Risk Exposure Overview
+----------------------
+
+ - expert opinion
+
+Risk Exposure description
+-------------------------
```
alibow: Maybe this heading is unnecessary?

docs/source/gbd2017_models/risks/risk_exposure_template.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Risk Exposures Description in GBD
+------------------------------
+
+  - Type of risk (categorical, continuous, etc.)
+
+  - What is this a risk for?
```
alibow: Maybe we could rephrase "what is this a risk for?" to "what causes are affected by this risk?" for clarity, but it might not need to be done for this PR.

docs/source/gbd2017_models/risks/risk_exposures_gbd.rst
```diff
@@ -0,0 +1,15 @@
+.. _2017_risk_exposures_gbd:
+
+=====================
+Risk Exposures in GBD
+=====================
+
+..	todo::
+	This page will follow the analogous vivarium_research\docs\source\gbd2017_models\causes\causes.rst page.
```
NathanielBlairStahn: I think both of these pages ("Risk Exposures in GBD" and "Cause Models in GBD") are in the wrong folders. I think they should both go in the "Understanding GBD Data" section, i.e. https://vivarium-research.readthedocs.io/en/latest/data.html  (`vivarium_research/docs/build/html/data.html` - we'll have to create a new folder there).  What do other people think?

docs/source/gbd2017_models/risks/risk_exposure_template.rst
NathanielBlairStahn: Yes, the underline is too short - invalid `.rst`!

Also, do we need both the "Risk Exposure description" section and the "Risk Exposures Description in GBD" section?

docs/source/gbd2017_models/risks/risk_exposure_template.rst
beatrixkh: risk exposure! thanks for the catch

docs/source/gbd2017_models/risks/risk_exposure_template.rst
beatrixkh: oops; thank you!

docs/source/gbd2017_models/risks/risk_exposure_template.rst
beatrixkh: agreed!

docs/source/gbd2017_models/risks/risk_exposures_gbd.rst
beatrixkh: I think probably yes, but i'll leave that for a different PR.

beatrixkh: Actually, I think definitely yes. Still leaving for a different PR.

docs/source/gbd2017_models/risks/risk_exposure_template.rst
NathanielBlairStahn: Right, there shouldn't be a cause model diagram. I don't think there's an analogue for risks, but yes, we should think of other things we need for risk exposures.

PR #120, opened 2020-02-11
Risk exposure template

beatrixkh:
I accidentally merged the risks_stub branch without incorporating your comments. They've been incorporated here.

I've also expanded the risk exposure template to have more detail. I didn't do it in @NathanielBlairStahn 's to-do form; I found it somewhat confusing to discern which todos were for the template page, vs for the specific cause page---so wanted to check out this version. happy to convert over if we prefer the todo format!

### DIFF
--- a/docs/source/gbd2017_models/risks/risk_exposure_template.rst
+++ b/docs/source/gbd2017_models/risks/risk_exposure_template.rst
@@ -84,35 +84,97 @@ in a different document.
 Risk Exposures Description in GBD
 ---------------------------------
 
-  - What type of risk model does Type of risk (categorical, continuous, etc.)
+Include a description of this exposure model in the context of GBD, involving
+but not limited to:
+
+  - What type of statistical model? (categorical, continuous?)
+
+  - How is the exposure estimated? (DisMod, STGPR?)
 
   - Which outcomes are affected by this risk?
 
-  - TMREL (high level overview. does it vary by outcome?)
+  - TMREL? (This should be a very high level overview. Namely, does the TMREL 
+  vary by outcome? The details of the TMREL will be included in the 
+  *Risk Outcome Relationship Model* section)
 
 Vivarium Modeling Strategy
 --------------------------
 
-Scope
-+++++
+Include here an overview of the Vivarium modeling section
+
+Restrictions
+++++++++++++
+
+.. list-table:: GBD 2017 Exposure Restrictions
+   :widths: 15 15 20
+   :header-rows: 1
+
+   * - Restriction Type
+     - Value
+     - Notes
+   * - Male only
+     -
+     -
+   * - Female only
+     -
+     -
+   * - Age group start
+     -
+     -
+   * - Age group end
+     -
+     -
+
+..	todo::
+
+	Determine if there's something analogous to "YLL/YLD only" for this section
 
 Assumptions and Limitations
 +++++++++++++++++++++++++++
 
+Describe the clinical and mathematical assumptions made for this cause model,
+and the limitations these assumptions impose on the applicability of the
+model.
+
 Exposure Model Diagram
 ++++++++++++++++++++++
+
+Include diagram of Vivarium exposure model.
 
 Data Description Tables
 +++++++++++++++++++++++
 
- - Constants table
+As of 02/10/2020: follow the template created by Ali for Iron Deficiency, copied 
+below. If we discover it's not general enough to accommodate all exposure types,
+we need to revise the format in coworking. 
 
- - Parameters table
+.. list-table:: Constants 
+	:widths: 10, 5, 15
+	:header-rows: 1
 
- *We discussed following Ali's example with iron deficiency.*
+	* - Constant
+	  - Value
+	  - Note
+	* - 
+	  - 
+	  - 
+
+.. list-table:: Distribution Parameters
+	:widths: 15, 30, 10
+	:header-rows: 1
+
+	* - Parameter
+	  - Value
+	  - Note
+	* - 
+	  - 
+	  -
 
 Validation Criteria
 +++++++++++++++++++
+
+..	todo::
+	Fill in directives for this section
 
 References
 ----------
--- a/docs/source/gbd2017_models/risks/risk_exposure_template.rst
+++ b/docs/source/gbd2017_models/risks/risk_exposure_template.rst
@@ -80,15 +80,13 @@ Note that this is only for the exposure; you will include information on the
 relative risk of the relevant outcomes, and the cause models for those outcomes, 
 in a different document.
 
-Risk Exposure description
--------------------------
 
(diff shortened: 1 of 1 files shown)


docs/source/gbd2017_models/risks/risk_exposure_template.rst
```diff
@@ -1,53 +1,180 @@
...
+---------------------------------
+
+Include a description of this exposure model in the context of GBD, involving
+but not limited to:
 
-  - Type of risk (categorical, continuous, etc.)
+  - What type of statistical model? (categorical, continuous?)
 
-  - What is this a risk for?
+  - How is the exposure estimated? (DisMod, STGPR?)
```
SantoniS: I think adding this section is very informative! I'm interested to see how 'What is this a risk for?' can be either a short answer or visualized using a hierarchy (if the risk affects many causes and their subcauses). Should there be a section to consider about 'risk-risk correlation', potentially? Or is that just based on a specific risk-cause pair?

docs/source/gbd2017_models/risks/risk_exposure_template.rst
```diff
@@ -1,53 +1,180 @@
...
 
-  - TMREL (high level overview. does it vary by outcome?)
+  - Which outcomes are affected by this risk?
+
+  - TMREL? (This should be a very high level overview. Namely, does the TMREL 
+  vary by outcome? The details of the TMREL will be included in the
```
NathanielBlairStahn: There's an indentation error here: The line starting with "vary by outcome?" and the line below that should be indented to line up with TMREL above so that they are part of the bullet list. Currently they are showing up as a new paragraph.

docs/source/gbd2017_models/risks/risk_exposure_template.rst
```diff
@@ -1,53 +1,180 @@
...
+      subdirectory is where you will put all the files for your exposure model
+      documentation, including this document, image files, .csv files, etc.
+
+
+   #. Copy this template into your subdirectory and rename
+      it :code:`index.rst`.
+
+   #. Replace the `internal hyperlink target
+      <https://docutils.sourceforge.io/docs/user/rst/quickref.html#internal-hyperlink-targets>`_
+      :code:`.. _2017_cause_model_template:` at the top of this file with a
```
NathanielBlairStahn: Copy-paste typo:
```suggestion
      :code:`.. _2017_exposure_template:` at the top of this file with a

```

docs/source/gbd2017_models/risks/risk_exposure_template.rst
```diff
@@ -1,53 +1,180 @@
...
+
+   To begin documenting a risk exposure in GBD 2017 for a Vivarium simulation, start by
+   following these steps (after you have :ref:`created a new git branch
+   <contributing>` to work in):
+
+   #. Make a subdirectory :file:`{exposure_name}/` in the folder
+      :file:`gbd2017/risks/` , where :file:`{exposure_name}` is replaced with the
+      (potentially shortened) name of the exposure you are modeling, such as :file:`unsafe_water_source/` or :file:`low_birthweight_short_gestation/`.  This
+      subdirectory is where you will put all the files for your exposure model
+      documentation, including this document, image files, .csv files, etc.
```
NathanielBlairStahn: I'd be in favor of explicitly saying "risk exposure" instead of just "exposure," and using "risk_name" to refer to the subdirectory.  Even though this document is supposed to be focused on just the "exposure" piece of the risk, I think it might be clearer to think of it as a document about the risk as a whole, even though we may have separate documents for the risk effects.

docs/source/gbd2017_models/risks/risk_exposure_template.rst
```diff
@@ -1,53 +1,180 @@
-.. _2017_risk_exposure_template:
+.. _2017_exposure_template:
 
-----------------------
+======================
 Risk Exposure Template
-----------------------
+======================
```
NathanielBlairStahn: If this title is at the same section level as the "Full Name of Risk Exposure in GBD 2017" section below (i.e. they are both decorated with the same `======` underline/overline), then the Risk Exposure Models index page (`/vivarium_research/docs/build/html/gbd2017_models/risks/index.html`) shows "Full Name of Risk Exposure in GBD 2017" in the table of contents.  Build the HTML to see what I mean.  That's why in the cause template I used a different, unique heading for "Cause Model Template," namely an overline/underline of `--------`.

docs/source/gbd2017_models/risks/risk_exposure_template.rst
beatrixkh: Do you happen to know how to wrap lines when making a bulleted list? My quick google made it sound like we might be able to add a css class that would allow this? For now I just haven't wrapped lines to fix this.

docs/source/gbd2017_models/risks/risk_exposure_template.rst
beatrixkh: Do you think it's overdoing it to go ahead and label everything with `risk_exposure` as opposed to just `risk` or `exposure`?
e.g., Make a subdirectory :file:`{risk_exposure_name}/` and `_2017_risk_exposure_template:`

?

docs/source/gbd2017_models/risks/risk_exposure_template.rst
NathanielBlairStahn: I don't know the easiest way to do this.  I use the autoflow package in Atom to automatically hard-wrap lines to 80 characters using a keyboard shortcut - sometimes it works how I want, and sometimes I have to manually fiddle with it to get the indentation right, particularly with bulleted or numbered lists.

docs/source/gbd2017_models/risks/risk_exposure_template.rst
NathanielBlairStahn: Possibly, but I think that might be clearer.

docs/source/gbd2017_models/risks/risk_exposure_template.rst
```diff
@@ -4,50 +4,180 @@
...
+      it :code:`index.rst`.
+
+   #. Replace the `internal hyperlink target
+      <https://docutils.sourceforge.io/docs/user/rst/quickref.html#internal-hyperlink-targets>`_
+      :code:`.. _2017_risk_exposure_template:` at the top of this file with a
+      unique reference label for your cause. The reference label should have the
+      form :samp:`.. _2017_exposure_{\{exposure_name\}}:`, where
+      :samp:`{\{exposure_name\}}` is replaced with a unique descriptive name or
+      abbreviation for your risk exposure, e.g.
+      :code:`.. _2017_risk_exposure_unsafe_water:`.
```
NathanielBlairStahn: Looks like some of these got changed to "risk_exposure" and some are still just "exposure."

PR #122, opened 2020-02-12
CKD sequelae and IKF categories mapping table

SantoniS:
I added a table in 'Vivarium Modeling Strategy' which maps CKD disease state, sequelae group (stage 3, 4, 5) to IKF risk exposure category. 

@davidbernotas The risk exposure categories include Stage 1-2, which we do not include in our model. The risk exposure categories do not include 'end stage renal disease' sequelae, like we included in our disease states. Is this an issue we need to address?


davidbernotas:
Does the inclusion of stage 1 and stage 2 mean the prevalences do not add up to what we will use in our model?
I don't think problem to exclude stage 1 and 2.

docs/source/gbd2017_models/risk_attributable_causes/chronic_kidney_disease/index.rst
```diff
@@ -106,12 +106,12 @@ Vivarium Modeling Strategy
...
 
 Vivarium Modeling Strategy for Risk Factor Impaired Kidney Function (IKF) 
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
-Based on the GBD 2017 PAF of 1 relationship between IKF and CKD, simulants are initiated in the model with an assigned value of IKF. Based on this, the simulant will be determined as either moderate or severe CKD. In the treatment ramp for intervention scenarios, this assigned IKF value will determine the care that the simulant receives.
+In this model, simulants are initialized through the following process: Simulants will start as 'susceptible', then based on the prevalence of 'With Condition with CKD' (sum of prevalence of all CKD sub cause sequelae), simulants will be assigned a CKD sequelae value and thus CKD disease state (based on CKD sub cause prevalence). Then, each simulant's sequela value will be mapped to a specific impaired kidney function (IKF) category. The model will include a function which will, based on this category, initialize each simulant with a specific reduced kidney function value.
```
davidbernotas: In this model, simulants are initialized as "susceptible" or "with condition" through the following process:  simulants will be assigned directly to a CKD sequelae ("with condition" state) based on each sequelae prevalence.  Those not assigned to a sequelae will be initialized to the "susceptible" state.  Each sequelae will then be mapped back to the distribution of IKF based on sequelae based severity splits.  The result will be an IKF value for each simulant that is consistent with sub-cause prevalence.

SantoniS: > In this model, simulants are initialized as "susceptible" or "with condition" through the following process: simulants will be assigned directly to a CKD sequelae ("with condition" state) based on each sequelae prevalence. Those not assigned to a sequelae will be initialized to the "susceptible" state. Each sequelae will then be mapped back to the distribution of IKF based on sequelae based severity splits. The result will be an IKF value for each simulant that is consistent with sub-cause prevalence.

I've updated this section with your note above. Is it sufficient and ready to be approved to be merged?

SantoniS:
@collijk

PR #123, opened 2020-02-14
new outline for modeling risk factors

NathanielBlairStahn:
I created a new file since the outline we discussed today seemed different from the existing outlines for "Modeling Risk Exposure" and "Modeling Risk Effects."

Once this gets merged into master, we can each make a new branch and work separately on the individual sections within this file, and hopefully we can avoid merge conflicts.

### DIFF
--- a/docs/source/model_design/risk_factors.rst
+++ b/docs/source/model_design/risk_factors.rst
@@ -1,0 +1,32 @@
+.. _models_risk_factors:
+
+=====================
+Modeling Risk Factors
+=====================
+
+Overview
+--------
+
+What is a risk factor?
+++++++++++++++++++++++
+
+What is a causal relationship?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+What is a risk exposure?
+++++++++++++++++++++++++
+
+What is a risk effect?
+++++++++++++++++++++++
+
+Definitions
+-----------
+
+TMREL/TMRED
++++++++++++
+
+Relative Risk
++++++++++++++
+
+Population Attributable Fraction (PAF)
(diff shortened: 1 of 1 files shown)


Ninicorn:
looks good!

docs/source/model_design/risk_factors.rst
```diff
@@ -0,0 +1,32 @@
...
+What is a risk effect?
+++++++++++++++++++++++
+
+Definitions
+-----------
+
+TMREL/TMRED
++++++++++++
+
+Relative Risk
```
Ninicorn:
```suggestion
Relative Risk (RR)
```

docs/source/model_design/risk_factors.rst
```diff
@@ -0,0 +1,32 @@
...
++++++++++++++
```
Ninicorn:
```suggestion
++++++++++++++++
```

NathanielBlairStahn:
> This looks good! I think we should also add in an intro (what's subjects are covered here) and learning objectives (what should you understand by the end of reading this)---those were really helpful to me in the Modeling Causes page.
> 
> We also have, in the Modeling Causes page, a 'why do we want a document that describes each {cause model} and {what dos a model document look like}, but perhaps we can reorganize where that info is after we've written up risk factors.

Good points, @beatrixkh .  I added a Todo to add an introduction and to add the sections you're referring to.  It looks like the "learning objectives" and "why do we want a document describing each model?" sections in the Modeling Causes page don't actually have any content specific to causes, so perhaps they should go in a separate document that applies to both causes and risks?  Or do you think each document should have its own separate learning objectives that are going to be different for different types of models?

PR #124, opened 2020-02-17
Update formulates for disability and prevalence

collijk:
- Add prevalence/dw for acute state based on sequelae prevalence.  
- Fix dw calc to scale by cause level prevalence.

### DIFF
--- a/docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
+++ b/docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
@@ -140,16 +140,16 @@ restrictions on the ages and sexes to which the cause applies.
      - False
      -
    * - YLL age group start
-     - 15 to 19
+     - 15
      - [15, 20), age_group_id=8
    * - YLL age group end
-     - 95 Plus
+     - 125
      - [95, 125 years), age_group_id=235
    * - YLD age group start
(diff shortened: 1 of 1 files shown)


docs/source/gbd2017_models/causes/ischemic_heart_disease/index.rst
```diff
@@ -140,16 +140,16 @@ restrictions on the ages and sexes to which the cause applies.
      - False
      -
    * - YLL age group start
-     - 15 to 19
```
collijk: The limits have to be scalars, not intervals.

collijk:
@SantoniS we need documentation about the choice for the excess mortality rates.  What are the names of those models?  Why did we pick them?  Did you verify with a researcher on GBD's CVD team that they were the appropriate models?

SantoniS:
> @SantoniS we need documentation about the choice for the excess mortality rates. What are the names of those models? Why did we pick them? Did you verify with a researcher on GBD's CVD team that they were the appropriate models?

I'll add documentation about the decision to the cause model. We had used those meids in hypertension. I believe we also discussed including these meids in the half-day session, but I never personally talked to the GBD modellers. @davidbernotas  Did you check with the GBD modellers about these EMR models, when we were doing hypertension?

PR #125, opened 2020-02-17
Bugfix/ischemic stroke data

collijk:
- Add prevalence to acute state
- Weight dw calc by total ischemic stroke prevalence
- Fix data definitions


collijk:
@SantoniS same thing about EMRs here.  There are 20+ dismod/epi models for ischemic stroke.  Document your selection process.

PR #127, opened 2020-02-19
Risk factor overview description

beatrixkh:
Risk factor overview explanation.

Specific notes/questions, in addition to soliciting general feedback:
 - should i make the low birth weight example more GBD-correct, and if so, @NathanielBlairStahn would you talk it through with me? Alternately, should I choose a less complex example?

- is it confusing or clarifying to go from individ --> population level?

- i need to figure out the indented bullet list spacing. it's annoying me right now but i thought i'd go ahead and PR so you can review the content.

### DIFF
--- a/docs/source/model_design/risk_factors.rst
+++ b/docs/source/model_design/risk_factors.rst
@@ -33,6 +33,60 @@ Overview
 What is a risk factor?
 ++++++++++++++++++++++
 
+A risk factor is any attribute whose measure is causally related to the measure 
+of an outcome. Such attributes can range widely, and are categorized by GBD into 
+environmental/occupational, behavioral, and metabolic risks. For example:
+
+  * Environmental/occupational
+
+    * The level of particulate matter in the air of an individual's environment
+
+    * An individual's access or lack thereof to clear water
+
+  * Behavioral
+
+    * An individual's alcohol intake
+
+    * A low level of physical activity
+
+    * A history or lack thereof of childhood abuse0
+
+  * Metabolic
+
+    * The blood pressure of an individual
+
+    * Having or not having a high BMI
+
+For our purposes, the outcomes of these risks will generally be a differential 
+probability of some cause of health loss. However, the outcome of a risk could 
+also be the differential probability of another risk exposure.
+
+  For example, we might say that an individual with the attribute "is a daily 
+  smoker" is :math:`x_1` times more likely to develop lung cancer than if that same 
+  individual had the attribute "has never smoked". We quantify this more 
+  rigorously in our explanation of relative risk (RR) and theoretical minimum risk 
+  level (TMREL).
+
+  We might also say that an individual with the attribute "mother had a BMI of 17 
+  during pregnancy" is :math:`x_2` times more likely to end up with the attribute "low 
+  weight" than if that same child, all other factors held constant, had the 
+  attribute "mother had a healthy BMI during pregnancy". We will then say that the 
+  attribute "low birth weight" causes the child to have a higher probability of 
+  experiencing a bout of diarrheal disease. We then attribute health loss to this 
+  bout of diarrheal disease.
+
+In the context of our models, risk factors will be attributes of populations. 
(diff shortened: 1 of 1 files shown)


docs/source/model_design/risk_factors.rst
```diff
@@ -33,6 +33,60 @@ Overview
...
+
+  We might also say that an individual with the attribute "mother had a BMI of 17 
+  during pregnancy" is :math:`x_2` times more likely to end up with the attribute "low 
+  weight" than if that same child, all other factors held constant, had the 
+  attribute "mother had a healthy BMI during pregnancy". We will then say that the 
+  attribute "low birth weight" causes the child to have a higher probability of 
+  experiencing a bout of diarrheal disease. We then attribute health loss to this 
+  bout of diarrheal disease.
+
+In the context of our models, risk factors will be attributes of populations.
```
collijk: In our models, they're not attributes of populations, but attributes of individuals averaged over the size of a time step.  E.g.

- The average level of particulate matter an individual is exposed to over the course of a week.
- The average blood pressure of an invidividual over a month.
etc.

The challenge is taking the population level exposures averaged over a demographic bin for a year and applying them to an individual for some subset of time.  We tend to think rather simply about these risk factors, but we don't have to.  Take smoking, for example: whether or not a person is currently a smoker has much less bearing on their health than their history of smoking.  We could make similar arguments about obesity and blood pressure (and many others).  The important features of a risk exposure might even be different for different outcomes. A person's current exposure to intimate partner violence will have a strong effect on acute outcomes like injuries or firearm deaths, but their history of exposure will be a strong factor in depressive disorders.

beatrixkh: ah; oops. I described a GBD risk factor there, yes?

ok interesting, i wasn't picturing this correctly and see how it could gain a lot of complexity. I can revise this before doc session tmr.

thanks!

docs/source/model_design/risk_factors.rst
```diff
@@ -33,6 +33,72 @@ Overview
...
+a risk effect and a risk exposure, the subset of a simulant's history 
+of exposure that ought to be associated with a risk effect will depend on the 
+risk factor. 
+
+	For example, consider the risk-outcome pairs *unsafe water 
+	source* and *diarrheal diseases*, versus *smoking* and *diabetes*. We see that 
+	only a simulant's recent exposure to an unsafe water source will affect their 
+	probability of suffering from diarrheal diseases in the next week. However, the 
+	probability of becoming diabetic in the next year will be affected by a
+	simulants entire history of smoking.
```
beatrixkh: simulant's*

docs/source/model_design/risk_factors.rst
```diff
@@ -33,6 +33,72 @@ Overview
...
+    * An individual's access or lack thereof to clear water
+
+  * Behavioral
+
+    * An individual's alcohol intake
+
+    * A low level of physical activity
+
+    * A history or lack thereof of childhood abuse0
+
```
alibow: Typo here with "0"

docs/source/model_design/risk_factors.rst
```diff
@@ -33,6 +33,72 @@ Overview
...
+  For example, we might say that an individual with the attribute "is a daily 
+  smoker" is :math:`x_1` times more likely to develop lung cancer than if that same 
+  individual had the attribute "has never smoked". We quantify this more 
+  rigorously in our explanation of relative risk (RR) and theoretical minimum risk 
+  level (TMREL).
+
+  We might also say that an individual with the attribute "mother had a BMI of 17 
+  during pregnancy" is :math:`x_2` times more likely to end up with the attribute "low 
+  weight" than if that same child, all other factors held constant, had the
```
NathanielBlairStahn: To be consistent, I think this should say "low birth weight" instead of just "low weight."

@beatrixkh, Based on our discussion at the documentation session today, I think this example is good, even though "low birth weight" doesn't correspond exactly to a risk in GBD.

PR #129, opened 2020-02-21
Updating the CKD cause model and update on data tables.

SantoniS:
I still have to add more updates, so this isn't the final commit.


SantoniS:
@davidbernotas  @collijk @rmudambi I need to update the CKD remission rates here.

davidbernotas:
Is it correct that prevalence of moderate = (prev of ALB + prev of III + prev IV)/(prev of CKD)?
Why not simply (prev of ALB + prev of III + prev IV)?
Since the stage-level prevalence add up to the total CKD prevalence, why do we divide by (prev of CKD)?

docs/source/gbd2017_models/risk_attributable_causes/chronic_kidney_disease/index.rst
```diff
@@ -148,11 +148,10 @@ Mapping CKD States to IKF Categories in Vivarium
...
 +++++++++++++++++++++++++++
 
-.. todo::
+* EMR = 0 for all states, except for Stage V.
+
+* Stage V is the only Stage which is fatal, where there will be YLLs and Deaths.
 
-  Describe the clinical and mathematical assumptions made for this cause model,
-  and the limitations these assumptions impose on the applicability of the
-  model.
```
davidbernotas: Presently, we are using prevalence for each stage of CKD to assign the each person in the population a CKD severity state.  We are assuming (for now) that there is no transition between states.  As a result, we should expect the prevalence for early stage CKD to swell as the simulation runs, since there is nowhere for these new incident cases to go.  If we had remission (which is actually the term used for "progression") rates, we could use these as inter-stage incidence rates.

docs/source/gbd2017_models/risk_attributable_causes/chronic_kidney_disease/index.rst
```diff
@@ -219,7 +218,7 @@ State and Transition Data Tables
      -
    * - C
      - prevalence
-     - prevalence_c589
+     - 0
      -
    * - M
      - prevalence
```
davidbernotas: See comment about whether dividing by prevalence of CKD is correct.

docs/source/gbd2017_models/risk_attributable_causes/chronic_kidney_disease/cause_model_ckd.svg
```diff
@@ -1 +1 @@
...
..."#y"/><use transform="matrix(0.05,0,0,0.05,15.949999999999998,0)" xlink:href="#aa"/><use transform="matrix(0.05,0,0,0.05,25.950000000000003,0)" xlink:href="#Z"/><use transform="matrix(0.05,0,0,0.05,35.95,0)" xlink:href="#E"/><use transform="matrix(0.05,0,0,0.05,45.95,0)" xlink:href="#I"/></g><path d="M127-220V0H93v-220H8v-28h204v28h-85" id="af"/><path d="M115-194c55 1 70 41 70 98S169 2 115 4C84 4 66-9 55-30l1 105H24l-1-265h31l2 30c10-21 28-34 59-34zm-8 174c40 0 45-34 45-75s-6-73-45-74c-42 0-51 32-51 76 0 43 10 73 51 73" id="ag"/><g id="w"><use transform="matrix(0.05,0,0,0.05,0,0)" xlink:href="#af"/><use transform="matrix(0.05,0,0,0.05,10.3,0)" xlink:href="#H"/><use transform="matrix(0.05,0,0,0.05,16.25,0)" xlink:href="#C"/><use transform="matrix(0.05,0,0,0.05,26.25,0)" xlink:href="#Z"/><use transform="matrix(0.05,0,0,0.05,36.25,0)" xlink:href="#z"/><use transform="matrix(0.05,0,0,0.05,45.25,0)" xlink:href="#ag"/><use transform="matrix(0.05,0,0,0.05,55.25,0)" xlink:href="#N"/><use transform="matrix(0.05,0,0,0.05,59.2,0)" xlink:href="#C"/><use transform="matrix(0.05,0,0,0.05,69.2,0)" xlink:href="#Z"/><use transform="matrix(0.05,0,0,0.05,79.2,0)" xlink:href="#D"/></g></defs></g></svg>
```
davidbernotas: I appreciate the new model diagram.  
We are not including kidney transplant in the simulation model, however.  Is this figure supposed to represent what we are actually planning on using, or a more general model?

SantoniS: It should be a generalized model that we are using. Should I update the cause model diagram to remove 'kidney transplant'?

collijk: Yes.  The diagram should correspond to the simulation model.  You can include other diagrams for comparison if you'd like.

SantoniS:
> Is it correct that prevalence of moderate = (prev of ALB + prev of III + prev IV)/(prev of CKD)?
> Why not simply (prev of ALB + prev of III + prev IV)?
> Since the stage-level prevalence add up to the total CKD prevalence, why do we divide by (prev of CKD)?
> 
> Also, per a conversation we had recently - please confirm that we do have prevalence for end stage renal disease.

I updated calculation for prevalence of moderate. I also confirm we have prevalence for end stage renal disease.

docs/source/gbd2017_models/risk_attributable_causes/chronic_kidney_disease/cause_model_ckd.svg
SantoniS: I updated the cause model diagram to remove 'kidney transplant'. I do not think it is complete yet, but is on the right track. @davidbernotas what do you recommend the cause model diagram to look like to connect between the cause, sequelae, and risk factor categories?

PR #130, opened 2020-02-21
add in headings for causality section

alibow:


docs/source/model_design/risk_factors.rst
```diff
@@ -36,6 +36,18 @@ What is a risk factor?
 What is a causal relationship?
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+Counfounding
```
SantoniS: I think this should just be 'Confounding', other than that this is great!

alibow: oops :). I will fix. Thanks!

PR #132, opened 2020-02-25
Iron deficiency modeling strategy

alibow:
This branch is accidentally named causality, but it is really for iron deficiency anemia : ) 

I updated a few things and marked where we need to make decisions as a research team regarding modeling implementation strategy. If anyone has any thoughts on the approaches I proposed in the "to-do" notes in this document (or elsewhere), please leave your feedback!

Thanks.

### DIFF
--- a/docs/source/gbd2017_models/risk_attributable_causes/iron_deficiency/index.rst
+++ b/docs/source/gbd2017_models/risk_attributable_causes/iron_deficiency/index.rst
@@ -64,7 +64,7 @@ sex-, and location-specific demographic group, the prevalence of anemia (by
 severity level) in each group is determined by the WHO hemoglobin thresholds 
 defined in the following table.
 
-.. _above:
+.. _`WHO hemoglobin tresholds table`:
 
 .. list-table:: WHO Hemoglobin Thresholds (g/L) [Kassebaum-et-al-2016]_
 	:widths: 15, 15, 15, 15
@@ -155,7 +155,7 @@ The sequela associated with the dietary iron deficiency cause in GBD 2017
 include mild iron deficiency anemia, moderate iron deficiency anemia, and 
 severe iron deficiency anemia. The severity of iron deficiency anemia is 
 determined by the WHO age- and sex- specific hemoglobin concentrations, as 
-described in the table above.
+described in the `WHO hemoglobin tresholds table`_.
 
 Iron Deficiency Risk Factor
 +++++++++++++++++++++++++++
@@ -202,6 +202,8 @@ Notably, not all causes of anemia in the GBD anemia impairment are considered
 iron responsive (i.e. will respond to iron supplementation). A list of causes 
 with iron responsive anemia health states along with their cause and 
 anemia-afflicated sequelae IDs are included in the table below.
+
+.. _`iron responsive anemia causes table`:
 
 .. list-table:: Iron Responsive Anemia Causes 
 	:widths: 40 40 40
@@ -274,7 +276,10 @@ The modeling strategy detailed in this document aims to evaluate *all* iron
 responsive anemias rather than dietary iron deficiency anemia, specifically. 
 However, the modeling strategy can be modified to include only dietary iron 
 deficiency (PAF of 1 cause), if desired by assuming that dietary iron 
-deficiency anemia is the only iron responsive cause of anemia.
+deficiency anemia is the only iron responsive cause of anemia, rather than all 
+of the causes listed in the `iron responsive anemia causes table`_ and that the 
+dietary iron deficiency anemia sequelae are the only iron responsive anemia 
+sequelae in the `anemia sequelae IDs table`_.
 
 .. note:: 
 
@@ -325,6 +330,8 @@ distribution will **not** vary as the simulant ages.
 
 The ensemble distribution of population hemoglobin concentrations can be 
 recreated with the following equations and code: 
+
+.. _`population hemoglobin parameters table`:
 
 .. list-table:: Population Hemoglobin Parameters
 	:widths: 10, 5, 15
@@ -396,6 +403,8 @@ Where *prevalence_iron_responsive_anemia* and *prevalence_total_anemia* are
 equal to the severity-, age-, sex-, and location-specific prevalence (from COMO)
 summed across all iron responsive anemia and all total anemia sequela IDs, 
 respectively. Sequela IDs for each category are listed in the table below. 
+
+.. _`anemia sequelae IDs table`:
 
 .. list-table:: Sequela IDs 
 	:widths: 5, 30, 20
@@ -590,7 +599,7 @@ The overall prevalence and YLDs of anemia should be equal between:
 The prevalence of anemia using the population hemoglobin distribution can be 
 calculated using the code below using the parameters defined earlier in this 
 document and assuming age- and sex- specific *anemia_threshold* values as 
-defined in the table above_:
+defined in the `WHO hemoglobin tresholds table`_ and using the parameters defined in the `population hemoglobin parameters table`_:
 
 .. warning::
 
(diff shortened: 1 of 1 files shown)


docs/source/gbd2017_models/risk_attributable_causes/iron_deficiency/index.rst
```diff
@@ -509,7 +560,9 @@ and age groups.
 Because we are not modeling individual causes of anemia (and their associated 
 mortality), we assume that all simulants have the same mortality rate regardless 
 of their hemoglobin value, when in reality, those with lower hemoglobin values 
-will have higher mortality rates than those with higher hemoglobin values.
+will have higher mortality rates than those with higher hemoglobin values. 
+Notably, deaths due to causes with iron responsive anemia sequelae account for 
+approximately 1.1% of deaths in the first five years of life globally.
```
NathanielBlairStahn: @alibow It would be good to include a reference for this number. For example, you could put your notebook in the vivarium data analysis repo and make a link to it, or add a link to GBD compare.

docs/source/gbd2017_models/risk_attributable_causes/iron_deficiency/index.rst
```diff
@@ -464,7 +505,14 @@ on the methodology described here.
 
 .. todo::
 
-	Describe how to handle the changing prev_ira/prev_total_anemia across age groups.
+	Describe how to handle the changing prev_ira/prev_total_anemia across age 
+	groups, which generally decreases by about 2% between the early NN and 1-4
```
NathanielBlairStahn: Similarly, it would be good to cite where you got this figure of 2% - either from your own code or from GBD compare.

docs/source/gbd2017_models/risk_attributable_causes/iron_deficiency/index.rst
```diff
@@ -268,7 +270,16 @@ Model Scope
 The scope of the Vivarium modeling strategy detailed in this document is to 
 sample the hemoglobin concentration for an individual simulant (who is not a 
 woman of reproductive age) and evaluate if that simulant's hemoglobin 
-concentration will respond to iron supplementation.
+concentration will respond to iron supplementation (iron responsive). 
+
+The modeling strategy detailed in this document aims to evaluate *all* iron
```
beatrixkh: Maybe 'the modeling strategy detailed **in this section**', to distinguish from the GBD strategy described elsewhere

docs/source/gbd2017_models/risk_attributable_causes/iron_deficiency/index.rst
```diff
@@ -289,6 +300,10 @@ intialize these parameters are included in the following sections.
 Notably, the initialization of a simulant's hemoglobin concentration should 
 occur *before* the initialization of iron responsiveness.
 
+.. todo::
```
beatrixkh: once you've confirmed this, i think it would be worth including the reasoning!

docs/source/gbd2017_models/risk_attributable_causes/iron_deficiency/index.rst
```diff
@@ -268,7 +270,16 @@ Model Scope
...
+concentration will respond to iron supplementation (iron responsive). 
+
+The modeling strategy detailed in this document aims to evaluate *all* iron 
+responsive anemias rather than dietary iron deficiency anemia, specifically. 
+However, the modeling strategy can be modified to include only dietary iron 
+deficiency (PAF of 1 cause), if desired by assuming that dietary iron 
+deficiency anemia is the only iron responsive cause of anemia, rather than all 
+of the causes listed in the `iron responsive anemia causes table`_ and that the 
+dietary iron deficiency anemia sequelae are the only iron responsive anemia 
+sequelae in the `anemia sequelae IDs table`_.
```
beatrixkh: I'm assuming i'll understand this once i've read to the end of this section, or if i read this top to bottom. but note that this is confusing---

You write both that:
- the modeling strategy in this doc aims to evaluate all iron responsive anemias
- the vivarium modeling strategy is is for dietary iron deficiency

does that mean, either, that
- you _have_ assumed_, in the way you describe above, that that dietary iron deficiency anemia is the only iron responsive cause of anemia? or,
- when you say 'modeling strategy detailed in this document' on line 275, you mean the entire document, not the vivarium modeling strategy?

docs/source/gbd2017_models/risk_attributable_causes/iron_deficiency/index.rst
```diff
@@ -473,12 +521,15 @@ Other Model Notes/Strategies
 Neonatal Age Groups
 ^^^^^^^^^^^^^^^^^^^
 
+Neonatal age groups should be excluded from the process described in this
```
beatrixkh: I think you said at one point this is due to lack of data from gbd? i think it would be helpful to include the reasoning here.

docs/source/gbd2017_models/risk_attributable_causes/iron_deficiency/index.rst
```diff
@@ -509,7 +560,9 @@ and age groups.
 Because we are not modeling individual causes of anemia (and their associated 
 mortality), we assume that all simulants have the same mortality rate regardless 
 of their hemoglobin value, when in reality, those with lower hemoglobin values 
-will have higher mortality rates than those with higher hemoglobin values.
+will have higher mortality rates than those with higher hemoglobin values. 
+Notably, deaths due to causes with iron responsive anemia sequelae account for 
+approximately 1.1% of deaths in the first five years of life globally. 
 
 Because hemoglobin concentrations are not directly modeled among the early and 
 late neonatal age groups in GBD, the prevalence of mild, moderate, and severe
```
beatrixkh: lines 576-580, you have:

The modeling strategy both as conducted by the GBD modelers and as described in this document assume a constant shape and standard deviation in the hemoglobin distribution throughout the modeling process. This is a limitation of our modeling strategy in that we assume the distribution before a shift is applied maintains the same shift after a shift due to the intervention is applied.

is this 'the same *shape* after a shift is applies?

i still don't understand the implication of this, 😂 so the suggestion might be in error.

docs/source/gbd2017_models/risk_attributable_causes/iron_deficiency/index.rst
alibow: Good point! I will add

docs/source/gbd2017_models/risk_attributable_causes/iron_deficiency/index.rst
alibow: Ah, I think this is me being lazy with my wording. This document is a modeling strategy to model all iron responsive anemia rather than the singular dietary iron deficiency anemia cause; however, it can be restricted so that it only includes the singular dietary iron deficiency anemia cause by ignoring the other iron responsive anemia causes. I will clarify this in the document, but let me know if this still doesn't make sense.

docs/source/gbd2017_models/risk_attributable_causes/iron_deficiency/index.rst
alibow: Yes!! Thank you for catching this :)

alibow: The implication is that the shift applies uniformly to all individuals regardless of their position in the hemoglobin distribution (imagine just pushing a normal distribution to the right, but its shape stays the same). However, we might actually expect the left end of the distribution to get pushed farther than the right end of the distribution in reality, so the shape would change from a normal distribution to a more lop-sided distribution if this were to happen (we assume it doesn't). I should add the the shape is only maintained in our simulation when stratified by iron responsiveness (because non-iron-responsive individuals will not get pushed at all).

docs/source/gbd2017_models/risk_attributable_causes/iron_deficiency/index.rst
NathanielBlairStahn: I just pushed a new commit to resolve the other concerns, but I wasn't sure how to address this comment, so if necessary, this part could still be reworded for clarity.

beatrixkh: oh oops! i never looked back at this, apologies.

Yes! The wording is now clear to me. Looks good @NathanielBlairStahn @alibow

PR #133, opened 2020-02-25
Causality introduction text

alibow:


docs/source/model_design/risk_factors.rst
```diff
@@ -36,6 +36,76 @@ What is a risk factor?
...
+  cancer, such as differences in air pollution due to automobiles and the rise 
+  of electronic cigarettes. 
+
+  Therefore, while the comparison between 1954 and 2020 US may be interesting 
+  and useful, it is not a true counterfactual comparison. Instead, a 
+  counterfactual scenario could be conceptualized as "what would the lung 
+  cancer rate in the US be in 1954 *if no one smoked* and **all else was equal**?" 
+  Then, we could evaluate the independent effect of smoking on lung cancer
+  without interference from any other factors.
+
```
collijk: Another example I like:

Suppose you've finished work and you're trying to determine how to get down to the international district for dinner.  You go to your maps application and ask **What is the quickest route to my destination?**.  It tells you that based on the current time, weather, day of the week, road work conditions, etc.  that you should take the light rail.  

So you leave the office, walk a few blocks to the westlake station, and catch the train.  Two stops in, your train gets stopped behind a broken down bus. You decide to get off the train and walk the remaining distance.  Arriving ten minutes late to dinner, you think to yourself **What would have happened if I took the bus instead?**.

collijk: I like this one because there's a clear difference between the availability of the information when you ask the two questions, which is an important feature of answering counterfactual Qs.

collijk:
Also, some copy edits: https://github.com/ihmeuw/vivarium_research/pull/141

docs/source/model_design/risk_factors.rst
alibow: Yeah, I like this example too! I will add it in :)

yongqx: As mentioned above that counterfactual means only one variable changed and all others remain the same. What if we adjusted two parameters compare our intervention scenario to baseline? Is this intervention scenario not considered as counterfactual scenario anymore?

NathanielBlairStahn: Hmm, good point. Seems like the important thing is not necessarily that we are considering only one variable, but that we are considering what happens if we change a fixed set of variables while keeping everything else the same. If we liberally interpret "single variable" to include multi-dimensional variables, then perhaps the current wording already captures this idea :)

collijk: We can be very crisp about this, I think.  

You are doing counterfactual analysis if:
1. The **exact** same set of decisions are made in each of your scenarios (except for decisions explicitly screened or added by your intervention).
2. All variables exogeneous to the model (in a simulation, this is the random numbers) are **exactly** the same for each decision shared across the models.  (This is actually implied by condition 1, but worth stating explicitly).

This is contrasted with normal intervention analysis where you might split a population into multiple subgroups with approximately identical **distributions** of the exogenous variables and apply each of your scenarios to one of the subgroups.

PR #140, opened 2020-02-27
Updated description for CKD moderate EMR.

SantoniS:
I updated the description for the CKD model of the State Data table for EMR for moderate CKD. I forgot to delete the old description and found it after I merged to master.

### DIFF
--- a/docs/source/gbd2017_models/risk_attributable_causes/chronic_kidney_disease/index.rst
+++ b/docs/source/gbd2017_models/risk_attributable_causes/chronic_kidney_disease/index.rst
@@ -128,7 +128,7 @@ The aspects of the disease this cause model is designed to simulate is the basic
 Vivarium Modeling Strategy for Risk Factor Impaired Kidney Function (IKF) 
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
-In this model, simulants are initialized as "susceptible" or "with specific sequelae-level condition" through the following process: simulants will be assigned directly to a CKD sequelae ("with condition" state) based on each sequelae prevalence. Those not assigned to a sequelae will be initialized to the "susceptible" state. Each sequelae will then be mapped back to the distribution of IKF based on sequelae based severity splits. The result will be an IKF value for each simulant that is consistent with sub-cause prevalence.
+In this model, simulants are initialized as "susceptible" or "with specific sequelae-level condition" through the following process: simulants will be assigned directly to a CKD sequelae ("with condition" state) based on each sequelae prevalence. Those not assigned to a sequelae will be initialized to the "susceptible" state. Each sequelae will then be mapped back to the distribution of IKF based on sequelae based severity splits. The result will be an IKF value for each simulant that is consistent with sub-cause prevalence. This structure means there's no impact of SBP on CKD.
 
 Mapping CKD States to IKF Categories in Vivarium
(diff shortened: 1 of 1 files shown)


docs/source/gbd2017_models/risk_attributable_causes/chronic_kidney_disease/index.rst
```diff
@@ -250,7 +250,7 @@ State and Transition Data Tables
    * - M
      - excess mortality rate (EMR) of moderate CKD
      - :math:`\frac{\text{CSMR_c589}}{\text{prevalence_sequelae_mod.sub_causes.c589}}`
-     - EMR for moderate CKD is 0, given the assumption that moderate CKD is non-fatal only.
```
aflaxman: I thought EMR for M and Sev would be the same, both equal to \frac{\text{CSMR_c589}}{\text{prevalence_sequelae_sev.sub_causes.c589 + prevalence_sequelae_mod.sub_causes.c589}}

docs/source/gbd2017_models/risk_attributable_causes/chronic_kidney_disease/index.rst
```diff
@@ -128,7 +128,7 @@ The aspects of the disease this cause model is designed to simulate is the basic
...
 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
-In this model, simulants are initialized as "susceptible" or "with specific sequelae-level condition" through the following process: simulants will be assigned directly to a CKD sequelae ("with condition" state) based on each sequelae prevalence. Those not assigned to a sequelae will be initialized to the "susceptible" state. Each sequelae will then be mapped back to the distribution of IKF based on sequelae based severity splits. The result will be an IKF value for each simulant that is consistent with sub-cause prevalence.
+In this model, simulants are initialized as "susceptible" or "with specific sequelae-level condition" through the following process: simulants will be assigned directly to a CKD sequelae ("with condition" state) based on each sequelae prevalence. Those not assigned to a sequelae will be initialized to the "susceptible" state. Each sequelae will then be mapped back to the distribution of IKF based on sequelae based severity splits. The result will be an IKF value for each simulant that is consistent with sub-cause prevalence. This structure means there's no impact of SBP on CKD.
```
aflaxman: I understand that David has said he is fine with no impact of SBP on CKD, but I don't understand *why* we need to impose this limitation.  There is still incidence of CKD, and it seems to me that we have a way to make this incidence higher for people with higher SBP.

SantoniS: @davidbernotas Would you like to answer this?

docs/source/gbd2017_models/risk_attributable_causes/chronic_kidney_disease/index.rst
SantoniS: I read @davidbernotas  email and the note says 'As for the issue concerning EMR and the propensity for people to exit from severe stages more frequently than from mild/moderate stages – Abie recommends, and I agree, that we simply use the CSMR for each cause and apply this to all severity splits symmetrically.' 

I probably didn't understand how to update the EMR caulculations and misunderstood 'applying to severity splits symmetrically'. I will update EMR for Mod and Sev calculations with what your comment says above. If it needs to be revisited, I can update at a later time to something else.

PR #145, opened 2020-02-29
Add data tables section on cause model template

kiranklc:


docs/source/gbd2017_models/causes/cause_model_template.rst
```diff
@@ -276,6 +276,325 @@ Cause Model Diagram
...
+     - birth_prevalence_cid
+     - 
+   * - I
+     - excess mortality rate
+     - :math:`\frac{\text{deaths_cid}}{\text{population} \times \text{prevalence_cid}}`
+     - = (cause-specific mortality rate) / prevalence
+   * - I
+     - disability weights
+     - :math:`\displaystyle{\sum_{s\in \text{sequelae_cid}}} \scriptstyle{\text{disability_weight}_s \,\times\, \text{prevalence}_s}`
+     - = average disability weight over all sequelae
```
yongqx: if you mean average disability weight over all sequelae, then you need to divide by sum(prevalence_s). Otherwise the formula means total disability weight over all sequelae in this disease state.

yongqx: if you add this denominator to your equation, the more precise definition is prevalence weighted average disability weight over all sequelae

kiranklc: I think the formula is correct. So I will change the wording as per your suggestion

PR #149, opened 2020-03-05
Vitamin A Deficiency: Start Vivarium modeling section

NathanielBlairStahn:
It would be great to get some feedback about the following:

1. Is my description of the "propensity model" accurate?  What about the different terms I used to refer to it ("exposure model" and "prevalence-only model")?
2. Does my rationale for using this model sound accurate?
3. What needs to go in the data tables and what should the table format be?


docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
```diff
@@ -7,15 +7,19 @@ Vitamin A Deficiency
...
 ------------------------------------
-In Global Burden of Disease (GBD) 2017, VAD exposure definition is proportion of the population with serum retinol concentration <0·7 μmol/L.
-Like iron deficiency, the cause VAD is also a population attributable fraction (PAF) of 1 cause with the VAD risk factor. That is, 100% of the VAD cases are attributable
-to the VAD risk factor. VAD Risk exposure and VAD cause prevalence data are the same.[2]
+
+In Global Burden of Disease (GBD) 2017, VAD exposure definition is proportion of
+the population with serum retinol concentration <0·7 μmol/L. Like iron
+deficiency, the cause VAD is also a population attributable fraction (PAF) of 1
+cause with the VAD risk factor. That is, 100% of the VAD cases are attributable
+to the VAD risk factor. VAD Risk exposure and VAD cause prevalence data are the
```
beatrixkh: Does this mean that all data that goes into the GBD exposure model is the same as the data that goes into the GBD prevalence model, or that everyone with VAD exposure also has the VAD cause? (i.e., is this about GBD model inputs or outputs?)

NathanielBlairStahn: This means the GBD outputs are the same. There is only one VAD model, not separate models for the risk and cause. I can add some wording to clarify this.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
```diff
@@ -7,15 +7,19 @@ Vitamin A Deficiency
...
 
 GBD 2017 Modeling Strategy
 ------------------------------------
-In Global Burden of Disease (GBD) 2017, VAD exposure definition is proportion of the population with serum retinol concentration <0·7 μmol/L.
-Like iron deficiency, the cause VAD is also a population attributable fraction (PAF) of 1 cause with the VAD risk factor. That is, 100% of the VAD cases are attributable
-to the VAD risk factor. VAD Risk exposure and VAD cause prevalence data are the same.[2]
+
+In Global Burden of Disease (GBD) 2017, VAD exposure definition is proportion of
+the population with serum retinol concentration <0·7 μmol/L. Like iron
+deficiency, the cause VAD is also a population attributable fraction (PAF) of 1
```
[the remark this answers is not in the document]
NathanielBlairStahn: Ok, I can edit this.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
```diff
@@ -81,21 +89,21 @@ The Vitamin A deficiency risk factor in GBD 2017 is a **dichotomous variable** .
...
-	* - Risk factor exposure
-	  - gbd_id = reiid(96)
-	  - como, use get_measure
+  :widths: 20 45 40
+  :header-rows: 1
+
+  * - Measure
+    - ID
+    - Data source
+  * - Remission
```
collijk: 1. Incidence and remission are measures for the cause.
2. I thought we decided not to use them at all.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
```diff
@@ -141,44 +149,148 @@ Restrictions
 Relative Risks
 ^^^^^^^^^^^^^^
 
-The affected causes with the Vitamin A deficiency cause in GBD 2017 include
+The causes affected by the Vitamin A Deficiency risk in GBD 2017 include
 :ref:`lower respiratory infections <2017_cause_lower_respiratory_infections>`,
-:ref:`diarrhoeal diseases <2017_cause_diarrhea>`, :ref:`measles
+:ref:`diarrhoeal diseases <2017_cause_diarrhea>`, and :ref:`measles
 <2017_cause_measles>`. The relative risks for these causes appear in Table 4 on
```
collijk: I think I would put a big note here that LRI is dropped in the most recent round of GBD due to insufficient causal criteria and mention it again in your limitations.  It's worth pointing out to the client.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
```diff
@@ -141,44 +149,148 @@ Restrictions
...
-	  - No (not significant)
-	* - Malaria
-	  - 
-	  - 3.65 (2.23 - 5.97)
-	  - No (only one study)
-	  
+  :widths: 15 13 15 15
+  :header-rows: 1
+
+  * - Cause
```
collijk: You're welcome to include the actual data means and uncertainties here, but unless you want to do something differently that use the draw level rrs from GBD, put the rei_id and cause_id associated with the risk outcome pair you're using in the table.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
```diff
@@ -141,44 +149,148 @@ Restrictions
...
+  * - Meningitis
+    -
+    - 3.2 (0.69 - 14.75)
+    - No (not significant)
+  * - Malaria
+    -
+    - 3.65 (2.23 - 5.97)
+    - No (only one study)
+
+The above relative risks for GBD 2017 should be interpreted as rate ratios for
```
collijk: This is good.  I think you should have the interpretation as a column in the table though.  A single risk factor may have different kinds of effects on different outcomes and should be specified pairwise.  Also include the numerator and denominator or a link to what rate ratio means (I think we made a glossary in the documentation).

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
```diff
@@ -141,44 +149,148 @@ Restrictions
...
+the incidence rates of diarrhea, measles, and LRI.
+
 Vivarium Modeling Strategy
 --------------------------
 
+We will use an **exposure model** (or **prevalence-only model** or **propensity
```
collijk: I think exposure model (vs. cause model) is an appropriate way to refer to this.  We should put a description of this in the general risk documentation.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
```diff
@@ -141,44 +149,148 @@ Restrictions
...
+model**) for a vitamin A deficiency, in which each simulant is initialized with a "propensity" for vitamin A deficiency, and the simulant's vitamin A status is determined by comparing this
```
collijk: I don't know why we call this propensity except out of habit.  I think it might be clearer just to call it a percentile.

collijk: For use with our standard sampling technique: https://en.wikipedia.org/wiki/Inverse_transform_sampling

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
```diff
@@ -141,44 +149,148 @@ Restrictions
...
+propensity to the overall VAD exposure/prevalence in the population.
+Such
+propensity/exposure models have been used in Vivarium for other risk factors and
```
collijk: They are used for all other risk models.  It is the standard.  They are *infrequently* used for cause models because we usually trust the dynamic disease parameters more or we care about counting cases.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
```diff
@@ -141,44 +149,148 @@ Restrictions
...
+risk-attributable causes, such as child stunting, :ref:`child wasting/PEM
+<2017_cause_pem>`, and :ref:`iron deficiency anemia
+<2017_cause_iron_deficiency>`.
+
+In more detail, the basic strategy is to initialize each simulant with a
+propensity score distributed uniformly in [0,1], then compare this propensity
+score with the (location/age/sex/year/intervention-status)-dependent prevalence
```
collijk: Use exposure throughout.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
```diff
@@ -141,44 +149,148 @@ Restrictions
...
+of vitamin A deficiency at each time step to determine whether the simulant has
+VAD during that time step. Each simulant's propensity is assigned only once, but
+the underlying prevalence distribution can change throughout the course of the
+simulation, which may result in a change in the simulant's vitamin A status.
+
+In particular, our modeling strategy will **not** explicitly use incidence or
+remission data for vitamin A deficiency, but only *prevalence* (which is the
+same as the exposure data for the VAD risk factor). The rationale for this approach is twofold:
+
+1.  We want to guarantee that the simulated baseline prevalence of vitamin A
```
collijk: We are only doing this because of option 2.  We expect the cause version of the model to get incidence and remission correct in addition to getting prevalence correct.

collijk: However, it is more important for us to use the intervention data correctly than it is to get the dynamic parameters of the disease correct.  Details about the limitations and the expected impact to be found in <assumptions and limitations section>.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
```diff
@@ -141,44 +149,148 @@ Restrictions
...
+
+.. todo::
+
+  Verify that effect sizes on VAD should actually be interpreted as described
+  above, and that the prevalence-only model is a good way to accurately
+  represent these numbers.
+
+  Explain why the prevalence-only model is a reasonable strategy, citing
+  incidence, remission, and prevalence data, as well as expert opinions about
+  VAD. (Perhaps this explanation should come later, e.g. in the Assumptions and
```
collijk: Assumptions and limitations is right, I think.  That's where I'd look for the justification.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
```diff
@@ -141,44 +149,148 @@ Restrictions
...
+
+Determining Vitamin A Status
+++++++++++++++++++++++++++++
+
+At each time step, Vivarium needs to determine whether each simulant has vitamin
+A deficiency. To do so, follow these steps:
+
+1.  **Initialize:** When simulant :math:`i` enters the simulation (e.g. at the
+    start of the simulation or at the time step when the simulant is born),
+    assign the simulant a random number :math:`v_i \sim
```
collijk: I think we should standardize the notation for these things.  I follow `scipy` conventions in the code and I think they're reasonable:

x_i - the random variable (the exposure)
q_i - the percentile or propensity of the exposure x_i in the distribution (called q because the inverse cdf is the quantile function, though also called the percent point function).

collijk: Initialize should also describe how we actually get `x_i`.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
```diff
@@ -141,44 +149,148 @@ Restrictions
...
+        :math:`\text{subpop}(i,t)`.
+
+    b)  Look up or compute the prevalence
+        :math:`p_\text{VAD}(\text{subpop}(i,t))` of vitamin A deficiency for the
+        simulant's updated subpopulation.
+
+    c)  If :math:`v_i < p_\text{VAD}(\text{subpop}(i,t))`, the simulant has
+        vitamin A deficiency on the next time step; otherwise, they don't.
+
+To address a point of potential confusion in the above algorithm, note that a
```
collijk: Just leaving notes.  I think this section is great. I think we want to pull it out into the general risk model for the **standard way to sample from categorical distributions**.  In order to do so, we have to generalize this section a bit.  Technically what we do is start with a distribution

p = [p1, p2, ..., pN, 1 - (p1 + ... +pn)]

We then take the cumulative sum over the distribution

p_cum = [p1, p1 + p2, ..., p1 + ...+ pN, 1]

to form the right bounds of a partition of the interval [0, 1] with each subinterval mapping to a risk exposure category (and associated probability).  

We then select the category j by  arg_max_j (q_i < p_cum[j]) and assign that category as x_i.

This means the interpretation of *propensity* is dependent on the sort order of the categories.

collijk: The default sort order is worst to best.

collijk: w/r/t risk effect.  We have not had to deal with unordered categorical risks.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
NathanielBlairStahn: What do you mean?  Do you mean I should call this the "(location/age/sex/year/intervention-status)-dependent exposure distribution," which in this case happens to represent prevalence of VAD?

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
NathanielBlairStahn: Ugh, it's not a percentile -- it's the inverse of the percentile/quantile under the quantile function, i.e. the probability of lying below the corresponding quantile.  I can't find a standard terminology for this. Perhaps "continuous [rank statistic](https://en.wikipedia.org/wiki/Ranking#Ranking_in_statistics)"?

See https://en.wikipedia.org/wiki/Quantile and https://en.wikipedia.org/wiki/Quantile_function

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
NathanielBlairStahn: With regard to (1), there's also the concern mentioned in @aflaxman 's comment above: If we use the incidence and remission data from GBD, then basically the whole population will end up getting VAD if we run the simulation for 5 years, but we suspect this would not be representative of reality.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
NathanielBlairStahn: @collijk In this case `x_i` would be a binary variable, "has VAD"/"does not have VAD", correct? How would you initialize this before following the procedure in the "Update" step? I assumed the "update" part would first happen during the same time step as initialization.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
collijk: Yeah, I think in a risk-based model, we should explicitly use exposure since the disease prevalence is derived from it.  As opposed to a PAF of 1 model of VitA in which the disease model was the source of truth and the risk exposure was derived from prevalence.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
collijk: Man, I've been thinking about percentile wrong for a while.  My stats professors would be ashamed.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
collijk: The procedure is the same as in update, but it has to happen before a time step takes place.  The value of other attributes the simulant is initialized with (e.g. whether or not they are receiving vitamin a fortification) may be dependent on their initial vitamin a deficiency status.  This is a fencepost error (https://en.wikipedia.org/wiki/Off-by-one_error#Fencepost_error).  All attributes of a simulant must be assigned an initial value before the first time step starts or you introduce extremely hairy issues into the order you must update simulant state each time step.

Here's the procedure:

1. Initialization:
 - Sample propensity and store it for all time.
 - Use attributes initial distribution is conditional on (age, sex, year) to construct probability mass function
 - Use propensity to sample pmf and assign initial status.
2. Update:
 - Your procedure looks good here.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
collijk: There really isn't a name for this.  Huh.

NathanielBlairStahn: Aha! Apparently it's called the [percentile rank](https://en.wikipedia.org/wiki/Percentile_rank) (or presumably quantile rank if we're talking about values in [0,1] instead of [0,100]).

So if Q is the quantile function and p is a probability, then q = Q(p) is a quantile with quantile rank p, i.e. a p-quantile. Or in reverse, if F is the CDF and q is an arbitrary value of the random variable, then the probability p = F(q) is the quantile rank of q, so q is a p-quantile. (I'm ignoring some subtleties about nonuniqueness when F is discontinuous or has constant pieces.)

I've been confused about this too.  But it's the statistics professors' fault!  For not laying out the definitions clearly.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
NathanielBlairStahn: Ok, this brings up another point I wanted to check in about: Does this mean the VAD documentation should **not** include a cause model diagram because the underlying algorithm is different from a standard cause model?  Or is a cause model diagram still useful?

I think a simple SIS diagram actually still makes sense in terms of representing what states a simulant can be in (I = has VAD / S = does not have VAD) and what transitions are possible (incidence = simulant develops VAD due to change in exposure distribution / remission = simulant recovers from VAD due to change in exposure distribution).

Is it useful to have a diagram to visualize this, or would that be confusing because it's not a rate-based model?

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
NathanielBlairStahn: Adding a todo to address this.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
NathanielBlairStahn: Adding a note about this.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
NathanielBlairStahn: Adding a todo to reformat the RR table.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
NathanielBlairStahn: Adding a todo to reformat the RR table.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
NathanielBlairStahn: Adding a todo to reword this.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
NathanielBlairStahn: Added some wording and a todo to reword everything to ensure consistent terminology.

I guess I was thinking about this from the GBD perspective - my impression is that they ran a DisMod model to get the cause data, and just copied the prevalence data over to the risk factor to get exposure.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
NathanielBlairStahn: I added a todo to clarify this.

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
collijk: I think a diagram with risk categories that map into disease states makes sense.  I would hesitate to put transitions on the diagram (or if we put them, we need to be super clear that they're implicit and based on resampling risk exposure).

docs/source/gbd2017_models/risk_attributable_causes/vitamin_a_deficiency/index.rst
collijk: Makes sense.  As long as your consistent, it seems fine either way.

PR #153, opened 2020-03-06
Modeling Risk Factors: Start TMREL/D section

NathanielBlairStahn:
Note: All the content I have added is in the TMERL/D section and the new References section. The other changes that git shows are due to my editor automatically removing whitespace.


docs/source/model_design/risk_factors.rst
```diff
@@ -38,11 +38,11 @@ What is a risk factor?
   Update this section once this page has been filled in further.
 
 .. todo::
-
```
collijk: Protip for other reviewers.  Click the gear icon at the top and select the box to hide whitespace changes.  Nathaniel uses a good editor that strips erroneous whitespace by default and so his commits have a bunch of changes that aren't content changes. 
image

docs/source/model_design/risk_factors.rst
```diff
@@ -134,8 +134,131 @@ Definitions
...
+exposure distribution** used for the causal attribution of disease burden to a
+given risk factor in a population (see `Population Attributable Fraction
+(PAF)`_). Other choices of counterfactual include the *plausible* minimum risk,
+*feasible* minimum risk, and *cost-effective* minimum risk, each of which can
+obviously depend on specific attributes of the population under consideration.
+On the other hand, Murray et al. state
+[Comparative-quantification-health-risks]_:
+
+  Biological principles as well as considerations of equity would necessitate
+  that, **although the exposure distribution for theoretical minimum risk may
```
collijk: I want to diagram this sentence.

docs/source/model_design/risk_factors.rst
```diff
@@ -134,8 +134,131 @@ Definitions
...
+  depend on age and sex, it should in general be independent of geographical
+  region or population.**
+
+However, they go on to add:
+
+  Exceptions to this are however unavoidable. An example would be the case of
+  alcohol consumption, which in limited quantities and certain patterns, has
+  beneficial effects on cardiovascular mortality, but is always harmful for
```
collijk: I thought they changed this claim: http://www.healthdata.org/news-release/new-scientific-study-no-safe-level-alcohol

docs/source/model_design/risk_factors.rst
```diff
@@ -134,8 +134,131 @@ Definitions
...
+  other diseases such as cancers and accidents. In this case, the composition of
+  the causes of death as well as drinking patterns in a region would determine
+  the theoretical minimum distribution.
+
+.. todo::
+
+  Is there an updated/better example of TMRED depending on population, since the
+  latest research says that that there is no safe amount of alcohol?
```
collijk: Seems you caught that.  Yes.  Iron deficiency.

collijk: I think BMI is protective against some cancers as well.  You could make an argument that the TMRED/TMREL should be defined separately for different risk outcome pairs.  In fact, I think it's important to point out specifically that the TMREL is the level at which risk of **all** outcomes is minimized.

NathanielBlairStahn: Do you know of a reference that discusses iron deficiency?

What about a reference that says that TMREL minimizes risk for all outcomes?

collijk: For BMI, it's mostly a thing I've heard in discussions with research people, though some literature talks about this: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5830139/

The relationship is pretty complicated though cause there's reverse causality at work (e.g. chemo makes you lose weight).  Kiran has been working on the obesity washout stuff which is directly related and might have more info.

collijk: Iron deficiency specifically has draw level TMREDs that vary by location.  I don't know what that means, but I think Ali looked into it some.

collijk: I'm inferring the global nature of the TMREL from its operational definition.  The GBD only defines one TMREL/TMRED for the risk factor without consideration to the affected outcomes and apply it to all risk-outcome pairs in paf calculation.

NathanielBlairStahn:
More to say about counterfactual: We took this population in particular, and set this one risk to the TMRED, but we left everything else about that population the same. As opposed to a comparison analysis, where we look at another population and they happened to all be the TMRED.

Highlight why these are different. Most important point: Other underlying risk factors remain the same, vs. comparing to a different population in reality where other risk factors would be different as well.

What about mediation? For obesity, if everyone were in the TMRED, then blood pressure would also be different because SBP mediates BMI.

BMI-->SBP-->mortality
BMI--------->mortality

Adjustment is risk exposure and all causally related outcomes.

So the population in the counterfactual (TMRED) for BMI would have both different BMI and different SBP, because of the mediation. But GBD does not tell us how to figure out what the different distribution of SBP would be. Instead, this difference is captured in the change in relative risk to mortality (the long arrow above).

Notes:
- Someday, we should do the work to figure out how to incorporate this in our models; until then steer clients away from this point.
- This mediation relationship is distinct from correlation.

NathanielBlairStahn:
Would it be worth it to mention other approaches?
- No adverse outcome exposure level (NAOEL)
- LOWEL
Industrial hygiene and dose-response toxicology literature, evidence-based approach to minimum risk level. Only possible in dose-response situations. Often show dose-response trend which can be helpful for visualization.

NathanielBlairStahn:
Note: Other people who do cost-effectiveness analysis do so by choosing different distributions other than TMRED, and compare actual population to what would happen if we shifted it to this other distribution.

NathanielBlairStahn:
Possible examples of TMRED depending on population

Hb level: people living at higher altitudes vs. at sea level 

Reverse causality for iron deficiency - you can give people iron, but that won't cure all the anemia...
Or not reverse causality but just...
GBD says risk is iron deficiency, but exposure distribution is Hb, so they define TMRED in terms of the risk, i.e. what amount of the burden of iron deficiency can they address by shifting Hb levels ????

difference between definition of risk and def. of exposure, issue is units

Doesn't make sense to James that there's only one TMRED for each risk factor, rather than one for each risk-outcome pair.

Note: Framing in GBD (and Vivarium) for TMRED assumes risk is monotonic. But this is not always true, e.g. high BMI is bad, but so is low BMI. This is about how we compute relative risks as a function of exposure, and about how we compute PAFs using that definition.

Another example would be sodium - e.g. we need a certain amount, but too much is bad.

docs/source/model_design/risk_factors.rst
```diff
@@ -134,8 +134,131 @@ Definitions
...
+[Comparative-quantification-health-risks]_:
+
+  Biological principles as well as considerations of equity would necessitate
+  that, **although the exposure distribution for theoretical minimum risk may
+  depend on age and sex, it should in general be independent of geographical
+  region or population.**
+
+However, they go on to add:
+
+  Exceptions to this are however unavoidable. An example would be the case of
```
beatrixkh: I'm a little confused by this.

Would the exceptions be due to biological differences that are region-specific, or feasible differences that are geographically specific?

For example, maybe going for a 30 min jog outside every day is good for your heart and respiratory health, unless there are wildfires and SMOG, in which case it ends up being bad for you?

So here the TMRED would be geographically dependent, but any given population, if transplanted to the same geography, would presumably have the same TMRED

Which is different from, say, if there was some specific population with a biological immunity to skin cancer, which then maybe has a different TMRED for sun exposure than the rest of the world?

docs/source/model_design/risk_factors.rst
```diff
@@ -134,8 +134,131 @@ Definitions
...
+    <https://en.wikipedia.org/wiki/Becquerel>`_/m\ :superscript:`3`, which is
+    equivalent to the average outdoor concentration of radon [ICRP]_.
+2.  The :ref:`Low Birth Weight and Short Gestation <2017_risk_lbwsg>` risk
+    factor has multiple TMREL categories since healthy babies have many
+    different birth weights and gestational ages.
+
+These examples illustrate some complexities in defining the TMREL and TMRED for
+a given risk factor. For continuous risk exposure variables --- such as radon
+exposure, or hemoglobin concentration, or systolic blood pressure --- it may be
+impossible to define a single TMREL for the population, as we expect different
```
beatrixkh: Is there anything in the literature about distributions of TMREDs (or TMRELs for an individual?)

Ninicorn (approved):

PR #159, opened 2020-03-09
Updated CKD data tables, cause model diagram, limitations.

SantoniS:
Updated:
- CKD data table
- cause model diagram

To-do: 
- update some of the tasks related to disease state and IKF cat mapping (if needed?)
- assumptions


docs/source/gbd2017_models/risk_attributable_causes/chronic_kidney_disease/index.rst
```diff
@@ -128,7 +128,16 @@ The aspects of the disease this cause model is designed to simulate is the basic
...
+Initialization:
+- In this model, simulants are initialized as "susceptible" or "with specific sequelae-level condition" through the following process: simulants will be assigned directly to a CKD sequelae ("with condition" state) based on each sequelae prevalence. Those not assigned to a sequelae will be initialized to the "susceptible" state. Each sequelae will then be mapped back to the distribution of IKF based on sequelae based severity splits. The result will be an IKF value for each simulant that is consistent with sub-cause prevalence. This structure means there's no impact of SBP on CKD.
+
+Progress:
+- In this model, simulants will not transition into later states over time. Simulants will stay in the disease states they were initialized in.
```
davidbernotas: Technically, simulants may not stay in the disease state into which they were initialized.
"As simulants age, their risk exposure will change, which may result in them progressing into a different disease state over time.  Also, simulants will experience mortality based on their risk exposure."

docs/source/gbd2017_models/risk_attributable_causes/chronic_kidney_disease/index.rst
```diff
@@ -166,7 +175,11 @@ Mapping CKD States to IKF Categories in Vivarium
 Assumptions and Limitations
 +++++++++++++++++++++++++++
 
-Presently, we are using prevalence for each stage of CKD to assign the each person in the population a CKD severity state. We are assuming (for now) that there is no transition between states. As a result, we should expect the prevalence for early stage CKD to swell as the simulation runs, since there is nowhere for these new incident cases to go. If we had remission (which is actually the term used for "progression") rates, we could use these as inter-stage incidence rates.
+- Presently, we are using prevalence for each stage of CKD to assign the each person in the population a CKD severity state. We are assuming (for now) that there is no transition between states. As a result, we should expect the prevalence for early stage CKD to swell as the simulation runs, since there is nowhere for these new incident cases to go. If we had remission (which is actually the term used for "progression") rates, we could use these as inter-stage incidence rates.
```
davidbernotas: Transition rates (progression rates) between states are not available from the GBD model.  As such, we are using evolution of risk exposure over time (changes with simulant age) to proxy for progression between CKD states - as a simulant ages, they may move to a different part of the IDF distribution, thereby landing them in a more advanced CKD state.
The reason we are modeling CKD this way is because it is a condition for treatment of LDL-C, which is the intervention in this model.  Thus, we need to get the prevalence at each severity (mild/moderate v. severe) correct.  CKD is not a cause of interest in this project, so the severity specific prevalence is our priority.

SantoniS:
@davidbernotas  I added your updates. Do you believe this is good to be merged to master? If so, please let me know by 'approving' this PR.

docs/source/gbd2017_models/risk_attributable_causes/chronic_kidney_disease/index.rst
```diff
@@ -134,11 +134,6 @@ Initialization:
 Progress:
 -As simulants age, their risk exposure will change, which may result in them progressing into a different disease state over time. Also, simulants will experience mortality based on their risk exposure.
 
-.. todo::
-Confirm with the RT and SE team how this model affects other models.
-
-- This model affects other models – disability weights, excess mortality weight for each state, cause specific mortality rate. 
-
```
davidbernotas: Since the treatment for LDL-C does not impact IKF, the relevant disabilities for this project will not be affected.  However, disability due to CKD will be affected by the methodology described above.  If simulants progress into subsequently worse CKD states too slowly, there may be a reduction of disability for the population (caveat: it is possible that slowing the progression into worse states may lessen the disability if said progression results in faster death).  Punchline: if morbidity is lower than GBD, we should expect mortality to be higher.
If mortality from CKD is elevated due to these assumptions, we should see a correlated reduction in IHD and IS deaths since the simulants will die from CKD instead of the CV causes.

docs/source/gbd2017_models/risk_attributable_causes/chronic_kidney_disease/index.rst
```diff
@@ -178,11 +173,16 @@ Assumptions and Limitations
...
+- Presently, we are using prevalence for each stage of CKD to assign the each person in the population a CKD severity state. We are assuming (for now) that there is no transition between states. As a result, we should expect the prevalence for early stage CKD to swell as the simulation runs, since there is nowhere for these new incident cases to go. Transition rates (progression rates) between states are not available from the GBD model. As such, we are using evolution of risk exposure over time (changes with simulant age) to proxy for progression between CKD states - as a simulant ages, they may move to a different part of the IDF distribution, thereby landing them in a more advanced CKD state. The reason we are modeling CKD this way is because it is a condition for treatment of LDL-C, which is the intervention in this model. Thus, we need to get the prevalence at each severity (mild/moderate v. severe) correct. CKD is not a cause of interest in the current project it is being modeled in, so the severity specific prevalence is the current priority.
```
davidbernotas: I don't think this is true anymore - there is transition to subsequent disease states, and is based on IKF risk exposure.

docs/source/gbd2017_models/risk_attributable_causes/chronic_kidney_disease/index.rst
```diff
@@ -178,11 +173,16 @@ Assumptions and Limitations
...
 
 - Simulants are in each disease state longer than they should be, compared to GBD 2017. 
 
-- This model effects other models – disability weights, excess mortality weight for each state, cause specific mortality rate
+- This model assumes there is no impact of SBP nor FPG on CKD.
+
+Limitations
++++++++++++
+
+- This model is consistent with prevalence in population. The following relationships between CKD/SBP and CKD/FPG will be modeled using correlation. The iniitial distribution will be correct, but will change over time and become inaccurate due to mitigating factors.
```
davidbernotas: Also, given that prevalence is our priority in the larger model (since CKD status impacts treatment of high LDL-C patients), we are willing to sacrifice other precision here to ensure prevalence is correct.  That the simulaiton is only 5 years long should minimize the variance in prevalence when compared to GBD.

PR #170, opened 2020-03-18
LSFF #2: coverage algorithm

NathanielBlairStahn:
Sequel to https://github.com/ihmeuw/vivarium_research/pull/169 (i.e. read that first)

PR #169, opened 2020-03-18
LSFF #1: add population coverage data for folic acid

NathanielBlairStahn:

PR #171, opened 2020-03-18
LSFF #3: determining effect of folic acid fortification

NathanielBlairStahn:
Sequel to 
https://github.com/ihmeuw/vivarium_research/pull/170 (read second), and
https://github.com/ihmeuw/vivarium_research/pull/169 (read first)

### DIFF
--- a/docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
+++ b/docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
@@ -107,6 +107,10 @@ Folic Acid Fortification
 Population Coverage Data
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
+.. todo::
+
+  Define a, b, and c prior to displaying their values in the table below.
+
 The following table provides data for three parameters, :math:`a`, :math:`b`,
 and :math:`c`, that will be used in our `coverage algorithm`_ for the folic acid
 intervention. The food vehicle referred to in the table is wheat flour.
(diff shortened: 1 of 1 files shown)


docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -156,13 +160,13 @@ Exchange <https://tinyurl.com/rdm4wza>`_.
...
 
-For all the locations other than Ethiopia, use a Beta distribution with mean
+**For all the locations other than Ethiopia**, use a Beta distribution with mean
 equal to the central estimate, and variance equal to the variance of a normal
 distribution with the same mean and 95% confidence interval.
 
 .. todo::
 
-  Describe how to get the correct Beta distribution.
+  Show how to get the correct Beta distribution. Maybe even include some graphs.
```
aflaxman: 👍 for including graphs

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -236,12 +241,54 @@ In words, our intervention algorithm does the following:
...
+
+1.  **Define variables:** Each simulant needs a boolean attribute
+    `mother_ate_folate_fortified_food`, which will be `True` if the simulant's
+    mother ate folic-acid-fortified food starting at least a year before the
+    simulant was born, and `False` otherwise.
+
+2.  **Initialize the simulation:** At the start of the simulation, set
+    `mother_ate_folate_fortified_food = False` for all simulants; this attribute
+    will not be used after the simulant is born, so the value is irrelevant for
+    all simulants already alive at the start of the simulation.
```
aflaxman: Ah, I think I realized why James wanted to get this part right... although this value is irrelevant to DALYs averted, setting it to False for all will result in an *over-estimate* of NTD DALYs in the baseline scenario.  The magnitude of this over-estimate will depend on the baseline coverage level, and we must remember that we have it when we do our standard validations or if we decide to calculate percent of DALYs averted.

collijk: Setting it to false will not produce an over estimate.  The effect is on `birth_prevalence`, whereas sims initialized will use `prevalence`.  We should still be able to properly validate NTDs in the baseline and new cases coming in should properly be affected by the baseline coverage of folic acid fortification.  The limitation here is that we will simply not know the status of the mother for existing sims.  We can be explicit (and I will in the implementation) by saying `mother_ate_folate_fortified_food` is in the set {`True`, `False`, `Unknown`}.

NathanielBlairStahn: @collijk Sounds good. "Unknown" is more accurate than arbitrarily assigning True or False. Technical question: How would you implement this variable as a 3-value data type?  Do you just use strings?

@aflaxman Does James's explanation here sound accurate, i.e. that we won't overestimate DALYs in baseline?

It sounds like there are some issues here that should go in the limitations section.  Is there anything relevant here that should also be described in the algorithm?

collijk: Yep, just use strings.

PR #172, opened 2020-03-18
Vitamin A effect size

alibow:
So far, this is mostly getting our assumptions/references on the research side written down. I will be adding additional info about the vivarium modeling strategy regarding age-dependent effects, sampling distribution, and intervention effect lag in subsequent PRs.


docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -98,6 +98,183 @@ Interventions
...
+
+While the supplementary meta-analysis shown above contains more studies and 
+data than the Cochrane review, it relies on results that were not directly 
+reported in the individual studies (and in some cases visaully estimated 
+values). **Therefore, we will conservatively use the results from the Cochrane 
+review, with increased certainty in the results based on the confirmatory 
+results from the supplementary meta-analysis.**
+
+Notably, all of these studies included in the supplementary analysis were 
+conducted among children and study locations included Guatemala, South Africa,
```
NathanielBlairStahn: Typo? There are two "included"s in this sentence... I'm not sure exactly what it's trying to say.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -98,6 +98,188 @@ Interventions
...
+
+	Notably, the effect can occur either through the direct consumption of 
+	vitamin A fortified foods or through the consumption of breastmilk from 
+	mothers who consume vitamin A fortified foods (Sandjaja et al. 2015; WHO 
+	Guidelines).
+
+For individuals aged between 0 and six months, we made the following 
+assumptions:
+
+	1. Maternal consumption of vitamin A fortified foods as no effect on
```
NathanielBlairStahn: typo: as -> has

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: I don't think it's a typo but maybe could be more clear. I mean to say that the study populations for each of the studies that we used for the supplementary meta-analysis were children from one of the countries in that list. Is that more clear?

alibow: Just noting that the studies looked specifically at children and listed the countries so that we could see they were largely LMICs, which supports the theory of generalizability to our study populations.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: fixed

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Oh, maybe you're just missing a comma before the and: "children, and study locations..."

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -98,6 +98,188 @@ Interventions
...
+
+	3. The scale-up in effect size from birth to six months of age occurs in a 
+	linear fashion. We made this assumption in the absence of supporting data.
+
+.. todo::
+
+	Add citations for all of these studies/references.
+
+	Add descriptions of strategy and assumptions regarding differences between baseline and intervention coverage (intervention effect lag, non-age-dependent, SOURCES for these)
+
```
alibow: This is an outstanding to-do on my part. Will update tomorrow!

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -98,6 +98,188 @@ Interventions
...
+In this model, the vitamin A fortification intervention affects the 
+**prevalence of vitamin A deficiency**. The effect size for this intervention 
+was obtained from a Cochrane review performed by Hombali et al. (2019) on the 
+fortification of staple foods with vitamin A for vitamin A deficiency. 
+Notably, the relative risk for vitamin A foritification on vitamin A 
+deficiency from this review only included data from two randomized controlled 
+trials and the authors of the review assessed the certainty of the evidence to 
+be "very low" (Hombali et al. 2019). The relative risk of vitamin A deficiency 
+prevalence among the population exposed to vitamin A fortification relative to 
+the population unexposed to vitamin A fortification from this review was
```
Ninicorn: The relative risk of vitamin A deficiency prevalence among an intervention population exposed to vitamin A fortified staple foods relative to a control population given the same staple foods not fortified with vitamin A....

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -98,6 +98,188 @@ Interventions
...
+	2. There is a scale up in effect size from no effect (RR=1) at birth to 
+	the full effect size (RR=0.45) at six months of age. While there was no 
+	data identified for the effect of vitamin A *fortification* in children 
+	less than six months of age, we found that there were mixed results on the 
+	impact of maternal vitamin A *supplementation* (via breastmilk) on three 
+	month old infants (as discussed by Martins et al. 2010) and based our 
+	assumption on this finding.
+
+	3. The scale-up in effect size from birth to six months of age occurs in a 
+	linear fashion. We made this assumption in the absence of supporting data.
```
NathanielBlairStahn: Technically, I think we're assuming a linear scale-up of the _reciprocal_ of the relative risk stated above.  If that's correct, maybe we should change this to something like "The scale-up in the reciprocal of the effect size from birth to six months of age occurs in a linear fashion," just to be clear.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Right! Updated :)

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Yeah, agreed. I will update this.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Ok. I will update this.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -98,6 +98,188 @@ Interventions
...
+is applied continuously from the start of the simulation to the end of the 
+simulation and only depends on age, such that:
+
+.. code-block:: Python
+
+  if age_i < 0.5:
+  	rr_i = 1 + (rr - 1) * a_i / 0.5
+  else:
+  	rr_i = rr
+
```
NathanielBlairStahn: Two comments:
1. Make sure the variable names are consistent, i.e. should a_i be age_i?
2. It would be good to explicitly say what the variable rr is before you use it. Continuing the code from where the lognormal distribution was defined, you could precede this if/else statement by rr = rr_distribution.rvs() to generate a sample value of the relative risk and store it in the variable rr.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -98,6 +98,188 @@ Interventions
...
+  if coverage_time_i < 0.5:
+  	rr_i = 1 + (rr - 1) * coverage_time_i / 0.5
+  else:
+  	rr_i = rr
+
+.. note::
+
+	Any ideas from the research team on how we might want to combine
+	the impact of age_i < 0.5 and coverage_time_i < 0.5 here?
```
NathanielBlairStahn: Hmm, I'll have to think about this some more. I was assuming we'd so something like: If we have reached p_1% of the age scale-up and p_2% of the coverage time scale-up, then the overall effect will be p_1*p_2% of the total effect (the "percents" should really be proportions, but it's easier to say percent.).

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -98,6 +98,188 @@ Interventions
...
+
+  if age_i < 0.5:
+  	rr_i = 1 + (rr - 1) * a_i / 0.5
+  else:
+  	rr_i = rr
+
+For simulants covered by the *intervention scale-up*, the effect of the 
+intervention will scale up linearly from zero at the start of coverage to the 
+full effect after six months of coverage and does *not* depend on age (
+research assumption), such that:
```
NathanielBlairStahn: Why are we making this assumption? Is this just because we don't know how to combine the effects, as mentioned below, or is this based on research?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -98,6 +98,188 @@ Interventions
...
+
+Where,
+
+	- **age_i** = age of simulant in years
+
+	- **coverage_time_i** = time since start of intervention coverage for a simulant in years
+
+	- **rr_i** = relative risk to be applied to an individual simulant
+
+	- **rr** = relative risk for intervention, sampled as described above based on a random number between 0 and 1, assigned to a simulant *for the entirety of the simulation*
```
NathanielBlairStahn: I'm confused by this description:

- What is "sampled as described above based on a random number between 0 and 1" referring to?
- I thought rr_i was what gets assigned to the simulant, and rr would be a global parameter, not a property of a simulant. Do I have the wrong interpretation here?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: updated locally

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Yeah, I was thinking this too, but then we get some weird cases on the margins
. I will draw it out and share it in the LSFF slack channel.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: It's not based on research. Rather because we do not have data to guide us and this sis a simplistic implementation.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: I clarified this in the latest commit! Let me know if it still doesn't make sense.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -98,6 +98,224 @@ Interventions
...
+
+For simulants covered by *baseline coverage*, the effect of the 
+vitamin A fortitication is determined as follows:
+
+.. code-block:: Python
+
+  if age_i < 0.5:
+  	rr_i = 1
+  else:
+  	rr_i = rr_distribution.rvs()
```
NathanielBlairStahn: Aha, I see why I was confused by your description of rr - I think we are thinking about it differently.

The way this code is written, each simulant will get a different relative risk sampled from the lognormal distribution above. While this seems like a perfectly reasonable strategy, it is *not* what the lognormal distribution is intended to capture.

In particular, the lognormal distribution for the RR is intended to capture the *parameter uncertainty* of our estimate. That is, we imagine that there is one true value of the RR in the real world, but all these different studies measured different values that got combined in a meta-analysis, and we want to model the uncertainty in how far away we think the real value is from the average value reported by the meta-analysis. When we run the simulations, each draw should sample the lognormal distribution _once_ and use that value of the RR as the universal globally true value of the RR in the world simulated by that draw. (It is probably worth spending a little more time thinking about the theory of the statistics involved in parameter estimation to confirm that a lognormal distribution is in fact the appropriate thing to use here.)

On the other hand, it looks like you are using the lognormal distribution to model *heterogeneity between simulants*, i.e. each simulant will get a different random value for their relative risk. Usually our models have very little individual heterogeneity between simulants, because we don't have enough data to know what to do...

----
Ok, we just discussed this with James, but I'll go ahead and post this comment anyway.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -98,6 +98,224 @@ Interventions
...
+	distribution of response times follows a lognormal distribution with a 
+	median value of five months, a 0.025 percentile of 2 months, and a 0.975 
+	percentile of 12 months.
+
+ 	2. If an individual was covered by baseline coverage of vitamin A 
+ 	fotification, we assumed that the individual was covered (via breastmilk 
+ 	or direct consumption) for long engough to exhibit a response (at least 12 
+ 	months).
+
+**Vivarium Modeling Strategy**
```
NathanielBlairStahn: I think this section is too broad in scope. The entire "Vitamin A Fortification" section is describing the Vivarium modeling strategy. (This is different from cause models where there is a separate GBD modeling strategy.)

I think it would helpful to break up our strategy into discrete chunks and make a separate section and a separate PR for each of those, similar to what I did for folic acid, e.g.

- coverage data
- effect size
- intervention scale-up
- effect of the intervention on simulants
- research considerations

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -98,6 +98,224 @@ Interventions
...
+  rr_distribution = lognorm(s=sigma, scale=median)
+
+.. note::
+
+	I copied this from Nathaniel's documentation for the folic acid RR, but I 
+	think that the same approach is appropriate. Perhaps we can eventually 
+	create a separate page that lists similar strategies that we can reference 
+	via links.
+
+Further, the time-to-response to vitamin A fortification in years should also
```
NathanielBlairStahn: I think this is worth discussing more.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Yes, agreed to all of this. So I'll rework it to:

# draw level
rr = rr_distribution.rvs()

# individual level
 if age_i < 0.5:
  	rr_i = 1
  else:
  	rr_i = rr

Is this more clear/accurate?

PR #179, opened 2020-03-23
Iron effect on birth weight

alibow:


docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -446,9 +444,123 @@ Where,
...
+.. note::
+
+	This distirbution is slightly shifted to the right (0.025 percentile is 
+	equal to -0.1, rather than -0.2) due to rounding approximations in the 
+	meta-analysis for the effect size causing a slightly non-symmetrical 
+	confidence interval around the mean.
+
+**Birth Weight**
+
+The effect of maternal consumption of iron fortified food on infant birth weight was obtained directly from Haider et al. (2013). According to this data source, mean birth weight among babies born to mothers who consumed iron fortified foods had a mean birth weight **15.1 grams (95% CI: 6.0, 24.2)** higher than babies born to mothers who did not consume iron fortified foods.
```
NathanielBlairStahn: This is not quite right.  The Haider number was a dose-response: +15.1 g per additional 10 mg iron the mother gets.  To get the actual effect size for our model, we need to estimate how much additional iron women are getting from fortified food.

NathanielBlairStahn:
@alibow Is the only new part here the **Birth Weight** section? Should the base branch be changed to `iron_effects` so that only the differences with that branch show up?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Oh, my bad, I copied from the word doc concept model so thanks for checking. Have we already looked into how to do this or is this an outstanding task? @Ninicorn

alibow:
> @alibow Is the only new part here the **Birth Weight** section? Should the base branch be changed to `iron_effects` so that only the differences with that branch show up?

Yes. I accidentally did this incorrectly at first and now I'm not sure how to change it.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Ok, I just pushed a commit that proposes a strategy to handle this, although there are still a couple of outstanding todos within the documentation.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -525,6 +525,123 @@ as follows:
...
+	operated under the assumption that 10 mg of daily iron consumption, as 
+	referenced in the Haider paper, represented 10 mg of *elemental* iron.
+
+Therefore, we investigated the amount of daily iron consumption that a 
+pregnant mother would likely consume through iron fortification of staple 
+foods in our locations of interest in order to scale the effect size 
+accordingly. See the table below:
+
+.. list-table:: Daily Iron Consumption Parameters
+  :widths: 5 5 5ight
```
NathanielBlairStahn: This line causes a syntax error

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -525,6 +525,123 @@ as follows:
...
+	NaFeEDTA for the locations in this table, we conservatively chose to  use 
+	the concentrations of NaFeEDTA because elemental iron is uesd at a *lower* 
+	concentration in flour when NaFeEDTA is used than for other forticants.
+
+[1] https://tinyurl.com/wka9mgh
+
+[2] https://www.ephi.gov.et/images/pictures/National%20Food%20Consumption%20Survey%20Report_Ethiopia.pdf (table 18; women)
+
+* In the absence of better data, we assumed that individuals in Nigeria and 
+India consumed the same amount of fortifiable flour per day as individuals in
```
NathanielBlairStahn: I get a syntax error here:

/Users/ndbs/repos/vivarium_research/docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst:596: WARNING: Bullet list ends without a blank line; unexpected unindent.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -525,6 +525,123 @@ as follows:
...
+
+	# bw_md_per_10_mg_iron : defined above (random sample from effect size distribution)
+	# loc_specific_elemental_iron_consumed_per_day : defined above (location-specific iron consumed per day in mg)
+
+	location_specific_bw_shift = loc_specific_elemental_iron_consumed_per_day * bw_md_per_10_mg_iron / 10
+
+See the following section to see if/how to apply the 
+*location_specific_bw_shift* parameter to individual simulants.
+
 Determining Whether A Simulant is Affected
```
NathanielBlairStahn: Since we're duplicating section heading titles, it looks like we now have to be more careful with our references to sections:

/Users/ndbs/repos/vivarium_research/docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst:837: WARNING: Indirect hyperlink target "reduces the birth prevalence"  refers to target "effect size", which is a duplicate, and cannot be used as a unique reference.
/Users/ndbs/repos/vivarium_research/docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst:837: WARNING: Duplicate target name, cannot be used as a unique reference: "effect size".
/Users/ndbs/repos/vivarium_research/docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst:942: WARNING: Duplicate target name, cannot be used as a unique reference: "determining whether a simulant is affected".

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -525,6 +525,123 @@ as follows:
...
+
+	Find better data and replace values for flour consumption per day for 
+	Nigeria and India.
+
+The amount of elemental iron consumed daily, in miligrams per person, (at the 
+draw level) should be calculated as such:
+
+.. math::
+
+	\frac{\text{X mg NaFeEDTA}}{\text{kg flour}} * \frac{\text{Y g flour consumed daily}}{\text{person}} * \frac{\text{1 kg flour}}{\text{1,000 g flour}} * \frac{\text{55.845 g elemental iron}}{\text{367.047 g NaFeEDTA}} = \frac{\text{Z mg elemental iron consumed daily}}{\text{person}}
```
NathanielBlairStahn: This equation is too long and gets cut off... there may be some options for making it a multi-line equation.

Also, not a big deal, but you can use \times instead of * for multiplication.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -525,6 +525,123 @@ as follows:
...
+  :header-rows: 1
+
+  * - Location
+    - Concentration of forticant in flour
+    - Amount of fortifiable flour consumed daily
+  * - Ethiopia
+    - 30 mg NaFeEDTA / kg flour  [1]
+    - 100 g (IQR: 77.5, 200) [2]
+  * - India
+    - 14 to 21.25 mg NaFeEDTA / kg flour [1]; assume midpoint value of **17.6**
```
NathanielBlairStahn: We could model this with a distribution - maybe a symmetric triangular distribution on the interval?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: If you're going to use variable names in this equation (X, Y, Z), I would define them earlier, and use the same variable names in the names of the table columns above so it's clear which numbers correspond to what quantity.  It took me a couple times reading through for me to figure out how to match things up.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -525,6 +525,123 @@ as follows:
...
+.. note::
+
+	While there was data for the concentration of iron forticants other than 
+	NaFeEDTA for the locations in this table, we conservatively chose to  use 
+	the concentrations of NaFeEDTA because elemental iron is uesd at a *lower* 
+	concentration in flour when NaFeEDTA is used than for other forticants.
+
+[1] https://tinyurl.com/wka9mgh
+
+[2] https://www.ephi.gov.et/images/pictures/National%20Food%20Consumption%20Survey%20Report_Ethiopia.pdf (table 18; women)
```
NathanielBlairStahn: It would be better to display the title of the reference, and make the text clickable to these links. (The second url is super long and goes off the edge of the page.)

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -525,6 +525,123 @@ as follows:
...
+The amount of elemental iron consumed daily, in miligrams per person, (at the 
+draw level) should be calculated as such:
+
+.. math::
+
+	\frac{\text{X mg NaFeEDTA}}{\text{kg flour}} * \frac{\text{Y g flour consumed daily}}{\text{person}} * \frac{\text{1 kg flour}}{\text{1,000 g flour}} * \frac{\text{55.845 g elemental iron}}{\text{367.047 g NaFeEDTA}} = \frac{\text{Z mg elemental iron consumed daily}}{\text{person}}
+
+.. code-block:: Python
+
+	## WRITE CODE THAT SAMPLES FROM AMOUNT OF FLOUR CONSUMED PER DAY (BETA DISTRIBUTION??)
```
NathanielBlairStahn: I will think about an appropriate distribution to use here. Do we have some idea of what should be the _maximum_ amount of daily flour consumed?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Or maybe just a uniform distribution?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: We only have 25th, 50th, and 75th percentiles.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: I think I fixed this. I have been struggling to get the "make html" function to work on my loaner computer, but I will try harder to get this working and I apologize for my syntax errors in the meantime.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Fixed

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: I think I fixed this. I am having trouble getting the "make html" function to work on my loaner computer, but I will try harder to get it running and I apologize for my syntax errors in the meantime.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: I made the iron headings specific to iron

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: I got rid of the final = statement. Like I said I am having trouble rendering, so I'm not sure if this fixed it but I will revisit sometime soon.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Sure. Do you think it is worth it?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Done.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Done

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: I think we have a couple options here.

The data we have says that there is a range of fortification levels in India, so either:

1. We could use a single number for the mean -- we could either just say this is the midpoint, or we could use e.g. a beta distribution or triangular distribution to model parameter uncertainty, since we're not sure how much fortification happens at each level so we don't really know where the mean is,
2. OR we could we could actually model the range of fortification levels among different households, and use e.g. a uniform distribution to randomly assign a fortification level to each simulant who gets fortified food.

What do you think?

alibow: I think that we should model it as a population parameter, not an individual parameter since the fortification at a certain concentration is presumably happening at an industrial level.

I was just thinking that because we do not know anything about the underlying distribution that we might want to avoid guessing at it, but if we want to prioritize including the uncertainty in our model, then I would say that option 1 with a uniform distribution would make the most sense to me.

alibow: Assuming a triangular or beta distribution seems odd to me because for all we know, all of the industries might fortify at the low range of the recommended range (to save money) or the high end (maximize impact). Which is why I would prefer a uniform distribution, I think.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: The situation here is similar to what I wrote above about our data for fortification levels in India: This data is telling us about the heterogeneity between households in Ethiopia.

In this case, however, it is unclear to me how to pull out a single parameter to use for the model. E.g. do we want the mean or the median? And if we want the mean, how do we estimate it?

Therefore, I think we should use the data to directly estimate the distribution of flour consumption among the population, i.e. to model individual heterogeneity rather than just using a single parameter with uncertainty.

The simplest way I can think of to estimate the distribution is to use a mixture of uniform distributions supported between the quartiles (this is also the [maximum entropy](https://en.wikipedia.org/wiki/Maximum_entropy_probability_distribution) strategy for choosing the least informative distribution subject to the constraints). That is, for simulants who eat fortified flour, assign their daily wheat flour consumption using the following procedure:

- Sample 25% of the simulants uniformly between 0 and Q1=77.5
- Sample 25% of the simulants uniformly between Q1=77.5 and Q2=100
- Sample 25% of the simulants uniformly between Q2=100 and Q3=200
- Sample 25% of the simulants to have consumption >Q3=200. Either we can choose some maximum (maybe Q4=500 g?) and sample uniformly on [Q3,Q4], or we can choose some distribution supported on [Q3, infinity) and sample from that (e.g. a shifted [exponential distribution](https://en.wikipedia.org/wiki/Exponential_distribution) or [Pareto distribution](https://en.wikipedia.org/wiki/Pareto_distribution), which are also maximum entropy distributions subject to certain constraints).

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Well, if we just choose a single number, that **is** still guessing at the underlying distribution - it's saying that we're 100% sure it's the number we pick. Speaking of which, do we just want to use single numbers for Nigeria and Ethiopia, or should we model uncertainty in those?

I'm not sure I see the argument for using a single parameter rather than assigning different levels to different simulants, other than that it's computationally cheaper. The first option (single parameter) says that everyone gets exactly the same concentration of iron in their fortified food. The second option (different levels to different simulants) says that different factories are fortifying at different levels in the specified range, and each person will get an amount of iron that depends on which factory they get their food from.  To me the second option seems closer to reality.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Ok. Notably, the reported figure was the amount of fortifiable wheat eaten among those who reported that they ate fortifiable wheat, so the lower bound should be > 0. What if we took the minimum 25th percentile and maximum of the 75th percentile and used those as the upper and lower bounds? 
image

So...
Sample 25% of the simulants uniformly between **50** and Q1=77.5
Sample 25% of the simulants uniformly between Q1=77.5 and Q2=100
Sample 25% of the simulants uniformly between Q2=100 and Q3=200
Sample 25% of the simulants uniformly between Q3=200 and **350.5**

Obviously we will be truncating a little bit here, but at least we will capture some regional variation?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: I guess in the second option, we are saying we are 100% sure what the distribution of fortification levels among factories looks like, which is perhaps what you're objecting to.  I guess to avoid that, we'd have to model uncertainty in the *distribution itself*, e.g. by choosing a different beta distribution for each draw (uniform is a special case of beta, so that could be included, but we could also include scenarios where the distribution among factories is asymmetric).

So if we want to avoid that level of complexity, we can say that we really only want to guess at the *mean*, not the entire distribution, and model uncertainty in our estimate of the mean, and then just assign all simulants the mean level of fortification for simplicity.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -525,8 +525,128 @@ as follows:
...
+    - 100 g (IQR: 77.5, 200) [2*]
+  * - Nigeria
+    - 40 mg NaFeEDTA / kg flour [1]
+    - 100 g (IQR:77.5, 200) [2*]
+
+.. note::
+
+	While there was data for the concentration of iron forticants other than 
+	NaFeEDTA for the locations in this table, we conservatively chose to  use 
+	the concentrations of NaFeEDTA because elemental iron is uesd at a *lower*
```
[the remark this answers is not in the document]
Ninicorn: While there was data for the concentration of iron forticants other than NaFeEDTA for the locations in this table, we conservatively chose to use the concentrations of the NaFeEDTA forticant. This is because the concentration of elemental iron in flour is lower when the NaFeEDTA forticant is used compared to other forticants.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -525,8 +525,128 @@ as follows:
...
+	# random sample from effect size distribution
+	bw_md_per_10_mg_iron = bw_md_distribution.rvs()
+
+.. note::
+
+	The Haider et al. 2013 paper did not report if the dose of iron was 
+	measured as elemental iron or as an iron compound such as NaFeEDTA. We 
+	operated under the assumption that 10 mg of daily iron consumption, as 
+	referenced in the Haider paper, represented 10 mg of *elemental* iron.
+
```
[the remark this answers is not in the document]
Ninicorn: this was answered in slack chat!

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -525,8 +525,128 @@ as follows:
...
+India consumed the same amount of fortifiable flour per day as individuals in 
+Ethiopia.
+
+.. todo:: 
+
+	Find better data and replace values for flour consumption per day for 
+	Nigeria and India.
+
+The amount of elemental iron consumed daily, in miligrams per person, (at the 
+draw level) should be calculated as such (where X is mg or NaFeEDTA per kg of
```
Ninicorn:
```suggestion
draw level) should be calculated as such (where X is mg or (of?) NaFeEDTA per kg of 
```

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Right, I mean, it's saying in the absence of better data we're making this simple assumption, but yeah.

I don't think we should model uncertainty for Nigeria and Ethiopia because those values are supposed to be national standards (presumably all iron fortified foods should be at that concentration)

And yeah, I agree with what you're saying in the second half there. I guess the key is that the variation will be happening between groups of individuals, not necessarily an individual level, but an individual level is the closest thing we have to that, so I'm fine with doing a uniform distribution at the stimulant level for India.

Also, I just looked back into the source and for non-NaFeEDTA forticants they report an average value for india of 35.25, which is the midpoint of the range. So it's probably reasonable to assume the midpoint is the average for NaFeEDTA forticants too.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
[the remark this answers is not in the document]
alibow: Yes. This is another thing that I think will be documented in the research section when time allows!

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -525,8 +525,128 @@ as follows:
...
+foods in our locations of interest in order to scale the effect size 
+accordingly. See the table below:
+
+.. list-table:: Daily Iron Consumption Parameters
+  :widths: 5 5 5
+  :header-rows: 1
+
+  * - Location
+    - Concentration of forticant in flour
+    - Amount of fortifiable flour consumed daily
```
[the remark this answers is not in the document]
alibow: I avoided that because there will be some variation in uncertainty intervals that will be computed in vivarium using their random seeds, so I'll leave the computation out of this table for now : )

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Fixed this, thanks :)

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: In the absence of more data, I don't necessarily agree that the lower bound should be >0, because increasing it makes our estimates less conservative.  Since the uniform distribution assigns probability 0 to the value 0 (or any other value), a lower bound of 0 is consistent with the data being from people who eat fortifiable wheat. It just says that there's some small probability that they eat, say <1g per day.

If we increase the lower bound to 50, we know we will be overestimating some people's wheat flour consumption (e.g. 25% of people in Amhara), so I think it is better to go with the conservative lower bound of 0.

For the upper bound, it seems fine to go with the maximum 75th percentile. We know we will be underestimating some people's wheat consumption, but that makes this a conservative estimate.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Haha, I bet they just took the midpoint of 28 and 42.5 to get their "average" of 35.25, rather than estimating it by collecting data. So yeah, I suppose we can do the same thing.

Ok, let's just make it clear what assumptions we're making.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Ok, yeah that sounds good to me. Will update!

PR #185, opened 2020-03-27
Develop/effect mod

beatrixkh:


beatrixkh:
Note from doc session:
- add in ali's diagram, posted in docs channel
- for gbd/vivarium applications: gbd effect sizes typically global. if ever vivarium is looking at specific populations, esp populations in which we expect a differential effect of some exposure, need to be aware of whether or not gbd's data will have this sensitivity.

docs/source/model_design/risk_factors.rst
```diff
@@ -203,6 +203,85 @@ Intermediates
...
+exposure :math:`E` on disease :math:`D` varies for different values of 
+:math:`M`. Effect modification is sometimes also called *interaction*. We 
+illustrate this relationship below.
+
+.. figure:: effect_mod_arrow_diagram.svg
+  :align: center
+
+If :math:`M` is some dichotomous effect modifier, then :math:`B\neq C`, and to 
+understand the full story of how :math:`E` acts on :math:`O`, we cannot simply 
+consider :math:`A`.
```
alibow, line 216: I think that we could expand upon the "we cannot simply consider *A*" statement and note that this might be a useful measure if we are looking at results for an aggregate model population (no sub-group analyses) that has similar prevalence of the effect modifier from the study population where A was measured. And then additionally note that we might want to use B and C rather than A in situations where we are interested in evaluating associations in specific subgroups OR in cases where we are evaluating model populations with variable levels of exposure to the effect modifier.

docs/source/model_design/risk_factors.rst
```diff
@@ -203,6 +203,85 @@ Intermediates
...
+.. figure:: with_em_illustration.svg
+  :width: 400
+
+Here, we see that the risk attributable to our exposure is higher in our older group than in our younger group; thus age is an effect modifier for this risk outcome.
+
+Finally, we emphasize that when dealing with a confounding variable, in order to best understand the effects of our exposure, we seek to *remove* the influence of the confounder. By contrast, if variable B is an effect modifier for exposure A, then this interaction is an important property of the relationship between A and B, and their influence on the disease. Rather than remove, we thus try to *capture and describe* effect modification in the greatest detail possible. (Cite Nicole's textbook)
+
+.. todo::
+  Add section on GBD. They briefly discussed mediation in the risk factors training this year; no reference to effect modification. I would guess that very few GBD risk factors take this into account. Main challenge is that risk factors are modeled by different teams that are often siloed: if a risk factor on one team is an effect modifier for a risk factor on another team, it's difficult for them to coordinate workflows. Need to do some digging.
+
```
alibow, line 284: I think we noted this in the documentation session last week, but just to note it here, I think it would be great to note that GBD rarely has non-universal relative risks and that there are not separate relative risks by sex, for example. And then maybe also include a note about how we should consider effect modification by location on the sim science team (does the efficacy of an intervention depend on the infrastructure quality of a particular country? etc.).

docs/source/model_design/risk_factors.rst
beatrixkh: thank you! that's a good idea and a highly relevant clarification

PR #186, opened 2020-03-27
confounding documentation

Ninicorn:


docs/source/model_design/risk_factors.rst
```diff
@@ -195,7 +195,68 @@ relationships directionally between a cause/exposure and effect/outcome.
   Discuss criteria (sim science and GBD) for modeling a risk outcome pair
 
 Confounding
-""""""""""""
+"""""""""""
+**What is confounding?**
```
Ninicorn: 1) first define confounding
2) then describe RCTs vs observational studies and the role of confounding in each.

docs/source/model_design/risk_factors.rst
```diff
@@ -195,7 +195,118 @@ relationships directionally between a cause/exposure and effect/outcome.
...
+
+Much of epidemiology is concerned with establishing associations between exposures and health outcomes. This usually involves comparing representative groups from two populations that are as similar as possible in all respects except the exposure(s) under study. **Confounding occurs when an estimate of the association between an exposure and a health outcome is mixed up with the effect of another exposure on the same health outcome, and the two exposures are associated**. When present, it results in a biased estimate of the effect of exposure on the health outcome. The bias can be negative (resulting in underestimation of the exposure effect), or positive (overestimation of the exposure effect), or can even reverse the apparent direction of effect. Confounding should always be addressed when assessing causality and failure to account for confounding leads to incorrectly concluding that an effect or its magnitude is due to one, rather than another variable.
```
beatrixkh, line 201: I think you can still winnow this down a bit, and start right off with the definition in this first sentence. is there a way to state it without using 'mixed up'?

docs/source/model_design/risk_factors.rst
```diff
@@ -198,68 +198,115 @@ Confounding
...
+In order for a factor to be a confounder (C), it must meet the following criteria:
 
-**Confounding occurs when an estimate of the association between an exposure and a disease is mixed up with the effect of another exposure on the same disease, and the two exposures are associated.** Failure to account for confounding leads to incorrectly concluding that the effect is due to one, rather than the other variable. 
+  1. Must be a risk factor, or the surrogate of a risk factor, for the outcome (O) of interest in those who are unexposed to the exposure (E) being investigated.
+  2. Must be correlated, positively or negatively, with the exposure (E). If the study population is stratified into exposed and unexposed groups, this means that (C) has a differential distribution in the two groups.
```
beatrixkh, line 206: this seems still a little misleading--i think from what ali knew and the docs nathaniel found, it sounds like confounding _has_ to be causal, so we might want to avoid saying 'correlated'

docs/source/model_design/risk_factors.rst
```diff
@@ -198,68 +198,115 @@ Confounding
...
+  3. In addition, (C) should not be an intermediate step on the causal pathway between the exposure (E) and outcome (O) being investigated (else the factor is considered an :ref:`intermediate <Intermediates>`). In other words, (C) should not be caused by the exposure (E).  
 
-In order for a factor to be a confounder, it must meet the following criteria:
-1. The factor (C ) must be associated with the exposure (E) being investigated,
-2. The factor (C) must be a risk factor for the outcome of interest in those who are not exposed to the explosure (E) being investigated,
-3. In addition, the factor (C) should not be on the causal pathway between the exposure (E) and disease (D) being investigated.
+.. todo::
+  insert link to intermediates
+
+.. image:: confounding_triangle.svg
```
beatrixkh, line 212: i think to illustrating that confounding is causal, there should be a solid arrow from C to E

docs/source/model_design/risk_factors.rst
```diff
@@ -198,68 +198,115 @@ Confounding
...
+
+**An example**:
+Imagine that you are interested in knowing whether smoking (our risk factor of interest) influences the risk of coronary heart disease (CHD) in men aged 18 to 64 years. You conduct a cohort study with an exposed group of male smokers in this age group from the general population, and compare their risk of CHD after a number of years to that of an unexposed group of non-smoking males in the same age group. You would expect to find a higher incidence of CHD among smokers. However, males who smoke are also likely to have a higher alcohol intake, which also increases the risk of CHD. Thus, the effect that you observe for the smoking-CHD association is **mixed up** with the effect of the association between alcohol intake and CHD. Without taking account of differences in alcohol intake between smokers and non-smokers, the magnitude of the smoking-CHD association estimated from the study may be higher than the **true** value, leading to incorrect conclusions.
```
beatrixkh, line 219: Here, if i understand correctly, for alcohol to be a confounder, it needs to be the case that their effects aren't multiplicative, but that drinking more also causes people to smoke more.

docs/source/model_design/risk_factors.rst
```diff
@@ -198,68 +198,115 @@ Confounding
...
-3. In addition, the factor (C) should not be on the causal pathway between the exposure (E) and disease (D) being investigated.
+.. todo::
+  insert link to intermediates
+
+.. image:: confounding_triangle.svg
+
+In **randomized experimental studies**, the process of randomising individuals (or groups) to different exposures generally ensures that the different groups are equally balanced with respect to all relevant factors that might influence the risk of the outcome. In such randomised studies, 'exposure' usually refers to a treatment or other intervention that is being compared to another, or to no intervention. Randomisation ensures that every treatment group has a similar risk of the outcome at the beginning of the study. Provided the study is conducted rigorously and is sufficiently large, if we see a difference in the incidence of the outcome between treatment groups at the end of the study, then we can conclude that this difference is caused by the treatment. For this reason, experimental studies provide the strongest evidence of a causal association between an exposure and disease.
```
beatrixkh, line 214: I think that these two sections (on randomized experimental and observational studies) can be cut, and you could go straight into your example. what do others think? @alibow ? @NathanielBlairStahn ?

docs/source/model_design/risk_factors.rst
yongqx: When we calculate the summary exposure value, we are evaluating the impact of multiple risk factors on the same health outcome. Is confounding happens only if those factors are in causal relationship rather than simply correlated?

beatrixkh:
Ok---sorry about this, but disregard my comments from a week ago.

I've now looked at multiple papers with different definitions of confounding, some of which stipulate a causal relationship, and some of which don't. There's a thread in the docs channel where this was discussed further, and Nicole you pointed out/wrote this:

"we will need to deal with confounding only when we interpret our effect sizes from the literature...cuz all gbd models (the RRs for the risk cause-models) have assumptions of causal, with no confounding right?"

which seems like what this documentation should inform.

You also posted this link: https://www.theanalysisfactor.com/what-is-a-confounding-variable/, which talks about (/corroborates) the existence of multiple definitions of confounding.

So ultimately I think it's important to acknowledge somewhere in this PR that there _are_ these multiple definitions of confounding, and it might also be nice to choose what relationship we will call "confounding" within the framework of a vivarium model.

docs/source/model_design/risk_factors.rst
beatrixkh: I take this comment back, after further discussion!

docs/source/model_design/risk_factors.rst
beatrixkh: Actually---i see how this might be helpful when trying to find data, and assess the studies from which you're pulling effect sizes.

docs/source/model_design/risk_factors.rst
```diff
@@ -195,7 +195,118 @@ relationships directionally between a cause/exposure and effect/outcome.
...
+
+In order for a factor to be a confounder (C), it must meet the following criteria:
+
+  1. Must be a risk factor, or the surrogate of a risk factor, for the outcome (O) of interest in those who are unexposed to the exposure (E) being investigated.
+  2. Must be correlated, positively or negatively, with the exposure (E). If the study population is stratified into exposed and unexposed groups, this means that (C) has a differential distribution in the two groups.
+  3. In addition, (C) should not be an intermediate step on the causal pathway between the exposure (E) and outcome (O) being investigated (else the factor is considered an :ref:`intermediate <Intermediates>`). In other words, (C) should not be caused by the exposure (E).
```
alibow, line 207: I think we should also mention mediation here. Maybe something like...

3. In addition, (C) should not be an intermediate on the a causal pathway between the exposure (E) and outcome (O). If this is the case, (C) may in fact be an intermediate [LINK] or a mediator [LINK]. In other words, (C) should not be caused by the exposure (E).

docs/source/model_design/risk_factors.rst
alibow: I agree, but I think that this text is really useful to have elsewhere in our documentation. I'm not exactly sure where we should put it, but maybe we can save it and then discuss in an upcoming section where we would like to put general epi stuff like this? I think that it's a little distracting in our risk factors section as is!

docs/source/model_design/risk_factors.rst
```diff
@@ -195,7 +195,118 @@ relationships directionally between a cause/exposure and effect/outcome.
...
+
+**An example**:
+Imagine that you are interested in knowing whether smoking (our risk factor of interest) influences the risk of coronary heart disease (CHD) in men aged 18 to 64 years. You conduct a cohort study with an exposed group of male smokers in this age group from the general population, and compare their risk of CHD after a number of years to that of an unexposed group of non-smoking males in the same age group. You would expect to find a higher incidence of CHD among smokers. However, males who smoke are also likely to have a higher alcohol intake, which also increases the risk of CHD. Thus, the effect that you observe for the smoking-CHD association is **mixed up** with the effect of the association between alcohol intake and CHD. Without taking account of differences in alcohol intake between smokers and non-smokers, the magnitude of the smoking-CHD association estimated from the study may be higher than the **true** value, leading to incorrect conclusions.
```
alibow, line 219: I wounder if we could say "confounded by" rather than "mixed up"? Maybe just a personal preference, but it might be nice to be specific with our language (we could say "mixed up" in reference to effect modification, mediation, etc.).

docs/source/model_design/risk_factors.rst
```diff
@@ -195,7 +195,118 @@ relationships directionally between a cause/exposure and effect/outcome.
...
+
+.. image:: smoking_triangle.svg
+
+Note that the dotted line between smoking and alcohol indicates that we do not expect a 'causal' link between the smoking and alcohol. Rather, it indicates that the two are associated in the population.
+
+**How to identify confounding?**
+
+We look for confounding by stratifying (splitting) the data according to the proposed confounding factor and then examining the measures of effect of the exposure on the outcome in the different strata separately. If the stratum-specific measures of effect are similar to each other, but different from the crude measure of effect, this is evidence for confounding. We can now consider a numerical example to demonstrate how you can look for confounding in data from an epidemiological study.
+
+**An example**:
```
beatrixkh, line 235: I think, because in this case, coffee is not causally related to the outcome, this is a confusing example that we might not want to include.

While this does fit the three points of the definition (1, smoking is a risk factor for cancer, 2. smoking is correlated with coffee consumption, and 3., smoking does not cause coffee consumption), given that coffee isn't a risk factor for cancer, would we ever put it in one of our models? and does this still fall under the purview of confounding?

docs/source/model_design/risk_factors.rst
```diff
@@ -195,7 +195,118 @@ relationships directionally between a cause/exposure and effect/outcome.
...
+
+Having established that smoking appears to confound the association between coffee consumption and pancreatic cancer, the final step is to combine, or pool, the data across strata of smoking to obtain a combined, or pooled, estimate of the effect of coffee consumption on pancreatic cancer adjusted for the confounding effect of smoking. This pooled estimate is an average of the effect across all strata of smoking weighted by the size of each stratum. There are statistical methods for obtaining such pooled estimates (e.g. Mantel-Haenszel, regression). There are a number of strategies to minimise or deal with confounding, both when designing a study and in the analysis phase.
+
+.. todo::
+  How does GBD handle confounding
```
alibow, line 309: I think that we covered this a bit in the general causality section, so it might be fine not to add more detail here unless there's anything else specific that you would like to add!

docs/source/model_design/risk_factors.rst
```diff
@@ -195,7 +195,118 @@ relationships directionally between a cause/exposure and effect/outcome.
...
+
+  1. Alcohol intake is a risk factor for CHD in non-smokers. In other words the relationship between alcohol and CHD is not dependent on smoking status.  
+  2. Smokers generally have a higher alcohol intake (in most populations). 
+  3. Finally, alcohol intake is not on the causal pathway between smoking and CHD (smoking does not in itself cause people to drink more alcohol). 
+
+.. image:: smoking_triangle.svg
+
+Note that the dotted line between smoking and alcohol indicates that we do not expect a 'causal' link between the smoking and alcohol. Rather, it indicates that the two are associated in the population.
+
+**How to identify confounding?**
```
alibow, line 231: I wonder if it would be valuable to move this to a general epi section like the study design stuff (since stratification is really valuable for effect modification also). Doesn't need to be done now necessarily, but we can maybe talk about it at our next docs session.

docs/source/model_design/risk_factors.rst
alibow: Also, it might be helpful to add a section about what we should think about as sim science team members to watch out for confounding. Like when looking for data sources we should check that the studies adjusted for age/sex/other important variables when appropriate.

docs/source/model_design/risk_factors.rst
beatrixkh: The coffee example actually made me a bit confused about this definition---is there any definitional requirement that the exposure be causally related (or plausibly causally related) to the outcome?

Because if not, you could take any variable (C) which is causally related to some variable (O) and associated with some (E), where E and O are otherwise unrelated, and say that C confounds the relationship between E and O. (Which may well be the case, but I wasn't previously thinking of this as something that would fit under the definition of confounding)

docs/source/model_design/risk_factors.rst
alibow: I see what you're saying here, Beatrix. I think examples like these are shown in intro epi lectures to demonstrate that if you don't account for confounding, you can find *spurious* relationships and conclude that coffee causes pancreatic cancer even if that's not the case (as a warning to be carefully when analyzing primary data collected in epi studies!). I agree though that it is a little less relevant for our purposes in that we will pretty much never (shouldn't) be dealing with exposures that aren't causally related to the outcome. I'm fine with keeping the coffee example because it's already written out, but maybe we could just include a note about this somewhere?

docs/source/model_design/risk_factors.rst
Ninicorn: I thought summary exposure value is for one risk factor (that has many categories). so the SEV measures the effect of one risk-outcome pair (adjusted for confounders- either correlated or causally linked to the outcome? don't know how GBD handles confounders...might be good to find out!)

docs/source/model_design/risk_factors.rst
Ninicorn: Ok! ill leave it in here, we can copy and paste as we see fit later!

NathanielBlairStahn: We should add a `.. todo::` to remind us to do this later.

docs/source/model_design/risk_factors.rst
NathanielBlairStahn: Yeah, good point. Does the first requirement, that (C) "Must be a risk factor, or the surrogate of a risk factor, for the outcome (O)" imply that (C) must be causally related to (O)?  Because our definition of "risk factor" is "any attribute whose measure is causally related to the measure of an outcome."

Or if (C) is just a "surrogate" of the risk factor, does this mean it's _not_ necessarily causally related to (O)?  What exactly does "surrogate" mean?

docs/source/model_design/risk_factors.rst
Ninicorn: mixed together? I don't want to use the word 'confound' in the definition.

docs/source/model_design/risk_factors.rst
Ninicorn: I think it would be good to have a study design page for our documentation and the important things to consider when pulling effect sizes from different study designs. I've made this into a todo note.

docs/source/model_design/risk_factors.rst
Ninicorn: In the coffee example, coffee is the EXPOSURE, and smoking is the CONFOUNDER, cancer is the OUTCOME. I don't think there is any requirement that the is exposure is causally related to the outcome- as this is the big unknown and what we are trying to prove or disprove! In our example, after controlling for smoking we see NO effect on cancer from coffee so we can say that there is no causal relationship between our exposure and outcome. 

I think we need to think of it as there is an exposure and there is an outcome. People are saying E causes O for some reason (like, vaccines causes autism). We as epidemiologists need to prove yes or no, eg. whether the relationship between vaccines and autism is real (causal) or spurious (acting through confounders only).

Ninicorn: @NathanielBlairStahn that is correct- the confounder must be causally related (risk factor) for the outcome for it to be considered. 

I think we need to think of it as first we have an exposure outcome pair that we are unsure of (like this new claim that coffee causes cancer, or vaccines causes autism). Then some epidemiologists are like...that's curious, we are unconvinced!! so they postulate, that the relationship that is being observed between coffee and cancer OR vaccines and autism must be _caused_ by "something else" and this "something else" is just _mixed-up_ with the exposure.  To figure out what this "something else" is, they must then come up with a list of things that they know must cause the outcome. This "something else" must also be correlated with the exposure, otherwise we won't observe this spurious relationship between the exposure and outcome.

Ninicorn: by surrogate i think it means the risk is measured by a proximal factor. like if tar in lungs is the real cause of lung cancer in smokers, and you can't measure tar in lungs, you have to approximate it by measuring number of cigarettes smoked.

docs/source/model_design/risk_factors.rst
Ninicorn: ahaha, mixed up is the thesaurus definition of confound. I thought it would be easier to think of confounding (itself a confounding topic!) with a more familiar term. duno...open to suggestions!

docs/source/model_design/risk_factors.rst
Ninicorn: ya! :) 

In our models, we never are involved with proving causality. but we need to make an assessment of whether the reported effect sizes are real or not. Or whether the reported effect sizes are as big/small as they claim- maybe some of that effect is coming from a confounder that they failed to adjust for (which happens often!) and we are biasing our models results.

docs/source/model_design/risk_factors.rst
Ninicorn: yes exactly!!

docs/source/model_design/risk_factors.rst
Ninicorn: ok sounds good :)  I've addressed all the comments. I'm gona merge it so we have it up. We can go through it and see how we can address some of the more structural things like have a general epi section.

docs/source/model_design/risk_factors.rst
NathanielBlairStahn: Thanks for all these clarifying comments, Nicole!

PR #196, opened 2020-04-01
Update concept_model.rst

davidbernotas:
Updated Table 5 to reflect the probability that doctors don't make any change to medication in a follow up visit.  I don't know how to remove the original Table 5 though.
Also, I'm sure it's clear from Table 5 - increasing dose, adding 2nd drug and changing potency are mutually exclusive (which is why probabilities in Table 5 sum to 1).  Only one factor changes at a time.


SantoniS:
@davidbernotas  I see that you have merge conflicts that will need to be resolved. Do you know how to resolve them?

PR #200, opened 2020-04-03
Risk effects

beatrixkh:


alibow:
> I started to write this up, and then realized I'm not sure of the definition, so right now it's a stub with questions that I also posted to the docs channel

(Copying here from slack):

Hmm, good question. My assumption is that we might want to frame this in terms of the relative risks for a given risk factor and what they apply to (for instance, incidence versus prevalence versus excess mortality rate for a given cause... maybe they differ by cause OR it could be direct units of the outcome, like mmHg for blood pressure). This makes the most sense to me, but James might have had something else in mind in terms of specific GBD risk architecture, so if you or anyone else has other ideas we should consider those as well!

docs/source/model_design/risk_factors.rst
```diff
@@ -339,6 +339,33 @@ complexity into the exposure component.
...
+For example, we might model the risk effect *blood pressure* as an outcome of 
+*sodium intake*. For each measurement of the exposure sodium intake, this model 
+would map to a probability distribution of possible blood pressures.
+
+.. note::
+  Wrting this up, I realized I'm not sure what the definition is. In the above, 
+  is the risk outcome "blood pressure"? is it a given measurement of blood 
+  pressure? or is it a differential probability of a given measure of blood 
+  pressure?
+
```
yongqx, line 354: For risk effect outcome, we usually measure the proportion or burden of disease attributable to specific risk factor. Talking about _blood pressure_, it reminds me the casual relationship of high sodium intake -> high blood pressure -> ischemic heart disease. To model a risk-outcome pair, we have to find evidence from epidemiological studies supporting a causal relationship between risk factor exposure and the outcome, as well as available data to estimate effect sizes per unit of exposure. Evidence that these effects can be generalized to a general population.

PR #201, opened 2020-04-03
Causal intro update

alibow:
I'm planning to add in a sim science specific section as well, but wanted feedback on what is included so far before starting. Thanks in advance!

### DIFF
--- a/docs/source/model_design/risk_factors.rst
+++ b/docs/source/model_design/risk_factors.rst
@@ -206,6 +206,57 @@ Effect Modification
 Mediation
 """""""""
 
+Causal Inference in the Global Burden of Disease Study
+""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+Notably, GBD researchers use an evidence scoring system that is based off of a subset of the Bradford Hill Criteria to evaluate the quality of evidence regarding causal relationships between risk-outcome pairs in GBD. Specifically, before computing the relative risks for a GBD risk factor, GBD researchers evaluate the *risk of bias* among individual studies that investigate the relationship between a risk-outcome pair. Then, GBD researchers additionally evaluate the strength (as a direct result of the relative risk curve they compute), consistency (through evluating between study heterogeneity), and dose-response (through the shape of the relative risk curve) for the computed relative risks for a given risk factor. Using these criteria, GBD researchers create a quantitive quality of evidence score for each risk-outcome pair in GBD.
+
+Specifically, GBD researchers evaluate the risk of bias within individual studies based on the following characteristics:
+
+1) Representativeness of the study population
+
+2) Exposure measurement
+
+  a) Individual versus population
+
+  b) Objective versus self-report
+
+  c) Multiple prospective versus baseline prospective versus retrospective
+
+3) Outcome measurement
+
+  a) Death certificatie/physician diagnosis/medical records versus self-report
+
+  b) Blind outocme assessment versus not
+
+4) Reverse causation: low, medium, high
+
+5) Control for confounding 
+
+  a) Randomized controlled trial
+
+  b) Age, sex, tobacco, income, education, other critical determinants for a specific outcome not on the causal pathway
+
+  c) Age, sex, tobacco, other critical determininants for a specific outcome not on the causal pathway
+
+  d) Age, sex
+
+6) Selection bias
+
+  a) High follow-up (95%), not opportunity for selection
+
+  b) Moderate follow-up (85-95%), limited opportunity for selection
+
+  c) Low follow-up (<85%), considerable opportunity for selection
+
+.. note::
+
+  This information was obtained from a science seminar presented by Ryan Barber and Chris Murray on March 11, 2020; a recording is available `here <https://hub.ihme.washington.edu/display/GBD2020/GBD+Science+Seminar+series>`_. Documentation for GBD's evidence scoring system is available `here <https://hub.ihme.washington.edu/display/GBD2020/Evidence+score>`_.
+
+.. note::
+
+  It may be helpful to add sections for bias (selection bias, measurement bias, confounding bias, etc.) and generalizability to the above causal relationships section for additional context here
+
 What is a risk exposure?
 ++++++++++++++++++++++++
 
--- a/docs/source/model_design/risk_factors.rst
+++ b/docs/source/model_design/risk_factors.rst
@@ -124,7 +124,25 @@ An term that is often used in causal inference is the **counterfactual**. The
 counterfactual refers to an alternate reality in which only a single variable 
 has changed and all else has remained exactly the same. 
 
-  For instance, say that we wanted to evaluate the causal realtionship between 
+  For instance, suppose you've finished work and you're trying to determine 
+  how to get down to the international district for dinner. You go to your 
+  maps application and ask **What is the quickest route to my destination?** 
+  It tells you that based on the current time, weather, day of the week, road 
+  work conditions, etc. that you should take the light rail.
+
+  So you leave the office, walk a few blocks to the westlake station, and 
+  catch the train. Two stops in, your train gets stopped behind a broken down 
+  bus. You decide to get off the train and walk the remaining distance. 
+  Arriving ten minutes late to dinner, you think to yourself **Would I have 
+  arrived on time if I took the bus instead?**
+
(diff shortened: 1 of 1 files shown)


docs/source/model_design/risk_factors.rst
```diff
@@ -206,6 +217,77 @@ Effect Modification
...
+researchers evaluate the *risk of bias* among individual studies that 
+investigate the relationship between a risk-outcome pair. Then, GBD 
+researchers additionally evaluate the strength (as a direct result of the 
+relative risk curve they compute), consistency (through evluating between 
+study heterogeneity), and dose-response (through the shape of the relative 
+risk curve) for the computed relative risks for a given risk factor. Using 
+these criteria, GBD researchers create a quantitive quality of evidence score 
+for each risk-outcome pair in GBD.
+
+Specifically, GBD researchers evaluate the risk of bias within individual
```
beatrixkh, line 236: Do you know if they're just documenting this score, perhaps with the hope of being able to use this info later, or if they do anything with it? in other words, if something just barely clears their threshold criteria for addition as a relative risk, but they believe it to have some confounder, do you know if they do anything to account for the confounder that they've documented?

beatrixkh:
I think we also discussed that if GBD is using some risk exposure A that has confounding variable, or mediator B, and B is not accounted for in their effect size, that they can use this as long as B is also not accounted for in their relative risk. However, in this scenario, the effect size and relrisk may or may not be usable by us.

I think that should be noted somewhere within the scope of the risk factor page---do you think this would be a good place for that? I think the other reasonable place would be within a vivarium-specific section, noting that that's something to watch out for when pulling GBD RRs and relative risks.

You, @alibow might've already noted this and included it in one of your sections---I don't remember in which context we discussed this!

docs/source/model_design/risk_factors.rst
```diff
@@ -206,6 +217,77 @@ Effect Modification
...
+
+6) Selection bias
+
+  a) High follow-up (95%), not opportunity for selection
+
+  b) Moderate follow-up (85-95%), limited opportunity for selection
+
+  c) Low follow-up (<85%), considerable opportunity for selection
+
+.. note::
```
beatrixkh, line 275: thank you for laying this out; it's very clear/nicely organized!

docs/source/model_design/risk_factors.rst
alibow: > I think we also discussed that if GBD is using some risk exposure A that has confounding variable, or mediator B, and B is not accounted for in their effect size, that they can use this as long as B is also not accounted for in their relative risk. However, in this scenario, the effect size and relrisk may or may not be usable by us.
> 
> I think that should be noted somewhere within the scope of the risk factor page---do you think this would be a good place for that? I think the other reasonable place would be within a vivarium-specific section, noting that that's something to watch out for when pulling GBD RRs and relative risks.
> 
> You, @alibow might've already noted this and included it in one of your sections---I don't remember in which context we discussed this!

Ah, good point! I will definitely add this in here.

alibow: I wonder

> This is a really clear explanation of their evidence scoring---thank you for writing it!
> 
> I am still unclear (both from your page and from trying previously to dig into this a bit) on if/how they take any of this into account when modeling a risk factor. like if they simply don't use something that has a strong confounder, if they _do_ somehow take some of these relationships into accounts, or if they find studies for their effect sizes/rel risks that are representative for year/locaton/sex/age, and if that's sufficient for their purposes, because all estimates are at the year-loc-age-sex population level.

Hmm, good question. I am a little unclear on this as well, but my understanding/suspicion is that they account for potential biases/confounding in their MR-BRT analyses. Basically, the data extraction template has a bunch of indicator variables for "adjusted_for_sex = yes/no" and then that will be added as a covariate for that study when they are running the meta-regression for the R-O pair, BUT I am not certain that this is what their process is and I will try to investigate! Let me know if you have any thoughts/ideas about this :)

yongqx: The way i understand how GBD select RO pairs is based following criteria:

1. The likely importance of risk factor to a disease.
2. Data and methodology for scaling the risk exposure.
3. epidemiological evidence to support its causal relationship.
4. The effects can be generalized to a general population.

alibow: I think this would also be important to add, thanks @yongqx! I was focusing just on #3 here (and I guess a little bit on #4), but that would be good to add for context.

docs/source/model_design/risk_factors.rst
```diff
@@ -124,7 +124,25 @@ An term that is often used in causal inference is the **counterfactual**. The
...
+  So you leave the office, walk a few blocks to the westlake station, and 
+  catch the train. Two stops in, your train gets stopped behind a broken down 
+  bus. You decide to get off the train and walk the remaining distance. 
+  Arriving ten minutes late to dinner, you think to yourself **Would I have 
+  arrived on time if I took the bus instead?**
+
+As illustrated here, by definition the counterfactual question is impossible 
+to *directly* answer because it is a purely hypothetical question without 
+direct supporting evidence; we can only use what we know to make a *guess* at 
+what might have happened in the counterfactual scenario. Consider the
```
Ninicorn, line 142: pompous literary quote suggestion: 

“We can never know what we want, because, living only one life, we can neither compare it with our previous lives nor perfect it in our lives to come." -Milan Kundera The Unbearable Lightness of Being

docs/source/model_design/risk_factors.rst
```diff
@@ -137,14 +155,13 @@ has changed and all else has remained exactly the same.
...
-evaluate, which is the core challenge in causal inference. Luckily, there are 
-several strategies that can be used to attempt to indirectly answer the 
-counterfactual question and perform causal inference. Some examples include
-randomized control trials and adjustment for confounding variables in 
-epidemiology studies.
+  However, we cannot wave a magic wand and explore the lung cancer rate in 
+  this alternative universe in 1954 to answer this question. Luckily, there 
+  are several methods in the field of epidemiology that attempt to answer 
+  conterfactual questions regarding causal relationships through randomized 
+  controlled trials and other types of studies.
```
Ninicorn, line 165: The section starts off by talking about causal inference, then counter-factual, then causal inference again using the bradford hill criteria. The counterfactual paragraph feels slightly out of place. Maybe it can be called out separately in a note box?

docs/source/model_design/risk_factors.rst
```diff
@@ -206,6 +217,77 @@ Effect Modification
...
+
+  This information was obtained from a science seminar presented by Ryan 
+  Barber and Chris Murray on March 11, 2020; a recording is available `here <https://hub.ihme.washington.edu/display/GBD2020/GBD+Science+Seminar+series>`_. Documentation for GBD's evidence scoring system is available `here <https://hub.ihme.washington.edu/display/GBD2020/Evidence+score>`_.
+
+.. note::
+
+  It may be helpful to add sections for bias (selection bias, measurement 
+  bias, confounding bias, etc.) and generalizability to the above causal 
+  relationships section for additional context here.
+
```
Ninicorn, line 285: ya agree! and for the generalisability part, we could make our own assessment criteria to score/rank how generalisable the effect sizes we pick from literature apply it to our sim population.

Ninicorn: we can also add sections on study design? cuz each study design will have their own flaws/bias?

docs/source/model_design/risk_factors.rst
```diff
@@ -206,6 +217,77 @@ Effect Modification
...
+""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+Notably, GBD researchers use an evidence scoring system that is based off of a 
+subset of the Bradford Hill Criteria to evaluate the quality of evidence 
+regarding causal relationships between risk-outcome pairs in GBD. 
+Specifically, before computing the relative risks for a GBD risk factor, GBD 
+researchers evaluate the *risk of bias* among individual studies that 
+investigate the relationship between a risk-outcome pair. Then, GBD 
+researchers additionally evaluate the strength (as a direct result of the 
+relative risk curve they compute), consistency (through evluating between
```
Ninicorn, line 230:
```suggestion
relative risk curve they compute), consistency (through evaluatingbetween 
```

PR #211, opened 2020-04-15
LSFF: Write intro to stratifying on baseline coverage

NathanielBlairStahn:
Once again, my editor deleted a bunch of whitespace, so git shows lots of extraneous changes. To see only the substantive changes, click the gear icon at the top and select the box that says "Hide whitespace changes":

image

The new content is primarily in the section "Stratifying Exposure Variables by Baseline Intervention Coverage."

### DIFF
--- a/docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
+++ b/docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
@@ -30,12 +30,12 @@ Model Versions
 
 .. list-table::
    :header-rows: 1
-   
+
    * - Version
      - Description
      - Done
    * - v1.0_disease_only
-     - Model with mortality, morbidity, diarrhea, LRI, measles, and 
+     - Model with mortality, morbidity, diarrhea, LRI, measles, and
        neural tube defects.
      - Yes
    * - v2.0_lbwsg
@@ -57,11 +57,11 @@ Model Versions
    * - v7.0_scenarios
      - Full model with final stratification and all scenarios.
      - Not started.
-   
-Model results can be located at 
-``/share/costeffectiveness/results/vivarium_conic_lsff`` in the 
+
+Model results can be located at
+``/share/costeffectiveness/results/vivarium_conic_lsff`` in the
 appropriate model version directory.
-   
+
 
 Model Components
 ----------------
@@ -127,17 +127,71 @@ PAF-of-1 Cause/Risk Pairs
 Risk-Outcome Relationships
 ++++++++++++++++++++++++++
 
-Coverage Gap Framework
-++++++++++++++++++++++
-
 Effect size stratification for baseline population
---------------------------------------------------
+++++++++++++++++++++++++++++++++++++++++++++++++++
 
-From GBD we obtain mean population values for prevalence of vitamin A deficiency, birth prevalence of neural tube defects, and mean haemoglobin levels. Because we are interested in the effect of fortification and there exists baseline coverage of fortification, we must first stratify our population into those who were covered vs not covered by the forticant of interest. We then need to calculate the risk (prevalence) of vitamin A deficiency, risk (birth prevalence) of neural tube defects, and mean haemoglobin levels by coverage strata. 
+From GBD we obtain overall population estimates for our outcomes of interest
+(prevalence of vitamin A deficiency, birth prevalence of neural tube defects,
+distribution of haemoglobin levels, and distribution of birth weights). Because
+there exists baseline coverage of our interventions -- i.e. there is already
+some food fortification occurring in the locations we are simulating -- these
+overall population estimates include both people who are eating fortified food
+and those who are not.
 
-This method applies to exposures with dichomotous outcomes such as Vitamin A deficiency or neural tube defects:
+When we introduce additional food fortification into the population, only those
+who are *not* already eating fortified food should receive the benefit.
+Therefore, in order to properly measure the effects of our interventions, we
+must first stratify our population into those who were already covered vs. not
+covered by the forticant of interest. We then need to calculate the risk
+exposure in each coverage stratum to get a complete picture of the baseline
+population. For example, we need to know the prevalence of vitamin A deficiency
+or the distribution of haemoglobin in the subpopulation already eating fortified
+food as well as in the subpopulation not eating fortified food.
 
-We always define the exposure as bad to match GBD 2017 definitions, so relative risks are always >1 
+These coverage-stratified risk exposures are calculated so that
+
+1.  Their population-weighted average matches the overall population estimate
+    from GBD, and
+2.  The differential exposure between the subpopulations matches the
+    corresponding effect size for our intervention.
+
+For example, the ratio of the prevalence of vitamin A deficiency between the
+fortified and unfortified groups should match the risk ratio of vitamin A
+fortification on vitamin A deficiency from the literature, and the difference in
+mean haemoglobin level between the fortified and unfortified groups should match
+the mean shift in haemoglobin from iron fortification found in the literature.
+When we stratify the population in this way, we are assuming that the effect
+sizes of fortification found in the literature are generalizable to the
+populations we are simulating (note if this assumption is *not* true, we should
+not be simulating the interventions in theses populations at all).
+
+.. todo::
+
+  Come up with some examples to show what would happen if we *didn't* first
+  stratify by existing coverage. I.e., why is this baseline stratification
+  important? It is not totally obvious to me what the effect would be if we
+  didn't do it (e.g. would we end up overestimating or underestimating the true
+  effect?) -- I think the net effect depends on the level of existing coverage,
+  the shape of the underlying exposure distribution, and the magnitude of the
+  effect size.
+
+The details of stratifying the baseline population depend on form of the risk
+exposure distribution, in particular whether it is categorical or continuous.
+The simplest case is for a dichotomous risk (e.g. vitamin A deficiency or neural
+tube defects). This situation has been dealt with in a number of Vivarium models
+under the title "Coverage Gap Framework." The strategy for continuous risk
+exposure distributions (e.g. haemoglobin or birth weight) is less well-developed
+-- the current approach only approximates a true stratification of the
+continuous distribution, and it has mathematical limitations that may be
+problematic in certain situations. We discuss the dichotomous and continuous
+cases in separate sections below.
+
+Coverage Gap Framework for Dichotomous Risk Exposures
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This method applies to exposures with dichotomous distributions such as Vitamin A deficiency or neural tube defects:
+
+We always define the exposure as bad to match GBD 2017 definitions, so relative risks are always >1.
 
 :math:`C_{vita_{baseline}}`: coverage of vitamin A fortified food in the population from the literature that is applied to our sim population at baseline
 
@@ -157,7 +211,13 @@ PAF= :math:`\frac{P_{exposure_{baseline}}(RR-1)}{1+P_{exposure_{baseline}}(RR-1)
 
 1-PAF= :math:`\frac{1}{1+P_{exposure_{baseline}}(RR-1)}`
 
-*Important assumptions and limitations* This equation for PAF is valid under the assumption of no confounding. An alternate equation for PAF should be used when to get an unbiased PAF in the presence of confounding; however, we will need the attributable fraction in the exposed which we do not readily have. Hence this is a limitation. The RRs we use, and the exposure % we use are approximating the PAFs. We make the assumption that the RRs pulled from literature is generalizable. 
+*Important assumptions and limitations* This equation for PAF is valid under the
+assumption of no confounding. An alternate equation for PAF should be used when
+to get an unbiased PAF in the presence of confounding; however, we will need the
+attributable fraction in the exposed which we do not readily have. Hence this is
+a limitation. The RRs we use, and the exposure % we use are approximating the
+PAFs. We make the assumption that the RRs pulled from literature is
+generalizable.
 
 .. todo::
 
@@ -168,7 +228,7 @@ risk in (1-:math:`C_{vita}`), exposed group: :math:`ϴ_{1}= ϴ_{GBD}*(1-PAF)*RR`
 
 risk in (:math:`C_{vita}`), unexposed group: :math:`ϴ_{0}= ϴ_{GBD}*(1-PAF)` … equation 2
 
-**How to apply the intervention**: the intervention increases the population coverage of vitamin A fortified food, this value --> :math:`C_{vita}`, and shifts the amount of people who receive equation 1 to equation 2. 
+**How to apply the intervention**: the intervention increases the population coverage of vitamin A fortified food, this value --> :math:`C_{vita}`, and shifts the amount of people who receive equation 1 to equation 2.
 
 Interventions
 +++++++++++++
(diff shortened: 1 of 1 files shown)


docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -127,17 +150,72 @@ PAF-of-1 Cause/Risk Pairs
...
+some food fortification occurring in the locations we are simulating --- these
+overall population estimates include both people who are eating fortified food
+and those who are not.
+
+When we introduce additional food fortification into the population, only those
+who are *not* already eating fortified food should receive the benefit.
+Therefore, in order to properly measure the effects of our interventions, we
+must first stratify our population into those who are already covered vs. not
+covered by the forticant of interest. We then need to calculate the risk
+exposure (or other relevant outcome) in each coverage stratum to get a complete
```
alibow, line 169: maybe we could include an example here for clarity "(e.g. mean birth weight or prevalence of VAD)".

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -127,17 +150,72 @@ PAF-of-1 Cause/Risk Pairs
...
 
-Effect size stratification for baseline population
---------------------------------------------------
+  Come up with some examples to show what would happen if we *didn't* first
+  stratify by existing coverage. I.e., why is this baseline stratification
+  important? It is not totally obvious to me what the effect would be if we
+  didn't do it (e.g. would we end up overestimating or underestimating the true
+  effect?) -- I think the net effect depends on the level of existing coverage,
+  the shape of the underlying exposure distribution, and the magnitude of the
+  effect size.
```
alibow, line 201: I think that the most obvious example of this is for birth weight. The additional 10 grams of birth weight from an intervention will be much more valuable to you if you are born in the 0-500 gram category than the 2500 to 2750 category. 

So, if we DO calibrate to baseline coverage, the uncovered group will have a lower baseline mean birth weight (say 2500) than the covered group (say 2750). The difference in RR between the 2500 g group and the next birth weight group >> the difference in RR between the 2750 group and the next birth weight group. Essentially, we are helping those who really need it!

If we do NOT calibrate for baseline coverage, both groups have the same starting mean birth weight. In this case, the uncovered group will be starting slightly "better off" and therefore have less to gain from the intervention (we might underestimate the effect of the intervention).

A similar situation is true for dichotomous exposures like VAD. If overall population prevalence of VAD is 20% (baseline covered VAD prevalence = 10%, baseline uncovered = 30%) and the RR for vitamin A fortification is 0.5. Also let's say that baseline coverage is 50%. 

So if we DO account for baseline coverage differences, we will reduce VAD by 50% in the baseline uncovered group... 30% to 15% so that 15% of the population remits.

If we DO NOT accoun for bsaeline coverage differences, we will still reduce VAD by 50% in the baseline uncovered group, but there will be *less* VAD there to reduce so that VAD prevalence changes from 20% to 10% so that 10% of the population remits.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -127,17 +150,72 @@ PAF-of-1 Cause/Risk Pairs
...
+
+For example, the ratio of the prevalence of vitamin A deficiency between the
+fortified and unfortified groups should match the risk ratio of vitamin A
+fortification on vitamin A deficiency from the literature, and the difference in
+mean haemoglobin level between the fortified and unfortified groups should match
+the mean shift in haemoglobin from iron fortification found in the literature.
+When we stratify the population in this way, we are assuming that the effect
+sizes of fortification found in the literature are generalizable to the
+populations we are simulating (note that if this assumption is *not* true, we
+should not be simulating the interventions in theses populations at all).
```
alibow, line 190: We are assuming that the proportion of our outcome attributable to our exposure is the same in our model population as it is in our study population. For something like vitamin A deficiency, it's pretty much all attributable to vitamin A intake. For something like anemia, it is *not* necessarily all attributable to iron intake (i.e. non-iron responsive anemias). The assumption is valid for situations in which there are not many other factors that cause the outcome other than the exposure at hand. However, it should be more carefully evaluated when there are multiple factors that can cause the outcome (*many* things can cause low birth weight).

We generally get around this issue in our vivarium simulations by directly using the PAF for a given exposure and outcome. However, we don't use the PAF in this approach, so that is why it requires stronger assumptions.

I don't know that we need to write this here, but I don't think we've written it down, so I just wanted to make a note : )

NathanielBlairStahn: Thanks for this explanation, @alibow !  When you say we don't use the PAF in this approach, are you referring specifically to continuous outcomes like hemoglobin or birth weight?  Because I thought our approach for dichotomous outcomes like VAD was essentially equivalent to computing the PAF.  Is that accurate, or am I missing something?

alibow: Ah... yes, the point I was trying to make was that we're not using GBD's PAF. GBD's PAFs are calculated after going through a process that vets the RRs and adjusts for potential confounders etc. in MRBRT and assesses generalizability so we are relying on this analysis and don't have to make the assumption on our own; we just have to assume that GBD is correct : ).

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -127,17 +150,72 @@ PAF-of-1 Cause/Risk Pairs
...
+covered by the forticant of interest. We then need to calculate the risk
+exposure (or other relevant outcome) in each coverage stratum to get a complete
+picture of the baseline population. For example, we will need to know the
+prevalence of vitamin A deficiency or the distribution of haemoglobin in the
+subpopulation already eating fortified food as well as in the subpopulation not
+eating fortified food.
+
+These coverage-stratified exposures are calculated so that:
+
+1.  Their population-weighted average matches the overall population estimate
```
Ninicorn, line 177:
```suggestion
1.  Their population-weighted average of the risk exposure matches the overall population estimate
```

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -127,17 +150,72 @@ PAF-of-1 Cause/Risk Pairs
...
+    from GBD, and
+2.  The differential exposure between the subpopulations matches the
```
Ninicorn, line 179:
```suggestion
2.  The differential risk exposure between the subpopulations matches the
```

Ninicorn: just wanted to call them the same thing as they are referring to the same thing

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -127,17 +150,72 @@ PAF-of-1 Cause/Risk Pairs
...
+there exists baseline coverage of our interventions --- i.e. there is already
+some food fortification occurring in the locations we are simulating --- these
+overall population estimates include both people who are eating fortified food
+and those who are not.
+
+When we introduce additional food fortification into the population, only those
+who are *not* already eating fortified food should receive the benefit.
+Therefore, in order to properly measure the effects of our interventions, we
+must first stratify our population into those who are already covered vs. not
+covered by the forticant of interest. We then need to calculate the risk
```
Ninicorn, line 168:
```suggestion
covered by the forticant of interest. We then need to calculate the prevalence of the risk
```

Ninicorn: since earlier it was referred to as prevalence

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
Ninicorn, line 169:
```suggestion
exposure (eg. vitamin A deficiency, haemoglobin means or other relevant outcome) in each coverage stratum to get a complete
```

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -127,17 +150,72 @@ PAF-of-1 Cause/Risk Pairs
...
 ++++++++++++++++++++++++++
 
-Coverage Gap Framework
-++++++++++++++++++++++
+Stratifying Exposure Variables by Baseline Intervention Coverage
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+From GBD we obtain overall population estimates for our outcomes of interest
+(prevalence of vitamin A deficiency, birth prevalence of neural tube defects,
+distribution of haemoglobin levels, and distribution of birth weights). Because
```
Ninicorn, line 158: is it distribution or means ?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -127,17 +150,72 @@ PAF-of-1 Cause/Risk Pairs
...
+and those who are not.
+
+When we introduce additional food fortification into the population, only those
+who are *not* already eating fortified food should receive the benefit.
+Therefore, in order to properly measure the effects of our interventions, we
+must first stratify our population into those who are already covered vs. not
+covered by the forticant of interest. We then need to calculate the risk
+exposure (or other relevant outcome) in each coverage stratum to get a complete
+picture of the baseline population. For example, we will need to know the
+prevalence of vitamin A deficiency or the distribution of haemoglobin in the
```
Ninicorn, line 171: do we use the means of the distributions?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
Ninicorn: I thought we are using GBD's paf and that's why we run into limitations cuz we are using an RR from one study population and the PAF from our country location from GBD?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -127,17 +150,72 @@ PAF-of-1 Cause/Risk Pairs
...
 
-Effect size stratification for baseline population
---------------------------------------------------
+  Come up with some examples to show what would happen if we *didn't* first
+  stratify by existing coverage. I.e., why is this baseline stratification
+  important? It is not totally obvious to me what the effect would be if we
+  didn't do it (e.g. would we end up overestimating or underestimating the true
+  effect?) -- I think the net effect depends on the level of existing coverage,
+  the shape of the underlying exposure distribution, and the magnitude of the
+  effect size.
```
Ninicorn, line 200: oh!! I have an example I've written out for myself for this. I can add it once this is merged as a new branch?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -127,17 +150,72 @@ PAF-of-1 Cause/Risk Pairs
...
+cases in separate sections below.
 
-This method applies to exposures with dichomotous outcomes such as Vitamin A deficiency or neural tube defects:
+Baseline Coverage Stratification -- Dichotomous Variables (Coverage Gap Framework)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-We always define the exposure as bad to match GBD 2017 definitions, so relative risks are always >1 
+This method applies to exposures with dichotomous measures such as Vitamin A deficiency or neural tube defects:
+
+We always define the exposure as bad to match GBD 2017 definitions for risk exposures, so relative risks are always >1.
```
Ninicorn, line 218:
```suggestion
We always define the exposure as deleterious to match GBD 2017 definitions for risk exposures, so relative risks are always >1.
```

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: I don't think we have GBD PAF's for the RR's we get from external literature for our interventions. Is there a specific example you're thinking of @Ninicorn ?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: I started editing things to just say "exposure" instead of "risk exposure" because it's not always risks we're talking about, but possibly other outcomes like neural tube defects.

But I can reword this sentence to explicitly say "exposures" instead of implicitly referring to them with "Their".

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Again, I was shying away from explicitly saying "risk" because we're not always talking about risks. Does that make things more confusing, or is it ok to just say "exposure" if I make sure to keep things consistent?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: I was avoiding saying "prevalence" except for specific examples, because the general strategy applies to continuous variables as well as discrete ones. For categorical variables the exposure distribution is specified by the prevalence in each category, but for continuous variables we need the probability density function, which is like a "prevalence density" at each exposure level. I could explicitly say "exposure *distribution*" here though.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: I give these examples in the next sentence. I'll reword the sentence as noted in the previous comment, to just say "exposure distribution" instead of "risk exposure," and then I won't need the parenthetical remark.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: We need the entire distribution, not just the means. Our current strategy is to assume that the stratified distributions have the same shape as the original distribution, but our current documentation is incomplete because it does not make this clear, it just explains how to get the correct means. That's why I'm rewriting this part, to try to make things more explicit.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Cool!

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Cool, thanks Ali.  I'll paste this into the todo so we can edit it later.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: @alibow I incorporated your description above, and added some additional comments in a todo.  I'm still a bit hazy about what exactly our assumptions mean and when they are appropriate, so I think it's worth continuing to flesh this out as much as possible.

PR #213, opened 2020-04-17
LRI birth prevalence

alibow:


docs/source/gbd2017_models/causes/lower_respiratory_infections/index.rst
```diff
@@ -72,20 +72,13 @@ There are three sequelae associated with LRI. We are not tracking the long-term
...
 
-  Initial validation of the LRI cause model has indicated that prevalence (and 
-  therefore cause-specific mortality) of LRI is significantly underestimated for 
-  the early neonatal and late neonatal age groups (age group IDs 2 and 3). This 
-  may be because DISMOD models allow for birth prevalence of LRI, although birth 
-  prevalence is not a reported outcome of GBD. This issue should be investigated 
-  to see if we can incorporate birth prevalence into the LRI cause model to avoid 
-  underestimation of LRI in the neonatal age groups. Note: according to GBD 
-  modeler (Chris Troeger) LRI birth prevalence can be found using age_group_id 164.
+   Describe more assumptions and limitations of the model.
```
Ninicorn, line 81: age group id 164 refers to birth.

Ninicorn: maybe good to add a note saying that the remission is longer than 6 days (the prior bound on remission in the models is 7.3 days), so birth prevalence and 0-6 day prevalence should be nearly identical. 





12:05
Yep, the prior bound on remission in the models is 7.3 days.

alibow: We don't see this in the data though I think because LRI has a high EMR... babies with LRI are dying in the first 6 days of life which causes LRI birth prevalence > LRI early neonatal prevalence

Ninicorn: oh awesome! would be nice to note.

docs/source/gbd2017_models/causes/lower_respiratory_infections/index.rst
```diff
@@ -167,6 +160,10 @@ Data Description
      - Sources
      - Description
      - Notes
+   * - birth_prevalence_meid_1258
+     - epi
+     - Birth Prevalence of LRI
+     - get_draws('modelable_entity_id', 1258, source='epi', age_group_id=164, measure_id=5)
```
NathanielBlairStahn, line 166: I think we need to add birth_prevalence to the State Data table as well as the Data Sources table. I.e. state I should have measure "birth prevalence" with value `birth_prevalence_meid_1258`, and state S should have measure "birth prevalence" with measure `1 - birth_prevalence_meid_1258`.

See the neural tube defects documentation for an example: https://vivarium-research.readthedocs.io/en/latest/gbd2017_models/causes/neural_tube_defects/index.html

docs/source/gbd2017_models/causes/lower_respiratory_infections/index.rst
NathanielBlairStahn: Agreed, that sounds like a good note to add.  Not sure where to put it -- maybe under Validation Criteria?

PR #214, opened 2020-04-18
new lbwsg branch with revisions

Ninicorn:


docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -51,3 +52,162 @@ described below, and relative to the joint TMREL.
...
+
+   |  i = low birth weight short gestation category
+   |  mr_i = mortality hazard in early and late neonatal period for category i
+   |  rr_i = relative risk for all cause mortality in category i
+   |  state = either 1 with condition or 0 without condition 
+   |  PAF* = this is the PAF of the most detailed cause affected by lbwsg
+
+Hence, the mortality hazard for an individual in category i is:  
+
+:math:`\sum\limits_{\text{salmon}}\text{CSMR} + (ACMR - \sum\limits_{\text{salmon}}\text{CSMR} - \sum\limits_{\text{green}}\text{CSMR} + \sum\limits_{\text{green}}\text{EMR_state})\cdot\text{(1-PAF*)}\cdot rr_i`
```
NathanielBlairStahn, line 165: I think it would be worth writing out a multi-line derivation of this equation, so the reader can easily follow all the steps and understand where each piece comes from. I'd be willing to work on that for a future PR -- I think the present version is fine for now.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -51,3 +52,162 @@ described below, and relative to the joint TMREL.
...
+
+
+.. important :: 
+
+   PAF in the above equation represents PAFs for most-detailed-cause (they are all roughly equal) affected by LBWSG (or as calculated in vivarium from LBWSG relative risks and exposure). This approach assumes that relative risks for LBWSG applies only to causes that GBD considers to be affected by LBWSG (green and blue causes).
```
NathanielBlairStahn, line 170: Why is this important?  What are the alternatives (e.g. what PAF would we use besides the one for most-detailed causes)?

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -51,3 +52,162 @@ described below, and relative to the joint TMREL.
...
++------------------+-------------------------------------------------------+-----------+
+| Restriction type | Value                                                 | Notes     |
++==================+=======================================================+===========+
+|  Male only       | False                                                 |           |
++------------------+-------------------------------------------------------+-----------+
+|  Female only     | False                                                 |           |
++------------------+-------------------------------------------------------+-----------+
+|  Age group       | early neonatal (0-6 days)                             | id 2      |
+|                  | late neonatal (7-28 days)                             | id 3      |
++------------------+-------------------------------------------------------+-----------+
```
NathanielBlairStahn, line 190: For a simple table like this, I think it would be better to use a list-table, because that is easier to reformat if something changes.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -51,3 +52,162 @@ described below, and relative to the joint TMREL.
...
+
+.. important :: 
+
+   PAF in the above equation represents PAFs for most-detailed-cause (they are all roughly equal) affected by LBWSG (or as calculated in vivarium from LBWSG relative risks and exposure). This approach assumes that relative risks for LBWSG applies only to causes that GBD considers to be affected by LBWSG (green and blue causes). 
+
+.. todo ::
+   link notebook that ali made on obtaining most-detailed-cause PAF lbwsg PAFs and abie's notebook on continuous conversion of the categories 
+
+
+Restrictions
```
NathanielBlairStahn, line 176: I think these restrictions are built into GBD, so this restrictions table should go in the GBD modeling section instead of the Vivarium modeling section.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -51,3 +52,162 @@ described below, and relative to the joint TMREL.
...
+|  382     | neonatal encephalopathy due to birth asphyxia and trauma|
++----------+---------------------------------------------------------+
+|  383     | neonatal sepsis and other neonatal infections           |
++----------+---------------------------------------------------------+
+|  384     | hemolytic disease and other neonatal jaundice           |
++----------+---------------------------------------------------------+
+|  385     | other neonatal disorders                                |
++----------+---------------------------------------------------------+
+|  686     | sudden infant death syndrome                            |
++----------+---------------------------------------------------------+
```
NathanielBlairStahn, line 126: Again, I think a list-table would be easier to format here, but perhaps it's not worth changing at this point.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -51,3 +52,162 @@ described below, and relative to the joint TMREL.
...
+
+Because the relative risks from GBD are for all-cause mortality in the early and late neonatal period, we first define all-cause mortality rate (ACMR) as the sum of: 
+
+   - mortality from causes that are affected by lbwsg and modelled in the sim (green)
+   - mortality from causes that are affected by lbwsg but not modelled in the sim (blue)
+   - mortality from causes that are unaffected by lbwsg, modelled or not modelled (salmon)
+
+An example of these causes from the large-scale-food fortification concept model diagram is shown below: 
+
+.. image:: causes_equation.svg
```
NathanielBlairStahn, line 145: I like how this table lets you visualize what's going on.  Nice job!

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -51,3 +52,162 @@ described below, and relative to the joint TMREL.
...
+.. note ::
+    That this is likely biasing towards overestimating extreme birthweights or gestational ages. For example, in the 0-500g category, most babies are probably pretty close to 500g, not equally probable to be 1 gram versus 499 grams.
+
+Because the relative risks from GBD are for all-cause mortality in the early and late neonatal period, we first define all-cause mortality rate (ACMR) as the sum of: 
+
+   - mortality from causes that are affected by lbwsg and modelled in the sim (green)
+   - mortality from causes that are affected by lbwsg but not modelled in the sim (blue)
+   - mortality from causes that are unaffected by lbwsg, modelled or not modelled (salmon)
+
+An example of these causes from the large-scale-food fortification concept model diagram is shown below:
```
NathanielBlairStahn, line 143: You could add a reference to make the text an actual link to the LSFF concept model.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
alibow: all-cause PAF, which is a different value. WE have to use the most-detailed PAF because this is the one that is calculated directly from the relative risks, which is what we want to apply to the entire group of LBWSG affected causes

Ninicorn: ill include link to the notebook that calculates the all-cause PAF and most-detailed-cause PAF

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: where do I find that how the links are referenced? for example I know its this: 

:ref:`intermediate <_where do I find how this is called?_>`

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: ya...maybe we can change the aesthetics of it later ....

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: how do I reference the notebook? 

:ref:lbwsg_paf_notebook <_where do I find how this is called?_>

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: ok

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
NathanielBlairStahn: > all-cause PAF, which is a different value. WE have to use the most-detailed PAF because this is the one that is calculated directly from the relative risks, which is what we want to apply to the entire group of LBWSG affected causes

I think some version of this explanation should appear within the `.. important::` directive, as well as linking to the notebook.  I'm not sure where the notebook is - @alibow is it somewhere in the vivarium_data_analysis repo?

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: I attempted to write out how the equation is derived.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
NathanielBlairStahn: Here's the Sphinx documentation for how to reference different locations within the documentation: https://www.sphinx-doc.org/en/1.5/markup/inline.html#cross-referencing-arbitrary-locations

If you go to the source code for the [LSFF concept model](https://vivarium-research.readthedocs.io/en/latest/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.html), you should see this after the comments at the top:

```
.. _2017_concept_model_vivarium_conic_lsff:

=============================================
Vivarium CoNIC Large Scale Food Fortification
=============================================

Model Overview
--------------

Objective
+++++++++

Intervention Definitions
++++++++++++++++++++++++

Questions of Interest
+++++++++++++++++++++

Scope of Modeling
+++++++++++++++++


Concept Model Diagram
---------------------

.. image:: lsff_concept_model_diagram.svg
```

The first line above shows you that the reference label for the title of the document is `2017_concept_model_vivarium_conic_lsff`. Thus, to link to the concept model document, you would do :ref:`large-scale-food fortification concept model <2017_concept_model_vivarium_conic_lsff>`. (I'm not sure how to escape backticks, so this is not quite rendering correctly...)

If you wanted to link to the concept model diagram directly, you would have to edit the concept model `.rst` document and add a new reference label before the diagram, e.g.

```
.. _lsff_concept_model_diagram:

Concept Model Diagram
---------------------

.. image:: lsff_concept_model_diagram.svg
```

And then you could link to it like :ref:`large-scale-food fortification concept model diagram <lsff_concept_model_diagram>`.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -51,3 +52,173 @@ described below, and relative to the joint TMREL.
...
+
+= :math:`\sum\limits_{\text{salmon}}\text{CSMR} + (ACMR - \sum\limits_{\text{salmon}}\text{CSMR} - \sum\limits_{\text{green}}\text{CSMR} + \sum\limits_{\text{green}}\text{EMR_state})\cdot\text{(1-PAF*)}\cdot rr_i`
+
+
+.. important :: 
+
+   PAF in the above equation represents PAFs for most-detailed-cause (they are all roughly equal) affected by LBWSG (or as calculated in vivarium from LBWSG relative risks and exposure). This approach assumes that relative risks for LBWSG applies only to causes that GBD considers to be affected by LBWSG (green and blue causes). 
+
+.. todo ::
+   link notebook that ali made on obtaining most-detailed-cause PAF lbwsg PAFs and abie's notebook on continuous conversion of the categories
```
NathanielBlairStahn, line 201: I think I have Abie's notebook somewhere - I'll look for it and make a PR in the vivarium_data_analysis repo if no one else has added it.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -51,3 +52,173 @@ described below, and relative to the joint TMREL.
...
+Assumptions and limitations
++++++++++++++++++++++++++++
+ 
+Strengths
+
+   o  This approach is consistent with GBD methodology and avoids artificially decreasing the mortality rate for individual causes that are not affected by improvements in LBWSG (due to reverse causality or other concerns).
+   
+Limitations
+
+   o  This approach applies the LBWSG relative risks in an inconsistent manner with what they represent (ratios of ACMRs). This implies that the relative risks may be different as they relate to the subset of causes that GBD considers affected by LBWSG rather than all causes.
```
NathanielBlairStahn, line 213: I had to re-read this a couple times to understand what it was saying. I think the point is that the RR's are supposed to represent _all cause_ mortality, but we are only applying them to _some_ of the causes (the ones GBD says are affected), correct?  I think it would be clearer if the "all vs. some" distinction were emphasized sooner, rather than at at the end of the comment.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -51,3 +52,173 @@ described below, and relative to the joint TMREL.
...
+   
+Notably, we are not sure which direction this may bias the results (We would need to evaluate stratified microdata results.) If the studies from which we obtain our intervention effect sizes includes mortality data due to causes that GBD considers unaffected by LBWSG, then we may be underestimating the impact of the intervention in our model.
```
NathanielBlairStahn, line 215: I think this could use a concrete example to help clarify what the potential problem is.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: Yes it is!! it's in the data analysis repo

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -51,3 +52,173 @@ described below, and relative to the joint TMREL.
...
+.. todo ::
+   link notebook that ali made on obtaining most-detailed-cause PAF lbwsg PAFs and abie's notebook on continuous conversion of the categories 
+
+   
+Assumptions and limitations
++++++++++++++++++++++++++++
+ 
+Strengths
+
+   o  This approach is consistent with GBD methodology and avoids artificially decreasing the mortality rate for individual causes that are not affected by improvements in LBWSG (due to reverse causality or other concerns).
```
NathanielBlairStahn, line 209: I think this comment (and several others) make the most sense in the context of knowing all the different possible approaches we were considering for modeling LBWSG, i.e. that document Ali wrote up describing three different approaches. Without that context, the reader is left wondering, "Well why would you do something that _does_ artificially decrease the mortality rate for individual causes?" or something similar.

I'm not sure if this is the right place to do it, but somewhere we should have a more complete explanation of the different approaches we considered taking, and why we went with this one.  Oh, and also some documentation of the problems we ran into with the PAFs and RRs (I've already forgotten what exactly what the problem was and how we decided to fix it).

All of that is perhaps out of scope for this PR, but something we should figure out how to incorporate.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: ok thanks!

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
NathanielBlairStahn: Great, then to reference it, you would just do :ref:`LBWSG PAF notebook <https://github.com/ihmeuw/vivarium_data_analysis/blob/master/pre_processing/lbwsg/LBWSG%20exposure%2C%20rrs%2C%20pafs.ipynb>` (or replace the text before <...> with whatever you want).

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: ok, I can add that to a todo and work on that.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: I've tried to make this clearer: In the risk appendix of GBD 2017, it says the relative risks for lbwsg apply to all-cause mortality rates. However, we are choosing to apply the RRs *only* to the causes that GBD believe lbswg affects (the list of affected causes). Our approach may be inconsistent with the GBD intended for the RRs.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
NathanielBlairStahn: Cool, thanks, that helps!

A question occurred to me while reading this:

Does this method require us to pull CSMRs for _every single cause_ in GBD _except_ for the "blue" ones (the un-modeled causes affected by LBWSG)? Because we're summing up CSMRs for all the "salmon" causes, which include _all_ of the unaffected causes that we are _not_ modeling? And if that's the case, why not just compute the sum of the blue CSMRs directly too, instead of computing it from ACMR, salmon, and green?

If we want to minimize the number of things we need to explicitly compute or pull data for, it seems like it would make more sense to solve for the salmon CSMRs in the original equation (instead of the blue ones), then write everything in terms of the green and blue CSMRs instead of the green and salmon ones.

Am I understanding this correctly?

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -51,3 +52,173 @@ described below, and relative to the joint TMREL.
...
+
+.. image:: causes_equation.svg
+
+All-cause mortality is sum of all cause-specific mortalities: 
+
+   ACMR =  :math:`\sum\limits_{\text{green}}\text{CSMR} + \sum\limits_{\text{blue}}\text{CSMR} + \sum\limits_{\text{salmon}}\text{CSMR}`
+
+The mortality from unmodelled causes affected by lbwsg (blue) is thus: 
+
+   :math:`\sum\limits_{\text{blue}}\text{CSMR}` = ACMR - :math:`\sum\limits_{\text{salmon}}\text{CSMR} - \sum\limits_{\text{green}}\text{CSMR}`
```
NathanielBlairStahn, line 168: Thanks for making this into an equation instead of just an expression :) I think that makes things easier to read.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
NathanielBlairStahn: Well, I think we _are_ using the RR's consistently with what GBD intended -- we're applying them exactly to the causes GBD says to apply them to.  The potential inconsistency is from the way the data for the RR's was collected -- if our understanding is correct, these RR's were computed for _all_-cause mortality rates, but then GBD excluded all the causes for which they couldn't plausibly establish causality, i.e. everything except the blue causes.

So yes, I think your above explanation is clearer, except for the last sentence.

Ninicorn: ah yes-ok, I think that is correct. So, the data is collected for all-cause mortality, then we back-trace to what could be the cause of this mortality and came up with a list of 15 things, excluding the reverse causality stuff.  Hence GBD (and us) are using the RRs inconsistent with the data.

Ninicorn: revised as: 

In the risk appendix of GBD 2017, it says the relative risks for lbwsg apply to all-cause mortality rates. However, we are choosing to apply the RRs *only* to the causes that GBD believe lbswg affects (the list of affected causes). Ours approach and GBD's may be inconsistent with what the RRs from the data are for.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: let me know if this hypothetical example makes sense and if what @ali meant:

for example: suppose we have a nutritional supplement that impacts lbwsg. This supplement was tested in an RCT in western Kenya where malaria is prevalent. Suppose there is some causal link in both directions between birthweight and malaria. While malaria causes lbw, there might also be some causal pathway that lbw babies are more susepticble malaria infection. If we improve birthweight in this population due to the supplement, we also decrease incidence of malaria, and decrease mortality from malaria. However, this will not captured in our model, so our modelled effect size might be less than the empirical effect of this supplement on neonatal mortality.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: salmon includes all the unaffected causes that we are both modelling and not modelling. 

I guess we should use:

 cause_group_with_the_most_causes = ACMR - cause_groupS_with_the_least_causes 

to save on computation ? 

Not sure why we decided to derive blue causes from ACMR rather than the other ones...

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: the :ref: doesnt work to link a website. Instead, I used: 

`lbwsg PAF notebook <https://github.com/ihmeuw/vivarium_data_analysis/blob/master/pre_processing/lbwsg/LBWSG%20exposure%2C%20rrs%2C%20pafs.ipynb>`__.

https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/WritingReST/CheatSheet.html#cheat-sheet-intersphinx

NathanielBlairStahn: Oh, right -- sorry, I got confused.  The :ref: syntax is specific to Sphinx, but the backtick ... backtic_ syntax is pure .rst

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
NathanielBlairStahn: Yeah, I don't know.  I didn't totally follow the discussion between @alibow  and @collijk about the different equation options...

alibow: I don't think we addressed this specific issue directly. I think that either way works and that whatever makes the most sense from a software perspective will work here.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
alibow: Hmm... I don't think that GBD is actually making the assumption that "all of the excess mortality that was measured in the RR's are theoretically all coming from these plausible blue causes"... rather they are making the conservative decision not to include these causes in their LBWSG PAF because they did not have enough evidence to conclude that they were causally related (because of concerns with reverse causality)... I think we just phrase this too strongly as is and instead could say something like "GBD evaluated the causes for which LBWSG causes increased mortality and concluded that this was the cases for the blue causes. They assumed that the relative risks for all-cause mortality rates by LBWSG category applied equally to mortality rates from each of these blue causes only and did not apply to any other GBD causes in order to calculate the population attributable burden due to LBWSG; in other words, the conservatively ignored the potential impact of LBWSG on mortality due to causes that did not meet their causal criteria."

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
NathanielBlairStahn: @Ninicorn I just made a PR here for converting LBWSG categories into a continuous distribution: https://github.com/ihmeuw/vivarium_data_analysis/pull/42

Feel free to merge it into master so you can link to the notebook.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -4,50 +4,251 @@
...
+
+Bias
+
+   Notably, we are not sure which direction this may bias the results (We would need to evaluate stratified microdata results.) If the studies from which we obtain our intervention effect sizes includes mortality data due to causes that GBD considers unaffected by LBWSG, then we may be underestimating the impact of the intervention in our model.
+
+    o   One way of bias could be from not including the reverse-causality causes: suppose we have a nutritional supplement that impacts LBWSG. This supplement was tested in an RCT in western Kenya where malaria is prevalent. Suppose there is some causal link in both directions between birthweight and malaria. While malaria causes lbw, there might also be some causal pathway that lbw babies are more susepticble to malaria infection. If we improve birthweight in this population due to the supplement, we also decrease incidence of malaria, and decrease mortality from malaria. However, this effect through malaria will not be captured in our model, so our modelled effect on neonatal mortality might be less than the empirial effect of this supplement on neonatal mortality.
```
NathanielBlairStahn, line 225: This seems to be glossing over some details a bit, which could be confusing:
> While malaria causes lbw, there might also be some causal pathway that lbw babies are more susepticble to malaria infection.

When we say "malaria causes lbw," presumably we mean malaria in the pregnant _mother_ causes a low birth weight _baby_ (unless you're saying that if a fetus somehow contracts malaria in utero, they are more likely to be low birth weight - but based on some preliminary googling, it seems pretty rare for babies to contract malaria in utero). So I think what's missing is an assumption that malaria in the mother could also cause malaria in the baby, either in utero or, more likely, during the birth (which there seems to be some evidence for).  With that extra assumption, it makes sense that then we wouldn't be able to distinguish which was the cause and which was the effect if a low birth weight baby dies of malaria.  But as written, it is unclear whether we are talking about malaria in the mother or malaria in the baby, which is confusing.

Ninicorn: what I am saying here is that both have the potential to happen- one is malaria in the mother caused a lbw baby (reverse causality). then the other is that a lbw baby caused it to have malaria (don't know if there is actual evidence for this, but lets say there is...like it makes it more likely the baby will get an infectious bite and develop disease). We do not model malaria as an affected cause but in this case lbw caused malaria and which caused increased mortality. I'll try to clarify a bit.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -4,50 +4,251 @@
...
+Assumptions and limitations
++++++++++++++++++++++++++++
+ 
+Strengths
+
+   o  This approach is consistent with GBD methodology and avoids artificially decreasing the mortality rate for individual causes that are not affected by improvements in LBWSG (due to reverse causality or other concerns).
+   
+Limitations
+
+   o  The risk appendix of GBD 2017 says that the data available to compute the relative risks for the risk exposure LBWSG are for the outcome of all-cause mortality. GBD then evaluated the relative risk of all-cause mortality across all available sources.  Based on criteria of biologic plausibility, a list of causes were selected. Some causes, most
```
alibow, line 218:
```suggestion
   o  The risk appendix of GBD 2017 says that the data available to compute the relative risks for the risk exposure LBWSG are for the outcome of all-cause mortality. GBD then evaluated the relative risk of all-cause mortality across all available sources.  Based on criteria of biologic plausibility, a list of causes for which LBWSG causally impacts mortality due to that cause were selected. Some causes, most
```

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -4,50 +4,251 @@
...
+   notably congenital birth defects, haemoglobinopathies, malaria, and HIV/AIDS, were excluded based on the criteria that reverse causality could not be excluded. GBD assumed that the relative risks for all-cause mortality rates by LBWSG category applied equally to mortality rates from each of these blue causes only and did not apply to any other GBD causes in order to calculate the population attributable burden due to LBWSG; in other words, the conservatively ignored the potential impact of LBWSG on mortality due to causes that did not meet their causal criteria. We are choosing to apply the RRs only to this list of LBWSG-affected causes. We believe this is consistent with GBD's approach but may not fully reflect what the RRs capture.
+
+Bias
+
+   Notably, we are not sure which direction this may bias the results (We would need to evaluate stratified microdata results.) If the studies from which we obtain our intervention effect sizes includes mortality data due to causes that GBD considers unaffected by LBWSG, then we may be underestimating the impact of the intervention in our model.
```
alibow, line 223:
```suggestion
   Notably, it is uncertain if this approach will cause an exaggeration or underestimation of the impact of LBWSG on mortality in the neonatal age groups because it requires an evaluation of the relative risks of mortality by LBWSG exposure category stratified by affected and unaffected causes and these data are not readily available to us.
```

alibow: I edited the first sentence for clarity and took out the second sentence because I generally the intervention effect sizes will be able the intervention's impact on birth weight, not on the impact of birth weight on mortality (this was an issue I was initially confused about and should have been taken out of my word-doc). Let me know if this does or does not make sense to you guys!

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -4,50 +4,251 @@
...
+    :math:`RR_{acmr}` = :math:`\frac{\text{(LBW_deaths_affected + LBW_deaths_unaffected)/LBW_births}}{\text{(NBW_deaths_affected + NBW_deaths_unaffected)/NBW_births}}`
+
+                      = :math:`\frac{\text{(LBW_deaths_affected + LBW_deaths_unaffected)}}{\text{(NBW_deaths_affected + NBW_deaths_unaffected)}} \times \frac{\text{NBW_births}}{\text{LBW_births}}`
+
+    :math:`RR_{aacmr}` = :math:`\frac{\text{LBW_deaths_affected/LBW_births}}{\text{NBW_deaths_affected/NBW_births}}`
+
+                       = :math:`\frac{\text{LBW_deaths_affected}}{\text{NBW_deaths_affected}} \times \frac{\text{NBW_births}}{\text{LBW_births}}`
+
+   Since we do not know the ratio of the number of :math:`\text{LBW_deaths_unaffected}` to the number of :math:`\text{NBW_deaths_unaffected}`, we do not know the direction of bias. We would need to analyse the stratified microdata. 
+
```
alibow, line 246: Maybe we should additionally note that because we are applying the same RR_ACMRs to *all* affected causes, we are not able to evaluate the impact of LBWSG on *cause-specific* mortality accurately. Otherwise this all looks good to me!

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -4,50 +4,251 @@
...
+Risk Exposure Model Diagram
++++++++++++++++++++++++++++
+
+Data Description Tables
++++++++++++++++++++++++
+
+Validation Criteria
++++++++++++++++++++
```
alibow, line 254: We could just quickly note here that our baseline scenario should compare to GBD data with regard to:

- LBWSG exposure categories (note that maybe we want to consider a proxy for this so that we don't need to observe person time in each category... perhaps mean BW or mean RR?)
- All cause mortality rates in the early neonatal and late neonatal categories

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -4,50 +4,251 @@
...
+
+    :math:`RR_{acmr}` = :math:`\frac{\text{(LBW_deaths_affected + LBW_deaths_unaffected)/LBW_births}}{\text{(NBW_deaths_affected + NBW_deaths_unaffected)/NBW_births}}`
+
+                      = :math:`\frac{\text{(LBW_deaths_affected + LBW_deaths_unaffected)}}{\text{(NBW_deaths_affected + NBW_deaths_unaffected)}} \times \frac{\text{NBW_births}}{\text{LBW_births}}`
+
+    :math:`RR_{aacmr}` = :math:`\frac{\text{LBW_deaths_affected/LBW_births}}{\text{NBW_deaths_affected/NBW_births}}`
+
+                       = :math:`\frac{\text{LBW_deaths_affected}}{\text{NBW_deaths_affected}} \times \frac{\text{NBW_births}}{\text{LBW_births}}`
+
+   Since we do not know the ratio of the number of :math:`\text{LBW_deaths_unaffected}` to the number of :math:`\text{NBW_deaths_unaffected}`, we do not know the direction of bias. We would need to analyse the stratified microdata.
```
NathanielBlairStahn, line 245: @Ninicorn @alibow  Do you have any sense of whether we expect

(LBW_deaths_unaffected / NBW_deaths_unaffected) < (NBW_deaths_affected / NBW_deaths_affected)

or the reverse inequality?

Like you say, we can't know for sure without looking at the data, but to me the above inequality seems more likely than the reverse, but I could be wrong.  If the above inequality is true, it implies RR_acmr < RR_aacmr (I did some algebra to check this).

alibow: I would agree with you if the unaffected categories did not include the reverse causality causes which seem to significantly complicate the story here. So I'm not really comfortable guessing which is more likely without doing some real digging in the lit, but we could put a todo note about looking into this further.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: ok!

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -4,50 +4,251 @@
...
+
+   
+Assumptions and limitations
++++++++++++++++++++++++++++
+ 
+Strengths
+
+   o  This approach is consistent with GBD methodology and avoids artificially decreasing the mortality rate for individual causes that are not affected by improvements in LBWSG (due to reverse causality or other concerns).
+   
+Limitations
```
NathanielBlairStahn, line 216: I think Ali mentioned a separate limitation yesterday, which was that GBD assumed that the RR's for CSMR for each affected cause were the same as the overall RR for ACMR. This won't matter for the blue causes that we aren't modeling explicitly, but for the green causes that we _are_ modeling, it could throw off our results depending on whether the RR's for that cause is larger or smaller than the overall RR for all causes.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: ok

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
NathanielBlairStahn: What I'm saying is that "malaria in the mother causes LBW baby" is **not** the reverse of "Being LBW causes the baby to get malaria," because the malaria is happening in two different people.  There are three outcomes to consider here, not just two.  It seems like the only way we can call this "reverse causality" is if we also say that "malaria in the mother causes malaria in the baby."  Or another way to say it might be that malaria in the mother is a confounder for the effect of LBW on malaria in the baby.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: can we just look at the birth prevalence of each exposure lbswg category?

alibow: Yeah we can, there are just so many that it might be too computationally intense so I thought it could be good to list some potential alternatives

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
alibow: Yeah. I think we could say something like hemoglobinopathies may cause a baby to be LBWSG AND it is also possible that a neonate is more likely to die from hemoglobinopathies if they are LBWSG.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: i just added that the bias is between our models (plus GBDs) and real life.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: good point, have clarified now!!

PR #215, opened 2020-04-22
LSFF improve baseline calibration equations

alibow:


docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1257,48 +1257,36 @@ effect size will then be used to determine the simulant's LBWSG risk category.
...
-
-Step 1: Solve for ∆_π_0 and ∆_π_1 using the equations above
+.. math::
 
-Step 2: Apply ∆_π_0 as the additive effect size for simulants covered by
-baseline coverage and ∆_π_1 as the additive effect size for simulants not
-covered by baseline coverage (apply in the same fashion as described in the
-sections above; notably, this is only done in the iron_responsive_i = 1
-population for the hemoglobin outcome).
+  π_i =	\hat{π}_{GBD(i)} - αp + αx_i
```
NathanielBlairStahn, line 1264: The notation GBD(i) makes it seem like GBD is a function of i, which is weird.  If you're following James's notation, it should be `\hat{π}_{GBD}(i)`, though James's notation is also a bit weird.  The hats should be unnecessary.  Why not:

`Y_i = Y_GBD - \alpha p + \alpha x_i`,

where

Y_i = individual hemoglobin level or birthweight
Y_GBD = hemoglobin level or birthweight drawn from the distribution in GBD

and \alpha, p, and x_i are defined as below.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1257,48 +1257,36 @@ effect size will then be used to determine the simulant's LBWSG risk category.
...
+  * - :math:`p`
+    - Proportion covered by iron fortification in the baseline scenario
+  * - :math:`π_{GBD}`
+    - Mean birthweight or hemoglobin from GBD
+  * - :math:`α`
+    - Effect size (age-dependent, sampled from distribution as described above)
+  * - :math:`π_{i}`
+    - Individual simulant birthweight/hemoglobin value
+  * - :math:`\hat{π}_{GBD}`
+    - Distribution of birthweight/hemoglobin according to GBD
```
NathanielBlairStahn, line 1283: We don't actually need to refer to the distribution explicitly. But we *do* need to define the random variable drawn from that distribution, which was `\hat{π}_{GBD}(i)` in James's notation, or `Y_i^\text{GBD}` in my notation above.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: I think that the hat notation by convention in statistics denotes a sample of a given population parameter, so the idea there is that it is a given draw GBD distribution but I can definitely clarify that. 

Then \hat{π}_{GBD(i)} would be the value for an individual simulant taken from a given draw of the GBD distribution.

So given that I think my preference is keeping the hat notation, but let me know if you still prefer the Y. I can clarify the existing notation to match what I put in this comment and also change \hat{π}_{GBD(i)} to \hat{π}_{GBD}(i)

NathanielBlairStahn: Or I guess to be consistent for our notation for random variables, we should either use X,Y (both capital) or x,y (both lower case).

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1257,48 +1257,36 @@ effect size will then be used to determine the simulant's LBWSG risk category.
...
+  :widths: 5 15
+  :header-rows: 1
 
-  Improve formatting/layout in this section
+  * - Parameter
+    - Definition
+  * - :math:`p`
+    - Proportion covered by iron fortification in the baseline scenario
+  * - :math:`π_{GBD}`
+    - Mean birthweight or hemoglobin from GBD
```
NathanielBlairStahn, line 1277: We don't actually need to refer to the mean in order to tell the software engineers what to do, just to prove that our approach does what we expect.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1257,48 +1257,36 @@ effect size will then be used to determine the simulant's LBWSG risk category.
...
+  * - :math:`π_{i}`
+    - Individual simulant birthweight/hemoglobin value
+  * - :math:`\hat{π}_{GBD}`
+    - Distribution of birthweight/hemoglobin according to GBD
+  * - :math:`x_{i}`
+    - Whether an individual simulant is covered (1=yes, 0=no)
+
+See the proofs for this approach below.
+
+.. image:: baseline_calibration_proofs.png
```
NathanielBlairStahn, line 1289: I could write the proof up in LaTeX so we don't have to use an image.

alibow: Sure! I just did this to save time because it was already nicely formatted instead of having to copy over the latex code.

alibow: Feel free to work directly on this branch and add this in and change the variable notation to whatever makes sense from your perspective. I kept the notation consistent in my table with Jame's so that I could use the screenshot, but it can be edited to whatever makes sense :)

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Haha, probability vs. statistics. Wait, I thought a hat denoted an _estimate_ of a population parameter from data.  E.g. if pi_{GBD} is the mean hemoglobin in the population, then \hat{pi}_{GBD} would be an estimate of the population mean, e.g. by measuring Hb levels in 100 people and then computing the average.

https://en.wikipedia.org/wiki/Hat_operator#Estimated_value
https://en.wikipedia.org/wiki/Notation_in_probability_and_statistics#Statistics

But that's not what we're doing here - we're sampling from a distribution with mean pi_{GBD} (well actually, this would be \hat{pi}_{GBD}, because GBD's value is an estimate of the true population parameter), and we don't expect the sampled value to be near the mean, but rather to vary according to the ensemble distribution.

alibow: Yes! I've generally heard /hat{p} referred to as the sample proportion and p as the population or parameter proportion (generally because you take a *sample* of the population to make an estimate for that population parameter in epi). So I am mixing my "sample" vocab, but I think the concept is similar in that the sample/estimate is a value that is close to the "true value."

Overall I think either way is fine as long as we get the right equation documented. Are you planning to make an updated commit with the latex equations instead of the screenshot or should I plan to make some clarifying edits and keep as is?

NathanielBlairStahn: Why don't you go ahead and keep James's notation for now (just move the (i) out of the subscript) and keep the image, and I'll make a new PR to edit things after this is merged into master.

The main thing I think needs to be fixed before merging is to remove the definition of `\hat{π}_{GBD}` from the table below, and replace it with the definition of `\hat{π}_{GBD}(i)` (see my comment below).

Ninicorn: should we have a separate repo page to document our standard notations so we are consistent and clear moving forward how to represent things?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: I think it's relatively low priority to change the image to text or to change the notation, so let's go ahead and use the current version, after fixing the definition of the GBD Hb level/birthweight drawn for an individual (see comments above).

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Ok, it is updated now : )

NathanielBlairStahn: @alibow I don't see the updates.

With regard to notation, from an epidemiology perspective, the hemoglobin level or birthweight (what I called Y_i) would be an exposure or outcome, depending on context. So is the indicator x_i for whether someone gets fortification.  @alibow and @Ninicorn, is there any standard epi convention for naming variables?

I think I've see capital letters used, which matches the convention in probability for random variables, which is how you mathematically model these things, so that's why I chose "Y" above.  In Python the naming convention is to use capital letters for constants or class names, and lower case letters for variables, so that's probably why James uses lower case letters like x_i.  But for this documentation we're aiming for readability of text, not code, so we should choose variable naming conventions that make sense to us.

alibow: Oops, sorry! My push didn't go through. It is updated now.

alibow: And yeah, I'd say it varies somewhat but generally is expressed as
Y_i = B0 + B1 * x1_i + B2 * x2_i + ....

where 
Y = outcome
B0 = intercept
B1 = slope for exposure
X1 = exposure
B2 = slope for first covariate
X2 = first covariate

^ This is generally for regression modelling, though.
In more conceptual contexts (like causal diagrams) it's common to see E for exposure and O for outcome and other intuitive letters for additional variables.

It really doesn't matter to me what variable notation we use as long as they are always clearly defined :)

PR #216, opened 2020-04-22
LSFF add effect size randomness instructions

alibow:


docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -391,6 +391,21 @@ below.
   This will make the mean slightly larger than it should be, but so will any
   strategy we pick.
 
+Model Randomness
+~~~~~~~~~~~~~~~~
```
NathanielBlairStahn, line 395: Is this supposed to be a subsection of "Stratifying Exposure Variables by Baseline Intervention Coverage" (which it currently is), or a separate section (in which case the section level should be +++++++)?

Ninicorn: I was just thinking whether we fixed this for the BEP costs in our bep model?

alibow: Good catch! I updated to "+++++"

PR #218, opened 2020-04-24
Vivarium best practices initial draft

alibow:


docs/source/model_design/concept_models.rst
```diff
@@ -0,0 +1,18 @@
...
+
+- Common random numbers 
+		- Between model scenarios
+		- Between model locations
+		- Between model parameters
+- Statistical distributions of uncertainty
+	- Uniform
+	- Normal/Truncated normal
+	- Beta
+	- Ensemble
```
NathanielBlairStahn, line 17: For the section on probability distributions, I was thinking it would be useful to organize it according to what we are using them for rather than by distribution type. For example:
* What types of distributions could I use to model uncertainty in
    - a relative risk?
    - a mean difference?
    - a proportion?
    - a cost estimate?

And so on.  And then also address questions like:
* What do I do if someone gives me a confidence interval that's very asymmetric (this can pose problems because you may not be able to find a distribution that matches your assumptions well)?
* How do we think about modeling uncertainty based on belief, with very little data available, vs. uncertainty based on combining multiple estimates from published papers?
* How do I model uncertainty if I am extrapolating a subnational estimate to use as a national estimate?

And whatever other methodological questions we run into.

For quick reference, it may also be useful to compile a list of distributions we find ourselves commonly using, link to the Wikipedia page for a description of the distribution, and list some common use cases for each one.

NathanielBlairStahn: Another important piece is choosing a strategy for coupling the parameters you're modeling, i.e. specifying an appropriate joint distribution, not just the marginals.  This is related to the bit about common random numbers, but I think it's a more general problem (I haven't thought carefully about how to do this in a systematic way within the Vivarium framework).

alibow: I added an outline for all of these topics in the latest commit. Thanks for the ideas!

docs/source/model_design/concept_models.rst
```diff
@@ -0,0 +1,245 @@
...
+Other Considerations
+~~~~~~~~~~~~~~~~~~~~
+
+.. todo:: 
+
+	- How to handle very asymmetric confidence intervals
+	- How to handle uncertainty in data source(s) rather than statistical uncertainty from a single high quality data source?
+		- Ex: combining multiple estimates from published papers with their own statistical uncertainty
+	- How to handle uncertaity when extrapolating a subnataional estimate to a national estimate?
+	- How to handle uncertainty distribution in the case of joint distributions
```
NathanielBlairStahn, line 83: I think this might deserve to be its own section, because we are *always* modeling joint distributions -- if there is more than one random parameter in our model, then they must have a joint distribution.  The "default" is to use the joint distribution where everything is independent, unless we specifically tell the software engineers to couple some of the parameters in some other way.  But I think we should be putting more emphasis on thinking about what joint distribution we are actually using, rather than relegating it to an afterthought.

docs/source/model_design/concept_models.rst
```diff
@@ -0,0 +1,245 @@
...
+
+Double Counting Issues
+++++++++++++++++++++++
+
+.. todo::
+
+	This section
+
+Incompatibilities Between Disease States Needed for Model and GBD Disease States
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
```
NathanielBlairStahn, line 200: Is this section underline long enough?

docs/source/model_design/concept_models.rst
alibow: Ah, are you referring to the total effect of the uncertainty intervals for all of the parameters in our model? I would classify this concept as uncertainty in monte carlo analysis, which has its own section currently (at the end!). Is that what you mean or am I misinterpreting?

docs/source/model_design/concept_models.rst
alibow: Fixed

docs/source/model_design/concept_models.rst
davidbernotas: Typically we model parameters, risk factors, etc. to be independent.  As a result, the joint distribution exists, but it's just the product of the marginal distributions, and uncertainty is, obviously, independent.  As such, I don't see added value in interpreting this case as a "joint" distribution, but maybe I'm not looking far enough ahead.
If we have conditionally dependent variables, however, this gets more complicated.  My understanding is that we are just beginning to include things like correlated risks into the simulation models.

PR #222, opened 2020-04-29
revised lbwsg equation to include modelled unaffected causes EMR state

Ninicorn:


alibow:
This looks good to me! It would be good to hear from Kjell too that this is everything he needs!

PR #225, opened 2020-05-01
LSFF: add pseudocode summary of iron intervention

NathanielBlairStahn:


docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1314,6 +1314,62 @@ See the proofs for this approach below.
...
+  # At beginning of simulation
+  mother_fortified_i = "unknown"
+  mother_fortified_baseline_i = "unknown" # For stratification
+  birth_weight_i = birthweight_gbd_i
+  child_fortified_i(0) = p_i < coverage(0)
+  hb_i(0) = hb_gbd(age_i(0)) + (-coverage(0)+child_fortified_i(0)) * hb_shift *  hb_age_fraction(age_i(0))
+
+  # When simulant is born in the simulation
+  mother_fortified_i = p_i < coverage(t - gestational_age_i + 20_weeks)
+  mother_fortified_baseline_i = p_i < baseline_coverage(t - gestational_age_i + 20_weeks) # For stratification
```
alibow, line 1364: We might need to put a note here about what to do if a baby has a gestational age less than 20, which should be rare, but is possible based on our exposure distribution.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1314,6 +1314,62 @@ See the proofs for this approach below.
...
+  bw_shift_i = bw_response * iron_concentration * daily_flour_i
+  birth_weight_i = birthweight_gbd_i + (-coverage(t)+mother_fortified_i) * bw_shift_i
```
alibow, line 1366: birth_weight_i = birthweight_gbd_i + (-coverage(t)+mother_fortified_i) * bw_shift_i

This looks appropriate to me

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1314,6 +1314,62 @@ See the proofs for this approach below.
...
+
+  # At each simulation time step (including when simulant is born)
+  child_fortified_i(t) = p_i < coverage(t)
+  time_covered_i = argmin_t(child_fortified_i(t) == True) # =float('inf') if simulant never gets covered
+  hb_i(t) = hb_gbd(age_i(t)) + (-coverage(t)+child_fortified_i(t)) * hb_shift *  hb_age_fraction(age_i(t)) * hb_lag_fraction(t-time_covered_i)
```
NathanielBlairStahn, line 1371: @aflaxman @alibow I think this equation has an error. I think instead it should be:

`hb_i(t) = hb_gbd(age_i(t)) + (-coverage(t)+child_fortified_i(t) * hb_lag_fraction(t-time_covered_i)) * hb_shift *  hb_age_fraction(age_i(t))`

Basically, this changes `child_fortified_i(t)` from a step function, which jumps immediately from 0 to 1 at `t=time_covered_i`, to a continuous function that ramps up from 0 to 1 over 6 months, starting at `t=time_covered_i`.

I think the current version jumps from `hb_gbd(age_i(t)) - coverage(t) * hb_shift *  hb_age_fraction(age_i(t))` to `hb_gbd(age_i(t))` at time `t=time_covered_i`, then ramps up to the full effect size over 6 months from that point.  That doesn't seem like what we want. This piece is supposed to implement the hemoglobin effect described in the section `Determining Whether A Simulant is Affected - Iron`.

I'm going to go ahead and make this change before merging to master. Since you already approved the PR, let me know if you think the new version looks correct.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Good point. Do we also need to worry about making sure the mother gets iron for at least 7 weeks? E.g. if they are born between 20 - 27 weeks.  I'll add a piece to try to handle those situations.

NathanielBlairStahn: @alibow So basically, should we just say that `mother_fortified_i = False` in either of the two situations above, i.e. if the gestational age is less than 27?  What if the mother had been eating iron-fortified food for the entire pregnancy up to that point?  Is it only the critical window 20-27 weeks that counts, or would, e.g. 19-26 weeks count as well?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Hmm, actually I think the issue here is that we are not differentiating between baseline and intervention coverage.

My thought is that baseline coverage should be defined as:
hb_baseline_i(t) = hb_gbd(age_i(t)) + (-coverage(t)+child_fortified_i(t)) * hb_shift *  hb_age_fraction(age_i(t)) 
^ No need to include the hb_lag_fraction here because we are assuming that time since coverage is sufficient for everyone.

And intervention coverage should be defined as:
hb_intervention_i(t) = hb_baseline_i(t) + new_coverage_i * hb_shift * hb_age_fraction(age_i(t)) * hb_lag_fraction(t-time_covered_i)

Where new_coverage_i is an indicator variable for delta coverage (would need to be a function of time).

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Hmm, I'm not quite sure about that, but based on the Haider studies (https://drive.google.com/drive/u/1/folders/1X3Mx8mmPBuAi7wjTMppWuoSjxiZ64-S0), I think we should not require that the critical range is 20-27 weeks, but rather than fortification exposure needs to occur at least before 20 weeks of gestation (although we could go up to 28 weeks, which is the latest start in the Haider review) and require that duration of exposure before birth is at least 7 weeks.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Should your baseline coverage equation be:
`hb_baseline_i(t) = hb_gbd(age_i(t)) + (-baseline_coverage(t)+child_fortified_baseline_i(t)) * hb_shift * hb_age_fraction(age_i(t))` ?

If so, I think the new version I wrote down is equivalent to this and works in both the baseline and intervention scenarios, because if the child is covered in baseline, then `time_covered_i = float('-inf')` and hence `hb_lag_fraction(t-time_covered_i) = `hb_lag_fraction(float('inf')) = 1`.  But perhaps it would be clearer to separate out the baseline level explicitly as you suggest...

alibow: Oh, ok! I didn't understand that at first pass. And yes, about child_fortified_baseline_i(t) in my equation.

If those equations are equivalent then I think it looks good!

NathanielBlairStahn: Actually, my version was wrong because I used `coverage(t)` instead of `baseline_coverage(t)`. This is corrected in the latest commit.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Actually, it turns out there were two problems with this:
1. The equation should use `baseline_coverage(t)` instead of `coverage(t)`. This is corrected in the latest commit.
2. James says the downward shift to the treatment-deleted birthweight should be a population-level parameter rather than an individual-level parameter - i.e. it should be 
`birth_weight_i = birthweight_gbd_i - coverage(t) * average_bw_shift + mother_fortified_i * bw_shift_i`,
where `average_bw_shift` is computed by taking the mean over the population distribution of `daily_flour_i`. I'm not actually sure why this matters, but I added a Todo to change this.

PR #226, opened 2020-05-09
LSFF: Add draft of tables for desired outputs

NathanielBlairStahn:


docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1846,6 +1846,62 @@ The procedure here uses the standard coverage gap framework:
...
+  :header-rows: 1
+
+  * - Priority
+    - Definition
+  * - High
+    - Needed for client
+  * - Med
+    - Needed for primary verification & validation
+  * - Low
+    - Needed for additional V&V or research
```
alibow, line 1860: I wonder if there should be an additional column in this table that indicates what stage of the model building process an output for V&V is needed. I know we had previously discussed that once we validated a cause model that it likely isn't necessary to report the state person time for that cause any more. I'm not sure what the best way to include this information might be, but maybe some columns that have the status of V&V for each model build by the software engineers and then we can update the table as "verified; no longer needed"? And others that we might want to keep in as "verified; still needed". Just a thought! Otherwise I think this looks good.

NathanielBlairStahn: Interesting idea.  I added your comment to the Todo so maybe we can incorporate it later.

PR #227, opened 2020-05-12
DAGs, with backdoor paths and colliders

alibow:


docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
 Directed Acyclic Graphs (Causal Diagrams)
 ------------------------------------------
 
-.. todo::
+.. important::
+
+  The figures represented in this section represent directed acyclic graphs/causal diagrams. These figures are distinct from compartamental model diagrams (such as SIR models). For our purposes we will represent directed acyclic graphs as *circles* connected by arrows and compartamental model diagrams as *squares* connected by arrows.
```
beatrixkh, line 684: could you clarify here if 'for our purposes' applies just to this section, or to the full repo?

docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
...
+
+  Additionally, for the purposes of directed acyclic graphs, variables will be labeled with capital letters (A, B, C, etc.) and arrows will be labeled with lower case letters (a, b, c, etc.).
+
+Introduction to Directed Acyclic Graphs
++++++++++++++++++++++++++++++++++++++++
+
+Directed acyclic graphs (DAGs), also termed causal diagrams, are composed of variables that are linked together by directional arrows. Critically, the solid arrows in DAGs represent a causal relationship between an exposure (start of the arrow) and outcome (node of the arrow), representing the *directional* aspect of DAGs. Notably, a directional arrow in a DAG does not offer any information about whether the exposure causes in increased or decreased probability of the outcome, only that there is some causal realtionship bewteen the exposure and outcome.
```
beatrixkh, line 691:
```suggestion
Directed acyclic graphs (DAGs), also termed causal diagrams, are composed of variables that are linked together by directional arrows. Critically, the solid arrows in DAGs represent a causal relationship between an exposure (start of the arrow) and outcome (node of the arrow), representing the *directional* aspect of DAGs. Notably, a directional arrow in a DAG does not offer any information about whether the exposure causes an increased or decreased probability of the outcome, only that there is some causal realtionship bewteen the exposure and outcome. 
```

docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
...
+
+Introduction to Directed Acyclic Graphs
++++++++++++++++++++++++++++++++++++++++
+
+Directed acyclic graphs (DAGs), also termed causal diagrams, are composed of variables that are linked together by directional arrows. Critically, the solid arrows in DAGs represent a causal relationship between an exposure (start of the arrow) and outcome (node of the arrow), representing the *directional* aspect of DAGs. Notably, a directional arrow in a DAG does not offer any information about whether the exposure causes in increased or decreased probability of the outcome, only that there is some causal realtionship bewteen the exposure and outcome. 
+
+The *acyclic* nature of DAGs is that there cannot be feedback loops between any variables in the graph. In this way, the arrows in DAGs can be interpreted as the passage of time so that the exposure (cause) must always occur temporally *before* the outcome (effect). See the diagram below for an example of an inappropriate and appropriate DAG with regard to temporality.
+
+.. image:: acyclic_dags.svg
```
beatrixkh, line 695: This is really clear, i like this diagram

docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
...
+
+Applications, Assumptions, and Limitations of Directed Acyclic Graphs
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+DAGs are used in epidemiology to represent a network of relationships between several variables, which in turn helps to guide the selection of variable that will need to be measured or considered in a given study/project. Further, the exact relationship between variables in a DAG will dictate *how* to treat or consider a specific variable in that study/project by helping to indentify potential confounding, intermediate, or mediating variables between a given exposure and outcome of interest (see the `Causal Relationships`_ section for definitions).
```
beatrixkh, line 700:
```suggestion
DAGs are used in epidemiology to represent a network of relationships between several variables, which in turn helps to guide the selection of variables that will need to be measured or considered in a given study/project. Further, the exact relationship between variables in a DAG will dictate *how* to treat or consider a specific variable in that study/project by helping to indentify potential confounding, intermediate, or mediating variables between a given exposure and outcome of interest (see the `Causal Relationships`_ section for definitions).
```

docs/source/model_design/general_research.rst
beatrixkh: When you describe what solid arrows represent, is that universal, or highly common, or just for the DAGs we will be using?

docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
...
+Directed paths are those in which variables are connected through arrows that flow all flow in the same direction. See below for an example of a directed path between variables A and D.
+
+.. image:: directional_dag.svg
+
+**In the case of a directed path between two variables, those variables will be causally related.**
+
+Undirected Paths
+~~~~~~~~~~~~~~~~
+
+Undirected paths are those in which variables are connected through arrows that do not all flow from head to tail. The two cases of undirected paths covered here include backdoor paths and colliders.
```
beatrixkh, line 723: Maybe, 'do not all flow in the same direction', because each individual arrow flows in it's own defined direction? (that direction being, in my head, is from tail to head; though perhaps that's not a point of contention to anyone else, haha)

docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
...
+.. list-table:: DAG Pathways
+   :widths: 5 5 5 5 5
+   :header-rows: 1
+
+   * - Path Name/Term
+     - Description
+     - Direction
+     - Open or closed
+     - Association
+   * - Direct path
```
beatrixkh, line 757: Is this a 'direct' or a 'directed' path, or are both terms used? if it's one or the other, i think it would be helpful to make this chart align with the section from line 711-714!

docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
...
+
+.. image:: unknown_var_association.svg
+
+The second approach is to represent the association between the two variables with a *dashed* line. For our purposes, dashed lines in DAGs will represent a non-causal association between two variables as shown in the figure below. While this format does not allow us to easily identify open backdoor paths in the way described in this section, it still allows us to identify confounding using the definition provided in the Confounding_ section. Additionally, this method allows us to represent non-casual associations that we may want to model explicitly, such as a correlation coefficient between two variables.
```
beatrixkh, line 801: I think, again, it would be helpful to specify if this schematic representation decision (dashed arrow = association) will be global to our repo (which seems like a good idea, and which i think we had decided on!) or specific to this section

docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
...
+
+"Blocking" an open backdoor path can be done via randomization of that variable between exposure groups in the context of an randomized controlled trial or via conditioning (adjustment via stratification or other methods) in the context of an observational study  on that variable. In the following diagrams, variables that are blocked in this way will be represented as squares rather than circles; when this is the case, assume that a pathway with a square variable is a *closed* pathway. Notably, it is not necessary to condition on all variables on a backdoor path; the path can be closed by conditioning on a minimal set of variables and there may be more than one possible minimal set of variables. See below for possible options for closing backdoor paths by conditioning a *minimal set* of variables (in this case a single variable).
```
beatrixkh, line 782:
```suggestion
"Blocking" an open backdoor path can be done via randomization of that variable between exposure groups in the context of a randomized controlled trial or via conditioning (adjustment via stratification or other methods) in the context of an observational study  on that variable. In the following diagrams, variables that are blocked in this way will be represented as squares rather than circles; when this is the case, assume that a pathway with a square variable is a *closed* pathway. Notably, it is not necessary to condition on all variables on a backdoor path; the path can be closed by conditioning on a minimal set of variables and there may be more than one possible minimal set of variables. See below for possible options for closing backdoor paths by conditioning a *minimal set* of variables (in this case a single variable).
```

docs/source/model_design/general_research.rst
NathanielBlairStahn: Yeah, this is slightly confusing because there are no compartmental model diagrams in this section.  Perhaps eventually it would be good to switch the entire repo to this convention, but currently all the cause model diagrams are drawn with circles, which is inconsistent with the description here.

Perhaps we could change the wording here to something like: "DAGs for causal diagrams are commonly represented as *circles* connected by arrows, and compartmental model diagrams are commonly represented as *squares* connected by arrows; we will follow this convention here."

docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
...
+.. important::
+
+  The figures represented in this section represent directed acyclic graphs/causal diagrams. These figures are distinct from compartamental model diagrams (such as SIR models). For our purposes we will represent directed acyclic graphs as *circles* connected by arrows and compartamental model diagrams as *squares* connected by arrows.
+
+  Additionally, for the purposes of directed acyclic graphs, variables will be labeled with capital letters (A, B, C, etc.) and arrows will be labeled with lower case letters (a, b, c, etc.).
+
+Introduction to Directed Acyclic Graphs
++++++++++++++++++++++++++++++++++++++++
+
+Directed acyclic graphs (DAGs), also termed causal diagrams, are composed of variables that are linked together by directional arrows. Critically, the solid arrows in DAGs represent a causal relationship between an exposure (start of the arrow) and outcome (node of the arrow), representing the *directional* aspect of DAGs. Notably, a directional arrow in a DAG does not offer any information about whether the exposure causes an increased or decreased probability of the outcome, only that there is some causal realtionship bewteen the exposure and outcome.
```
NathanielBlairStahn, line 691: I mostly like how this is written, but my one complaint is that DAGs are a general mathematical concept with many different applications, and the causal diagrams described here are a specific application of DAGs.  I can't think of a good way to reword this part right now, but these two Wikipedia pages provide some context, including related models and other applications of DAGs:

https://en.wikipedia.org/wiki/Directed_acyclic_graph
https://en.wikipedia.org/wiki/Causal_model#Causal_diagram

NathanielBlairStahn: Perhaps something like:

>[Directed acyclic graphs](https://en.wikipedia.org/wiki/Directed_acyclic_graph) (DAGs) are composed of nodes that are linked together by directional arrows, in such a way that the arrows do not form any [directed cycles](https://en.wikipedia.org/wiki/Cycle_graph#Directed_cycle_graph). For our purposes here, DAGS will represent [causal diagrams](https://ftp.cs.ucla.edu/pub/stat_ser/r332.pdf). Critically, the solid arrows in DAGs represent a causal relationship between an exposure variable (start of the arrow) and outcome variable (end of the arrow), representing the *directional* aspect of DAGs.

Note: I changed "node of the arrow" to "end of the arrow" - did you mean to write "node" (I'm not familiar with that usage)?

aflaxman: A term-of-art for the directional arrows is "arcs", which we could also bring in, e.g. "that are linked together by directional arrows (arcs of the graph)"

Two alternative/additional links https://en.wikipedia.org/wiki/Directed_graph https://en.wikipedia.org/wiki/Glossary_of_graph_theory_terms

docs/source/model_design/general_research.rst
NathanielBlairStahn: I like it too, though perhaps it's not totally clear from the diagram that you also can't have feedback loops that involve more than two variables. But I think the interpretation in terms of flow of time helps clarify that even if we don't spell it out explicitly.

docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
...
+
+.. image:: acyclic_dags.svg
+
+Applications, Assumptions, and Limitations of Directed Acyclic Graphs
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+DAGs are used in epidemiology to represent a network of relationships between several variables, which in turn helps to guide the selection of variables that will need to be measured or considered in a given study/project. Further, the exact relationship between variables in a DAG will dictate *how* to treat or consider a specific variable in that study/project by helping to indentify potential confounding, intermediate, or mediating variables between a given exposure and outcome of interest (see the `Causal Relationships`_ section for definitions).
+
+However, DAGs are limited in that they require *prior causal knowledge* about the variables that may be causally related to a given exposure and outcome of interest. Therefore, there may be uncertain, unmeasured, or unknown variables that cannot be represented in a given DAG. Further, DAGs do not allow for easy representation of direction or magnitude of a causal relationship or for effect modification.
```
NathanielBlairStahn, line 702: Beatrix had indicated that the phrase "direction or magnitude of a causal relationship" could be confusing because "direction" could be interpreted as the direction of the causal arrow, rather than whether the arrow implies an increased or decreased probability of the outcome. I think it helps that you previously mentioned this "increased or decreased probability," but perhaps there's a different way to word this to make that even more clear?

docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
...
+
+For the purpose of this section of our documentation, we will assume that each DAG is completely accurate, represents **all** relevant variables in a given causal network and that there are no unrepresented variables.
```
NathanielBlairStahn, line 704:
```suggestion
For the purposes of this section of our documentation, we will assume that each DAG is completely accurate, represents **all** relevant variables in a given causal network and that there are no unrepresented variables.
```

docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
...
+
+Types of Pathways Within Directed Acyclic Graphs
+++++++++++++++++++++++++++++++++++++++++++++++++
+
+A path in a DAG is defined as any route drawn through arrows that link two variables **regardless of the direction of the arrows**.
+
+Directed Paths
+~~~~~~~~~~~~~~
+
+Directed paths are those in which variables are connected through arrows that flow all flow in the same direction. See below for an example of a directed path between variables A and D.
```
NathanielBlairStahn, line 714: I believe the standard definition of directed path requires the arrows all to be oriented in the positive direction relative to the path. On the other hand, I think the current wording implies that there is a directed path from D to A (as well as a directed path from A to D), which (I think) is nonstandard usage. Here's a suggested change to clarify this:

```suggestion
Directed paths are those in which variables are connected through arrows that all flow in the same direction as the path. See below for an example of a directed path from A to D.
```

I think we could also use the phrase "directed path *between* A and D" to mean either a directed path from A to D *or* a directed path from D to A, but I don't know if it's necessary to spell that out.

docs/source/model_design/general_research.rst
NathanielBlairStahn: Agreed, I think 'do not all flow in the same direction' would be better.

docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
...
+
+**In the case that there is an undirected path (and no directed path) bewteen two variables, those variables will NOT be causally related, although they may be associated in some way.**
+
+Backdoor Paths
+^^^^^^^^^^^^^^
+
+Backdoor paths are those in which two variables are connected via a common cause. In other words, two variables that are linked via a variable that has two arrows flowing *away* from it. See below for an example in which variables Y and Z are connected via a backdoor path through variable X.
+
+.. image:: backdoor_path_dag.svg
+
```
NathanielBlairStahn, line 733: I like this - I had seen backdoor paths mentioned in papers but didn't know what they were. This made it really clear.

I think to make it even clearer, it would be good to add one more sentence after the image clarifying that a backdoor path can have more than one variable between the endpoints. E.g. if we added some more variables between X and Z, it would still count as a backdoor path between Y and Z.

docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
...
+**In the case of a backdoor path between two variables (and no directional path), there will be an association (due to a common cause between them), but no causal relationship between the two variables.**
```
NathanielBlairStahn, line 734: We should keep the terminology consistent:

```suggestion
**In the case of a backdoor path between two variables (and no directed path), there will be an association (due to a common cause between them), but no causal relationship between the two variables.**
```

docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
...
+
+Colliders
+^^^^^^^^^
+
+Colliders are variables that have more than one cause; in other words, where pathways collide and a variable has two arrows flowing *toward* it. See below for an example in which variable Z is a collider on the path between X and Y.
+
+.. image:: collider_dag.svg
+
+**In the case that two variables are linked through a pathway with a collider (and there is no directional or backdoor pathway between them), there will be no association or causal realtionship between the two variables.** This is also referred to as a **"closed path."**
```
NathanielBlairStahn, line 743: Again, consistency, and clarify what "this" refers to (please correct my interpretation if I got it wrong):

```suggestion
**In the case that two variables are linked through a pathway with a collider (and there is no directed or backdoor path between them), there will be no association or causal realtionship between the two variables.** A path with a collider is also referred to as a **"closed path."**
```

NathanielBlairStahn: Also, I think it would be good to mention that conditioning on a collider introduces dependence between its parents. See e.g.

https://en.wikipedia.org/wiki/Collider_(statistics)
https://en.wikipedia.org/wiki/Berkson%27s_paradox

This topic is probably out of scope for this PR, and perhaps should go in a separate section, but maybe add a Todo.

docs/source/model_design/general_research.rst
NathanielBlairStahn: This should be "Directed path".

docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
...
+     - Directed
+     - Open
+     - Causal relationship
+   * - Open backdoor path
+     - Path with arrows flowing in different directions without a collider
+     - Undirected
+     - Open
+     - Non-causal association
+   * - Closed path
+     - Path with arrows flowing in different directions with a collider
```
NathanielBlairStahn, line 768: I think saying "with arrows flowing in different directions" is redundant - how about just saying "Path with a collider"?

docs/source/model_design/general_research.rst
```diff
@@ -679,11 +679,131 @@ Conducting Systematic Literature Searches/Reviews
...
+     - Undirected
+     - Closed
+     - No association
+
+Identifying Confounding Using Directed Acyclic Graphs
++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
```
NathanielBlairStahn, line 775: I like the colors for the graphs in this section!

docs/source/model_design/general_research.rst
NathanielBlairStahn: Actually, based on [this reference (pp. 6-7)](https://ftp.cs.ucla.edu/pub/stat_ser/r332.pdf), it seems that the key property of a backdoor path is that it has to start with an arrow pointing into Y. This is not necessarily implied by your definition (e.g. put a collider between X and Y in your diagram), although I think it *is* implied by your definition of *open backdoor path* below - if the path has no colliders *and* has arrows pointing in different directions, then both the initial and final arrows must be pointing "outward" instead of "inward."

docs/source/model_design/general_research.rst
NathanielBlairStahn: Also it might be good to clarify that when we say that "a path has a collider" and hence is closed, we mean that the path contains the collider *and* its parents (i.e. we are abusing terminology by using the word "collider" to refer either to the single vertex Z or to the 3-vertex structure X -> Z <- Y, depending on context)...

For example, a path could still be open if it contains Z but does not contain X and Y.

NathanielBlairStahn: In fact, I think it might be a good idea to create a new subsection called "Open and Closed Paths" between thee "Colliders" section and the "Summary." You could move that final sentence defining open and closed paths to the new section, and clarify exactly what we mean when we say that a collider is or is not on a path.

docs/source/model_design/general_research.rst
NathanielBlairStahn: I think perhaps to resolve both these issues, you could start by saying something like:

> Backdoor paths are those in which two variables are connected via a common cause. The simplest example is when two variables are linked via a third variable that has two arrows flowing *away* from it. See below for an example in which variables Y and Z are connected via a backdoor path through variable X.

Then show the diagram, then afterwards, give a more general definition of backdoor paths.

PR #228, opened 2020-05-13
Covid-19 most basic outline

alibow:
Here is a skeleton outline of the most basic approach (completely random mixing) without any actual data sources yet except for the referenced IHME forecast model

### DIFF
--- a/docs/source/gbd2017_models/causes/covid_19/index.rst.rst
+++ b/docs/source/gbd2017_models/causes/covid_19/index.rst.rst
@@ -1,0 +1,258 @@
+.. _2017_covid_19:
+
+====================================
+Covid-19 Custom Vivarium Cause Model
+====================================
+
+.. contents::
+   :local:
+   :depth: 1
+
+Disease Overview
+----------------
+
+.. todo::
+
+   Add a general clinical overview of the cause.
+
+   Define :math:`R_{0}`.
+
+
+GBD 2017 Modeling Strategy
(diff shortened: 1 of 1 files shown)


docs/source/gbd2017_models/causes/covid_19/index.rst.rst
```diff
@@ -0,0 +1,265 @@
...
+     - XXX
+     - Age-dependent... use IHME data source
+   * - I
+     - traced
+     - XXX
+     - This will be an experimental variable under various scenarios
+   * - I 
+     - symptomatic
+     - XXX
+     - Age-dependent... does IHME have data source?
```
aflaxman, line 179: Maybe we can derive this from the age pattern of severe symptoms, by assuming the fraction of people in age group _a_ with any symptoms is similar to the fraction of people in that age group who end up in the ICU among those with symptoms severe enough to seek care.  I was hoping that by writing that out, I would come up with some justification for it...

docs/source/gbd2017_models/causes/covid_19/index.rst.rst
```diff
@@ -0,0 +1,265 @@
...
+
+- Simulant is in the S state
+- self_quarentine = 0
+
+Data Sources
+""""""""""""
+
+.. todo::
+
+  Define data sources/tables
```
aflaxman, line 257: Population will need some thought here, too.

PR #229, opened 2020-05-13
LBWSG math edits

NathanielBlairStahn:


docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -226,120 +231,233 @@ An example of these causes from the :ref:`large-scale-food fortification concept
...
+Note that since :math:`\text{state}_c(i)` implicitly depends on the time
+:math:`t`, the individual mortality hazard will also depend on time. Using the
+above definitions, we will define the following individual mortality rates
+below:
+
+.. math::
+  :nowrap:
+
+  \begin{align*}
+  &\text{csmr}_c(i) &&= \text{unstratified cause-specific mortality hazard from cause $c$ for $i$}\\
```
alibow, line 371: "unstratified" was initially confusing to me at first because I wasn't clear on what strata were being referenced (we always talk about age/sex strata, which are actually relevant here). Maybe "overall CSMR for a given age/sex/year/location group" would work?

Ninicorn: or we can say un-binned/uncategorized? (lbwsg bins/categories)

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -226,120 +231,233 @@ An example of these causes from the :ref:`large-scale-food fortification concept
...
+  & := \sum_{c\,\in\, \text{causes}} \text{csmr}_c^*(i) \\
+  &= \sum_{c\,\in\, \text{pink}}
+    \text{CSMR}_c
+    + \sum_{c\,\in\, \text{salmon}}
+    \text{EMR}_{\text{state}_c(i)} \\
+    &\qquad\qquad + \left(\sum_{c\,\in\, \text{blue}}
+    \text{CSMR}_c
+    + \sum_{c\,\in\, \text{green}}
+    \text{EMR}_{\text{state}_c(i)}\right)
+    \cdot (1-\text{PAF})\cdot \textit{RR}_{\text{cat}(i)},
```
alibow, line 422: since you were able to get a multi-line equation here maybe we can use "modelled, unaffected", etc. rather than color names now? either way works though

NathanielBlairStahn: Yeah, I think I could get it to look decent using "affected, modelled", etc. in the subscripts. But I thought it would be good to explicitly make the connection to the colors defined in the table above, unless we want to omit the colors entirely.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
NathanielBlairStahn: Good points, I was struggling with what to call these. The name "overall CSMR for a given age/sex/year/location group" would not be accurate -- that's the capital CSMR's, whereas the lowercase csmr's are additionally stratified on what cause states the individual is in. Then the starred version is further stratified by what LBWSG category the individual is in, so I was trying to highlight that distinction, but I agree that "unstratified" is confusing.

NathanielBlairStahn: @alibow @Ninicorn How about "conditional cause-specific mortality hazard"? (see new commit)

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
alibow: sounds good! I'm happy with it as is

Ninicorn: You better keep the colors! I spent ages on it :) hahaha

PR #230, opened 2020-05-14
LBWSG - edits for clarity

NathanielBlairStahn:


docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -331,16 +335,19 @@ All-cause mortality is the sum of all the cause-specific mortality rates
...
 
-To compute the individual mortality hazard :math:`\text{mr}(i)`, for the
-modelled causes (green and salmon) we will need the cause-state-dependent excess
-morality rates (EMR) instead of the CSMR, which is the average EMR over all
-cause states. For example, the excess mortality rates for a two-state cause
-(with condition / without condition) would be:
+We now describe our strategy for defining the individual mortality hazard
+:math:`\text{mr}(i)`, taking an individual's LBWSG category into account. For
+the modelled causes (green and salmon) we will use the cause-state-dependent
+excess morality rates (EMRs) instead of the CSMR, which is the average EMR over
```
Ninicorn, line 344:
```suggestion
excess morality rates (EMRs) instead of the CSMR (the CSMR is the average EMR over
```

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -331,16 +335,19 @@ All-cause mortality is the sum of all the cause-specific mortality rates
...
+all cause states (including the "without condition" state). For example, the
+excess mortality rates for a two-state cause (with condition / without
+condition) would be:
```
Ninicorn, line 347: I'm wondering if we can clarify these sentences a bit: 
original:
For the modelled causes (green and salmon) we will use the cause-state-dependent
excess morality rates (EMRs) instead of the CSMR, which is the average EMR over
all cause states (including the "without condition" state). For example, the
excess mortality rates for a two-state cause (with condition / without
condition) would be.

suggested:
For the modelled causes (green and salmon) we will use the excess morality rates (EMRs) instead of the CSMR. The EMR is cause-state dependent while the CSMR is the average EMR over all cause-states. For example, the excess mortality rates for a two-state cause (with condition / without condition) would be:

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -387,12 +394,18 @@ hazard for individual :math:`i` to be
   \end{cases}
 
 The descriptor "conditional" here means that the above individual csmr's can be
-interpreted as the cause-level expected CSMR's `conditioned <conditional
-expectation_>`_ (i.e. `stratified <stratification_>`_) on all the individual
-cause states observed in the simulation. Now we additionally stratify by LBWSG
-category: Define the LBWSG-stratified cause-specific mortality hazard of
+interpreted as the cause-level expected CSMR's `conditioned <conditioning_>`_
+(i.e. `stratified <stratification_>`_) on all the individual cause states
+observed in the simulation. In other words, :math:`\text{csmr}_c(i)` is the
```
Ninicorn, line 399:
```suggestion
observed in the simulation (for modelled cause only as we do not observe cause-states for unmodelled causes). In other words, :math:`\text{csmr}_c(i)` is the
```

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
```diff
@@ -450,6 +469,8 @@ because
    `LBWSG PAF notebook <https://github.com/ihmeuw/vivarium_data_analysis/blob/master/pre_processing/lbwsg/LBWSG%20exposure%2C%20rrs%2C%20pafs.ipynb>`__.
 
 .. todo::
+
+   - add a proof that the expected value of :math:`\text{mr}(i)` equals the ACMR.
```
Ninicorn, line 473: oooo !!I like this!

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
NathanielBlairStahn: Sure, that sounds good. What about keeping the "without condition" clarification in, like this:

> For the modelled causes (green and salmon) we will use the excess morality rates (EMRs) instead of the CSMR. The EMR is cause-state dependent while the CSMR is the average EMR over all cause-states (including the "without condition" state). For example, the excess mortality rates for a two-state cause (with condition / without condition) would be:

I just thought it was a good idea to really highlight this point because (1) it might not be clear that "cause-state" includes the state where you _don't_ have the cause, and (2) I think it helps clarify the distinction between the EMRs and the CSMRs.

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
NathanielBlairStahn: What about:

> observed in the simulation (note that we can only observe cause states for _modelled_ causes).

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: oh yes, thank you for the clarification. I originally read 'with' condition state, sorry. It's very clear now :)

docs/source/gbd2017_models/risks/low_birthweight_short_gestation/index.rst
Ninicorn: yes, excellent, I like it.

PR #233, opened 2020-05-15
Risk effects template

alibow:


beatrixkh:
This looks great, thanks Ali! My one comment is that it seems like I can't navigate to this page via the left-hand menu.

I think it would make sense to create this page:
/vivarium_research/docs/build/html/gbd2017_models/risk_effects/index.html

Modeled after this one:
/vivarium_research/docs/build/html/gbd2017_models/causes/index.html

Which could link to the template, and then all specific risk effect pages that get made for models.

alibow:
> This looks great, thanks Ali! My one comment is that it seems like I can't navigate to this page via the left-hand menu.
> 
> I think it would make sense to create this page:
> /vivarium_research/docs/build/html/gbd2017_models/risk_effects/index.html
> 
> Modeled after this one:
> /vivarium_research/docs/build/html/gbd2017_models/causes/index.html
> 
> Which could link to the template, and then all specific risk effect pages that get made for models.

Thanks, Beatrix! I made this change :)

PR #235, opened 2020-05-18
Covid model overview

alibow:
Here is my brief overview of the IHME covid model SEIR component so that we can reference it for our own covid cause model


docs/source/gbd2017_models/causes/covid_19/index.rst
```diff
@@ -11,35 +11,285 @@ Covid-19 Custom SEIR Vivarium Cause Model
...
+:math:`R_{t}` (or effective :math:`R`) will be defined here as the population average number of infections a single individual with covid-19 will transmit over the entire infectious period at a given point in time.
+
+Notably, while :math:`R_{0}`, :math:`R_{0}'`, and :math:`R_{t}` are population level average parameters, the distribution of these parameters at the *individual level* will be highly right skewed, indicating that many infectious individuals will transmit fewer infections than the population mean R value, while some individuals will transmit many more infections that the population mean value (e.g. mass transmission events).
+
 IHME Covid Team Modeling Strategy
 ---------------------------------
 
+Covid-19 is not included in currently available GBD study data. Rather, we will use custom data for all parameters in this cause model using available data from the existing IHME covid-19 forecasting model.
+
+IHME's existing covid-19 models have fit covariate estimates for **population density,  relative mobility, testing capacity, and temperature.** values over time. Below is a brief overview of the SEIR component of the IHME forecast model.
```
beatrixkh, line 33: Maybe anything in this section should have a timestamp on it? Aren't they now also using seasonal flu patterns, or something similar, as a covariate?

docs/source/gbd2017_models/causes/covid_19/index.rst
```diff
@@ -11,35 +11,285 @@ Covid-19 Custom SEIR Vivarium Cause Model
...
+
+#. Calculate deaths from infections and infection fatality ratio (IFR)
+
+Notably, the IHME forecast SEIR component includes a dual infectious state, :math:`I_{1}` and :math:`I_{2}`. The following equations represent the differential equations for each state in the SEIR component used in the IHME forecast model:
+
+			:math:`S` --> :math:`E` --> :math:`I_1` --> :math:`I_2` --> :math:`R`
+
+:math:`dS/dt = -(\beta(t)S(I_1+I_2)^\alpha)/N`
+
+:math:`dE/dt = (\beta(t)S(I_1+I_2)^\alpha)/N - \sigma E`
```
beatrixkh, line 55: Let A and B be the two groups of susceptible people that infectious people, person_a and person_b come into contact with. Do these assume that A and B are disjoint?

docs/source/gbd2017_models/causes/covid_19/index.rst
```diff
@@ -11,35 +11,285 @@ Covid-19 Custom SEIR Vivarium Cause Model
...
+
 Scope
 +++++
 
-Assumptions and Limitations
-+++++++++++++++++++++++++++
+The immediate intention of this cause model will be to aid in modeling various scenarios relating to varied population density, mobility, and other potential covid-19 control measures at the University of Washington and immediately surrounding areas/populations. This covid-19 Vivarium cause model is intended to make use of the data and framework of the existing IHME covid-19 forecast model SEIR component (briefly described above). The scope of this cause model document is intended to develop a basic strategy for an infectious SEIR cause model and will start with minimal complexity and can be expanded upon if it becomes necessary.
+
+Notably, this is currently intended to be a **mortality-only** cause model and does not indend to measure years lived with disability (YLDs).
```
beatrixkh, line 114: It would be helpful to me to state the outcomes of interest here, very clearly. Is infections something we're interested in? Or only deaths?

docs/source/gbd2017_models/causes/covid_19/index.rst
```diff
@@ -11,35 +11,285 @@ Covid-19 Custom SEIR Vivarium Cause Model
...
+Simulants will be initialized into covid-19 disease model states based on the forecasted IHME covid-19 model estimates. 
+
+.. todo::
+
+	Detail specific data source and values
+
+Mortality
+"""""""""
+
+We will model covid-19 mortality using an age-dependent infection fatality ratio (IFR), as consistent with the IHME forecast model. There will be two steps to this piece of the model, the first being the determination of if an individual dies due to covid-19 (described here), and the being determining *when* they exit their current model state (decribed in the `Transition Data`_ section).
```
beatrixkh, line 165:
```suggestion
We will model covid-19 mortality using an age-dependent infection fatality ratio (IFR), as consistent with the IHME forecast model. There will be two steps to this piece of the model, the first being the determination of if an individual dies due to covid-19 (described here), and the second being determining *when* they exit their current model state (decribed in the `Transition Data`_ section).
```

docs/source/gbd2017_models/causes/covid_19/index.rst
```diff
@@ -11,35 +11,285 @@ Covid-19 Custom SEIR Vivarium Cause Model
...
+
+In the same time-step for which a simulant is initialized into the :math:`E`, :math:`I_1`, or :math:`I_2` covid-19 model states or transitions into the :math:`E` model state, the age-specific IFR should be used to determine if that simulant will die of covid-19, such that the IFR shown in the table below represents the probability that an infected individual dies from covid-19.
```
beatrixkh, line 167: This is suuuper nit picky (feel free to disregard), but it took me a minute to process this first sentence, and it would've been clearer, to me, if you either said "covid-19 model state" or "model state", instead of using both phrases. That way I don't think through if there's a difference / some other type of model being layered on top.

docs/source/gbd2017_models/causes/covid_19/index.rst
```diff
@@ -11,35 +11,285 @@ Covid-19 Custom SEIR Vivarium Cause Model
...
+	Detail specific data source and values
+
+Mortality
+"""""""""
+
+We will model covid-19 mortality using an age-dependent infection fatality ratio (IFR), as consistent with the IHME forecast model. There will be two steps to this piece of the model, the first being the determination of if an individual dies due to covid-19 (described here), and the being determining *when* they exit their current model state (decribed in the `Transition Data`_ section).
+
+In the same time-step for which a simulant is initialized into the :math:`E`, :math:`I_1`, or :math:`I_2` covid-19 model states or transitions into the :math:`E` model state, the age-specific IFR should be used to determine if that simulant will die of covid-19, such that the IFR shown in the table below represents the probability that an infected individual dies from covid-19. 
+
+.. csv-table:: Infection Fatality Ratios
```
beatrixkh, line 169: Chris indicated that the higher burden of COVID among BIPOC is due to a higher proportion of BIPOC being essential workers, rather than, say, lower quality or lack of access to healthcare---but it would be interesting to see if there's any data out there on IFR stratified by race.

docs/source/gbd2017_models/causes/covid_19/index.rst
```diff
@@ -11,35 +11,285 @@ Covid-19 Custom SEIR Vivarium Cause Model
...
+   * - :math:`N`
+     - Total number of simulants in the model at time-step :math:`t`
+   * - :math:`\beta(t)`
+     - Taken from IHME forecast model under stated assumptions regarding covariates
+   * - :math:`\alpha`
+     - Taken from IHME forecast model under stated assumptions
+
+.. note::
+
+	This method assumes random mixing of the population and assumes homogeneous transmission probability across all demographic groups.
```
beatrixkh, line 213: It seems like these assumptions would ensure way too much mixing. Am I thinking through this correctly?

If the time steps are smaller than the average period of time for which someone is infectious, then each time step, this seems to assume that the people that person_A sees is a set independent of the people person_A saw during the previous time step. It sounds like it would inherently overestimate, as people are likely to see a lot of the same people every day, I would think!

docs/source/gbd2017_models/causes/covid_19/index.rst
```diff
@@ -11,35 +11,285 @@ Covid-19 Custom SEIR Vivarium Cause Model
...
+
+Method Two
+~~~~~~~~~~
+
+This method of modeling the susceptible to exposed state transition in the covid-19 model uses the :math:`\beta(t)` parameter *scaled to the average duration of the infectious period* OR :math:`R_t` from the IHME forecast model under stated assumptions regarding coviariates as the simulation's :math:`R_0` value under those same stated assumptions.
```
beatrixkh, line 218: When you say 'OR', does that mean that those two things are actually equivalent, or that there is a 'Method 2a' and a 'Method 2b' embedded here? They sound like they both aim to be the true R_t, but might be two methods of estimating R_t/might yield different values (?)

docs/source/gbd2017_models/causes/covid_19/index.rst
```diff
@@ -11,35 +11,285 @@ Covid-19 Custom SEIR Vivarium Cause Model
...
+
+.. warning::
+
+	This method requires careful scrutiny of the value that will be used for the simulation's :math:`R_0` to ensure that it is being applied in a consistent way
+
+At each time-step, for each simulant in the :math:`I_1` and :math:`I_2` covid-19 model states, sample the following number of simulants from the *entire simulant pool* (see *Important* note below for exception):
+
+.. math:: n_\text{simulants to sample} = R_0 * (\gamma_1 + \gamma_2) * duration_{\text{time step}}
```
beatrixkh, line 226: I don't understand this. Could you expand on it a little? Why does gamma_1 need to be there?

docs/source/gbd2017_models/causes/covid_19/index.rst
```diff
@@ -11,35 +11,285 @@ Covid-19 Custom SEIR Vivarium Cause Model
...
+     - :math:`1/\sigma`
+     - :math:`\sigma` obtained from IHME forecast model
+   * - :math:`I_1`
+     - :math:`1/\gamma_1`
+     - :math:`\gamma_1` obtained from IHME forecast model
+   * - :math:`I_2`
+     - :math:`1/\gamma_2`
+     - :math:`\gamma_2` obtained from IHME forecast model
+
+If it **was** determined that a simulant will die due to covid-19 (as described in the Mortality_ section), assume that the individual dies a the duration of the :math:`1_2` model state instead of transitioning them to the :math:`R` state.
```
beatrixkh, line 264: There's a typo here, and I'm not sure I know what it ought to have been. Either dying during I_1, I_2, or in either?

docs/source/gbd2017_models/causes/covid_19/index.rst
```diff
@@ -11,35 +11,285 @@ Covid-19 Custom SEIR Vivarium Cause Model
...
+	Determine if these values from IHME forecast model include cases of mortality and update methodology accordingly
+
+Assumptions and Limitations
++++++++++++++++++++++++++++
+
+There are several assumptions used in this cause model:
+
+- Recovered individuals are no longer susceptible to covid-19
+
+- There is no birth prevalence of covid-19
```
beatrixkh, line 277: Whoa! I hadn't thought about this. Have there been cases of children being born with COVID (or the flu? or another virus?) thus far?

docs/source/gbd2017_models/causes/covid_19/index.rst
```diff
@@ -11,35 +11,285 @@ Covid-19 Custom SEIR Vivarium Cause Model
...
+
+- These is random mixing of the population (unless :math:`\alpha` parameter is not equal to one or correlated mixing structure is introduced)
+
+- The transmission probability is homogenous across demographic groups
```
beatrixkh, line 281: How hard would it be to model sub-populations?

docs/source/gbd2017_models/causes/covid_19/index.rst
```diff
@@ -11,35 +11,285 @@ Covid-19 Custom SEIR Vivarium Cause Model
...
 
+Modeling covid-19 in Vivarium will require a different strategy than our standard Vivarium cause models. The standard methodology for Vivarium cause models assumes event independence in that one simulant's disease incidence does not affect another simulant's chance of disease incidence. However, given the infectious nature of covid-19, this cause model document will outline a strategy to model the dependent infectious nature of the disease transmission.
+
 Scope
 +++++
 
-Assumptions and Limitations
-+++++++++++++++++++++++++++
+The immediate intention of this cause model will be to aid in modeling various scenarios relating to varied population density, mobility, and other potential covid-19 control measures at the University of Washington and immediately surrounding areas/populations. This covid-19 Vivarium cause model is intended to make use of the data and framework of the existing IHME covid-19 forecast model SEIR component (briefly described above). The scope of this cause model document is intended to develop a basic strategy for an infectious SEIR cause model and will start with minimal complexity and can be expanded upon if it becomes necessary.
```
beatrixkh, line 112: I know you currently don't know, and I'm sure you were already planning on it, but as you develop a clearer idea, it would be good to list out the scenarios of interest here.

Even right now, it might be good to list categories of what it would be _possible_ for us to model. Specifically, would we consider doing sub-populations which do not mix homogeneously with one another?

docs/source/gbd2017_models/causes/covid_19/index.rst
```diff
@@ -11,35 +11,285 @@ Covid-19 Custom SEIR Vivarium Cause Model
...
+.. warning::
+
+	Determine if these values from IHME forecast model include cases of mortality and update methodology accordingly
+
+Assumptions and Limitations
++++++++++++++++++++++++++++
+
+There are several assumptions used in this cause model:
+
+- Recovered individuals are no longer susceptible to covid-19
```
beatrixkh, line 275: Could we estimate, as a function of 'reinfection rate' or something similar, how inaccurate this would make the model?

docs/source/gbd2017_models/causes/covid_19/index.rst
```diff
@@ -11,35 +11,285 @@ Covid-19 Custom SEIR Vivarium Cause Model
...
+
+.. todo::
+
+  Obtain precise definition of :math:`I_1` versus :math:`I_2`
+
+.. note::
+
+	In standard SEIR models, the parameter :math:`\beta(t)` defined here is often a *constant* parameter that represents the average number of contacts per person per time multiplied by the transmission probability of a contact between an infectious and susceptible individual (equal to :math:`R_0` when unit time is equal to the duration of the infectious period).
+
+	However, as described above, the :math:`\beta(t)` parameter is a function of time the considers the time-varying covariates used in the IHME forecasting model, which allows for the consideration of changes in relative mobility (which could be considered a proxy variable for the number of social contacts per person per time).
```
beatrixkh, line 102: For a given population/time, with a set mobility, temperature, etc. etc., does beta(t) yield a single value, or a distribution?

If we made beta(t) output a distribution, instead of the average value, given that this model is calculating cumulative deaths, would that make it any more accurate, or just more complex?

This then makes me want to ask if simulants could be initialized with the covariate values, with beta(t) calculated for each simulant... but again, given that we then aggregate the results, I don't have a good understanding of if that would make this more complicated & still estimate the same outcomes, on average.

docs/source/gbd2017_models/causes/covid_19/index.rst
alibow: Yes, this definitely needs to be updated as this is an evolving list. I will add a note about that, thank you :)

docs/source/gbd2017_models/causes/covid_19/index.rst
alibow: I'm not sure I understand the question. This wasn't in the version that you reviewed, but essentially I1 = pre-symptomatic and I2 = symptomatic, although both I1 and I2 have infectious potential. Does that help clarify?

docs/source/gbd2017_models/causes/covid_19/index.rst
alibow: I meant to say something along the line of the conclusion of the I_2 state, but I changed it to "at the instant that they would transition to the R state" instead. Does that make sense now?

docs/source/gbd2017_models/causes/covid_19/index.rst
alibow: We ended up abandoning this method, but I can still explain my thoughts here! I actually tried to simplify this equation and did it incorrectly. This is what it should be:

Both I1 and I2 are infectious states. 1/gamma_1 is the duration of the 1_2 state. 1/gamma_2 is the duration of the I_2 state.

R_0 is the value for the total number of people that a single infectious person will infect over the total amount of time they spend in the I_1 and I_2 states.

To scale this to a single time step (all duration measured in days)...

R_0 / (1/gamma_1 + 1/gamma_2) = number of infections per day during infectious period

R_0 / (1/gamma_1 + 1/gamma_2) * time-step duration = number of infectious per time-step during infectious period.

docs/source/gbd2017_models/causes/covid_19/index.rst
alibow: I've seen some studies that discuss vertical transmission of covid-19 although there doesn't seem to be great consensus. The GBD dismod model for LRI has birth prevalence though that Chris said was believed to be due to vertical transmission which is very interesting!

docs/source/gbd2017_models/causes/covid_19/index.rst
alibow: For the UW model (which will be in the upcoming PR), we are considering modeling the UW population explicitly and then allowing for some sort of constant covid-19 flow into the model from the surrounding population. Not quite a subpopulation, but gets at the concept that our model population is a subpopulation of the larger world!

I think that modeling subpopulations like dorms and classrooms would be really cool, but it's beyond our initial scope from my discussions with abie.

docs/source/gbd2017_models/causes/covid_19/index.rst
beatrixkh: My question is essentially:

I understand beta(t) to be an estimate of how many people the average infected person infects. I am then interpreting the people _entering_ E to be calculated by multiplying beta by the # of infectious individuals

That seems to assume that infectious people A and B don't both contribute to getting susceptible person C sick ---that people A and B each get their own, discrete sets of people sick

But it sounds like you're accounting for this perhaps with alpha? (How does alpha work/how is it chosen?)

docs/source/gbd2017_models/causes/covid_19/index.rst
beatrixkh: it still says "dies *a* the duration of the l_2 model state". But  "during the duration" or "at the conclusion of" both make sense and are clear to me!

PR #239, opened 2020-05-20
LBWSG: Assigning cause of death

NathanielBlairStahn:


Ninicorn:
Perhaps we can leave the coloring in the table cuz it helps visually separate out the 4 types of causes. Then when we refer to them we can call them mo|aff (**mo**delled-**aff**ected = green); un|aff (**un**modelled-**aff**ected=blue); mo|un ( **mo**delled-**un**affected=salmon); and un|un (**un**affected-**un**modelled=pink). The labelling system is that anything before the bar "|" describes whether its modelled or not , and after the bar, describes whether it's affected or not. We can also add these labels to the table with the color where they are first described.

NathanielBlairStahn:
I like having the colors in the table because, as Nicole says, it helps visually separate them out. It would be easy enough to change our variable names to something more descriptive, but if we don't refer to the colors anywhere, then we lose the advantage of having colors at all.  So I'm struggling with how to incorporate both rather than one or the other. Any ideas?

PR #243, opened 2020-05-28
LSFF: Iron algorithm updates

NathanielBlairStahn:
- Fixed description of birthweight shift to use mean for downshift
- Clarified that flour consumption is only needed if mother is fortified
- Changed duration of fortification to account for children not eating solids before 6 months
- Fixed hemoglobin shift to account for iron-responsive vs. non-responsive anemias
- Added a description of the iron algorithm summary section

### DIFF
--- a/docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
+++ b/docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
@@ -1370,6 +1370,7 @@ clarify the intent.
   p_i                     = propensity of simulant and mother for exposure to lack of iron fortification (~uniform(0,1))
   gestational_age_i       = gestational age of simulant drawn from GBD (in years)
   birthweight_gbd_i       = birthweight drawn from GBD for simulant
+  iron_responsive_i       = whether individual is responsive to iron fortification
   age_i(t)                = the simulant's age at time t
   hb_gbd(age_i(t))        = hemoglobin level drawn from GBD for simulant
     # These times are both float('-inf') if simulant is covered in baseline
@@ -1415,28 +1416,25 @@ clarify the intent.
                          + bw_shift_i if mother_fortified_i else 0)
 
   ## At each simulation time step (including t=0 or when simulant is born):
-  household_covered_i(t)    = time_covered_i <= t
-                            = p_i < coverage(t)
-    # Fortification starts either when the household receives coverage, or when
-    # the child turns 6 months old and starts eating solids, whichever is later.
-    # Recall that if the child is covered in baseline, then time_covered_i =
-    # float('-inf'), so coverage always starts at age 6 months in this case.
-  time_since_fortified_i(t) = min(t - time_covered_i, t - age(t) + 0.5)
-    # Effect on Hb - calibrate for baseline coverage, and take into account
-    # the age-dependent effect size and the 6 month lag time.
-  hb_i(t)                   = (hb_gbd(age_i(t))
-                               - baseline_coverage(t) *  hb_age_fraction(age_i(t)) * hb_shift
-                               + (household_covered_i(t)
-                                  * hb_lag_fraction(time_since_fortified_i(t))
-                                  * hb_age_fraction(age_i(t))
-                                  * hb_shift
-                                 )
-                              )
-
-.. todo::
-
-  Edit above summary to account for iron-responsive vs. iron-non-responsive
-  anemias in the Hb shift.
+    # Only adjust Hb level for iron-responsive individuals
+  if iron_responsive_i:
+    household_covered_i(t)    = time_covered_i <= t
+                              = p_i < coverage(t)
+      # Fortification starts either when the household receives coverage, or when
+      # the child turns 6 months old and starts eating solids, whichever is later.
+      # Recall that if the child is covered in baseline, then time_covered_i =
+      # float('-inf'), so coverage always starts at age 6 months in this case.
+    time_since_fortified_i(t) = min(t - time_covered_i, t - age(t) + 0.5)
+      # Effect on Hb - calibrate for baseline coverage, and take into account
+      # the age-dependent effect size and the 6 month lag time.
+    hb_i(t)                   = (hb_gbd(age_i(t))
+                                 - baseline_coverage(t) *  hb_age_fraction(age_i(t)) * hb_shift
+                                 + (household_covered_i(t)
+                                    * hb_lag_fraction(time_since_fortified_i(t))
+                                    * hb_age_fraction(age_i(t))
+                                    * hb_shift
+                                   )
+                                )
 
 Folic Acid Fortification
 ~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
+++ b/docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
@@ -1317,22 +1317,31 @@ See the proofs for this approach below.
 Summary of Iron Intervention Algorithm
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+The following Python-style pseudocode summarizes all the pieces of the iron
+intervention algorithm (for both birthweight and hemoglobin) and is intended to
+resolve any ambiguities that may remain in the descriptions above. If there are
+any inconsistencies between the above descriptions and the algorithm below, the
+algorithm in this section takes precedence. However, if any ambiguities or
+inconsistencies remain in the algorithm below, the above descriptions may help
+clarify the intent.
+
 .. code-block::
 
   # Definitions
-  t                     := current time in years since start of simulation
-  t_end                 := time when simulation ends, in years since start (e.g. 6 years)
-  n_weeks               := n weeks, measured in years (e.g. n*7/365.25)
-  time_covered_i        := time at which simulant's household first receives iron-fortified food
-  mother_fortified_i    := whether simulant's mother got sufficient iron-fortification during pregnancy
-  child_fortified_i(t)  := whether child is eating fortified food at time t
-  birthweight_i         := simulant's birthweight after adjusting for mother's iron consumption
-  hb_i(t)               := simulant's hemoglobin level after adjusting for fortification
+  t                       := current time in years since start of simulation
+  t_end                   := time when simulation ends, in years since start (e.g. 6 years)
+  n_weeks                 := n weeks, measured in years (e.g. n*7/365.25)
+  time_covered_i          := time at which simulant's household first receives iron-fortified food
+  mother_fortified_i      := whether simulant's mother got sufficient iron fortification during pregnancy
+  household_covered_i(t)  := whether child's household is receiving iron-fortified food at time t
+  duration_fortified_i(t) := amount of time the child has been eating iron-fortified food
+  birthweight_i           := simulant's birthweight after adjusting for mother's iron consumption
+  hb_i(t)                 := simulant's hemoglobin level after adjusting for fortification
     # Coverage groups are used only for stratification
-  iron_coverage_group   := population subgroup indicating in which scenario (if any) the
-                           child started eating iron-fortified food by the end of the simulation
-  mother_iron_group     := population subgroup indicating in which scenario (if any) the
-                           child's mother got sufficient iron-fortification during pregnancy
+  iron_coverage_group     := population subgroup indicating in which scenario (if any) the child's
+                             household received iron-fortified food by the end of the simulation
+  mother_iron_group       := population subgroup indicating in which scenario (if any) the child's
+                             mother got sufficient iron-fortification during pregnancy
 
   # Population level parameters
   hb_shift                = effect size on hemoglobin (normally disributed, mean 3.0 g/L)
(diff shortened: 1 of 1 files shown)


docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1316,43 +1317,57 @@ See the proofs for this approach below.
...
-  iron_concentration    = concentration X of iron in the country's fortified flour
-
-  # Individual level attributes
-  p_i                     = propensity of simulant and mother for exposure to unfortified flour (uniform(0,1))
-  daily_flour_i           = average daily flour consumption Y of simulant's mother (sampled from population distribution)
+    # Negative time_since_fortified indicates child has not yet started eating fortified food
+  hb_lag_fraction(time_since_fortified)
+                          = (0 if time_since_fortified < 0 else
+                             time_since_fortified/0.5 if time_since_fortified < 0.5 else
+                             1)
```
alibow, line 1361: The `0 if time_since_fortified < 0` portion is intended to apply to those who are not covered, right?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1391,25 +1403,39 @@ Summary of Iron Intervention Algorithm
...
-  child_fortified_i(t)  = time_covered_i <= t
-                        = p_i < coverage(t)
-    # If child is covered in baseline, then time_covered_i = float('-inf'),
-    # and hence hb_lag_fraction(t-time_covered_i) = 1, i.e. the child receives
-    # the full age-dependent effect in this case since the time lag has passed.
-  hb_i(t)               = hb_gbd(age_i(t)) + (
-                          -baseline_coverage(t) + child_fortified_i(t) * hb_lag_fraction(t-time_covered_i)
-                          ) * hb_shift *  hb_age_fraction(age_i(t))
+    # Effect on birthweight
+  daily_flour_i       = average daily flour consumption Y of simulant's mother
```
alibow, line 1407: if `mother_fortified_i = 1`?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1391,25 +1403,39 @@ Summary of Iron Intervention Algorithm
...
+                         + mother_fortified_i * bw_shift_i)
+
+  ## At each simulation time step (including t=0 or when simulant is born):
+  household_covered_i(t)    = time_covered_i <= t
+                            = p_i < coverage(t)
+    # Fortification starts either when the household receives coverage, or when
+    # the child turns 6 months old and starts eating solids, whichever is later.
+    # Recall that if the child is covered in baseline, then time_covered_i =
+    # float('-inf'), so coverage always starts at age 6 months in this case.
+  time_since_fortified_i(t) = min(t - time_covered_i, t - age(t) + 0.5)
```
alibow, line 1423: I'm confused about why we are taking the minimum of the time coverage and age components here for `time_since_fortified_i(t) = min(t - time_covered_i, t - age(t) + 0.5)`. My thought was that the `hb_lag_fraction(time_since_fortified_i(t))` should only be based on time since coverage and then the `hb_age_fraction(age_i(t))` takes care of the age component. Let me know if I'm misunderstanding what's going on here.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Yes, if `time_since_fortified < 0`, then either they'll never get coverage, or they haven't gotten it yet.

It doesn't actually matter whether this gets assigned to 0 here or not, because it gets multiplied by other things that will be 0, but I think assigning it to 0 makes the most sense for uncovered simulants.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Well, we only use it if `mother_fortified_i == True`, but the variable makes sense for all simulants, and we could actually assign its value if we wanted: It would be distributed according to Y if the mother eats flour (i.e. p_i<c, where p_i is the propensity for unfortified flour and c is the proportion of the population who eats flour, from our coverage data table), and it would be 0 otherwise.

So I _defined_ `daily_flour_i` as the mother's daily flour consumption for all simulants, but I only _assigned it a numerical value_ if `mother_fortified_i == True`, and otherwise assigned it the value `'irrelevant'` since we don't use it in that case.

Is that confusing?

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Well, I think that's how we were originally thinking about it, but after thinking about it some more, it seems that our original approach is slightly inconsistent with points 1-4 in the section [Determining whether a simulant is affected -Iron](https://vivarium-research.readthedocs.io/en/latest/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.html#determining-whether-a-simulant-is-affected-iron), and point 5 in that section is also slightly inconsistent with the other 4. Here's my reasoning:

We are trying to capture two effects:
1. Children don't start eating solid food until 6 months, at which point the amount of food slowly increases to a maximum at 2 years.
2. When someone starts eating iron-fortified food, the measured effect on hemoglobin increases from 0 to the maximum effect size over a period of 6 months.

The function `hb_age_fraction(age_i(t))` captures the first effect, while the function `hb_lag_fraction(time_since_fortified_i(t))` captures the second effect.

Now suppose we use our original strategy and set `time_since_fortified_i(t) = t - time_covered_i`, regardless of age, and suppose `time_covered_i` is before the simulant is born. Then we'd have `hb_lag_fraction(time_since_fortified_i(t))=1` when the child is six months old, which means we'd be saying that by age 6 months the simulant would already have been eating fortified food long enough to see the full effect (which happens to be 0 because of the age function, but that's beside the point), _even though they have not eaten any fortified food at that age!_ What I'm saying is that `time_since_fortified_i(t)` should be the actual amount of time you've been _eating_ fortified food, not just the time since you're household started receiving the food, because `hb_lag_fraction(time_since_fortified_i(t))` is supposed to capture a biological effect that depends on when you started eating the food.

Does that make sense?

Mathematically, the result is that for children whose household gets covered before they're  6 months old, their effect size will increase quadratically instead of linearly at first (since we're multiplying two linear functions), with slope 0 at 6 months, then curving upward until it hits the blue line at 1 year (1/3 the maximum effect size), then follows the blue graph from there.  I think slope 0 at 6 months seems more reasonable than positive slope at 6 months, though the difference in our results should be negligible.  Also, this is the same as what would have happened already if coverage is received at 6 months instead of birth. So basically I'm just saying there should be no difference between receiving coverage at 6 months and receiving coverage at birth since the child is not eating food during that time.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Cool, that sounds good

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1391,25 +1403,39 @@ Summary of Iron Intervention Algorithm
...
-                        = p_i < coverage(t)
-    # If child is covered in baseline, then time_covered_i = float('-inf'),
-    # and hence hb_lag_fraction(t-time_covered_i) = 1, i.e. the child receives
-    # the full age-dependent effect in this case since the time lag has passed.
-  hb_i(t)               = hb_gbd(age_i(t)) + (
-                          -baseline_coverage(t) + child_fortified_i(t) * hb_lag_fraction(t-time_covered_i)
-                          ) * hb_shift *  hb_age_fraction(age_i(t))
+    # Effect on birthweight
+  daily_flour_i       = average daily flour consumption Y of simulant's mother
+                      = (flour_consumption_dist.rvs() if mother_fortified_i else
```
alibow, line 1408:
```suggestion
                      = (flour_consumption_dist.rvs() if mother_fortified_i == True else
```

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: I made a suggested change below. If the suggested change is correct then I think I understand, but if not then I will need to revist.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Well, perhaps `mother_is_fortified` would be a better name for a boolean so that `if mother_is_fortified` reads even more naturally, but I was trying to save space...

https://softwareengineering.stackexchange.com/questions/12807/make-a-big-deal-out-of-true

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Yes, your suggested change is correct, but I'm avoiding explicitly writing `== True` for boolean variables (unless the SWEs think that would be clearer).

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Ah, ok... this is altering our assumptions slightly then I think (or at least the implicit assumptions that I made but maybe we never explicitly stated). My That sounds fine to me although we should definitely write that down in prose. I think that we had initially settled on the linear curve because we don't really know what happens in this situation and this was we were airing on the side of not being overly conservative (so that we can demonstrate that this will have minimal impact even without conservative constraints) as Abie had mentioned a few times. I don't really have a preference except I think that we need to clear up some of the contradictions between this section and the info in this screenshot. 
image

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Ok, gotcha. Wasn't initially clear to me because I tend to default to 0/1 rather than True/False, but I see now

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
alibow: Got it! This is just my lack of familiarity, so thanks for orienting me :)

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: I edited this to separate out the down-shift and put the code relevant to the up-shift into a single `if` block. I think that might make it clearer (though I still didn't explicitly write `==True` or `==1`).

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Ahh, I think I mistranslated your code when I split the age part and time lag part into separate functions.  I didn't quite understand what you were doing with the `age_of_coverage_i` variable, so when I reworked it, I wrote the `hb_age_fraction` function in terms of the simulant's current age.

But that has the effect of making `hb_age_fraction(age(t))` a linear function of t, whereas I think your variable `age_of_coverage_i` is a constant. So in my version, I am multiplying two linear functions together when `0.5<age(t)<2`, giving a quadratic function, whereas in your version you're multiplying a linear function by a constant, giving a linear function. (You'd think with a math PhD, I would have noticed this difference sooner, but I just noticed it yesterday...)

I think there were two reasons I was trying to split up the Hb shift into two pieces in the first place:

1. To make it easier to understand what's going on, and
2. Because for baseline calibration, I _think_ it makes sense to apply only the age part to the downshift but to apply both the age part and the time lag part to the upshift. (This will make our mean Hb slightly lower than GBD's, but it seems more logical than applying the time lag to the downshift as well.)

Does that second point seem correct to you, @alibow ? If so, I'm not sure how do do the baseline calibration correctly while preserving the overall linear scaleup in your version.

alibow: Ah, I see what you're saying... essentially it doesn't make sense to have the down shift be time-dependent, which is what happens in my version, but not yours. Is that the idea? If so then I think I agree with your approach.

NathanielBlairStahn: Yeah, I think that's the issue (though it's still a bit confusing for me to think about). Let's go with my version then, since it gives us a reasonable approach for baseline calibration.  I think I'll try drawing some graphs of both your function and my function to make sure everything looks how I think it should. Is your code for drawing the above graph on the vivarium_data_analysis repo?

NathanielBlairStahn: @KjellSwedin What's relevant for you here is that the research decision is to go with the version of the Hb shift (including baseline calibration) that's in the "Summary of Iron Intervention Algorithm" section in this PR, and *not* the version Ali posted above in the "Determining Whether A Simulant is Affected - Iron" section, which is slightly different and doesn't handle baseline calibration. Does that make sense?

I'm about to start updating the other sections to be consistent with the new version here. Let me know if you have any questions.

PR #248, opened 2020-06-03
LSFF: hemoglobin updates

NathanielBlairStahn:
Revised description of hemoglobin shift to be consistent with the "Summary of Iron Intervention Algorithm" section, reflecting the most recent research decisions.

### DIFF
--- a/docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
+++ b/docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
@@ -1163,7 +1163,7 @@ time since the simulant started eating iron-fortified food:
   """
   Multiplier on Hb effect size due to lag in response time to iron fortification.
   The argument `time_since_fortified` is the time (in years) since a simulant
-  started eating fortified food (a negative value of `time_since_fortified`
+  started eating fortified food (note that a negative value of `time_since_fortified`
   indicates the child has not yet started eating fortified food).
   """
     return (0                         if time_since_fortified < 0   else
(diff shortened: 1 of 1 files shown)


docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1115,63 +1115,165 @@ For the purposes of our simulation, we made a few assumptions:
...
+    return (0                         if time_since_fortified < 0   else
+            time_since_fortified/0.5  if time_since_fortified < 0.5 else
+            1)
+
+.. Note::
+
+  Since children don't start eating fortified food until 6 months of age, the
+  argument ``time_since_fortified`` for the time-lag function will be *either*
+  the time since the child's household got fortification coverage, *or* the time
+  since the child turned 6 months old, *whichever is smaller*.
```
KjellSwedin, line 1178: Is this consistent with the similar comment in the code block below?

> # Fortification starts either when the household receives coverage, or when
> # time_since_coverage_i	    # the child turns 6 months old and starts eating solids, whichever is later.
>

NathanielBlairStahn: Yes, I think so: Later time = smaller duration of time since it happened

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1115,63 +1115,165 @@ For the purposes of our simulation, we made a few assumptions:
...
+    they receive the full effect if they are at least one year old, or
+
+    b) If they are less than one year old, they have started or will start
+    eating fortified food as soon as they turn(ed) 6 months old, hence will
+    achieve the full age-dependent effect at exactly one year old.
+
+To summarize, we are trying to account for two factors that will decrease the
+reported effect size of iron fortification on hemoglobin in our simulants:
+
+1.  (**Age effect**) Children don't start eating (fortified) solid food until 6
```
aflaxman, line 1132: don't start eating _(much)_ (fortified) solid food

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1115,63 +1115,165 @@ For the purposes of our simulation, we made a few assumptions:
...
+.. Note::
+
+  When calibrating hemoglobin levels for baseline coverage (see the
+  :ref:`baseline calibration <baseline_calibration_hb_bw_section>` section
+  below), we will use **only** the age effect to compute the treatment-deleted
+  hemoglobin levels of the population (the down-shift), whereas we will use
+  **both** the age effect *and* the time-lag effect when applying the treatment
+  to an individual simulant (the up-shift). This will cause our population's
+  overall mean hemoglobin level to be slightly lower than the value in GBD, but
+  not by much, and this is the simplest strategy that seems reasonable.
```
aflaxman, line 1189: 👍

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1115,63 +1115,165 @@ For the purposes of our simulation, we made a few assumptions:
...
+
+.. Note::
+
+  When calibrating hemoglobin levels for baseline coverage (see the
+  :ref:`baseline calibration <baseline_calibration_hb_bw_section>` section
+  below), it is *possible* (but highly unlikely with the shift size we're dealing with) that a simulant's hemoglobin level can end up *negative* after the down-shift. **In the case of a negative hemoglobin level, we will arbitrarily reset the hemoglobin level to 10 g/L**.
```
aflaxman, line 1195: That is pretty uncool---if it is .1, you leave it, but if it is -.1, you make it 10?  Better to clip everything to 10, instead: `hb_level = np.clip(orig_hb_level, 10, np.inf)`.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -1115,63 +1115,165 @@ For the purposes of our simulation, we made a few assumptions:
...
+.. Note::
+
+  When determining whether a simulant gets covered by our iron fortification
+  program, the child's propensity for coverage (for the hemoglobin effect)
+  should be **the same** as their mother's propensity for coverage (for the
+  birthweight effect). In other words, when comparing the random propensity
+  against the iron fortification :ref:`coverage function
+  <coverage_algorithm_section>`, the same random number should be used to
+  determine whether the child's mother ate fortified food during pregnancy as is
+  used to determine whether the child eats fortified food.
```
aflaxman, line 1215: 👍  but we could consider making them positively correlated and not identical in a future iteration.  Maybe a sentence in a limitations section.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: Ok great, thanks for the suggestion @aflaxman . The number 10 was pretty arbitrary - is there any is there any better methodology to picking a cutoff?  Really we just need it to be positive, so what about using a cutoff of 0 (or some very small positive number)?  It seems like that would be more consistent with the *non*-iron-responsive individuals, whose Hb level doesn't get adjusted at all.

aflaxman: Google thinks 10 g/L is a very small positive number:
image

NathanielBlairStahn: @aflaxman Well yes, that's why I arbitrarily chose 10 - I figured once it's that low, the actual value doesn't really matter. But there are several issues here:

1. If 10 g/L is absurdly low, why not truncate at 50, or 75, or some other number?  I'm saying we have no good way of choosing a good cutoff (without doing extra research to make sure we get it right), so why not just go with the only cutoff we know for sure, which is 0?

2. If we're going to try to make the cutoff for the shifted distribution more realistic, then should we modify the unshifted GBD data to have more realistic values as well?  Namely, non-iron-responsive simulants will have Hb levels according to the GBD ensemble distribution (40% gamma, 60% mirror Gumbel), which I believe is supported on all real numbers, hence could theoretically be negative.  If we're truncating the shifted version of this at 10 g/L, should we also truncate the unshifted version at 10g/L?  Why not 50, or 75, or some other value?

3. Doing this truncation isn't necessarily about making our model more realistic - it's a hack to correct for our dumb method of applying the effect size and baseline calibration, to make sure our model doesn't give us blatantly wrong results.  For the LSFF model, I think the probability of seeing a negative Hb value will be so low that we will never actually run into this issue, but I remember James saying they *did* get some negative birthweights in the BEP model (which had a larger shift), and it should be the researchers' job to decide how to fix that rather than the SWEs'.  So I guess I'm trying to figure out what should be our general methodology.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
NathanielBlairStahn: I think saying "don't start eating *much* fortified solid food" would just muddle the description. I'm trying to clearly describe our algorithm to tell the SWEs what to build (and to remind the researchers what we decided to build), and our algorithm says children don't eat *any* fortified food until 6 months.  The discussion of how we arrived at this decision comes earlier in the section. If we don't think this approximation is close enough, then we should just change the function.

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
aflaxman: A principled way to approach it would be to look at the RR and see if it gives us an answer.

PR #257, opened 2020-06-17
LSFF: desired outputs

NathanielBlairStahn:


alibow:
This looks good to me - seems to be sufficient and clear.

alibow:
It appears to be consistent with my understanding of the outputs we have had so far except for the iron scenario which is missing the stratification by iron coverage group (maternal or child) and I have not seen any of the hemoglobin outputs yet.

NathanielBlairStahn:
Thanks @alibow . Here are some other things I think are different between what's here and what we actually have in the outputs:

1. My definitions of the stratification groups for fortification coverage do not match our actual outputs (fixed throughout simulation vs. current simulant state) - I should probably change my definitions to match the outputs
2. We do not have observers for state person time with fortification coverage
3. We do not have live births with LRI
4. Anemia is not stratified by iron responsiveness or child iron coverage group
5. I don't know whether Hb is stratified by year and sex
6. There are some differences in how means, variances/std devs, and low birthweight prevalence are reported
7. I haven't yet fully specified which stratification we want where

docs/source/gbd2017_models/concept_models/vivarium_conic_lsff/concept_model.rst
```diff
@@ -2018,41 +2018,74 @@ Desired Model Outputs
...
+      possible that this method can become `numerically unstable
+      <https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance>`_ if
+      the population is very large or the values of :math:`Y` are very large,
+      though typically this should not be a problem.
+
+  **To do:** Include more explicit formulas to make it clearer exactly how the
+  above two options work. Also, be clearer with notation for population
+  parameters vs. statistical estimates. Also, somewhere (perhaps not here)
+  describe the alternative definitions of mean and variance as
+  person-time-weighted averages.
```
aflaxman, line 2067: Perhaps we researchers should also figure out if the law of total variance (approach 1) is preferred to the moments approach (2) because of the numerical stability issue you have identified.

PR #258, opened 2020-06-17
outline of multiple myeloma

yaqi7:
Please help review if the file is located at right directory and the outline looks good.


SantoniS:
I think the outline looks good too!

PR #265, opened 2020-06-25
Cohort study overview

beatrixkh:


beatrixkh:
notes from docs session:
switch n to k
add 'polytomous ord/unord' terminology
compare in contrast to case control, good for rare outcomes
note can track people for a variety of different outcomes

rework awkward confounding sentence: 'wrt all variables that are potential confounders, known or unknown'

note historical studies used often (most often?) with occupational data. i.e.. mining exposures.

PR #269, opened 2020-07-01
GBD 2017 Modeling Strategy preliminary version

yaqi7:
Want to make sure the diagram is in right direction. 
The transition rate from initial MM infection to relapse/refractory (RR) is almost 100%, will check with the modeler if measures of RR should be treated the same as these of initial infection. And do you think it makes sense to branch treated RR group from initial MM (the diagram)?


yaqi7:
> I think we should deal with treatment separately. Do you need it in the cause model?
I don't think the treatment rectangular is necessary in the cause model for now. Let me remove the dashed part to avoid confusion.

yaqi7:
> I think we should deal with treatment separately. Do you need it in the cause model?

I don't think the treatment rectangular is necessary in the cause model for now. Let me remove the dashed part to avoid confusion.

PR #270, opened 2020-07-01
breast_cancer_cause_model_updates

yongqx:


docs/source/gbd2017_models/causes/neoplasms/breast_cancer_with_stage0/cancer_model.rst
```diff
@@ -277,23 +277,66 @@ GBD does not give us any information on the prevalence or incidence of DCIS or L
...
+      from marketscan to inform DCIS and LCIS prevalence in 2020-2040.
+    - A major assumption of this method of using marketscan ratios is that the
+      ratio of DCIS or LCIS to breast cancer in the US population is the same as
+      that in the Chinese population we are modelling. This could be a limitation
+      if breast cancer manifests differently among racial groups.
+    - We believe that the crude prevalence ratio of DCIS and crude prevalence
+      ratio of LCIS are independent, meaning no correlation between these two
+      ratios.
+    - The draw-level DCIS and LCIS prevalence ratios can be generated based on
+      a truncated normal distirbution (lower bound = 0) with given mean and
```
aflaxman, line 292: I think when the ratio is very close to the lower bound things will get weird, so let's keep an eye on this. If necessary, we can make the lower bound 0.1 to address weirdness.

PR #273, opened 2020-07-07
breast cancer output shell table update

yongqx:


aflaxman:
@rmudambi this PR makes it hard to see what has changed.  The key difference is that we recently realized that we also need to know how many _deaths were due to undetected breast cancers_.

yongqx:
> @rmudambi this PR makes it hard to see what has changed. The key difference is that we recently realized that we also need to know how many _deaths were due to undetected breast cancers_.

Will do list-table or csv in future documentation so that Sim Science folks can review it on github.

PR #275, opened 2020-07-09
bep concept model

Ninicorn:


docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
```diff
@@ -0,0 +1,353 @@
...
+
+Balanced energy protein (BEP): Under-nourished women are negatively associated with fetal growth and have increased risk of preterm birth. Balanced protein (where protein content provides less than 25% of total energy content) supplementation during pregnancy may help with gestational weight gain and subsequently avert adverse pregnancy outcomes, but the benefits are unclear. WHO recommends BEP for populations or settings with a high prevalence of undernourished pregnant women (not for individual pregnant women identified as being undernourished) to reduce the risk of stillbirths and small for-gestational-age neonates. Undernourishment is usually defined by a low BMI (i.e. being underweight BMI<18.5). For adults, a 20–39% prevalence of underweight women is considered a high prevalence of underweight and 40% or higher is considered a very high prevalence. MUAC may also be useful to identify protein–energy malnutrition in individual pregnant women and to determine its prevalence in this population. However, the optimal cut-off points may need to be determined for individual countries based on context-specific cost–benefit analyses.
```
alibow, line 69: I has been unclear to me whether MMN *inherently* contains IFA and whether BEP *inherently* contains IFA and MMN (as part of their ingredients) or whether it is *advised* that IFA is taken with MMN and that IFA and MMN are taken with BEP. Maybe this would be good to clarify here.

docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
```diff
@@ -0,0 +1,353 @@
...
+
+3.0 Causal framework
+++++++++++++++++++++
+
+.. _3.1:
+
+3.1 Causal diagram
+------------------
+
+  .. image:: causal_dag_bep_scenarios.svg
```
alibow, line 110: this is not rendering correctly. Not exactly sure why.

docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
```diff
@@ -0,0 +1,353 @@
...
+
+**Effect modifiers**:
+
+  - The effect of BEP on birthweight is differential according to maternal BMI
+
+:underline:`Current evidence scenarios`
+
+• Under the current evidence scenarios, the intervention BEP affects only birthweight. It does so differentially by maternal BMI status (Ota 2015). BMI is an effect modifier on the relationship between BEP and birthweight. 
+
+• BMI should have an effect on LAZ and WLZ scores suggested by the literature (and Gates) but we are not intervening on BMI, and so simply correlatng birthweight and LAZ / WLZ will capture the relationship between BMI and LAZ/WLZ.
```
alibow, line 127:
```suggestion
• Maternal BMI should have an effect on child LAZ and WLZ scores suggested by the literature (and Gates) but we are not intervening on maternal BMI, and so simply correlating birthweight and LAZ / WLZ will capture the relationship between maternal BMI and child LAZ/WLZ.
```

alibow: I think it would also be helpful to include a note about WHY it is important to correlate these things

docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
```diff
@@ -0,0 +1,353 @@
...
+• note x represents a crude effect size because we are not intervening on BMI and we only use this effect in baseline. (ideally crude mean shift birthweight by maternal BMI)
```
alibow, line 128:
```suggestion
• note x represents a crude effect size because we are not intervening on maternal BMI and we only use this effect to capture the *non-causal* baseline correlation between pre-pregnancy maternal BMI and child birthweight (ideally crude mean shift birthweight by maternal BMI)  
```

docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
```diff
@@ -0,0 +1,353 @@
...
+
+  - The effect of BEP on birthweight is differential according to maternal BMI
+
+:underline:`Current evidence scenarios`
+
+• Under the current evidence scenarios, the intervention BEP affects only birthweight. It does so differentially by maternal BMI status (Ota 2015). BMI is an effect modifier on the relationship between BEP and birthweight. 
+
+• BMI should have an effect on LAZ and WLZ scores suggested by the literature (and Gates) but we are not intervening on BMI, and so simply correlatng birthweight and LAZ / WLZ will capture the relationship between BMI and LAZ/WLZ.
+• note x represents a crude effect size because we are not intervening on BMI and we only use this effect in baseline. (ideally crude mean shift birthweight by maternal BMI)  
+• Originally Gates wanted us to model an effect size of RR=2 for BMI on LAZ and WLZ scores, which we did in our februrary model. However, since we are not intervening on BMI, we do not need to model this causal effect. We only want the baseline LAZ and WLZ scores by BMI status and this relationship should be captured by the correlation co-efficient between birthweight and LAZ and WLZ score.
```
alibow, line 129:
```suggestion
• Originally Gates wanted us to model a *causal* effect size of RR=2 for BMI on LAZ and WLZ scores, which we did in our Februrary model. However, since we are not intervening on BMI, we do not need to model this causal effect. We only want the baseline LAZ and WLZ scores by BMI status and this relationship should be captured by the correlation co-efficient between birthweight and LAZ and WLZ score, given that we also model the association between maternal BMI and birthweight.
```

docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
```diff
@@ -0,0 +1,353 @@
...
+• BEP only affects pregnancy weight-gain (given during pregnancy at ANC) and not pre-pregnancy BMI, hence no causal arrow from BEP to BMI (pre-pregnancy).
+
+:underline:`Hopes and dreams scenarios`
+
+• Current evidence does not show an effect between BEP and LAZ or WLZ scores (Ota 2015). But there is reason to believe it should affect child growth and hence we are modelling this effect in the hopes and dreams scenario. This is the effect that gates is trying to find in their current trials.
+• Potential reasons why current literature may not capture the effect of BEP on WLZ/LAZ
+  o While the literature has strong evidence there is some causal effect size between birthweight LAZ n WLZ (Harding 2017), the effect from BEP through birthweight might not be big enough to show an effect in LAZ or WLZ through birthweight or the studies might not be powerful enough to detect a difference. 
+  o Following up to child-growth failure outcomes require a long follow-up period, hence this outcome might not be measured accurately in current studies.
```
alibow, line 137: These bullets aren't rendering correctly. I think you need to use "-" only with the appropriate tab indents.

docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
```diff
@@ -0,0 +1,353 @@
...
+
+
+
+.. _4.0:
+
+4.0 Intervention
+++++++++++++++++
+
+• IFA: Iron and folic acid supplementation 
+- Dosage: Combined pill or separate pill (30-60 mg iron, 400 μg folic acid)
```
alibow, line 220: I think these need a tab indent to show up as intended!

docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
```diff
@@ -0,0 +1,353 @@
...
+
+4.1 Simulation scenarios
+------------------------
+
+• Baseline: Offering IFA to any women attending ANC (at empirical baseline coverage of IFA)
+• Scenario 1: Offering MMN to any women attending ANC (intervention at target 90% coverage at ANC)
+• Scenario 2: Offering BEP (current evidence effect size) to any women attending ANC (intervention at target 90% coverage at ANC)
+• Scenario 3: Offering BEP (current evidence effect size) to targeted undernourished women, and MMN to any other women attending ANC (intervention at target 90% coverage at ANC)
+• Scenario 4: Offering BEP (hope and dreams effect size) to any women attending ANC (intervention at target 90% coverage at ANC)
+• Scenario 5: Offering BEP (hope and dreams effect size) to targeted undernourished women, and MMN to any other women attending ANC (intervention at target 90% coverage at ANC)
```
alibow, line 246: So no target coverage of IFA alone?

docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
```diff
@@ -0,0 +1,353 @@
...
+
+5.0 Vivarium modelling components
++++++++++++++++++++++++++++++++++
+
+.. _5.1:
+
+5.1 Vivarium concept model 
+--------------------------
+
+  .. image:: vivarium_conceptdiagram_bep.svg
```
alibow, line 258: This isn't rendering correctly for me either :/

docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
Ninicorn: We just had an discussion with James, and he says that the correlation co-efficient between birthweight and LAZ/WLZ scores would **not** capture the relationship between BMI and LAZ/WLZ scores because of the way we are doing correlation (by the marginal distributions).  The way we had it in our february model, we _induced_ correlation between birthweight and LAZ/WLZ by means of the common cause of BMI so we should get the 'correct-ish' birthweight, LAZ, WLZ means by BMI subgroups.  

So the step forward for now is to check how well birthweight and LAZ/WLZ score correlate in our feb model (with updated LRI and LBW risk models) against the MAL-ED correlation coefficients that Chris produced. If it captures all the correlation, then good, else we might need to think again. 

Because we induced correlation by maternal BMI, we should only see the correlation structure when NOT stratified by BMI groupings. If we stratify by BMI groupings, the correlation should disappear. 

Nicole will run an interactive sim (with the help of Kjell) and produce individual level data at a time-point for those aged 6 months+, stratified by age (can collapse over sex as there is no effect of sex on bw/laz/wlz) and plot birthweight with laz and birweight with wlz to get the correlation coefficients and compare them with the MAL-ED data. 

@collijk @alibow @KjellSwedin @alibow @aflaxman Does this make sense? Let me know if I've missed anything!

docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
Ninicorn: Ya...if I remember correctly, we didn't do a scale up of IFA scenario...I need to update the DAGs titles

docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
Ninicorn: aww man...it is rendering correctly in my make html though...

alibow: Hmm, weird! Maybe it's just me then.

docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
alibow: Hmm, I am still confused by this. I think I am not understanding how we are using marginal distributions in a way that would not capture the association between maternal BMI and child stunting/wasting. Could you please explain that some more?

Ninicorn: image

I think...it's cuz the low BMI and normal BMI mothers have the different distributions with different means. so the low BMI mothers will draw from a birthweight distribution with lower means, and the high BMI mothers will draw from a birthweight dsitrubion with higher means, but the LAZ/WLZ distribution is the same. 

Now I'm wondering why we can't just draw at the lower end of the birthweight distribution (if low BMI) and higher end of the distribution (for high BMI mothers). As in we have one distribution for birthweight and just draw from different areas of the same distribution depending that is determined by the birthweight mean difference among BMI subgroups...?

alibow: so even if we have individual birth weights for each simulant (which will be based on their maternal BMI status), we cannot sample corresponding correlated WLZ and LAZ scores using the same methodology as above based on their existing birth weight?

docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
Ninicorn: yes, exactly

docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
Ninicorn: you are right! it is confusing. I've clarified in the new concept model update. Namely, I am referring to MMN as oMMN to mean micro-nutrients _other_ than folic acid and iron. Additionally, I have grouped things into 1) effect sizes (that of IFA, that of oMMN, and that of BEP) 2) interventions and 3) scenarios. The interventions are an additive combination of effect sizes and the scenarios are a combination of interventions.

docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
Ninicorn: BEP_correlation_vis

this has not yet been resolved and hoping to chat with Abie/James tomorrow (july 22 2020). I'm uploading an image of my understanding of it for the record.

PR #277, opened 2020-07-15
Mortality impacts - cause documentation

alibow:
This was something that I had worked on long ago for our cause model documentation and recently found it again when I got my old computer back after it had been fixed! This is a fairly low priority PR to review I believe, but maybe we can go over it briefly in an upcoming vivarium research documentation session. Thanks :)


docs/source/model_design/cause.rst
```diff
@@ -666,12 +666,148 @@ Severity Splits
...
+Notably, ACMR is used both for validation of Vivarium simulations, as well as 
+for estimating simulation mortality rates (see the `Mortality Hazards`_ 
+section for more detail).
+
 Cause-Specific Mortality
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
+Cause-specific mortality rate (CSMR) is a measure of deaths due to a 
+particular cause (or group of causes) per person-year in the overall age-, 
+sex-, time-, and location-specific population. Specifically,
```
aflaxman, line 697: Embedded in this innocuous statement is the pernicious assumption that every death has precisely one underlying cause.  I propose we call this assumption out around this point in the text, and link to the causality material for readers who want to dive down that rabbit hole.

alibow: Thanks, Abie! I included some notes about this in the latest update.

Ninicorn (approved):

PR #280, opened 2020-07-17
Create pages for GBD 2019

alibow:
Hello! For a pending rush BMGF project on maternal anemia, we are planning to use GBD 2019 results, so I made this PR to create pages for GBD 2019 models. I assume that there will be a lot of overlap between some of our 2017 and 2019 models, so if anyone has any opinions about how we should structure this, please let me know! I was thinking that we would have separate pages for LRI 2017 and LRI 2019 for example (even if we just copy and paste a lot of it over), but I am definitely open to suggestions!

Thanks :) Ali


alibow:
> Are concept models and intervention models specific to a GBD round? I see how the might be, but I also dream of them _not_ being.

Good point! I pull the concept model pages out of the GBD 2017 and GBD 2019 folders in the latest update.

alibow:
> Thanks for starting this Ali!
> 
> I see one 2017 left in the first sentence of the Cause Model Template - "To begin documenting a cause in GBD 2017 for a Vivarium simulation". Other than that looks great to start this off!

Thanks, Paulina! I updated this in the latest commit :)

PR #285, opened 2020-07-24
updated DAGs to reflect correlation models i,ii,iii using Abie's method

Ninicorn:


docs/source/gbd2017_models/concept_models/vivarium_gates_bep/concept_model.rst
```diff
@@ -432,11 +432,11 @@ Apply the following intervention shifts according to interventions recieved in e
 .. image:: intervention_shifts.svg
 
 .. important:: 
-  note the highlighted effect sizes, which are different from the previous feb_2020 model.
+  note-
    
-   1) oMMN size has changed
-   2) there is now a current evidence effect size for BEP
-   3) the shift in laz/wlz scores due to maternal BMI status can be kept in there if it is currently in there already, as it doesnt make any difference with or without it in scenarios A-E, only in F(i) targeted and F(ii) targeted. 
+   1) oMMN size has changed from feb_2020 model
```
KjellSwedin, line 437: Could I get the from and to values? The code currently has these:
```
MMN_BIRTH_WEIGHT_SHIFT_SIZE_MEAN = 22.4
MMN_BIRTH_WEIGHT_SHIFT_SIZE_LOWER = 8.3
MMN_BIRTH_WEIGHT_SHIFT_SIZE_UPPER = 36.4
MMN_BIRTH_WEIGHT_SHIFT_SIZE_LOWER_BOUND = 0
MMN_BIRTH_WEIGHT_SHIFT_SIZE_UPPER_BOUND = 45
```
Do you know where the 0, 45 upper and lower bounds came from?

aflaxman: The doc in the repo does not go back far enough to recover these values, but I have a word-doc version of this last edited 2020-02-03 that indicates these come from Fall (2009):
image

I suspect the 0 and 45 bounds were chosen to keep the extremes from making surprises.

Since Nicole is perhaps offline for the weekend, I propose we interpret the current doc to mean
```
MMN_BIRTH_WEIGHT_SHIFT_SIZE_MEAN = 51.24
MMN_BIRTH_WEIGHT_SHIFT_SIZE_LOWER = 43.71
MMN_BIRTH_WEIGHT_SHIFT_SIZE_UPPER = 58.76
MMN_BIRTH_WEIGHT_SHIFT_SIZE_LOWER_BOUND = 0
MMN_BIRTH_WEIGHT_SHIFT_SIZE_UPPER_BOUND = 75
```
This is a dramatic increase in the efficacy of MMN on birth weight!  @Ninicorn can you double check this when you are back from your weekend?

plinds:
In the future it would be great to have the output shell in .csv format rather than excel - James has mentioned this as being easier to see in PR's rather than having to download the excel file.

Ninicorn:
DELETED THIS BRANCH because of merge conflicts. new branch was made.

PR #298, opened 2020-09-04
Iron deficiency risk exposure

alibow:


aflaxman (approved):

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/risk_exposure/index.rst
```diff
@@ -7,3 +7,172 @@ Iron Deficiency Risk Exposure
...
+
+	The iron deficiency TMREL was estimated as the 95th percentile of mean population hemoglobin concentration among all GBD locations for each year-, sex-, and age-specific demographic group. If the observed mean population hemoglobin concentration for a given location was greater than the 95th percentile, then the observed population hemoglobin concentration was used as the TMREL for that location. See the specific code `here <https://stash.ihme.washington.edu/projects/MNCH/repos/anemia_causal_attribution/browse/iron_deficiency/calculate_normal_hgb.py>`_.
+
+	This value is stored as the TMREL for the iron deficiency risk factor in GBD 2019.
+
+#. Calculate iron deficiency risk exposure
+
+	The iron deficiency risk factor is calculated according to the following equation, as documented `here <https://stash.ihme.washington.edu/projects/MNCH/repos/anemia_causal_attribution/browse/iron_deficiency/nutrition_iron_exposure.py>`_ and externally validated `here <https://github.com/ihmeuw/sim_sci_maternal_anemia/blob/master/data_validation/maternal_disorders_burden/custom_paf_calculations.ipynb>`_.
+
+	:math:`exposure = TMREL - (p_* * (TMREL - mean_\text{Hgb}))`
```
NathanielBlairStahn, line 37: The outer set of parentheses is unnecessary (maybe a bit nitpicky, but it might be a bit more readable without them)

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/risk_exposure/index.rst
```diff
@@ -7,3 +7,172 @@ Iron Deficiency Risk Exposure
...
+    - 389
+  * - End-Stage Renal Disease on Dialysis
+    - Sequela
+    - 22989, 22990, 22991, 22992, 22993, 22999, 23000, 23001, 23002, 23003, 23009, 23010, 23011, 23012, 23013, 23019, 23020, 23021, 23022, 23023
+
+.. note::
+
+	There are some diseases that cause iron deficiency that are available in the necessary format as modelable entities or sequelae rather than causes, which is why several ID types are listed in the table above.
+
+#. Estimate iron deficiency risk exposure as the observed mean population hemoglobin concentration
```
NathanielBlairStahn, line 144: This second step is also numbered "1." in the rendered version. Maybe because your list-table and note are not indented to the same level as your numbered list?

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/risk_exposure/index.rst
```diff
@@ -7,3 +7,172 @@ Iron Deficiency Risk Exposure
...
+
+This estimation approach is described in the GBD 2019 Risk Factor Methods Appendix, but NOT used in GBD 2019 iron deficiency risk exposure calculations. Therefore, the following description does not reflect the values stored for the GBD 2019 iron deficiency risk factor, but represent an alternative estimation approach that is "conceptually/clinically more correct" than the Global TMREL exposure estimation approach according to a correspondence with Nick Kassebaum, although it did not meet the GBD 2019 requirement that TMRELs are not location-specific.
+
+The location-specific TMREL exposure estimation approach has the following steps:
+
+#. Estimate a location-specific TMREL, defined as the counterfactual population mean hemoglobin concentration in the absence of all diseases that cause iron deficiency
+
+	:math:`TMREL = mean_\text{Hgb} + \sum_{c=1}^{n} p_c * shift_c`
+
+	Where, :math:`\sum_{c=1}^{n} p_c * shift_c` represents the sum of the cause-specific prevalence multiplied by the cause-specific hemoglobin shift for all causes of iron deficiency and :math:`mean_\text{Hgb}` is the observed mean population hemoglobin concentration (MEID 10487).
```
NathanielBlairStahn, line 67: Looks like this (and the equation above) is indented farther than the text after #. in your first list item, so it shows up as a blockquote rather than just continuing the preceding paragraph. Not sure if that's what you intended.

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/risk_exposure/index.rst
```diff
@@ -7,3 +7,172 @@ Iron Deficiency Risk Exposure
...
+
+	An estimation of the TMREL using this method can be found `here <https://github.com/ihmeuw/sim_sci_maternal_anemia/blob/master/data_validation/maternal_disorders_burden/custom_paf_calculations.ipynb>`_.
+
+	The cause-specific hemoglobin shifts represent the average difference in hemoglobin concentration among individuals afflicted with that cause and the general population (from the literature) and can be found `here <https://github.com/ihmeuw/sim_sci_maternal_anemia/blob/master/data_validation/maternal_disorders_burden/hb_shifts.csv>`_. Note that these shifts are sex-specific but do not vary by age. Additionally, these shifts apply to mean hemoglobin concentration only; it is assumed that there is no change in hemoglobin distribution standard deviation, which is a limitation in this analysis due to a significant data gap.
```
NathanielBlairStahn, line 71: Where did you get the list of shifts? Is there a source you can cite?

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/risk_exposure/index.rst
NathanielBlairStahn: Can "TMREL - mean_{Hgb}" be interpreted as the hemoglobin shift corresponding to the "aggregate cause" consisting of all causes of anemia except iron deficiency?  I'm trying to understand how the global TMREL approach relates to the local TMREL approach described below.

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/risk_exposure/index.rst
```diff
@@ -7,3 +7,172 @@ Iron Deficiency Risk Exposure
...
+
+	:math:`exposure = TMREL - (p_* * (TMREL - mean_\text{Hgb}))`
+
+	Where, :math:`p_*` is the summed prevalence of all anemia caused by any disease except dietary iron deficiency (using the MEIDs listed below) and :math:`mean_\text{Hgb}` is the observed mean population hemoglobin value (MEID = 10487).
+
+	This exposure value is stored as the exposure for the iron deficiency risk factor in GBD 2019.
+
+	The iron deficiency risk exposure estimated in this way can be interpreted as the counterfactual population mean hemoglobin concentration in the absence of all diseases that do NOT cause iron deficiency, according to a correspondence with Nick Kassebaum, who leads the GBD team responsible for these estimates. However, it is important to note that the code used does not consider diseases that cause non-dietary iron deficiency.
```
NathanielBlairStahn, line 43: These two sentences are a bit hard to follow (e.g. the double negative in "absence of all diseases that do NOT cause iron deficiency") - I'm having trouble wrapping my head around which causes belong where (causing iron deficiency vs. not causing iron deficiency).

Based on the above description of "summed prevalence of all anemia caused by any disease except dietary iron deficiency," I would have expected the list of MEIDs for this sum to include the diseases from the location-specific section except for MEID 390, i.e. the diseases that cause iron deficiency (but not *dietary* iron deficiency), but it looks like those are *not* included.

I think the fact that the iron-deficiency-causing diseases are not included matches the statement that the exposure "can be interpreted as the counterfactual population mean hemoglobin concentration in the absence of all diseases that do NOT cause iron deficiency," but then it seems like p_* should be defined as "the summed prevalence of all anemia caused by any disease that do not cause iron deficiency," or something.

Also, the fact that the iron-deficiency-causing diseases are not included seems at odds with the statement "However, it is important to note that the code used does not consider diseases that cause non-dietary iron deficiency," since it seems like it's specifically excluding all these diseases.

I think I'm confused -- what am I missing???  I'm rather tired, which doesn't help...

NathanielBlairStahn (approved):
Well, I still don't totally understand their methodology, but your write-up is helping me get closer!

I'm approving so we can get a version of this merged into master, but see my comments above for potential things to address in a future version.

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/risk_exposure/index.rst
NathanielBlairStahn: Also, where is the assumption of no change in hemoglobin distribution standard deviation used?  And what are its implications?

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/risk_exposure/index.rst
alibow: Yeah, I am really confused by the GBD methods used here too... what they did really doesn't make sense to me, which is why I just quoted what Nick said for the interpretation. The location-specific approach they described in the appendix is a lot more intuitive to me

alibow: Essentially, I agree that the description of the approach and the explanation of what it represents seem contradictory

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/risk_exposure/index.rst
alibow: No, this is really just the difference in observed hemoglobin in a given location compared to the 97.5th percentile of all observed hemoglobin concentrations globally, which is attributable to all causes of low hemoglobin, including iron deficiency, non iron deficiency, pregnancy, and anything else. This is why this approach is a little non-sensical to me.

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/risk_exposure/index.rst
alibow: Came from the iron deficiency modelers and is hosted in their stash repo which is linked

alibow: The standard deviation assumption is used when applying the cause-specific shifts... it assumes that the entire shape of the hemoglobin distribution is preserved when a given cause of anemia is "removed" from the population, when in reality the shape and standard deviation of the distribution will likely change because hemoglobin observations from the lower end of the distribution will be more likely to be "removed." Does that help clarify? 

I believe (although I am not certain about this) that the implications are that we will underestimate the TMREL because of this. Do you agree?

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/risk_exposure/index.rst
alibow: Which is why we did not choose to follow this methodology for our previous project, we thought it was more valuable to have methods that were clear to explain and understand for the client

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/risk_exposure/index.rst
NathanielBlairStahn: I'm not sure.  If we're just talking about the equation for the TMREL, I don't see any place where the shape of the distribution comes into play, because means are additive regardless of the shape of the distribution.  How exactly do you think the assumption of constant variance will cause the TMREL to be underestimated?

I think if we wanted to compute the entire TMRED (i.e. the theoretical minimum risk distribution, not just its mean), then certainly the variance of this distribution would be smaller than the variance of the overall Hgb distribution before causes of anemia are removed.  But I think to calculate its mean (which is what it looks like they're calling the TMREL), the given equation should work (at least under the assumption that causes of low hemoglobin are mutually exclusive, which seems suspect... not sure if it should still work without that assumption).  Do you know where in their methodology the standard deviation(s) of the distribution(s) get used?

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/risk_exposure/index.rst
NathanielBlairStahn: OK :)

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/risk_exposure/index.rst
alibow: Oh good point, I think you're right.

So it's a continuous risk model, so when calculating the PAF, they use the ensemble distribution of individual values within a population compared to the TMREL. I agree that this will not matter when calculating the TMREL as a point value, but when a counterfactual exposure distribution is assumed, the constant variance assumption will impact the PAF calculation, I believe.

I'll update this to try to make it more clear.

alibow: Oh good point, I think you're right. I will update this.

PR #303, opened 2020-09-09
Re-opened hemoglobin model PR

alibow:
Sorry I accidentally deleted the old PR :(, so here it is again :)


alibow:
Abie had left this comment on the old PR before I accidentally deleted it: 

I love the "validation criteria" section. Everybody take a look at this, we should be doing them for everything!

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
```diff
@@ -6,4 +6,237 @@ Hemoglobin Model
...
+
+    Covariates for the standard deviation model included: Malaria Incidence, Haemoglobin C (sickle type C) trait, Haemoglobin S (sickle type S) trait, Sociodemographic Index, SEV for Impaired kidney function, Healthcare Access and Quality index, Education Relative Inequality (Gini), 50th percentile of haemoglobin (pooled across all microdata sources), and mean haemoglobin (results from mean [Hb] ST-GPR model)
+
+2. Calculation of ensemble weights
+
+    A set of two-parameter distributions (gamma, mirror gamma, Weibull, mirror lognormal, and mirror gumbel) were fit to the sample’s haemoglobin mean and variance for each location/year/age/sex group
+
+    The weights used for the GBD 2019 hemoglobin distribution model were 40% gamma and 60% mirror gumbel.
+
```
NathanielBlairStahn, line 31: I think I understand how the ensemble weights work because I've seen them before, but I don't think this provides enough clarity for an unfamiliar reader.

Here's my understanding of the steps for finding an ensemble distribution (correct me if I'm wrong):
1. Each of the two-parameter distributions is "fit" to the data in each location/year/age/sex group using the "[method of moments](https://en.wikipedia.org/wiki/Method_of_moments_(statistics))." I.e. you just choose the distribution that has the same mean and standard deviation as the data in each group, so there's not actually any complicated curve fitting going on here.

2. The distribution F of hemoglobin is assumed to be a [mixture distribution](https://en.wikipedia.org/wiki/Mixture_distribution) that also comes from a 2-parameter family (GBD calls this an ensemble), of the form
F(x; m_i,s_i) = w_1 F_1(x; m_i,s_i) + ... + w_k F_k(x; m_i,s_i),
where
m_i = mean Hb in group i
s_i = standard deviation of Hb in group i
F_1,...,F_k are the distributions in the list above (gamma, mirror gamma,etc.), parameterized by mean m and standard deviation s
F is the mixture distribution for Hb, also parameterized by mean m and standard deviation s
w_1,...,w_k are the **global** weights of the distributions in the mixture, used in all groups
(I think the important thing to note here is that the distribution weights are *global*, not depending on the location/year/age/sex, whereas it is only the mean and standard deviation that vary across groups. This is not totally clear from your description above.)

3. The global weights w_1,...,w_k are found by solving an optimization problem that somehow chooses the "best" weights that match the data in all the groups simultaneously. According to the YLD appendix, this "best fit" is defined by taking anemia prevalence data into account.  Whatever they did to solve this optimization problem, the weights they came up with are w_1 = 0.4 (for F_1 = gamma distribution) and w_2 = 0.6 (for F_2 = mirror gumbel distribution), with w_j = 0 for all the other distributions.

alibow: Yes, this is consistent with my understanding. I will add this detail to the existing documentation

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
```diff
@@ -6,4 +6,237 @@ Hemoglobin Model
...
+    ##weighting
+    prev = sum(
+      w[1] * pgamma(x, params_gamma$shape, params_gamma$rate), 
+      w[2] * pmgumbel(x, params_mgumbel$alpha, params_mgumbel$scale, lower.tail=T))
+    prev}
+
+Pregnancy Adjustment
+^^^^^^^^^^^^^^^^^^^^
+
+To sample hemoglobin values for pregnant/lactating women, use the same functions as above, but multiply the hemoglobin mean and standard deviation parameters used for those functions by the respective pregnancy adjustment factors listed below. Notably, the GBD 2019 assumes that the pregnancy adjustment factor applies to 40 weeks of gestation and 6 weeks post-gestation.
```
NathanielBlairStahn, line 197: Are the adjustment factors the ratios of pregnant women to all women, or pregnant women to non-pregnant women? This sentence makes it sound like the former, but when you first describe the ratios in the GBD section, it sounded like the latter.

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
```diff
@@ -6,4 +6,237 @@ Hemoglobin Model
...
+    - 1.032920188
+    - No uncertainty is used in the GBD 2019 code
+
+Validation Criteria
++++++++++++++++++++
+
+Hemoglobin samples should satisfy the following criteria:
+
+- all_samples > 0
+- all_samples < XMAX = 220
```
NathanielBlairStahn, line 219: Technically, I don't think these inequalities follow from the definition of the hemoglobin distribution. The [support](https://en.wikipedia.org/wiki/Support_(mathematics)) (i.e. the range of x values that occur with positive probability) of the gamma distribution is [0,infinity), and the support of the (mirror) gumbel distribution is (-infinity, infinity), so theoretically *any* x value is possible, similar to a normal distribution. Presumably, however, the probability of seeing a value outside the range [0,220] is very small...

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
```diff
@@ -6,4 +6,237 @@ Hemoglobin Model
...
+- mean(all_samples) ~= meid_10487
+- sd(all_samples) ~= meid_10488
+
+When the pregnancy adjustment is applied:
+
+- mean(general_population_samples) > mean(pregnant_population_samples)
+- standard_deviation(general_population_samples) < standard_deviation(pregnant_population_samples)
```
NathanielBlairStahn, line 226: Can't we be more specific and specify that the ratios should be approximately equal to (or can be calculated from) the pregnancy adjustments (see my question above about which populations the ratios actually apply to)?

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
NathanielBlairStahn: Cool. It's probably not necessary to give all the detail I wrote above, as most of that probably belongs on a separate page about ensemble distributions in GBD.  But I think we should at least write down the formula for what the hemoglobin distribution actually is, e.g.

F(x | mu,sigma) = 0.4 F_1(x | mu,sigma) + 0.6 F_2(x | mu,sigma),

where
F_1(x | mu,sigma) = gamma distribution parameterized by mean mu and standard deviation sigma,
F_2(x | mu,sigma) = mirror Gumbel distribution parameterized by mean mu and standard deviation sigma,
and mu and sigma depend on the location/year/age/sex (and pregnancy status).

(By "distribution" I mean "cumulative distribution function," but the same formulas also hold for the probability density functions.)

And ideally we would write down the mathematical formulas for the gamma and mirror Gumbel distributions, but probably don't worry about that yet since we still aren't sure what the correct formula for the mirror Gumbel distribution is.

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
NathanielBlairStahn: As you mentioned yesterday @alibow , having a negative hemoglobin level doesn't make any sense, so maybe reframe these as checks that should be built into the model by the software engineers: "If a sample is negative, resample!"

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
alibow: I added a description of the distribution and also added your walk through of ensemble distribution as a note that we could eventually move elsewhere but provides good context here for now.

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
alibow: I added a note clarifying this in the documentation. My understanding is that the adjustment factor is derived from nonpregnant versus pregnant women, but then is used in GBD to compare pregnant women with the general population.

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
alibow: Agreed! I worked this in.

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
alibow: I added a note clarifying this in the documentation. My understanding is that the adjustment factor is derived from nonpregnant versus pregnant women, but then is used in GBD to compare pregnant women with the general population.

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
```diff
@@ -6,4 +6,290 @@ Hemoglobin Model
...
+
+3. Generation of ensemble distributions for each location/year/age/sex group
+
+    Because anemia thresholds depend on pregnancy status, hemoglobin distributions were modeled separately for pregnant and non-pregnant females. The pregnancy model was identical to the non-pregnancy model except that the mean and variance were adjusted by the adjustment factor. The prevalence of anemia in pregnant women and non-pregnant women were then weighted by the pregnancy rate and combined to estimate population anemia prevalence. See the table below for the exact adjustment factors used.
+
+      The pregnancy rate was represented as :math:`(ASFR + SB) * 46/52`, where :math:`ASFR` is the location- and age-specific fertility rate and :math:`SB` is the location-specific stillbirth rate.
```
NathanielBlairStahn, line 85: You explain this below, but it might be worth also adding a quick comment here explaining that 46/52 represents (40 weeks of pregnancy + 6 weeks of lactation) / (52 weeks in a year).

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
```diff
@@ -6,4 +6,290 @@ Hemoglobin Model
...
+
+    And :math:`F_1(x)` is a mirror gumbel distribution parameterized by :math:`\mu` and :math:`\sigma`
+
+.. todo::
+
+	Include mathematical formulas for the gamma and mirror Gumbel distributions. For now, see the R code in the `Data Description Tables`_ section for details.
+
+.. note::
+
+	As summarized by Nathaneil, for context (details perhaps to be moved to a separate documentation page), the steps for finding an ensemble distribution are:
```
NathanielBlairStahn, line 58: Misspelled my name :) (ei -> ie)

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
```diff
@@ -6,4 +6,290 @@ Hemoglobin Model
...
+    - 
+  * - Mirror Gumbel distribution weight
+    - 0.6
+    - 
+
+Below is R code written to randomly sample hemoglobin concentration values from the hemoglobin distribution parameters and constants defined in the tables above. Additionally, the code block contains functions that will evaluate the proportion of the distribution below a given threshold. This code was adapted from the GBD stash code found `here <https://stash.ihme.washington.edu/projects/MNCH/repos/anemia/browse/model/envelope>`_, specifically the *DistList_mv2p.R* and *fit_ensemblemv2p_parallel.R* files. 
+
+.. note::
+
+  Attempts to move this code to python have not been successful in validating to GBD while using the R code has been successful. The recommended approach is to use the python package `rpy2` to run this R code within python.
```
NathanielBlairStahn, line 175: As Kjell noted in Slack, this is no longer accurate since he and James managed to figure out the Python version here:

https://github.com/ihmeuw/vivarium_gates_lsff/blob/master/tests/lsff_iron_exposure.ipynb

I think we should leave the R code, but update the comment and add a link to this notebook. And we could either add the Python code here as well or add a Todo to add it later.

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
```diff
@@ -6,4 +6,290 @@ Hemoglobin Model
...
+    params_gamma = gamma_mv2p(mn, vr)
+    params_mgumbel = mgumbel_mv2p(mn, vr)
+
+    ##weighting
+    sample_list = sum(
+      w[1] * rgamma(n, params_gamma$shape, params_gamma$rate), 
+      w[2] * rmgumbel(n, params_mgumbel$alpha, params_mgumbel$scale)
+    )
+    sample_list
+  }
```
NathanielBlairStahn, line 218: This is not actually the correct way to sample from an ensemble distribution. However, this *is* what Vivarium currently does, and it seems to give results that are fairly close to the right answer (except in the tails of the distribution).

I am in the process of writing up some documentation to describe how to correctly sample from ensemble distributions using propensities. Basically, the algorithm in this case would be:
1. Choose Gamma with probability 0.4 or mirror Gumbel with probability 0.6.
2. Sample from the chosen distribution.

In other words, each draw from the ensemble distribution samples from *either* Gamma or mirror Gumbel, but not both.

I'm not sure how to write the correct R code to do this. Until we have corrected sampling code, I think we should remove the incorrect version to avoid confusion.

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
```diff
@@ -6,4 +6,290 @@ Hemoglobin Model
...
+
+    ##weighting
+    prev = sum(
+      w[1] * pgamma(x, params_gamma$shape, params_gamma$rate), 
+      w[2] * pmgumbel(x, params_mgumbel$alpha, params_mgumbel$scale, lower.tail=T))
+    prev}
+
+.. note::
+
+	While not explicitly enforced by the code above, all hemoglobin values should be withn the bounds of 0 and 220 (XMAX). The probability of sampling a value outside of these bounds is small, but if it occurs, the value should be resampled until it falls within the bounds of 0 and 220.
```
NathanielBlairStahn, line 245: I agree that we should make sure all Hb values are >= 0 (do we want to resample, or just [clip](https://numpy.org/doc/stable/reference/generated/numpy.clip.html) the values to a small positive number?), but do we actually need to ensure that they're <= 220?  Is this assumption enforced anywhere in the GBD code?  My impression from a cursory look through the R code is that XMAX is really just an "approximate" maximum that is used to parameterize the distributions, but is not necessarily a hard maximum (though I could be wrong).

In particular, in [Kjjell's notebook](https://github.com/ihmeuw/vivarium_gates_lsff/blob/master/tests/lsff_iron_exposure.ipynb), sampling 10,000 draws from the ensemble gave values as high as 243. Do we actually need Vivarium to use rejection sampling to resample until all draws are <=220 (another option would be to just `np.clip` all the larger values to exactly 220), or is it fine to leave the large values?

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
NathanielBlairStahn: Also, did you intend this paragraph to be indented farther than the previous paragraph, or is that a mistake?

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
```diff
@@ -6,4 +6,290 @@ Hemoglobin Model
...
+
+2. Calculation of ensemble weights
+
+    A set of two-parameter distributions (gamma, mirror gamma, Weibull, mirror lognormal, and mirror gumbel) were fit to the sample’s haemoglobin mean and variance for each location/year/age/sex group
+
+    The weights used for the GBD 2019 hemoglobin distribution model were 40% gamma and 60% mirror gumbel, such that:
+
+    .. math::
+
+    	F(x|\mu,\sigma) = 0.4 * F_1(x|\mu,\sigma) + 0.6 * F_2(x|\mu,\sigma)
```
NathanielBlairStahn, line 38: You can use `\times` instead of `*` to get a standard multiplication symbol. Alternatively, it is also common to use the less conspicuous `\cdot` for multiplication.

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
alibow: I agree that we don't need to keep values under 220 and added that values could be clipped to 1 instead of being resampled.

alibow:
@aflaxman we've made some more updates and need another approval to merge. Could you take a look whenever you get a chance? Thanks!

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
NathanielBlairStahn: It looks like in the new version, all of the gamma draws will be placed before all of the mirror Gumbel draws - is that correct, or am I misunderstanding the R code?

If that's the case, I think there needs to be an extra step shuffling all the draws together so that we get an independent sequence of draws from the mixture.

NathanielBlairStahn: Also, if n=1, I think this updated method will always sample from mirror Gumbel, which is incorrect. So I think we need a different strategy for generating the draws.

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
```diff
@@ -6,4 +6,295 @@ Hemoglobin Model
...
+
+    ##weighting
+    prev = sum(
+      w[1] * pgamma(x, params_gamma$shape, params_gamma$rate), 
+      w[2] * pmgumbel(x, params_mgumbel$alpha, params_mgumbel$scale, lower.tail=T))
+    prev}
+
+.. note::
+
+	While not explicitly enforced by the code above, all hemoglobin values should be non-zero positive numbers. The probability of sampling a negative value is small, but if it occurs, the value should be resampled until it falls within the bounds of 0 and 220 or clipped to a value of 1.
```
NathanielBlairStahn, line 251: It's a bit confusing that this still mentions the upper bound 220 when now we're just talking about negative values. Maybe instead we could say "the value should be resampled until it is positive, or clipped to a value of 1."

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
NathanielBlairStahn: I should have said that **on average**, 40% of the draws will come from the Gamma distribution, and 60% of the draws will come from the mirror Gumbel distribution. The actual number from each won't exactly equal this ratio -- it will be random with a binomial distribution.

So one solution would be to sample the number of gamma draws as n1 = Binomial(n, 0.4), then take the number of mirror Gumbel draws to be n - n1, and then you'd have to randomly shuffle the draws to make sure they were in no particular order.

Or, a simple non-vectorized solution would be to just put the two steps I originally described above into a for loop of length n.

aflaxman (approved):

docs/source/model_design/risk.rst
```diff
@@ -13,9 +13,140 @@ Modeling Risk Exposure
 What is a risk factor?
 ----------------------
 
+What is a risk exposure?
+------------------------
```
NathanielBlairStahn, line 17: Are `risk.rst` and `risk_factors.rst` supposed to be in this PR, or was that an accident?

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
NathanielBlairStahn: Ok, looks good

docs/source/model_design/risk.rst
alibow: This was an accident I believe as a result of a merge conflict-like issue from a recent PR that restructured the risk factor documentation. I removed the substantive changes from this PR (the info that previously appeared in the risk.rst in this PR is all in the risk_factors.rst on the main branch)

NathanielBlairStahn (approved):

PR #306, opened 2020-09-29
update alternative scenario component

yongqx:


docs/source/concept_models/vivarium_swissre_cervicalcancer/cervical_cancer_concept_model.rst
```diff
@@ -537,18 +534,18 @@ Treatment for benign cervical cancer
...
      - 10%
      - 
      - The current HPV vaccination rates remain low in China, no data has
        identified.
    * - Target HPV vaccination coverage in 2030
-     - 90%
-     - WHO cervical cancer elimination strategy
+     - 40%
      - 
+     - by assumption
```
plinds, line 548: Flagging to look into literature again to guide these coverage numbers, I have it noted in Asana to keep track.

plinds (approved):
Noting that you'll need to update the concept model one more time to add a remission arrow from hrHPV state to susceptible state.

Andrei is scheduling a review meeting with Manoj, so we should be able to confirm with him in the near future whether we need the dotted line arrow from HPV vaccination to hrHPV.

aflaxman (approved):
New concept model diagram does not include treatment anymore, please add that back in

PR #308, opened 2020-09-30
LSFF: update stratification definitions

NathanielBlairStahn:
@KjellSwedin Here is my best guess at what the stratification groups currently mean or what we should make them mean. See in particular the notes for the vitamin A coverage groups and the child iron coverage groups.  The next step is for you to take a look through the code and see how closely the code matches the definitions here or not, and then we can update either the code or this table as appropriate.

@alibow It would be good to get your eyes on this too. In particular, what is your opinion on what would be the most useful way to stratify on vitamin A coverage and child iron coverage?


alibow:
A couple of things:

1. I remember that there was a miscommunication/understanding about the specific date for year stratification in one of these projects (I don't remember if it was LSFF or BEP), but I think it would be good to clarify that we mean 1/1/2020 - 12/31/2020 when we say "2020", etc. (If I'm remembering correctly it may have been implemented once as 7/1/2020 - 6/30/2021?).

2. The vitamin A coverage stratification looks good to me; my understanding is that we would expect to see no effect in the uncovered/covered groups and the full effect in the effectively covered group.

3. For iron coverage, if we would want to look into the scale up, we could perhaps stratify for the following groups: child uncovered by iron, child covered by iron for less than six months, child covered by iron for 6 months or more. Although in order for this to be directly useful, we would probably also want to stratify the 1-4 age group into 1-2 and 2-4 years... this way we could verify that the full effect was applied to the iron responsive population covered for 6 months or more in the 2-4 age group and then the effect should be lower in every other group. This is fairly involved, though, so I wonder if it could be checked using interactive simulation instead.

alibow (approved):

KjellSwedin (approved):
I still find the terminology "Simulant's household " versus simply simulant confusing. I vote for using "simulant" unless the notion of a "household" is actually different from a single simulant.

NathanielBlairStahn:
> I still find the terminology "Simulant's household " versus simply simulant confusing. I vote for using "simulant" unless the notion of a "household" is actually different from a single simulant.

@KjellSwedin I used the word "household" here because with the definition I wrote down, the simulant's household may be receiving fortified food, but the simulant is not eating it yet, namely if they're under six months old.  In other words, the the time when the *household* starts receiving fortified food depends only on the simulant's propensity for receiving fortification, but whether the simulant is actually *eating* the fortified food also depends on the simulant's age.

My current definitions are assuming a simple propensity-based stratification, without taking the simulant's age into account, hence the fortification group refers to the simulant's household and not necessarily to the simulant themself.

However, I need to think more about Ali's comments above to decide whether it would be worth it to use a more complicated stratification strategy that *would* take into account the simulant's age and/or whether they are experiencing the full effect of fortification.  If we change to a more complicated strategy, it may be more appropriate to refer to the simulant being fortified rather than their household being fortified.

NathanielBlairStahn:
> * The vitamin A coverage stratification looks good to me; my understanding is that we would expect to see no effect in the uncovered/covered groups and the full effect in the effectively covered group.
> *

@alibow This is why I want to get clarification from @KjellSwedin about what the current stratification actually means. Based on the definitions I wrote down (which were based on James' pseudocode), I think it is possible that an "effectively covered" simulant has passed the time lag for seeing the effect of fortification, but is still under 6 months old so is actually *not* getting the effect of fortification yet. Hence my question to Kjell about whether the coverage groups take the simulant's age into account or just the time lag (or maybe they just take age into account and *not* the time lag, or maybe it's both -- I don't know).

KjellSwedin:
@NathanielBlairStahn, @alibow  - here is the vitamin A answer. I've updated the jiri ticket with the same information.
Vitamin A:
> Called "covered" in current output, but that name is confusing. Is this definition accurate, or do simulants < 6mo old also fall in this category? 
Yes, simulants under 6 months will fall into this category

> Is this definition accurate, or does "effectively covered" also mean the simulant has to be > 6mo old? 
Yes, a simulant must be over 6 months of age before being considered effectively_covered.

This can be confusing when reading the code or even examining in an interactive simulation because the final binning takes place in the observer code. For example, examining an effectively_covered pipeline will show simulants that are both effectively_covered and under 6 months of age. The observer code deals with this. Look here for more info:

def vitamin_a_covered()
https://github.com/ihmeuw/vivarium_conic_lsff/blob/master/src/vivarium_conic_lsff/components/observers.py#L106

KjellSwedin:
@NathanielBlairStahn - Perhaps "household" makes sense in the description. I just want to make clear that there is no concept of a household in the code. For instance, I *think* the LTBI project modeled the concept of a household and so there was some sort of grouping and if someone in the grouping had latent TB then exposure went up for the group. There is no grouping in LSFF.

KjellSwedin:
> Should we define 3 coverage groups instead, similar to vitamin A, in order to track some of these effects?

This would have a fairly large code and verification footprint. I'd suggest against this for the bugfix release. If we want this in subsequent releases, we should implement early to get ample testing time.

NathanielBlairStahn:
> @NathanielBlairStahn, @alibow - here is the vitamin A answer. I've updated the jira ticket with the same information.
> Vitamin A:
> 
> > Called "covered" in current output, but that name is confusing. Is this definition accurate, or do simulants < 6mo old also fall in this category?
> > Yes, simulants under 6 months will fall into this category
> 
> > Is this definition accurate, or does "effectively covered" also mean the simulant has to be > 6mo old?
> > Yes, a simulant must be over 6 months of age before being considered effectively_covered.
> 
> This can be confusing when reading the code or even examining in an interactive simulation because the final binning takes place in the observer code. For example, examining an effectively_covered pipeline will show simulants that are both effectively_covered and under 6 months of age. The observer code deals with this. Look here for more info:
> 
> def vitamin_a_covered()
> https://github.com/ihmeuw/vivarium_conic_lsff/blob/master/src/vivarium_conic_lsff/components/observers.py#L106

Excellent, thanks for this clarification @KjellSwedin . I will update the table with the correct definitions for vitamin A coverage.

The observer code is very helpful for understanding what output we are getting.  Can you point me to similar code for what is happening with the mother and child iron coverage groups?

NathanielBlairStahn:
> > Should we define 3 coverage groups instead, similar to vitamin A, in order to track some of these effects?
> 
> This would have a fairly large code and verification footprint. I'd suggest against this for the bugfix release. If we want this in subsequent releases, we should implement early to get ample testing time.

Ok, I'll keep this in mind.  @KjellSwedin can you clarify exactly how the current iron stratification works (for both mother/birthweight and child/hemoglobin), similar to your explanation for vitamin A, so I can see whether my definitions in the table match what is actually happening?

NathanielBlairStahn:
> @NathanielBlairStahn - Perhaps "household" makes sense in the description. I just want to make clear that there is no concept of a household in the code. For instance, I _think_ the LTBI project modeled the concept of a household and so there was some sort of grouping and if someone in the grouping had latent TB then exposure went up for the group. There is no grouping in LSFF.

Thanks for the clarification @KjellSwedin . Yes, we do **not** need to explicitly model different simulants being in the same household like they did for LTBI.  In the context of LSFF, the concept of "household" is useful mainly for clarifying our thinking about how the intervention works: We are not directly delivering fortified food to the children; rather, we fortify food that their parents may buy, so the simulants are getting the intervention indirectly through their household (either because their mother eats the fortified food or because the children eat it themselves after their parents bring it home). But we do not need to explicitly model a "household" separately from the simulant - I am just using the concept to clarify what the individual simulants' attributes actually mean (e.g. their propensity for fortification represents the probability that their household receives fortified food).

NathanielBlairStahn:
> 1. I remember that there was a miscommunication/understanding about the specific date for year stratification in one of these projects (I don't remember if it was LSFF or BEP), but I think it would be good to clarify that we mean 1/1/2020 - 12/31/2020 when we say "2020", etc. (If I'm remembering correctly it may have been implemented once as 7/1/2020 - 6/30/2021?).

@alibow @KjellSwedin I updated the "Time" section in the concept model to specify the start and end dates instead of just the years. Does that seem sufficient, or do you think this also needs to be specified explicitly in the output tables?

alibow:
> > 1. I remember that there was a miscommunication/understanding about the specific date for year stratification in one of these projects (I don't remember if it was LSFF or BEP), but I think it would be good to clarify that we mean 1/1/2020 - 12/31/2020 when we say "2020", etc. (If I'm remembering correctly it may have been implemented once as 7/1/2020 - 6/30/2021?).
> 
> @alibow @KjellSwedin I updated the "Time" section in the concept model to specify the start and end dates instead of just the years. Does that seem sufficient, or do you think this also needs to be specified explicitly in the output tables?

That seems good to me as long as the SWEs are in understanding!

NathanielBlairStahn:
Relevant code sections from Kjell posted in Slack on 10/6/20:

> here are code pointers for items you referenced in the PR:
> // hemoglobin observer
> https://github.com/ihmeuw/vivarium_conic_lsff/blob/master/src/vivarium_conic_lsff/components/observers.py#L585-L628
> // maternal iron birthweight observer
> https://github.com/ihmeuw/vivarium_conic_lsff/blob/master/src/vivarium_conic_lsff/components/observers.py#L471-L531
> // maternal iron and child hemoglobin effect code
> https://github.com/ihmeuw/vivarium_conic_lsff/blob/master/src/vivarium_conic_lsff/components/fortification/folic_acid.py#L202-L361

PR #312, opened 2020-10-07
Lsff update iron intervention summary

NathanielBlairStahn:
I updated the "Summary of Iron Intervention Algorithm" section as follows:

- Incorporated corrected/clearer pesudocode for the Hb shift from the section "Determining Whether A Simulant is Affected - Iron > Hemoglobin Level," and added comments to both pseudocode blocks to better describe what is happening.
- Corrected a bug in the calculation of coverage duration in both pseudocode blocks:
```time_since_fortified_i(t) = min(t - time_covered_i, t - age_i(t) + 0.5)```
 became
```time_since_fortified_i(t) = min(t - time_covered_i, age_i(t) - 0.5)```
because `age_i(t) - 0.5` is the time since the simulant turned 6 months old (which is what we want), vs. `t - age_i(t) + 0.5`, which is the time at which the simulant turned 6 months old (which is **not** what we want).
- Removed the code dealing with stratification, as we are not using the method I previously described here.

I also added To-do's to update the other sections of the iron intervention that still conflict with the description in the "Summary," in order to clarify the remaining discrepancies and specify which version is correct.

@KjellSwedin @alibow  It would be good for both of you to look through this to verify that what is in the "Summary" section matches what we actually implemented (@KjellSwedin ) and also what we actually wanted to implement (@alibow ), and to try to catch any errors I may have still missed

### DIFF
--- a/docs/source/concept_models/vivarium_conic_lsff/concept_model.rst
+++ b/docs/source/concept_models/vivarium_conic_lsff/concept_model.rst
@@ -1465,6 +1465,8 @@ clarify the intent.
 
 .. code-block::
 
+  import numpy as np
+
   ## Definitions:
   t                         := current time in years (t=0 is start of simulation)
   t_end                     := time when simulation ends, in years since start (e.g. 6 years)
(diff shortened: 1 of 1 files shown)


aflaxman (approved):

alibow (approved):

alibow:
This looks good to me! I think that this layout/presentation is easier to follow than before.

PR #314, opened 2020-10-09
Update BEP child growth failure output shell data

alibow:
I think this is the appropriate update for the output table shell to achieve the child growth failure V&V that we need.


aflaxman (approved):
This seems fine, but it is hard to tell what changed (and what the colors mean).  Let's use csvs in the future or some format that GitHub can show diffs of more easily.

PR #321, opened 2020-10-20
update cervical cancer screening and detection

yongqx:


yaqi7 (approved):

aflaxman (approved):
I am confused --- what is the difference between the screening frequency and follow-up frequency columns in your table?

yongqx:
> I am confused --- what is the difference between the screening frequency and follow-up frequency columns in your table?

I split the test into two columns method and frequency in case our software engineers missed the frequency information. Say a woman aged 45 years who is recommended `co-test in 5 years` and tested positive for cytology from that co-test, then she will follow up with a `cytology alone test in 1 year`. The original screening frequency versus follow-up test frequency are 5 years and one year.

PR #331, opened 2020-11-02
Lung cancer preclinical state update

alibow:


docs/source/gbd2017_models/causes/neoplasms/lung_cancer/cancer_model.rst
```diff
@@ -190,19 +184,35 @@ State and Transition Data Tables
      - disabilty weights
      - 0
      -
-   * - I
+   * - PC
+     - prevalence
+     - 0
+     - Assumption; may revisit
```
aflaxman, line 190: I think this will make a big difference in the detection curves for the first five-or-so years of the sim, so let's think more about something more defensible than starting with p_PC = 0

aflaxman (approved):
I think a better label for the arrows in the svg would be i_pc instead of pc, etc.

docs/source/gbd2017_models/causes/neoplasms/lung_cancer/cancer_model.rst
alibow: Thanks, @aflaxman . I just pushed a commit with a rough concept to address this. Yongquan, Nicole, and I are also planning on talking through this during coworking today -- thought I would let you know in case you are available and interested in joining.

Ninicorn: I think the source state for R should be C not I right?

Ninicorn: after remission, do they go back to the S state or a remitted (R) state?

yongqx: I think they will stay in Recovered state rather than go back to Susceptible. And let's discuss the approach of using GBD incidence to inform i_PC.

yongqx (approved):

PR #333, opened 2020-11-04
Lung cancer assumptions, limitations

alibow:


docs/source/gbd2017_models/causes/neoplasms/lung_cancer/cancer_model.rst
```diff
@@ -126,12 +126,24 @@ This Vivarium modeling strategy is intended to simulate TBL cancer incidence/mor
...
 
-This model will assume the existence of a "recovered" cause model state in an attempt to be consistent with the GBD assumption that no morbidity due to TBL cancer occurs more than ten years past incidence of the *clinical* phase of TBL cancer. The assumption also asserts that there is no recurrance of TBL cancer.
+1. This model will assume the existence of a "recovered" cause model state in an attempt to be consistent with the GBD assumption that no morbidity due to TBL cancer occurs more than ten years past incidence of the *clinical* phase of TBL cancer. The assumption also asserts that there is no recurrance of TBL cancer.
 
 .. todo::
 
 	Confirm that the 10 year assumption should apply to the clinical phase of TBL cancer with GBD modeler.
 
+2. This model assumes that the GBD incidence rate corresponds to the incidence of asymptomatic preclinical/LDCT screen-detectable TBL cancer rather than *detected* lung cancer, which is a mix of preclinical and clinical detections. This assumption has a few notable downstream limitations, including:
+
```
aflaxman, line 136: For each of these limitations, let's speculate on the scale of the error it introduces.  For example, if the duration of the preclinical phase is 2 years, then we might expect 2 out of 100 simulants to die of other causes after preclinical incidence and before clinical incidence, leading to a 2% underestimate of TBL incidence.

For the second limitation, can we address this by shifting the incidence age-pattern by X years, where X is the mean sojourn time?

How is the third limitation different than the second limitation?

aflaxman (approved):

yongqx (approved):

docs/source/gbd2017_models/causes/neoplasms/lung_cancer/cancer_model.rst
alibow: Agreed, I will work on the quantification and include a to-do note for now.

Third assumption is similar to the second assumption but I wanted to specifically call out how we are estimating *prevalence* of the preclinical state, by a ratio approach rather than direct estimation from GBD.

Shifting the age pattern by X years is coming in another PR soon.

Thanks!

PR #334, opened 2020-11-06
Lung cancer initial concept model overview

alibow:


docs/source/concept_models/vivarium_swissre_lungcancer/concept_model.rst
```diff
@@ -49,32 +59,24 @@ Vivarium CSU Lung Cancer Simulation
 
 .. _1.1:
 
+The re-insurance company SwissRe is interested in the expected impact of increases in lung cancer screening coverage on the rates of detected lung cancer over the next 20 years among the insured population in select areas in China.
```
aflaxman, line 62: For some reason I like avoiding putting the names of CSU clients in our docs. can you remove?

aflaxman (approved):
Software engineers wanted a concept model diagram with more familiar columns, but this makes sense to me.  maybe if they are still asking about it we can change the arrows a little.

yongqx (approved):
Internally we may want to distinguish detections through two pathways, screening and symptomatic presentation, i think the output shell table we deliver to our clients doesn't need that stratification.

docs/source/concept_models/vivarium_swissre_lungcancer/concept_model.rst
alibow: Done!

alibow:
> Internally we may want to distinguish detections through two pathways, screening and symptomatic presentation, i think the output shell table we deliver to our clients doesn't need that stratification.

Thanks, I added this clarification in a notes column in the output table.

PR #341, opened 2020-11-10
update cervical cancer equations

yongqx:


docs/source/gbd2017_models/causes/neoplasms/cervical_cancer/cancer_model.rst
```diff
@@ -373,8 +373,8 @@ prev_ = prevalence; i_ = incidence; r_ = remission; RR_ = relative risk; PAF = p
      - derived from Abie's dismod
      - add filepath
    * - RR_hrHPV
-     - extracted from Chen et al.
-     - relative risk of developing BCC and/or ICC for hrHPV infected women versus without HPV infection = 16.2 (95%CI 9.6 to 27.3)
+     - extracted from Naucler et al.
+     - relative risk of HPV 16/18 causing CIN2+ = 27.4 (95%CI 19.7 to 38.0)
```
aflaxman, line 377: Why is Naucler preferred to Chen?

aflaxman (approved):
Let's get the engineers to re-run with these updated rates as soon as possible to see if it validates

docs/source/gbd2017_models/causes/neoplasms/cervical_cancer/cancer_model.rst
yongqx: Number extracted from Chen was associated with any type HPV infection but we prefer subtypes 16 and 18.

PR #247, opened 2020-06-02
draft concept model template

Ninicorn:

PR #345, opened 2020-11-14
Adding 'simulating Multiple Myeloma registries' concept model outline.

SantoniS:


aflaxman (approved):

alibow:
I think this looks good, but FYI there is another template that we've started using for the BEP and SwissRe cancer models that is pretty similar, although I like how the sections are numbered so that it is easier to navigate around what usually ends up being a really long document -- just something to consider :)

alibow (approved):

plinds:
Going off of Ali's comment - the template is still in a PR if you want to take a look (https://github.com/ihmeuw/vivarium_research/pull/247)! Or look at the template in action in the BEP and cancer concept models.

SantoniS:
I'd be happy to update the outline structure based on PR #247!

PR #360, opened 2020-12-02
CMD again since last once didn't update correctly with my new arrow descriptions

Ninicorn:


plinds (approved):
I'm not noticing any difference between these 2 figs...did you upload your updated version? Or maybe I'm just not catching the difference

PR #365, opened 2020-12-08
Smoking affected causes

alibow:


aflaxman (approved):
I don't understand how this model will work without states but *with* EMR...  will everyone have the SAC EMR?

alibow:
> I don't understand how this model will work without states but _with_ EMR... will everyone have the SAC EMR?

@aflaxman, there isn't actually an EMR for smoking affected causes. I included the EMR bit as an example to show how this would integrate when put into a model with other causes that are explicitly modeled (lung cancer), so that's what the EMR is referring to, not SACs.

alibow:
Here are the rendered versions for convenience :) 

image

image

alibow:
image

docs/source/gbd2017_models/causes/smoking_affected_causes/index.rst
```diff
@@ -0,0 +1,459 @@
...
+	List assumptions and limitations
+
+Cause Model Diagram
++++++++++++++++++++
+
+There is no cause model diagram for this cause model because it is a mortality-only model. Mortality using should be modeled as follows:
+
+.. math::
+
+	mr_i = ACMR - CSMR_\text{SACs} + CSMR_\text{SACs} * (1 - PAF) * RR_i - \sum_{c}^{n} CSMR_c + \sum_{c}^{n} EMR(i)_c
```
Ninicorn, line 333: I'm wondering if we build this equation up so its clear how this is derived (kind of like how we did for the LBWSG documentation)? My understanding is as thus (correct me if I'm wrong!):

unmod: unmodelled causes
unaff: unaffected causes
aff: affected (here it is the SACs)
mod: modelled (here it is lung cancer)

mr_i = ACMR 
        = sum_CSMR_unmod_unaff_causes + sum_CSMR_smoking_aff_causes + sum_CSMR_mod_causes (eg. lung cancer)

sum_CSMR_unmod_unaff_causes =  ACMR - sum_CSMR_smoking_aff_causes - sum_CSMR_mod_causes (eg. lung cancer)

sum_CSMR_smoking_aff_causes = CSMR*(1-PAF)*RR (does this need a summation or is it pooled somehow? this is lung cancer deleted right?)

sum_CSMR_mod_causes (eg. lung cancer) = sum_EMR(state)

(should there be a summation sign for CSMR_sacs ?)

docs/source/gbd2017_models/causes/smoking_affected_causes/index.rst
```diff
@@ -0,0 +1,459 @@
...
+    - 
+  * - RF
+    - Risk factor
+    - 
+
+Disease Overview
+----------------
+
+Mortality among smokers is much higher than mortality among non-smokers. Therefore, it can be important to consider increased mortality attributable to smoking when smoking is included in a model, even if specific causes affected by smoking are not necessary to explicitly include in the model. See the causal framework section in the :ref:`Vivarium CSU Lung Cancer Screening Concept Model Documentation <2017_concept_model_vivarium_swissre_lungcancer>` for an example of when this is important.
+
```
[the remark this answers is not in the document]
Ninicorn: I wonder if we can mention the relative risk here too because that's what is really conferring the increased mortality risk associated with the smoking exposure- we apply the (1-PAF)*RR to the CSMRsacs.

docs/source/gbd2017_models/causes/smoking_affected_causes/index.rst
```diff
@@ -0,0 +1,459 @@
+.. _2017_smoking_affected_causes:
+
+===============================================
+Smoking Affected Causes (Excluding Lung Cancer)
+===============================================
+
```
Ninicorn, line 6: since this is a mortality model, we can add mortality to the title of this documentation? "Modelling increased mortality associated smoking affected causes" or something like that...  ¯ \ _(ツ)_/¯

Ninicorn (approved):

PR #367, opened 2020-12-09
Lung cancer cause model updates discussed in meeting

alibow:


docs/source/gbd2017_models/causes/neoplasms/lung_cancer/cancer_model.rst
```diff
@@ -259,8 +259,8 @@ State and Transition Data Tables
      - 
    * - C
      - prevalence
-     - :math:`screening_\text{baseline} * prevalence_\text{c426} * \frac{1}{1+ODF} + (1 - screening_\text{baseline}) * prevalence_\text{c426}`
-     - 
+     - 0
+     - Assumed that there are no prevalent clinical (detected) lung cancers in the insured population; see the definition of :math:`prevalence_\text{C, general population}` for the prevalence of clinical lung cancers in the general population (Data Sources Table)
```
yongqx, line 263: Is this assumption based on all prevalent lung cancer cases are detected cases according to GBD definition?

yongqx (approved):

docs/source/gbd2017_models/causes/neoplasms/lung_cancer/cancer_model.rst
alibow: No, this is because we assume that no one in SwissRe's population already has detected cancer (they would only have asymptomatic preclinical or indolent lung cancer). Is that consistent with what we've done for the other cancer sites?

aflaxman (approved):

docs/source/gbd2017_models/causes/neoplasms/lung_cancer/cancer_model.rst
yongqx: In breast cancer and cervical cancer simulation, we modeled the cancer with benign and invasive states. There was no breakdown to preclinical and clinical phase for invasive state. Hence, we used forecasted prevalence for invasive state. There are undiagnosed prevalent cases can be found through routine screening and those cases are being caught rapidly in first couple of years.

alibow: Ok, I think this is the equivalent application of that assumption to the lung cancer model, but let me know if you disagree.

PR #368, opened 2020-12-10
Concept model notes for colorectal cancer

yongqx:


aflaxman (approved):
It is a lot easier to review a model if you do smaller PRs for each piece as you develop them.  When I get the whole thing at once, I often don't have time to give it all the careful read I would like to.

alibow (approved):

PR #371, opened 2020-12-11
update cervical cancer model

yongqx:


yongqx:
Two changes in this PR:
1. We removed regression of BCC from cause model document.
2. We add vaccination protection effect on transition from `BCC_S_hrHPV to BCC_C_hrHPV`, and assume same relative risk distribution as impact of vaccination on `Susceptible to BCC_S_hrHPV`. It will be influential on people who get screened.

aflaxman (approved):

plinds (approved):

PR #373, opened 2020-12-14
Stomach cancer cause model

Ninicorn:


docs/source/gbd2017_models/causes/neoplasms/stomach_cancer/cancer_model.rst
```diff
@@ -242,8 +244,8 @@ State and Transition Data Tables
      -
    * - PC
      - prevalence
-     - prev_PC = i_c414 (at age 'current age + MST') * MST
-     - Note: assumes all cancers in prevalence_c414 are in clinical phase
+     - prev_PC = (1-baseline screening %) x i_pc x MST
```
alibow, line 247: The "%" in this equation is distracting to me because I think that baseline_screening will be a proportion rather than a percentage.

alibow: Also, baseline screening shows up here but nowhere else... can we get away with this because there is no indolent state? Do we assume that the baseline screening coverage has no effect on incidence through treatment and note that as a limitation then?

alibow: Also, should we assume that the prevalence of PC for initialization is greater than this value given that we have no representation of the clinical state, so we would expect greater relative representation of the PC state relative to the general population? I would suggest dividing by (1-prevalence_c414) here.

alibow (approved):

docs/source/gbd2017_models/causes/neoplasms/stomach_cancer/cancer_model.rst
```diff
@@ -45,13 +45,11 @@ Stomach Cancer
...
-| AST   | Average Survival Time      |
-+-------+----------------------------+
 | PC    | Pre-Clinical Cancer        |
 +-------+----------------------------+
-| S     | Susceptible                |
+| S     | Susceptible in sim pop     |
 +-------+----------------------------+
-| C     | With Cancer condition      |
+| C     | Clinical cancer in sim pop |
 +-------+----------------------------+
```
yongqx, line 54: Can you add definition of `sim pop`, such as general population - (clinical stomach cancer cases +  pre-clinical stomach cancer cases detected through baseline screening).

yongqx (approved):

docs/source/gbd2017_models/causes/neoplasms/stomach_cancer/cancer_model.rst
aflaxman: or just strike out the phrase "in sim pop" --- I think it is clear enough without it

aflaxman (approved):

docs/source/gbd2017_models/causes/neoplasms/stomach_cancer/cancer_model.rst
Ninicorn: I've revised the equations to reflect no endoscopy screening in the general population which means no PC states in the general population

alibow:
I'm not seeing the assumption that there are no detected cancers upon initialization anymore. Should we put this back in?

PR #374, opened 2020-12-14
Smoking exposure document

alibow:


docs/source/gbd2017_models/risks/smoking/index.rst
```diff
@@ -4,58 +4,170 @@
...
+
+The below diagram displays the exposure possibilities within the Vivarium modeling strategy for the smoking risk exposure: simulants can transition between never, curent, and former smoking status groups (although the transition from current to never smokers is illogical). Additionally, current and former smokers are assigned pack-year exposures and former smokers are assigned years since quitting exposures.
+
+.. image:: smoking_exposure_diagram.svg
+
+Smoking Status
+++++++++++++++
+
+Each simulant should get a smoking status propensity score (smoking_status_propensity_i) that is a random value assigned based on a uniform distribution from 0 to 1. This propensity score for an individual should remain constant for the entire duration of the simulation. 
+
```
yongqx, line 71: The propensity score can change as simulant's smoking status changes, i assume constant number is used for initialization.

docs/source/gbd2017_models/risks/smoking/index.rst
```diff
@@ -4,58 +4,170 @@
...
+Pack-year exposure data are stored here: `/ihme/csu/swiss_re/forecast/py_forecast_draws.csv` and are age-, sex-, location-, and year-specific. The units of this file are number of individuals in each exposure bin per person-year of the *general population.* 
+
+Pack-years Among Former Smokers
++++++++++++++++++++++++++++++++
+
+Pack-years among former smokers should be assigned in a similar way to pack-years among current smokers, although the exposure should be sampled from the pack-year distribution among current smokers *the last year that the former smoker was a current smoker*. In other words, the year equal to the current year minus the simulant's years since quitting (see section below).
+
+The pack-year exposure data for former smokers should not change for the entire duration that a simulant is classified as a former smoker. 
+
+If a former smoker simulant becomes a current smoker, that simulant's pack-year exposure should be updated to reflect the pack-year exposure distribution in the current year (the year in which the former smoker becomes a current smoker).
```
yongqx, line 149: So that means we allow for simulants quit smoking with low pack-years , but a high pack-years when they become current smoker.

yongqx (approved):

docs/source/gbd2017_models/risks/smoking/index.rst
alibow: This propensity is for smoking status only, and not for pack-years or years since quitting; propensities for each of these variables should all be separate and independent. I will add some clarification here, thanks.

docs/source/gbd2017_models/risks/smoking/index.rst
alibow: Yes, we are not enforcing consistent exposure histories for individual simulants, but rather enforcing consistency with population level exposure forecasts.

docs/source/gbd2017_models/risks/smoking/index.rst
alibow: Was this what you meant?

aflaxman (approved):
You better try out a nano-sim of years since quitting to see how weird it is

alibow:
> You better try out a nano-sim of years since quitting to see how weird it is

@aflaxman You're right, we're fairly off from the forecasts after just four years (this plot looks at a single age group that has simulants aging in from the age group before it and also simulants aging out). Maybe we should take a similar approach of just assigning these values directly from the forecasts similar to the other exposures?

image

PR #381, opened 2020-12-17
Stomach cancer concept model

Ninicorn:


aflaxman (commented):
I think this PR got some extra material in it from your BEP project.  Was the goal to switch from a csv to an xlsx for one table?  I prefer csv, since it is easier to see in the PR.

PR #385, opened 2020-12-29
updated MST

Ninicorn:


plinds (approved):

docs/source/gbd2017_models/causes/neoplasms/stomach_cancer/cancer_model.rst
```diff
@@ -381,8 +380,8 @@ This parameter is be sampled *at the draw level* from the distribution detailed
   from scipy.stats import norm
 
   # mean and 0.975-quantile of normal distribution for mean difference (MD)
-  mean = 4.5
```
alibow, line 384: Because the confidence interval for the MST value here isn't symmetrical, the code below will create a slightly shifted distribution, I think. We should document whether we shift it up or down or come up with a different distribution that will accommodate the asymmetrical confidence interval

aflaxman (approved):

yongqx (approved):

docs/source/gbd2017_models/causes/neoplasms/stomach_cancer/cancer_model.rst
Ninicorn: Ah I see...I'll keep the upper limit as is, so the lower limit will be slightly lower at 1.78 meaning ppl can start transitioning 1.6 months earlier that the published lower limit. I think this is ok as this number is also an approximation that they derived using a formulae applied to retrospective cohort data.

Ninicorn: This would mean we might very slightly under-estimate preclinical cancers for ~5% of the population.  However, the shortest screening interval (branch 1 of 4) is 1 year. Hence we are only under-estimating for about 0.16% of the screened population (0.05*(1.6/12)/4).

PR #388, opened 2020-12-31
added validation, references for model 1 and 2

Ninicorn:


docs/source/concept_models/vivarium_swissre_stomachcancer/concept_model.rst
```diff
@@ -155,7 +157,10 @@ Scale-up of stomach cancer screening using ABC method with endoscopic follow-up
 
 :underline:`Alternative scenario`
 
-In the alternative scenario, there will be a scale up of ABC screening starting from 5% to 30% as indicated in the coverage figure below. We do not assume that the 5% who recieve H. pylori screening in the baseline is the same groups those who recieve screening in the alternative scenario. 
+In the alternative scenario, there will be a scale up of ABC screening starting from 5% to 30% as indicated in the coverage figure (orange line) below.
```
aflaxman, line 160: Note that this is different that other models from this line of research, because in the alternative scenario we have an additional screening modality, not just increased coverage

docs/source/concept_models/vivarium_swissre_stomachcancer/concept_model.rst
```diff
@@ -155,7 +157,10 @@ Scale-up of stomach cancer screening using ABC method with endoscopic follow-up
 
 :underline:`Alternative scenario`
 
-In the alternative scenario, there will be a scale up of ABC screening starting from 5% to 30% as indicated in the coverage figure below. We do not assume that the 5% who recieve H. pylori screening in the baseline is the same groups those who recieve screening in the alternative scenario. 
+In the alternative scenario, there will be a scale up of ABC screening starting from 5% to 30% as indicated in the coverage figure (orange line) below. 
+
+.. note::
+  We do not assume that the 5% who recieve H. pylori screening in the baseline is the same groups those who will recieve screening in the alternative scenario. (Hence H. pylori prevalence does not need to be split out between baseline screened or not screened).
```
aflaxman, line 164: Oh, I missed this.  Does it mean we need to do some sort of "treatment-deleted incidence rate" for the alternative scenario?

docs/source/concept_models/vivarium_swissre_stomachcancer/concept_model.rst
```diff
@@ -236,6 +241,37 @@ file paths for 2019 forecast data:
...
+Full references
+
+  - Bae J-M, Shin SY, Kim EH. Mean Sojourn Time of Preclinical Gastric Cancer in Korean Men: A Retrospective Observational Study. J Prev Med Pub Health 2014; 47: 201–5.
+
+Validation and verification
+
+  - i_pc should validate to ~ i_c414/(1-prev_c414) shifted by the 2.37 years forward. i_c414 and prev_c414 comes from forecast.
+  - i_c should validate to ~ 1/MST
+  - prevalence of pc ~ i_pc/(1-prev_c414) X 2.37 shifted by the 2.37 years forward. 
+  - prevalence of C should be 0 at the start and eventually catch up to forecast prevalence in later years
```
aflaxman, line 266: "Later years" meaning sooner than MST + 10

aflaxman (approved):

docs/source/concept_models/vivarium_swissre_stomachcancer/concept_model.rst
yongqx: For prevalence of PC state, do we need shift it by mean sojourn time forward if you apply `i_pc` directly from Sim?

yongqx (approved):

docs/source/concept_models/vivarium_swissre_stomachcancer/concept_model.rst
Ninicorn: I don't think so. There is no evidence available to say that those who had H. pylori in the past (baseline) is more likely to partake in stomach cancer screening when available or vice versa (in the alternative scenario). For simplicity, I think we assume that the population who partake in screening is the same as those who do not in relation to their past H. pylori screening status.

docs/source/concept_models/vivarium_swissre_stomachcancer/concept_model.rst
Ninicorn: @yongqx yes , you are right, the prevalence of PC doesn't need to be shifted!

PR #396, opened 2021-01-20
Smoking pack year and ysq exposures update to gbd 2019

alibow:
PY and YSQ exposure updates


aflaxman (approved):
I'm not clear on what is constant at the population level and what is constant at the individual level.  As long as Ali and Rajan agree, I am fine with it, though.

alibow:
> I'm not clear on what is constant at the population level and what is constant at the individual level. As long as Ali and Rajan agree, I am fine with it, though.

Thanks, Abie. I just made some hopefully clarifying edits.

PR #400, opened 2021-01-25
Update lung cancer assumptions and limitations

alibow:


aflaxman (approved):
Listing the assumptions is important, but there is more to say about what the implications of these assumptions might be, which we should also record before we forget about the details.

yongqx (approved):

PR #405, opened 2021-02-24
Adding GBD 2019 MM cause model and state definitions.

SantoniS:


aflaxman (approved):
All looks good, although perhaps the survival regressions noted in the cause model diagram will end up being included in the treatment model document and not in this cause model document.

yongqx (commented):
1. If screening is not necessary for this project, can we get rid of PC state since GBD didn't model such state.
2. Let's confirm with Manoj whether or not patient can go directly to recovered state from clinical MM state? If duration of recovery is long enough, simulants must go through RRMM then step into a recovered state.

yongqx (approved):

SantoniS:
> * If screening is not necessary for this project, can we get rid of PC state since GBD didn't model such state.
> * Let's confirm with Manoj whether or not patient can go directly to recovered state from clinical MM state. If duration of recovery is long enough, simulants must go through RRMM before they step into a recovered state.

I will add these two questions in the cause model as to-dos. Thank you!

PR #417, opened 2021-03-15
Updated state and transition data tables and model diagram

ndecleene:
I copied the model diagram and state and transition data tables from the 2017 documentation and updated them to match GBD 2019 and incorporated Greg's edits. 

Outstanding questions:
1) Do we want to have two parallel models for MI and angina, as pictured here? Or do we want to incorporate them into one model: one susceptible bucket with a transition to both AMI (which goes to post-MI IHD) and to angina, and a transition arrow between them?
2) How should we handle the incidence of angina (transition 4)? The incidence of cause 493 (IHD) is equal to the incidence of AMI. Do we want to leave this as the incidence of cause 493 or do we want to pull and process the incidence of angina from the dismod model (even though this incidence would be different from that of AMI)?


docs/source/gbd2019_models/causes/ihd/index.rst
```diff
@@ -116,331 +112,224 @@ Assumptions and Limitations
...
+     - Susceptible to IHD
+   * - A
+     - **A**\ cute Myocardial Infarction (MI)
+     - Simulant that experiences acute MI symptoms
+   * - P
+     - **P**\ ost-MI IHD
+     - Simulant that experiences angina and asymptomatic ischemic heart
+       disease following myocardial infarction; survival to 28 days following
+       incident MI
+   * - S2
```
aflaxman, line 149: It will be confusing to the engineers (and I confess it is also confusing to me) that there are two states labeled "Susceptible to IHD".  Is it possible to me in A and A2 simultaneously? If they are mutually exclusive, perhaps they should be in the same model diagram.

docs/source/gbd2019_models/causes/ihd/index.rst
```diff
@@ -116,331 +112,224 @@ Assumptions and Limitations
...
+     - duration-based transition from acute state then progress into post state
+   * - 3
+     - P
+     - A
+     - incidence_493
+     -
+   * - 4
+     - S2
+     - A2
+     - incidence_493
```
aflaxman, line 262: I'm thinking that this will end up with twice as much IHD as you have estimated in GBD, because people are moving from S and S2 to IHD states.

docs/source/gbd2019_models/causes/ihd/index.rst
```diff
@@ -116,331 +112,224 @@ Assumptions and Limitations
...
-     - Disability weight of sequela with id `sid`
-     - 
-   * - prevalence_s{`sid`}
+     - Disability weight of sequela with id {id}
+     -
+   * - incidence_493
      - como
-     - Prevalence of sequela with id `sid`
-     - 
+     - Incidence of ischemic heart disease
```
aflaxman, line 298: IHD is probably prevalent enough in older ages to warrant including the correction to go from "population rate" to "susceptible rate", i.e. i = incidence_493 / (1 - prevalence_493)

docs/source/gbd2019_models/causes/ihd/index.rst
```diff
@@ -116,331 +112,224 @@ Assumptions and Limitations
...
+     - {s380, s381, s382, s953}
+     -
 
 
 Validation Criteria
 +++++++++++++++++++
 
+.. todo::
+
+  Add in validation criteria
```
aflaxman, line 331: definitely include checks of the age-/sex-specific incidence, prevalence, excess mortality, cause-specific mortality, and all-cause mortality rates --- validate that the simulation comes up with rates within X% of the GBD estimates.

aflaxman (approved):

aliesally (approved):

docs/source/gbd2019_models/causes/ihd/index.rst
catherineojohnson: Edit pull request to reflect update to incidence calculation.

ndecleene:
Just added a commit to add the requested validation information and the update to the incidence calculation! The CVD team is still working on finalizing the cause diagram (which will likely result in some changes to the state and transition data tables), but for now I am merging this PR with the main branch.

PR #422, opened 2021-03-25
I edited the disease description section, GBD 2019 Modeling Strategy …

aliesally:
…section and included the updated cause hierarchy diagram.

A new column was added to the Severity Distribution table which is now a little too large for the page. I tried playing around with the column width sizes but it didn't seem to change anything. I also looked through the Quick reStructuredText page but didn't see anything about formatting the kind of table we are using.


ndecleene (approved):

docs/source/gbd2019_models/causes/ischemic_stroke/index.rst
```diff
@@ -33,165 +33,82 @@ Disease Overview
...
    :header-rows: 1
 
    * - Severity level
+     - Acute or Chronic
      - Lay description
      - Modified Rankin score
      - Cognitive status
      - DW (95% CI)
+   * - Stroke, asymptomatic
+     - Chornic only
```
aflaxman, line 72: typo: Chronic

aflaxman (approved):

docs/source/gbd2019_models/causes/ischemic_stroke/index.rst
aliesally: Oh good catch. Thank you!

PR #428, opened 2021-03-27
Added vivarium strategy

ndecleene:

PR #429, opened 2021-03-27
Added vivarium strategy

ndecleene:


aflaxman (approved):
Same comment as for #428 about the way the CSMR for D states will be confusing.

aliesally (approved):

PR #431, opened 2021-03-30
Feature/gbd2019 pad

aliesally:
Hi @aflaxman, we had a question for you: for causes where it is important to use the “susceptible rate” rather than the “population rate” is it better to use the cause-level incidence (which has undergone this transformation) and back transform? Or should we use the “hazard rate” from the DisMod which is the “susceptible rate”? 

Also, the death state (CSMR) is included because it is important to know where to pull CSMR from and we can update it later after more discussion or if there is confusion for people.


docs/source/gbd2019_models/causes/pad/cause_hierarchy_pad.svg
```diff
@@ -0,0 +1,3 @@
...
...witch><foreignObject style="overflow: visible; text-align: left;" pointer-events="none" width="100%" height="100%" requiredFeatures="http://www.w3.org/TR/SVG11/feature#Extensibility"><div xmlns="http://www.w3.org/1999/xhtml" style="display: flex; align-items: unsafe center; justify-content: unsafe center; width: 118px; height: 1px; padding-top: 462px; margin-left: 172px;"><div style="box-sizing: border-box; font-size: 0; text-align: center; "><div style="display: inline-block; font-size: 12px; font-family: Helvetica; color: #000000; line-height: 1.2; pointer-events: all; white-space: normal; word-wrap: normal; "><div>1 sequela:</div><div>asymptomatic peripheral arterial disease</div><div>s_1041</div></div></div></div></foreignObject><text x="231" y="466" fill="#000000" font-family="Helvetica" font-size="12px" text-anchor="middle">1 sequela:...</text></switch></g></g><switch><g requiredFeatures="http://www.w3.org/TR/SVG11/feature#Extensibility"/><a transform="translate(0,-5)" xlink:href="https://www.diagrams.net/doc/faq/svg-export-text-problems" target="_blank"><text text-anchor="middle" font-size="10px" x="50%" y="100%">Viewer does not support full SVG 1.1</text></a></switch></svg>
```
ndecleene, line 3: I think the sequela_id in this diagram should actually be 964, not 946

ndecleene (approved):

docs/source/gbd2019_models/causes/pad/cause_hierarchy_pad.svg
aliesally: Oh thank you! I will change that!

aflaxman:
> Hi @aflaxman, we had a question for you: for causes where it is important to use the “susceptible rate” rather than the “population rate” is it better to use the cause-level incidence (which has undergone this transformation) and back transform? Or should we use the “hazard rate” from the DisMod which is the “susceptible rate”?

I prefer you use the cause-level incidence and back transform; this handles any funny business that GBD introduced after DisMod, such as in the COMO step.

> Also, the death state (CSMR) is included because it is important to know where to pull CSMR from and we can update it later after more discussion or if there is confusion for people.

:thumbs up emoji:

aflaxman (approved):

PR #446, opened 2021-04-15
Added vivarium research strategy

ndecleene:


aflaxman (approved):
I don't know if this will work for the software engineers!  We usually have a "with condition" state, and excess mortality only for the simulants in that state.  @collijk : what do you recommend for a GBD cause that is considered fatal-only like AA?

aliesally (approved):

collijk:
Fatal only for our model just means that they won't accrue any disability from the condition.  That's fine.  
- who has AA at the start of the simulation (e.g. the prevalence)
- how people who don't have AA get AA as time passes, their risk profile changes, they get older, etc. (e.g. an incidence and relationships with risks defined by RRs)
- if people recover, how quickly they recover (a remission)

We can compute an excess mortality as csmr/prevalence

alibow:
Is the issue that there isn't any incidence/prevalence data to work with? If that is the case, would it work to set the incidence rate equal to the CSMR and then EMR=1? I think that how vivarium works that simulants would "get" AA one timestep before they died from it, but I believe that should be fairly inconsequential.

alibow:
We also never got around to building this model out, so it is not yet software engineer tested and approved, but this is an example of something we put together for a fatal-only model that affected the mortality rate directly and did not have a with condition state and EMR. https://vivarium-research.readthedocs.io/en/latest/gbd2017_models/causes/smoking_related_mortality/index.html

PR #448, opened 2021-04-15
Added vivarium modeling strategy

ndecleene:


aflaxman (approved):
"with COPD" state is sometimes "C" and sometimes "P"; I suggest you use "C".

ndecleene:
Thanks for flagging, Abie! I've fixed all the "with COPD" states to be "C" instead of "P"

aliesally (approved):

PR #450, opened 2021-04-15
Fixed the A/P state discrepency.

aliesally:
I accidently merged the last PR before making these changes, so I created a new PR.
I chose to use "A" instead of "P" or "C" for the prevalent state to match the cause model diagram, but noticed that Nikki chose to go with "C", and am happy to change it to "C" and update the diagram if people prefer them to match.


ndecleene (approved):
I went with "C" to stand for "with **C**OPD" so I think it makes sense to go with "A" for afib!

aflaxman (approved):

PR #451, opened 2021-04-17
Adding CKD IKF model GBD 2019

SantoniS:
I compared the GBD 2017 and GBD 2019 models for CKD cause model and IKF risk factor model. It looks like the CKD cause model is mostly the same. I added information within the PR and documentation of this page where the modeling strategy differed for GBD 2019. I updated the restrictions for IKF (yll age start update).


docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
```diff
@@ -50,9 +50,15 @@ Vivarium - CSU - Simulating Multiple Myeloma registries
 | MM    | Multiple myeloma           |
 +-------+----------------------------+
 | OS    | Overall Survival           |
+<<<<<<< HEAD
```
aflaxman, line 53: You need to manually resolve these "merge conflicts" by deciding if you want to keep that part between <<<< and === or the part between === and >>>

aflaxman (approved):
Are you sure that CKD/IKF matches the concept of Renal Impairment needed for the MM model? I'm starting to suspect that they are more interested in how MM leads to degraded renal function, and not the GBD concepts of the cause of CKD, or the risk factor of IKF.

Even if this is not needed for the MM work, it will likely be useful for the CVD work that @catherineojohnson and team are doing.

yongqx (approved):

PR #457, opened 2021-04-20
Added GBD modeling strategy for BMI

ndecleene:


aliesally (approved):
Again I think that using overweight and obesity together sounds awkward. I would change it to overweight and obese.

aflaxman (approved):

ndecleene:
For the GBD modeling strategy I left it as "obesity" since that is the what the team who models this risk uses in their write-ups and this section is pulled directly from there. But if everyone thinks it should be "obese" instead I can change it!

PR #462, opened 2021-04-21
Added Vivarium Modeling Strategy for LDL.

aliesally:
@alibow I was wondering if you could weigh in on whether the ldl model is complicated enough to need a risk exposure model diagram. Thank you!

### DIFF
--- a/docs/source/gbd2019_models/risk_exposures/ldl/index.rst
+++ b/docs/source/gbd2019_models/risk_exposures/ldl/index.rst
@@ -55,7 +55,11 @@ The TMREL for LDL-C is 0.7 to 1.3 mmol/L. This is the same for all risk-outcome 
 Vivarium Modeling Strategy
 --------------------------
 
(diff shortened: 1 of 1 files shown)


aflaxman (approved):

ndecleene (approved):

alibow:
I do not think that this needs a diagram and it looks good to me! The only other thing I would note for validation criteria (and it might not apply!) is to check and see if any illogical values are sampled from the ensemble distribution. For instance, if you generate a negative number on the very tail end of the distribution and then apply %reduction to the exposure or something similar, you might end up increasing the exposure values. It's unlikely to happen, but it could be worth throwing in the valiation criteria to make sure all sampled values are positive, if applicable!

alibow (approved):

ndecleene:
Thanks, Ali! I'm merging in this PR for now so we can keep moving forward, but we can update this with additional validation criteria later.

PR #463, opened 2021-04-21
Added Vivarium Modeling Strategy for FPG.

aliesally:
@alibow I have the same question as I had for LDL; wondering if the FPG model is complicated enough to need a model diagram. Thanks!


ndecleene (approved):

aflaxman (approved):
If this is an ensemble distribution, say that and include info on the ensemble weights.  Otherwise say what distribution.

alibow:
I also don't think that this needs a diagram!

alibow (approved):

PR #465, opened 2021-04-23
update mm concept model doc

yongqx:


aflaxman (approved):

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
```diff
@@ -295,22 +294,46 @@ developing MM. Instead, they are considered as determinants in patient and
...
 the choice of therapy is based on their GFR and cytogenetic risk. 
 
-**GFR** has a continuous risk exposure by age and sex (or race/ethnicity). 
-Simulants will be assigned to a CKD stage based on their GFR value. We consider 
-4 stages of CKD:
- - No CKD: GFR > 90 ml/min/1.73m^2
- - Mild CKD: GFR 60 to 90 ml/min/1.73m^2
- - Moderate CKD: GFR 30 to 60 ml/min/1.73m^2
- - Severe CKD: GFR < 30 ml/min/1.73m^2
+**eGFR** has a continuous risk exposure by age, sex, and race/ethnicity.
```
SantoniS, line 297: In GBD, CKD is defined as permanent loss of renal function. According to literature, treating the myeloma should reverse kidney impairment, sometimes even in a patient whose kidneys fail or a patient who requires dialysis. However, longer-term kidney failure is usually not reversible. Should we be explicit if we are going to incorporate eGFR and CKD in a way separate from GBD?

SantoniS (approved):

PR #467, opened 2021-04-28
Add page for Ensemble Distributions in Vivarium Model Considerations

NathanielBlairStahn:
This is an outline for a page about ensemble distributions in GBD. I put it under the "Vivarium Model Considerations and Best Practices" section -- let me know if that seems like a good location. Also let me know if you think there should be other sections in the document.


aflaxman (approved):
Eventually we could have info on all the dists included in the ensemble, but can you make just stubs for them, for now?

Ninicorn (approved):

PR #470, opened 2021-05-04
Ensemble distributions - start explaining math

NathanielBlairStahn:
Added the beginning of a section explaining how ensemble distribution modeling works in GBD. It would be good to get other people's eyes on this to make sure what I'm saying so far is accurate.

### DIFF
--- a/docs/source/model_design/concept_model/ensemble_distributions/index.rst
+++ b/docs/source/model_design/concept_model/ensemble_distributions/index.rst
@@ -68,9 +68,10 @@ distribution on :math:`\mathbb{R}` whose cumulative distribution function :math:
   F(x) = \sum_{i=1}^n w_i F_i(x)\quad \text{for } x\in \mathbb{R}.
 
 That is, the CDF of the mixture is a weighted average of the CDFs of the
-components, with the weight of :math:`F_i` being :math:`w_i`. If the
-:math:`F_i`'s all correspond to `continuous probability distributions`_ on
-:math:`\mathbb{R}`, then differentiating the above equation shows that the
+components, with the weight of :math:`F_i` being :math:`w_i`.
+
+If the :math:`F_i`'s all correspond to `continuous probability distributions`_
+on :math:`\mathbb{R}`, then differentiating the above equation shows that the
 mixture distribution is also continuous, with `probability density function
 <PDF_>`_ (PDF) given by the coresponding weighted average of the component
 densities, i.e. :math:`f(x) = \sum_1^n w_i f_i(x)`, where :math:`f=F'` and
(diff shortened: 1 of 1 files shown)


docs/source/model_design/concept_model/ensemble_distributions/index.rst
```diff
@@ -49,6 +49,113 @@ fits the data better than any of its individual components.
 Mathematical Definition of Ensemble Distributions
 -------------------------------------------------
 
+Mathematically, an ensemble distribution in GBD is just a `mixture
```
aflaxman, line 52: ~~just~~

docs/source/model_design/concept_model/ensemble_distributions/index.rst
```diff
@@ -49,6 +49,113 @@ fits the data better than any of its individual components.
...
+numbers :math:`w_1, w_2,\ldots, w_n`, called **weights**, such that
+:math:`\sum_1^n w_i = 1`. Then the **mixture distribution** with components
+:math:`\{F_i\}_1^n` and weights :math:`\{w_i\}_1^n` is the probability
+distribution on :math:`\mathbb{R}` whose cumulative distribution function :math:`F` is defined by
+
+.. math::
+
+  F(x) = \sum_{i=1}^n w_i F_i(x)\quad \text{for } x\in \mathbb{R}.
+
+That is, the CDF of the mixture is just a weighted average of the CDFs of the
```
aflaxman, line 70: ~~just~~

docs/source/model_design/concept_model/ensemble_distributions/index.rst
```diff
@@ -49,6 +49,113 @@ fits the data better than any of its individual components.
...
+
+Suppose we have a continuous risk exposure variable :math:`E` that we want to
+model. For example, :math:`E` could be body mass index, fasting plasma glucose,
+birthweight, or hemoglobin level. Suppose that we want to model :math:`E` for
+:math:`K` different population groups, e.g. indexed by location/age/sex/year as
+in GBD. Let :math:`k\in \{1,2,\ldots,K\}` specify which population group we are
+referring to. Suppose further that we have microdata on :math:`E` for at least
+some of the groups, and we are able to estimate the mean :math:`\mu_k` and
+standard deviation :math:`\sigma_k` of :math:`E` for each group :math:`k`;
+typically GBD teams do this using ST-GPR or other regression methods. Our goal
```
aflaxman, line 103: this aside seems out of place here---i suggest you first set up the problem, and then say how GBD researchers typically solve it.

aflaxman (approved):
A few suggestions to tighten the prose.  Let's get @collijk to read this and see if the technical parts match what he would expect.

docs/source/model_design/concept_model/ensemble_distributions/index.rst
NathanielBlairStahn: I guess there are two problems we could be talking about:

- **First problem:** How do you get means and standard deviations for E for all population groups, given microdata for some population groups, and covariates for all population groups?
- **Second problem:** Once you have means and standard deviations for all population groups, and microdata for some of them, how do you find a distribution to model the exposure of E?

The "problem setup" I'm trying to describe is for the second problem. I just added the comment as an aside about how the first problem is solved so the reader would have some idea how we got to the setup of the second problem. Do you think I should just omit the aside, or should I expand the discussion to include solving the first problem as well as the second problem?

aflaxman: I propose you state the first problem, and say that there is a solution (but don't say what it is, or link out to material on ST-GPR); then you can keep your text focused on the second problem.

NathanielBlairStahn: @aflaxman How does it look now?

aflaxman:
:thumbs up emoji:

docs/source/model_design/concept_model/ensemble_distributions/index.rst
```diff
@@ -49,6 +49,119 @@ fits the data better than any of its individual components.
...
+.. todo::
+  **Fill in this section.** Here are some unfinished snippets I started:
+
+  The family of functions :math:`F(x | \mu, \sigma)` will be defined as a
+  mixture of a fixed set of base distributions :math:`F_1(x | \mu, \sigma),
+  F_2(x | \mu, \sigma),\ldots, F_k(x | \mu, \sigma)`  parameterized by mean
+  :math:`\mu` and standard deviation :math:`\sigma`, using a fixed global set of
+  weights :math:`\{w_i\}_1^k` for all :math:`\mu` and :math:`\sigma`. GBD calls
+  :math:`F(x | \mu, \sigma)` an **ensemble distribution**.
+
```
yongqx, line 150: I wonder does GBD modeler always construct their ensemble distribution using 14 distributions (k=14)? How to determine the number of K and what are best fit using subset of certain distributions. I think the next step is to describe the methods of calculating distribution specific weight.

yongqx (approved):

docs/source/model_design/concept_model/ensemble_distributions/index.rst
yongqx: I noticed that you introduced a simple approach, only limited to distributions of 2-paramater family, what about the mixture of 3-parameter and 2-parameter?

NathanielBlairStahn: It looks like only 11 of the 14 distributions are actually used in the current version of the ensemble code -- in particular, neither of the 3-parameter distributions are used (though it looks like the generalized lognormal was previously used).  I don't know how the ensemble fitting would deal with 3-parameter distributions instead of 2-parameter.

Some teams (e.g. the anemia team) may use a custom version of ensemble distributions that use fewer than the 14 (or 11) listed here. If the modelers use the ensemble fitting code from the Central Comp repo (see https://github.com/ihmeuw/vivarium_research/pull/479 for a description of what's in this repo), then weights will be calculated for all 11 distributions, but some weights may be 0. Including extra distributions in the weight optimization shouldn't hurt because the weight can just be 0 or close to 0 if the distribution doesn't need to be used.

PR #472, opened 2021-05-07
Measles 2019, updated SIR model

alibow:
I started the page for the 2019 measles cause model. I attached a screenshot here of the problem we needed to solve from the previously documented SIR measles model (https://vivarium-research.readthedocs.io/en/latest/gbd2017_models/causes/measles/index.html). Essentially, the assumption of initializing zero simulants to the recovered state caused a shrinking of the susceptible population over time that caused our incidence and mortality rates to also decrease as the simulation progressed.

I made a suggestion for an updated value for the prevalence of the recovered state at model initialization that I think will fix this problem. Notably, I think that changing to an SIS model would also fix this problem, but I think we are set up better to incorporate vaccine coverage into this model than an SIS model in case we want to expand to that in the future.

There are still a couple outstanding to-dos in this PR (detailing the GBD modeling process and decomp steps) that I will update in a separate PR.

Thanks!

image


aflaxman (approved):

beatrixkh (approved):
Looks like a good start!

docs/source/gbd2019_models/causes/measles/index.rst
```diff
@@ -0,0 +1,398 @@
...
+vaccination campaign.
+
+2. This model uses country-level data, and cannot be used to model local measles
+outbreaks due to lack of vaccination in small communities.
+
+Some of the **assumptions** made in this model are:
+
+1. There is no data available for population in recovered state in GBD. Since the early and late neonatal age groups are not modeled in GBD, we made the assumption that there are no individuals in the recovered state at the start of the post-neonatal age group. We then calculated the proportion of the population in the recovered state in the 1-4 year age group using GBD measles incidence and mortality rates. Note that we performed this calculation only for the 1-4 age group because measles is most often of interest among the children under five population, but this assumption could be expanded to other older ages as well.
+
+2. There is no data avaialable for remission rate in GBD. So a constant remission rate is
```
Ninicorn, line 119: available typo

docs/source/gbd2019_models/causes/measles/index.rst
```diff
@@ -0,0 +1,398 @@
...
+     - GBD age group id is 15
+   * - YLD age group start
+     - Post Neonatal
+     - GBD age group id is 4
+   * - YLD age group end
+     - 50 to 54 years
+     - GBD age group id is 
+
+Notably, our cause model as described accounts for the number of simulants that enter the recovered state by becoming infected with measles and recovering, but does not consider simulants who enter the recovered state by receiving the measles vaccine without becoming infected with measles. Therefore, we are underestimating the prevalence of the recovered state in our model. This strategy was employed so that the size of the susceptible and recovered populations at model initialization will remain stable as the simulation runs and simulants progress through the SIR cause model (this is an improvement from our previous assumption that there was zero prevalence of the recovered state upon model initialization, which caused an decrease in the susceptible population and an increase in the recovered population over time, which affected calibration of simulation incidence and mortality rates to GBD rates).
+
```
Ninicorn, line 156: Maybe somewhere in here, add a little more clarification why it wasn't stable. something like:

incidence_gbd = #measles/(S+I+R)

In vivarium, we apply incidence to the S population only, so we must adjust the incidence_gbd appropriately so that we may get the correct number of measles cases 

incidence without prevR adjustment - the denominator is 1- prevI:  
 = #measles/(S+R)

incidence with prevR adjustment - the denominator is 1-prevI-preR:
 =  #meales/S

Without the prevR adjustment, the denominator is bigger, therefore the incidence is smaller, so we will underestimate measles cases. Hence, as we apply an under-estimated incidence rate to a shrinking S population, we get lower measles CSMR over time. 

~ something like that??

Ninicorn (approved):
This is very clear documentation! I added a little bit more detail to describe what is happening if we did not initialize an R prevalence. I hope I understood it correctly! Let me know if I am wrong >.<

docs/source/gbd2019_models/causes/measles/index.rst
```diff
@@ -0,0 +1,398 @@
...
+-------------------
+
+Measles_ is a highly contagious, serious disease caused by the measles virus
+(*Measles morbillivirus*). Symptoms usually develop 10-12 days after exposure to
+the virus, and last 7-10 days. Symptoms include fever, cough, runny nose,
+conjunctivitis, characteristic white spots inside the cheek (called Koplik's
+spots), and a red, flat, blotchy skin rash that develops on average 14 days
+after exposure to the virus (range, 7-21 days) and lasts 5-6 days. Recovery from
+measles confers lifelong immunity. [WHO]_, [CDC]_, [Wikipedia]_,
+[GBD-2017-YLD-Capstone-Appendix-1]_
```
NathanielBlairStahn, line 17: Looks like these are going to give us "duplicate citation" warnings and probably end up linking to the citations on a different page (like the 2017 Measles page). Should we rename the citations to be specific to this page?  Also the 2017 appendix should be updated to the 2019 appendix.  This could all be done in a separate PR, but just thought I'd point it out.

alibow: Thanks for pointing this out. I don't think that the duplicate citations bring you to another page, but I will update to the 2019 appendix in the other open PR I have for measles that adds detail on the GBD modeling strategy.

PR #477, opened 2021-05-10
Ensemble distribution sampling

NathanielBlairStahn:
Here's the section describing how to sample from a mixture distribution. It would be good to have both researchers and engineers take a look to make sure everything makes sense.

### DIFF
--- a/docs/source/model_design/concept_model/ensemble_distributions/index.rst
+++ b/docs/source/model_design/concept_model/ensemble_distributions/index.rst
@@ -29,7 +29,6 @@ Ensemble Distributions in GBD
 
 .. contents::
    :local:
-   :depth: 2
 
 What Is an Ensemble Distribution?
 ---------------------------------
@@ -223,6 +222,12 @@ specified.
 Sampling from Ensemble Distributions in Vivarium
 ------------------------------------------------
 
+Vivarium needs to sample values of a risk exposure variable :math:`E` from its distribution in order to assign an exposure value to each simulant. This contrasts with the usage of ensemble distributions in GBD, where computing the CDF or PDF of :math:`E` is sufficient. For example, a typical use case
+of ensemble distributions for a GBD team might be to estimate the prevalence of
+each exposure category of :math:`E` by computing areas under its PDF.
+
+
+
 As noted above, a GBD ensemble distribution is mathematically defined as a mixture distribution, so it is sufficient to describe how to sample from mixture distributions.
 
 Sampling from a Mixture Distribution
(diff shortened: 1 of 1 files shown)


aflaxman (approved):
Perhaps include some notes on how changing over time might be accomplished in the first of these approaches and perhaps not in the second?  No need to develop this line of inquiry fully---just get some notes down in case we come back to it one day in the future.

NathanielBlairStahn:
> Perhaps include some notes on how changing over time might be accomplished in the first of these approaches and perhaps not in the second? No need to develop this line of inquiry fully---just get some notes down in case we come back to it one day in the future.

@aflaxman What do you have in mind here?

aflaxman:
Here is some text, based on my recollection of the issues James raised when we discussed this last week:

In future simulations, we might want individual simulants to have risk factor exposures that are correlated over time, but not perfectly correlated over time. For example, this could be accomplished in a normally distributed risk by changing the risk factor propensity by a small amount every timestep, in a way that keeps the propensity uniformly distributed between zero and one.  The mixture interpretation of the ensemble distribution is amenable to a similar sort of imperfectly autocorrelated risk exposure, by changing the second propensity while leaving the first fixed.  This might result in unexpected features, however, and we will need to proceed with caution, if and when the time comes.

NathanielBlairStahn:
> Here is some text, based on my recollection of the issues James raised when we discussed this last week: In future simulations, we might want individual simulants to have risk factor exposures that are correlated over time, but not perfectly correlated over time. For example, this could be accomplished in a normally distributed risk by changing the risk factor propensity by a small amount every timestep, in a way that keeps the propensity uniformly distributed between zero and one. The mixture interpretation of the ensemble distribution is amenable to a similar sort of imperfectly autocorrelated risk exposure, by changing the second propensity while leaving the first fixed. This might result in unexpected features, however, and we will need to proceed with caution, if and when the time comes.

Ok, my plan is to discuss issues such as these in the final two subsections that I haven't written yet.  I'll add a Todo with the above text so we don't forget.

alibow (approved):
I think this looks good, but I think that adding a sentence or two describing the methodology of each option in lay terms could be valuable so that people can quickly pick up the differences without digging into all the mathematical details.

NathanielBlairStahn:
> I think this looks good, but I think that adding a sentence or two describing the methodology of each option in lay terms could be valuable so that people can quickly pick up the differences without digging into all the mathematical details.

Thanks @alibow .  What do you have in mind when you say "in lay terms"?  The existing two sections are intended to provide the mathematical background about why the sampling methods work, and my current plan is to fill in the final two subsections about propensity-based sampling with a description of how to actually implement the sampling methods in code (but I haven't written this part yet).  Would that be sufficient, or do you think it needs something else?

NathanielBlairStahn:
> Perhaps include some notes on how changing over time might be accomplished in the first of these approaches and perhaps not in the second? No need to develop this line of inquiry fully---just get some notes down in case we come back to it one day in the future.

Hi @aflaxman , what did you mean when you said "perhaps not in the second"?  I'm interpreting the "first approach" to be the two-propensity method and the "second approach" to be the one-propensity method -- is that what you meant here? If we implemented the one-propensity approach, then wouldn't changing the propensities work just like it would for any other risk factor?

aflaxman:
Hmm, I don’t remember what I was thinking at the time, and what you say makes sense to me now.  Ignore!

docs/source/model_design/concept_model/ensemble_distributions/index.rst
```diff
@@ -222,3 +221,170 @@ specified.
...
+Below, we describe two possible strategies for sampling from an ensemble
+distribution. The first strategy (:ref:`two-step sampling
+<two_step_sampling_of_mixtures>`) requires two random choices for each sampled
+value but is easy to implement if we already have a sampling strategy for all
+the base distributions. The second strategy (:ref:`inverse transform sampling
+<inverse_transform_sampling_of_mixtures>`) uses a single random choice for each
+sampled value but requires calculation of the ensemble distribution's quantile
+function, which is not straightforward. As noted above, a GBD ensemble
+distribution is mathematically defined as a mixture distribution, so it is
+sufficient to describe how to sample from mixture distributions.
```
NathanielBlairStahn, line 242: @alibow I updated this introductory paragraph with brief descriptions of the two sampling methods. Does this address your concern about being able to quickly pick up the differences between them?

alibow: Yes, this looks good to me, thank you!

NathanielBlairStahn: Great, thanks!

PR #480, opened 2021-05-13
GBD 2019 LRI: validation criteria

beatrixkh:


aflaxman (approved):
I expect some remitted cases during the early neonatal period.  Those brief age groups have always been hard to check for calibration, though.

plinds (approved):

PR #485, opened 2021-05-19
MM treatment model strategy detail

alibow:
This is a strategy for what I am considering the minimal treatment coverage algorithm.

I want to confirm with Manoj how we should handle subsequent treatment with isa if already treated with isa (I have proposed two different directions in this PR).

My main concern with this strategy as-is is our ability to apply reasonable efficacy data between the isa group and the "other" group. I am now shifting my focus to going to go through the different treatment comparisons that exist in the literature to see if we can tease out something reasonable.

### DIFF
--- a/docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
+++ b/docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
@@ -424,13 +424,15 @@ Treatment Modeling Strategy
 
 **Scope**:
 
-This modeling strategy allows us to track the proportion of MM/RRMM simulants that receive an isatuxamib-containing treatment. It also allows for coherent assignment of isatuxamib-containing treatments at the simulant level so that any given simulant is not assigned an isatuxamib-containing treatment an illogical amount of times, which will reduce noise in our estimations.
+This modeling strategy allows us to track the proportion of MM/RRMM simulants that receive isatuxamib as part of their regimen for a given treatment line. We assume that the scale-up of isatuxamib coverage will result in isatuxamib incorporation as an additional drug added to an existing treatment regimen that would have been delivered in the absence of an isatuxamib coverage scale-up. It also allows for coherent assignment of isatuxamib-containing treatments at the simulant level so that any given simulant is not assigned an isatuxamib-containing treatment an illogical amount of times, which will reduce noise in our estimations.
 
 This model strategy does NOT consider:
 
 - Differences between isatuxamib-containing treatment regimens (ex: isatuxamib + pomalidomide + dexamethasone versus isatuxamib + carfilzomib + dexamethasone)
-- Differences between the numerous treatment regimens that do NOT contain isatuxamib (note this has important implications for applying efficacy data to the appropriate reference group(s))
-- Specific eligibility requirements for isatuxamib-contianing treatments other than previous treatment with isatuxamib. Specifically, 
+- Differences between existing treatment regimens that would be eligible/ineligible for the addition of isatuxamib (For instance, isatuxamib logically would not be added to a treatment regimen that already contains dara. Additionally, isatuxamib is more likely to be added to a doublet treatment regimen than a triplet regimen.)
+- Differences between the numerous treatment regimens that do NOT contain isatuxamib 
+- Eligibility requirements for isatuxamib-containing treatments other than previous treatment with isatuxamib. Specifically, 
+
 	- We do not directly consider the requirement of previous treatment with lenalidomide and a PI, although this is likely minimally impactful as these are very common treatment regimens and "almost all" patients with myeloma develop disease that is eventually refracotory to lenalidomide and PIs, [Attal-et-al-2019]_.
 	- We do not consider previous treatment with other monoclonal antibody drugs such as daratumumab, which may impact probability of receiving an isa-containing treatment
 
@@ -441,7 +443,7 @@ We will consider the following treatment categories in our treatment model:
 
 .. note::
 
-	Other treatment categories to consider tracking include: reference treatment regimen from ISA RCT (polidamide + dexamethasone), daratumumab-containing regimens
+	Other treatment categories to consider tracking include: daratumumab-containing regimens
 
 Importantly, we will assume that a given percentage of patients who receive isa-containing regimen for a specific line of treatment will not receive an isa-containing regimen at a later line of treatment.
 
@@ -457,7 +459,7 @@ Importantly, we will assume that a given percentage of patients who receive isa-
 
 	Consider similarly assuming that patients who recieve dara-containing regimens will not receive isa-containing regimens at later lines of treatment.
 
-The overall proportion of patients who receive isa-containing treatment regimens at each line of treatment and in each scenario is shown in the table below.
+The overall proportion of incident cases that receive isa-containing treatment regimens upon diagnosis of MM or RRMM at each line of treatment and in each scenario is shown in the table below.
 
 .. list-table:: Isatuximab treatment coverage assumptions
    :header-rows: 1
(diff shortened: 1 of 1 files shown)


alibow:
Also, I just saw the discussion about the "burn-in" period, so we might need to make some assumptions about isa-coverage = 0% prior to 2020 (FDA approval date) to get the initialization right, but let me know if anyone thinks we'll have to do anything differently there.

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
```diff
@@ -419,14 +419,109 @@ As discussed in treatment guideline reviews, in general triplet regimens are pre
...
 
-Second-line treatment
-^^^^^^^^^^^^^^^^^^^^
+**Scope**:
+
+This modeling strategy allows us to track the proportion of MM/RRMM simulants that receive an isatuxamib-containing treatment. It also allows for coherent assignment of isatuxamib-containing treatments at the simulant level so that any given simulant is not assigned an isatuxamib-containing treatment an illogical amount of times, which will reduce noise in our estimations.
+
+This model strategy does NOT consider:
+
+- Differences between isatuxamib-containing treatment regimens (ex: isatuxamib + pomalidomide + dexamethasone versus isatuxamib + carfilzomib + dexamethasone)
```
yongqx, line 431: You may ignore the differences of efficacy for isa-containing regimens, but IsaPd is recommended to be used at 3rd line or later while IsaKd is recommended to be used at 2nd line or later.

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
```diff
@@ -419,14 +419,109 @@ As discussed in treatment guideline reviews, in general triplet regimens are pre
...
+- Differences between the numerous treatment regimens that do NOT contain isatuxamib (note this has important implications for applying efficacy data to the appropriate reference group(s))
+- Specific eligibility requirements for isatuxamib-contianing treatments other than previous treatment with isatuxamib. Specifically, 
+	- We do not directly consider the requirement of previous treatment with lenalidomide and a PI, although this is likely minimally impactful as these are very common treatment regimens and "almost all" patients with myeloma develop disease that is eventually refracotory to lenalidomide and PIs, [Attal-et-al-2019]_.
+	- We do not consider previous treatment with other monoclonal antibody drugs such as daratumumab, which may impact probability of receiving an isa-containing treatment
```
yongqx, line 435: No previous treatment with Dara doesn't mean you couldn't get Dara at same line as Isa, right?

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
```diff
@@ -419,14 +419,109 @@ As discussed in treatment guideline reviews, in general triplet regimens are pre
...
+We will consider the following treatment categories in our treatment model:
+
+- Isa-containing treatment regimen
+- Non-isa-containing treatment regimen
+
+.. note::
+
+	Other treatment categories to consider tracking include: reference treatment regimen from ISA RCT (polidamide + dexamethasone), daratumumab-containing regimens
+
+Importantly, we will assume that a given percentage of patients who receive isa-containing regimen for a specific line of treatment will not receive an isa-containing regimen at a later line of treatment.
```
yongqx, line 446: That is to say, Isa is a one-time treatment among myeloma patients. What about Dara and other types of regimen?

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
```diff
@@ -419,14 +419,109 @@ As discussed in treatment guideline reviews, in general triplet regimens are pre
...
+	
+	Confirm use of probability 1.0 for ineligibility for future isa-containing treatment regimen based on past treatment with isa.
+
+	Alternative approach: make ineligible if simulant progresses on or within 60 days seven months time OR with a probability of 60% if simulant progresses within seven months. This was chosen because the ISA trial defined non-response as progression on or within 60 days or progression within 6 months after achieving at least a partial response and reported at least partial response in 60% of subjects and a median time to response among this group was 35 days).
+
+.. note::
+
+	Consider similarly assuming that patients who recieve dara-containing regimens will not receive isa-containing regimens at later lines of treatment.
+
+The overall proportion of patients who receive isa-containing treatment regimens at each line of treatment and in each scenario is shown in the table below.
```
yongqx, line 460: I assume overall proportion stands for percent of prevalent cases plus percent of incident cases, correct?

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
```diff
@@ -419,14 +419,109 @@ As discussed in treatment guideline reviews, in general triplet regimens are pre
...
+
+	Finalize numbers currently marked as "XXX" based on commercial team input
+
+**How to assign treatment category to a simulant:**
+
+*Initialization*:
+
+	Each simulant initialized into a multiple myeloma state should be assigned to a treatment category. They should be assigned to the isa-containing treatment category or non-isa-containing treatment category with the probability of the isa-containing treatment category equal to the isatuxamib treatment coverage values for 2021 in the table above, specific to the cause model state that the simulant occupies.
+
+	For simulants assigned to the isatuxamib-containing treatment category, assign the attribute :code:`ineligible_for_future_isa = True` with a probability of 1. For all other simulants, set :code:`ineligible_for_future_isa` to :code:`False`.
```
yongqx, line 498: Under current scenario design, the probability of receiving Isa-containing regimen increases across lines of treatment, while keeping the probability of receiving non-Isa containing regimen as constant during 5 years of simulation. Notice that Pisa + Pnon-isa != 1.

yongqx (approved):

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
alibow: I meant if you had dara in a previous line, you're less likely to be prescribed isa in a subsequent line. I don't think dara and isa would ever be given at the same time.

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
alibow: It's a bit complicated... the recommendation is that at each line of treatment the patient should receive at least two drugs they have not previously received, preferable from another class (this is for all anti-myeloma drugs, not just isa). So previous treatment with dara/isa makes it less likely that a patient will be treated again with dara/isa, but not impossible. That's why I tried to structure it as a percent of patients, so that we could tweak this to 90% or some other value so that we allow for the possibility of retreatment after consultation with Manoj. Alternatively, I also suggested that we try to model a sort of refractory to isa model that would determine retreatment possibility (described in the to-do note)

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
alibow: Hmm, I think we want this to be incident cases. I'll update this.

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
alibow: Ohh, I see. I think I may have been misunderstanding this. So that means that isa would be added to existing treatments rather than swapped out? Or just that non-isa-containing treatments would decrease in a proportional manner?

aflaxman (approved):

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
alibow: I am confused by this. I don't see how the probability could be greater than one if it's just two binary categories where isa is or isn't part of a treatment regimen for a given line.

alibow: How I've set it up, isa-containing treatment doesn't mean that you only get isa, it means that isa is a part of your treatment (one part of a triplet, for example).

yongqx: Yes. Ideally, the probability treated with non-isa regimen would decrease when Isa coverage increases over time. But we don't know how fast it falls down so we made an assumption of constant coverage rate for all different type of non-isa regimens. Patients now have a new option Isa, that doesn't necessarily mean they prefer tp switch to this new type of treatment.

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
yongqx: Thanks for explaining!

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
alibow: Hmm, I am still not following. Can we chat about this when we're free?

yongqx: Sure!

PR #486, opened 2021-05-19
Wasting transitions equations update

beatrixkh:
Updated wasting exposure page to include
- new code with which we solve transition rates in terms of dur_cat3 and i3
- tables with new transition rates

Todo: 
- fix the tables (they need to wrap)
- should those transition rates be included anywhere else in the page?
- would it be nice to have a visualization of the matrix equation we're solving? it's implicit in the equations but not included, as of now

Todo later:
- include a death state and update all equations. right now we assume too many people are remitting who should be dying.


beatrixkh:
I branched off of this PR to add the wasting transitions + death in this PR: https://github.com/ihmeuw/vivarium_research/pull/491

Ninicorn (approved):

PR #500, opened 2021-05-28
LBWSG relative risk interpolation

NathanielBlairStahn:
Here's a description of the method I used to interpolate the Low Birthweight & Short Gestation relative risks for the large-scale food fortification project.

The section got kind of long because I was trying to describe the procedure in enough detail to reproduce what I did.  If it's too much to digest in one PR, let me know and I can try to split it up.

### DIFF
--- a/docs/source/gbd2019_models/risk_effects/low_birthweight_short_gestation/index.rst
+++ b/docs/source/gbd2019_models/risk_effects/low_birthweight_short_gestation/index.rst
@@ -326,7 +326,7 @@ LBWSG RRs for the large-scale food fortification project in March 2021.
 #.  **Take logarithms:** Since the LBWSG relative risks vary widely between categories (from 1.0 in
     the TMREL up to more than 1600 in the highest risk category in some draws), we will do
     the interpolation in log space to keep everything at a reasonable scale, and then exponentiate the results.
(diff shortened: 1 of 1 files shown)


docs/source/gbd2019_models/risk_effects/low_birthweight_short_gestation/index.rst
```diff
@@ -61,6 +61,19 @@ USA analyses were sex-specific.
 category of birthweight and gestational age.**
 [GBD-2019-Risk-Factors-Appendix-LBWSG-Risk-Effects]_
 
+.. note::
```
aflaxman, line 64: perhaps note this at the end, instead of the beginning.

aflaxman (approved):
👍 thanks for getting this written out!

docs/source/gbd2019_models/risk_effects/low_birthweight_short_gestation/index.rst
NathanielBlairStahn: I wanted to put the note right after the first mention of 500g and 2wk categories to clarify that it's an approximate description before the description gets used multiple times. Also, in terms of document structure, the note refers to all the subsequent sections, so I thought it made sense to put it in the intro section of the GBD modeling section. But I guess it could come later if you think that's better.  Where exactly would you put it instead? Maybe at the end of the "Details of Relative Risk Estimation" section? Or in the exposure category/TMREL section since that actually shows the categories?

aflaxman:
Makes sense.  Fine to keep where it is.

yongqx (approved):

NathanielBlairStahn:
I added a couple notes based on some questions that came up in the documentation session today.

PR #513, opened 2021-06-11
MM treatment coverage and effects

alibow:


docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
```diff
@@ -478,6 +541,54 @@ The treatment algorithm is described in this :download:`word document found here
...
+
+   * - Treatment category
+     - Progression free survival HR (95% UI)
+     - Overall survival HR (95% UI)
+   * - All categories combined
+     - 1 (reference)
+     - 1 (reference)
+   * - Isatuxamib-containing treatment category, :code:`retreated != True`
+     - 0.530 (0.356, 0.803)
+     - 1.159 (1.044, 1.185)
```
aflaxman, line 576: I am surprised that this is >1

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
```diff
@@ -478,6 +541,54 @@ The treatment algorithm is described in this :download:`word document found here
...
+   * - Isatuxamib-containing treatment category, :code:`retreated == True`
+     - 0.765 (0.678, 0.902)
+     - 1.318 (1.088, 1.37)
+   * - Daratumumab-containing treatment category, :code:`retreated != True`
+     - 0.217 (0.203, 0.231)
```
aflaxman, line 581: I am surprised that this is so low

aflaxman (approved):

yongqx (approved):

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
alibow: This is above one in reference to Flatiron, but it is below one in reference to the residual category.

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
alibow: Unfortunately, I think this may be because of the R (lenalidomide) in dara-Rd, which is a pretty effective drug. 

There is clinical trial data for dara-pomalidomide-dexamethasone, but it is not approved by the FDA for this application (trial data just came out in 2021). I think dara-Pd would be a reasonable selection instead of dara-Rd because the isatuxamib in the RR stetting is in combination with pomalidomide and dexamethasone and also, lenalidomide is one of the most commonly used drugs in early lines of treatment and pomalidomide is recommended for use after lenalidomide, so I think it is reasonable. 

I can test and see what the HRs would be with the dara-Pd regimen instead of dara-Rd, although I originally chose not to use dara-Pd due to the lack of current FDA approval.

alibow: The dara-Pd PFS HR = 0.66, which is pretty much equivalent to the isa-Pd PFS HR (0.67) relative to flatiron, as expected.

The dara-Pd OS HR is 1.03, which is slightly favored over isa-Pd of 1.16 relative to flatiron.

We could also pull in dara-carfilzomib-dexamethasone (dara-Kd) data, which we have done for isatuxamib (the recently published IKEMA trial), so that we are using the same combination regimens between isa/dara. Dara-Kd is approved for use by the FDA. The dara-Kd regiment has very low HRs (0.1 PFS, 0.67 OS)

aflaxman:
I don’t see the dara-Pd OS HR of 1.03 in the table.  Is it supposed to be there?

alibow:
@aflaxman, no it is not supposed to be there. We are currently only using dara-Rd survival data for the dara-containing category. This is because dara-Rd is the only FDA approved dara-containing moab+imid+dex treatment and we only have coverage data for moab+imid+dex. 

I was suggesting that we could update this to incorporate dara-Pd either instead or in addition to dara-Rd given the recently published trial results for dara-Pd. 

Notably, if the dara-containing treatment category becomes less effective, it will make the transition from the residual category to isa less impactful because the residual category will have better survival rates. However, I don't expect these changes to have a major impact on the differences between our scenarios which will primarily differ in the first line of treatment.

aflaxman:
Ok, I understand.  I think the way you have done it makes sense at this point.

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
```diff
@@ -478,6 +541,72 @@ The treatment algorithm is described in this :download:`word document found here
...
+
+.. code-block:: python
+
+	from numpy import log
+	from scipy.stats import norm, lognorm
+
+	q_975_stdnorm = norm().ppf(0.975)
+	mu = log(hr)
+	sigma = (log(hr_upper) - mu) / q_975_stdnorm
+	hr_distribution = lognorm(s=sigma, scale=hr)
```
alibow, line 600: @collijk @mattkappel, FYI the sampling distribution instructions are in this PR now and should be merged soon! I believe there is an outstanding to-do to incorporate this into the treatment effects model.

PR #516, opened 2021-06-18
Updating Risk factor disease

SantoniS:


alibow:
Serena and I are having some make html challenges that we will work on sorting out next week, so sorry for any formatting bugs, but we wanted to get the information posted for review without delay :)

alibow:
I also accidentally included some changes to the treatment algorithm document that shouldn't be in this PR. I will get that cleaned up next week as well!

alibow:
> w/r/t Race exposed group being "Black", shouldn't the unexposed group be "non-Black" rather than "White"?
> 
> EDIT: nevermind, I see the explanation below.

@mattkappel, I agree that this got a little buried though... I restructured this so that our reference group is non-Black and added some detail to the limitation section about it.

I also added some additional contextual information about the burn-in period for initialization. I am not sure if we have documentation on this yet for the burn-in strategy that was used for the cause model, but that same strategy will be fine for this and I am planning on writing it up in detail with some updates to the retreatment model that will be coming soon.

For the link removal, I was testing out a bunch of changes to see if I could pass the make html tests --  I think I put everything back to how it was supposed to be, but I can't see the specific line number that you were referencing, so let me know if this is still outstanding. 

Thanks!!

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
```diff
@@ -288,62 +288,198 @@ See :ref:`multiple myeloma cause model<2019_cancer_model_multiple_myeloma>`
...
+
+Notably, we only have risk factor exposure distribution data among newly diagnosed patients. Due to the differential survival rates among the different risk exposure groups, we do not expect that the risk exposure distribution among relapsed and refractory multiple myeloma patients to be the same as among newly diagnosed patients. Due to our lack of data to inform risk exposure initialization probabilities among RRMM patients at the beginning of the simulation, we will model a "burn-in" period prior to the official time-frame of the simulation (2021-2026) in which all MM patients are initialized to the first MM state, allowing these risk exposure distributions to shift along with disease progression according to the risk effects described in the next section.
+
+.. todo::
+
+  Add more details on the burn-in period... to be expanded upon in the retreatment section
```
mattkappel, line 304: Looking forward to more details on the burn-in period. Would we want this in the main sim or is this something we could maybe pre-compute?

mattkappel (approved):

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
yongqx: I suggest pre-compute and save the distribution of risk prevalence from different MM states. Then we can iteratively use the the risk exposure outcomes.

alibow: I think that the initial idea was that we could run a burn in period, record information from it, and then not have to keep running it and initialize using that recorded information instead. I believe that it has already been performed for the cause model to calibrate the initial prevalence of each of the RRMM states (Yongquan/Abie/James worked on this), but I think we will have to do it again after the incorporation of this new information. Anyway, I will be working on this today so we should have more details soon!

yongqx (approved):

PR #518, opened 2021-06-23
Added outline and disease description

ndecleene:


aflaxman (approved):
This will be PAF of 1 with FPG, right? I ask because HHD had that featured prominently in the title of the page, and we should be consistent about it to avoid future confusion (always say in the title or never say in the title---I lean towards always.)

plinds (approved):

ndecleene:
Thanks for flagging, Abie! I have updated the title to reflect that diabetes is a PAF of 1 with FPG.

PR #524, opened 2021-06-25
PEM GBD 2019: added cause hierarchy

beatrixkh:


aflaxman (approved):

Ninicorn:
oh I meant to say this here: minor detail- I just noticed that the sequalae ids for SAM sequelae in the Data Sources and Definition Table is reversed: severe with oedema should be = 2036 and that for severe without oedma =199. It is correct in the cause hierarchy diagram.

Ninicorn (approved):

PR #525, opened 2021-06-25
PEM GBD 2019: added restrictions table

beatrixkh:


aflaxman (approved):

Ninicorn (approved):
Should we update this to PEM 2020?

Ninicorn:
minor detail- I just noticed that the sequalae ids for SAM sequelae in the Data Sources and Definition Table is reversed: severe with oedema should be = 2036 and that for severe without oedma =199. It is correct in the hierarchy diagram.

PR #527, opened 2021-06-25
MM risk factor updates

alibow:


docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
```diff
@@ -468,19 +460,20 @@ We assume that the racial distribution of newly diagnosed MM patients does not v
...
+We assume that the hazard ratios for PFS and OS among Black individuals relative to white individuals reported in [Derman-et-al-2020]_ is similar to those among Black individuals relative to non-Black individuals. We make this assumption in the absence of reported hazard ratios adjusted for treatment for additional racial groups. Additionally, we assume that the effect of cytogenetic risk on PFS and OS among white individuals is similar to that among non-Black individuals. While [Derman-et-al-2020]_ did include the patients who reported being Hispanic/Latino in the white category and the majority of the non-Black population in the US is white, this remains a limitation of our analysis that could potentially be address through access to Flatiron microdata.
```
aflaxman, line 471: (if the data includes accurate race/ethnicity information --- this is often missing!)

aflaxman (approved):

yongqx (approved):

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
```diff
@@ -305,13 +305,11 @@ In this model, we implement risk exposures for simulants upon entry to the MM st
...
 
-.. todo::
-
-  Add more details on the burn-in period... to be expanded upon in the retreatment section
+After the initial 10 burn-in period run, prevalent risk exposure distributions for race, cytogenetic risk, and renal function should be recorded for each MM cause model state. In order to achieve this, person-time stratified by risk exposure and multiple myeloma cause model state is required as a simulation output. Given our assumpiton of independence of risk exposures, the prevalent risk exposure distributions should be evaluated at the population level; however, the age risk exposure distribution should be stratified by GBD 5 year-age group. For future runs in which a burn-in period is not run and simulants are initialized into RRMM states rather than solely the first MM state, simulants should be assigned with risk exposures for race/cytogenetic risk/renal function at diagnosis with probabilities corresponding to the recorded risk exposure distribution specific to the MM cause model state that the simulant is initialized into.
```
mattkappel, line 312: > After the initial 10 burn-in period run

10 year? And if 10 years, is that a new definition? My understanding is that burn-in starts before 1/1/2016.

mattkappel (approved):

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
alibow: So there's going to be two separate burn ins... the first is to get the correct initialization for the cause model and risk exposures, starting by initializing all simulants into the first MM state, letting them trickle through to the later RRMM states over 10 years, and then recording the prevalence of each MM/RRMM state as well as the risk exposure distribution at each state. The second burn in period will use the info from the first and will initialize simulants into MM and RRMM states with the corresponding risk exposure distributions but in 2016 so that all simulants will be initialized prior to dara/isa introduction into the treatment landscape. We'll then run the burn in period so that we can track how dara/isa exposure evolves over time as the simulants who have been treated survive/die and the background coverage increases over time.

Does this help clarify?

PR #532, opened 2021-06-29
Updated vivarium strategy

ndecleene:


aflaxman (approved):
It has been confusing in the past to have two susceptible states in one disease model.  I interpret this to mean that there are effectively two independent models, one for MI and another for SA, and this is really two cause model documents in one page.  If that does not sound right, or of the engineers find it to confusing, let's clarify.

PR #540, opened 2021-07-02
wasting: added paf eqns and moved wasting to 2020 folder

beatrixkh:
- I took the wasting page that was housed in gbd2019 models, and moved it into the gbd2020 models folder.
- I changed the "2019s" to "2020s" and deleted the 2019 RR information. 
- I then deleted to 2019 wasting page.

- i added (briefly) some info in the "data tables" section on pulling the RRs and calculating the PAFs. (these need updating; not in final form)

@Ninicorn will you make sure nothing i pulled from your 2019 wasting page and now have housed under 2020 is now incorrect?


aflaxman (approved):

Ninicorn (approved):
For the description of the variables in the 'Wasting transition rate equations' and also 'Variables for transition probabilities', we should call them _probabilities_ rather than _rates_ just to be clear we are talking about a probability of transition with **denominator 1** rather than denominator of person-time for a rate. 

Also `ap_0` is not a number, but rather the proportion of simulants of die _among those alive_ right?

PR #541, opened 2021-07-07
add survival observer description

yongqx:


aflaxman (approved):
What does time frame for survival observer period mean?  @mattkappel do you understand what this is asking for?

yongqx:
> What does time frame for survival observer period mean? @mattkappel do you understand what this is asking for?

Maybe i should call it time interval, indicating the start date and end date of survival observations.

mattkappel (approved):
@aflaxman I understand the ask; I'm less sure about the implementation approach but I can think of a couple ways to do it.

PR #545, opened 2021-07-09
Updated MM treatment effects

alibow:


yongqx (approved):

aflaxman (approved):
The difference here is really striking --- let's discuss to make sure we have a shared understanding of the implications.

PR #546, opened 2021-07-09
Updated MM HR sampling instructions

alibow:


docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
```diff
@@ -777,17 +777,28 @@ Each treatment category has a hazard ratio associated with it both for progressi
...
+  from scipy.stats import norm, lognorm
+  import random
+
+  def sample_paired_pfs_and_os_hazard_ratio_values(pfs_mid, 
+                                                   pfs_upper, 
+                                                   os_mid, 
+                                                   os_upper, 
+                                                   seed):
+    random.seed(seed)
+    pfs_value = lognorm(s=(log(pfs_upper) - log(pfs_mid)) / q_975_stdnorm, scale=pfs_mid).ppf(random.random())
```
yongqx, line 798: Are we supposed to set mu (scale) = log(hazard_mid) in a log-normal distribution?

aflaxman (approved):

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
alibow: I'm not quite sure what you mean. I think this is the correct way to sample from a lognormal distribution. Do you disagree?

This is code that we have used in the past that I simplified and it produces the expected values for the mean/median and 97.5/2.5 percentiles

```
from numpy import log
from scipy.stats import norm, lognorm

# median and 0.975-quantile of lognormal distribution for RR
median = 2.22
q_975 = 5.26

# 0.975-quantile of standard normal distribution (=1.96, approximately)
q_975_stdnorm = norm().ppf(0.975)

mu = log(median) # mean of normal distribution for log(RR)
sigma = (log(q_975) - mu) / q_975_stdnorm # std dev of normal distribution for log(RR)

# Frozen lognormal distribution for RR, representing uncertainty in our effect size
# (s is the shape parameter)
rr_distribution = lognorm(s=sigma, scale=median)
```

yongqx: My question is about the use of mean versus median to calculate standard deviation.

yongqx (approved):

PR #547, opened 2021-07-09
Updated PFS implementation, without data check

alibow:


aflaxman (approved):

docs/source/gbd2019_models/causes/neoplasms/multiple_myeloma/cancer_model.rst
```diff
@@ -279,24 +279,32 @@ data_dir = J:/Project/simulation_science/multiple_myeloma/data/cause_model_input
    * - incidence_MM_first_relapse
      - MM
      - MM_first_relapse
-     - data_dir/incidence First-line.csv
+     - data_dir/incidence First-line.csv - data_dir/mortality First-line.csv
```
yongqx, line 282: @alibow What i called `progression {X}-line.csv` is the result of `incidence First-line.csv - data_dir/mortality First-line.csv`. I have to rerun my code and make sure to use exact same random seed when producing the incidence and mortality hazard by draw, so that we can avoid a situation where incidence hazard < mortality hazard for a certain draw.

alibow: Yep! That's why I left a to-do for that. I just wanted to get the implementation details down in the meantime.

alibow: Ok, I think that it is not correct to do that subtraction outside of the sim because of how we need to apply the hazard ratios. So the incidence file should be the treatment duration rates and the mortality file should be the mortality rates and they should be subtracted within the sim after application of the hazard ratios to each measure separately.

alibow: Take a look at the changes made in the concept model to see what I mean about the hazard ratio application

mattkappel (approved):

docs/source/gbd2019_models/causes/neoplasms/multiple_myeloma/cancer_model.rst
alibow: Marking as resolved!

docs/source/gbd2019_models/causes/neoplasms/multiple_myeloma/cancer_model.rst
```diff
@@ -279,24 +279,28 @@ data_dir = J:/Project/simulation_science/multiple_myeloma/data/cause_model_input
...
```
yongqx, line 282: @mattkappel @alibow Since we have updated incidence and mortality csv files, do you want me to do the subtraction and output new files for relapse rate?

PR #560, opened 2021-07-20
CIFF SAM Concept Model: Add links to pages for model components

NathanielBlairStahn:
There's some superfluous whitespace changes, but the main thing is I added subsections under the concept model diagram that link out to the pages for the model components.

Note: I left out vitamin A deficiency because @Ninicorn said she's going to remove that.

### DIFF
--- a/docs/source/concept_models/vivarium_ciff_sam/concept_model.rst
+++ b/docs/source/concept_models/vivarium_ciff_sam/concept_model.rst
@@ -214,11 +214,11 @@ Scale up the coverage of treatment and prevention interventions to 90%.
 5.2.1 Population description
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-  - Location: Ethiopia
-  - Cohort type: Prospective open cohort of 0-5 years
-  - Size of largest starting population: 100,000 simulants
-  - Time span: Jan 1, 2022 to Dec 31, 2026
-  - Time step: 1 day
+- Location: Ethiopia
+- Cohort type: Prospective open cohort of 0-5 years
+- Size of largest starting population: 100,000 simulants
+- Time span: Jan 1, 2022 to Dec 31, 2026
+- Time step: 1 day
 
 
 .. _5.2.2:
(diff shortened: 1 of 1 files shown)


aflaxman (approved):

docs/source/concept_models/vivarium_ciff_sam/concept_model.rst
```diff
@@ -164,11 +164,45 @@ Scale up the coverage of treatment and prevention interventions to 90%.
...
+
+* :ref:`Child Stunting (GBD 2020) <2020_risk_exposure_child_stunting>`
+
+* Low Birthweight and Short Gestation (GBD 2019)
+
+5.1.4 Risk Effects Models
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* :ref:`Low Birthweight and Short Gestation Risk Effects (GBD 2019) <2019_risk_effect_lbwsg>`
+
```
Ninicorn, line 201: I'm wondering if we should add another section called 
1) "risk risk correlation" which is the LBWSG and stunting/wasting correlations and 
2) then another sections called "risk risk effect" which is the wasting to stunting arrow

Ninicorn (approved):
Hi! I added some suggestions to add two headings, one for 'risk risk correlation' and one for 'risk risk effect'

docs/source/concept_models/vivarium_ciff_sam/concept_model.rst
NathanielBlairStahn: @Ninicorn I'll go ahead and add a section for the risk-risk correlation.

I think risk-risk effects can just go in the 'risk effects' section  - the template was designed to be general enough to include both risk-cause and risk-risk. The wasting-stunting arrow should be one section in the "wasting risk effects" document (i.e. one of the listed outcomes would be stunting).

Also, I didn't know where to put the arrow for the vicious cycle, but I figured we could just decide once we actually have documentation.

Actually, since this PR has already been approved, I think I'll just merge it first and create a new PR with the new sections and new list items for the missing components, so that you and others will have a chance to review whether the new pieces look good.

PR #564, opened 2021-07-22
Added demographics and project overview sections

ndecleene:


docs/source/concept_models/vivarium_us_cvd/concept_model.rst
```diff
@@ -211,20 +220,32 @@ Variations on the above intervention increased support time or added meal replac
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-  - cohort type
-  - cohort length
-  - age and sex structure
-  - time step
-  - fertility
-  - stratifications 
+  - Cohort type: 
+  	- Prospective open cohort of adults aged 25 and above. New agents should age into the population at age_start
```
aflaxman, line 224: We will probably implement this as a _closed_ cohort, starting at age 3.  Then the youngest simulants will be turning 25 when the sim ends. Then you can analyze the 25+ population as if it were an open cohort.

aflaxman (approved):

docs/source/concept_models/vivarium_us_cvd/concept_model.rst
```diff
@@ -211,20 +220,32 @@ Variations on the above intervention increased support time or added meal replac
...
+  	- year_start: January 1, 2019
+  	- year_end: December 31, 2040
+  - Age and sex structure:
+  	- Sex: male/female/both
+  	- Age range: age_start=25, age_end=125
+  - Time step:
+  	- One month
+  - Fertility:
+  	- Not applicable
+  - Stratifications:
```
aflaxman, line 236: I don't understand this section

yongqx (approved):

docs/source/concept_models/vivarium_us_cvd/concept_model.rst
yongqx: I think it refers to simulants at different disease states/health conditions.

docs/source/concept_models/vivarium_us_cvd/concept_model.rst
catherineojohnson: starting at a range of 3-125
age range: 3 + duration of study - 125 + duration of study 
conscious of size of overall population vs size of initialized/observed group
doesn't matter that we don't have values for some inputs by age
modeling 3-25, but not observing; need to model correctly so that we end up with the right folks at 25

PR #565, opened 2021-07-22
MM sensitivity analysis runs and updated race assumptions

alibow:


yongqx (approved):

mattkappel (approved):

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
```diff
@@ -183,49 +182,34 @@ Among MM and RRMM patients, they expect to have:
...
+
+  1a. Flatiron Health based estimates
+
+  1b. Clinical trial based estimates
+
+2. Impact of race on survival outcomes (see the `5.3.2.2 Risk Factor Effects`_ section for more detail)
+
+  2a. Hazard ratios from the CoMMpass registry [Derman-et-al-2020]_
+
+  2b. Assumption of no impact of race on multiple myeloma survival outcomes
```
aflaxman, line 205: Is there some evidence out that that Black race has a protective effect on survival outcomes? If so, that would be an even more extreme sensitivity analysis that we could do instead of the no-impact version.

aflaxman (approved):

docs/source/concept_models/vivarium_sanofi_multiple_myeloma/concept_model.rst
alibow: Yes, there is some evidence of this, although it varies by study and does not always persist after adjustment for age.

Notably, and I can clarify this in the documentation, this assumption we are making is that there is no impact of race on MM survival outcomes *after adjustment for age.* So at the aggregate level in our simulation using this hazard ratio, black race will be protective for MM survival outcomes. 

I think this is the most appropriate value to run without doing a more systematic data extraction of the age-adjusted race HR estimates (there is also some evidence that it varies by age), but I can do that too!

alibow: @aflaxman, any updated thoughts based on this info?

aflaxman: Let's not make unnecessary work, and go with the alternative hypothesis that there is no impact of race on MM survival after adjusting for age.  Can you share an explanation of this with Jonathan, so he can confirm it is acceptable?

PR #571, opened 2021-08-02
updated concept model diagram

Ninicorn:


aflaxman (approved):

docs/source/concept_models/vivarium_ciff_sam/concept_model.rst
```diff
@@ -149,13 +149,68 @@ Historically, prevention research has primarily focused on stunting, and, as a r
...
+The baseline scenario will project GBD 2019 demographic and disease trends and GBD 2020 exposure trends out from 2022 to 2027 and coverage rates for all preventive and therapeutic interventions will be held constant across the 5 years of the microsimulation to simulate a business-as-usual treatment scenario.
 
 **Alternative scenario 1**
-Scale up the coverage of MAM management and MAM treatment interventions to 90%
+Scale up the 'effective-coverage' of GAM treatment from baseline level of effective-coverage to scenario level of effective-coverage. Those who are 'effectively covered' have a shorter duration of SAM and MAM episodes. In this first alternative scenario, a larger proportion of simulants will be effectivey covered. Keeping incidence of MAM (i2) and SAM (i1) the same as baseline, we expect the prevalence of SAM and MAM to decrease  with a shorter duration (prevalence ~ incidence x duration).    
+
+| Effective coverage = treatment coverage x treatment efficacy
```
alibow, line 157: What is the definition of treatment efficacy here? The proportion of treated individuals who respond to treatment? I think we should define this for clarity : )

docs/source/concept_models/vivarium_ciff_sam/concept_model.rst
```diff
@@ -149,13 +149,68 @@ Historically, prevention research has primarily focused on stunting, and, as a r
...
+    - Not effectively covered
+    - Reference
+    - Note
+  * - SAM baseline 
+    - 48.8% (95% CI: 37.4, 60.4)
+    - 70% (95% CI: 64, 76)
+    - 0.488 x 0.7 = 0.34
+    - 1 - 0.34 = 0.66
+    - [Isanaka_etal_2021]_ , [Zw_2020]_
+    - This is for SAM-OTP which is ~98% of SAM.
```
alibow, line 177: I'm not sure what SAM-OTP is or what the implication of this is

docs/source/concept_models/vivarium_ciff_sam/concept_model.rst
```diff
@@ -149,13 +149,68 @@ Historically, prevention research has primarily focused on stunting, and, as a r
...
+    - 1 - 0.675 = 0.325
+    - Sphere standards
+    - Sphere guideline for efficacy only 
+  * - MAM baseline 
+    - 48.8% (95% CI: 37.4, 60.4)
+    - 73.1% for RUSF
+    - 0.488 x 0.731 = 0.34
+    - 1 - 0.34 = 0.66
+    - [Ackatia_Armah_2015]_
+    - Baseline coverage of MAM needs to be updated, efficacy comes from trial and may be too optimistic
```
alibow, line 191: Why does the 48.8% coverage need updating? What is it based on now? Does it also need to be updated for SAM?

docs/source/concept_models/vivarium_ciff_sam/concept_model.rst
```diff
@@ -149,13 +149,68 @@ Historically, prevention research has primarily focused on stunting, and, as a r
...
-The baseline scenario will project GBD 2019 demographic and disease trends out from 2022 to 2027 and coverage rates for all preventive and therapeutic interventions will be held constant across the 5 years of the microsimulation to simulate a business-as-usual treatment scenario.
+The baseline scenario will project GBD 2019 demographic and disease trends and GBD 2020 exposure trends out from 2022 to 2027 and coverage rates for all preventive and therapeutic interventions will be held constant across the 5 years of the microsimulation to simulate a business-as-usual treatment scenario.
 
 **Alternative scenario 1**
-Scale up the coverage of MAM management and MAM treatment interventions to 90%
+Scale up the 'effective-coverage' of GAM treatment from baseline level of effective-coverage to scenario level of effective-coverage. Those who are 'effectively covered' have a shorter duration of SAM and MAM episodes. In this first alternative scenario, a larger proportion of simulants will be effectivey covered. Keeping incidence of MAM (i2) and SAM (i1) the same as baseline, we expect the prevalence of SAM and MAM to decrease  with a shorter duration (prevalence ~ incidence x duration).
```
alibow, line 155: I think Rajan has asked about this already, but just want to confirm that the general plan for the "scale-up" is to just have the coverage level be 90% in the alternative scenario rather than start at 48% and increase over time to 90%, correct? If so we should indicate that, I think.

docs/source/concept_models/vivarium_ciff_sam/concept_model.rst
```diff
@@ -167,7 +222,7 @@ Scale up the coverage of treatment and prevention interventions to 90%.
 5.1 Vivarium concept model diagram
 ----------------------------------
 
-.. image:: sam_concept_model_diagram.svg
+.. image:: am_concept_model_diagram.svg
```
alibow, line 225: Should we still have a correlation (dashed line) between wasting and stunting in this diagram even though we took the causal arrow out?

alibow (approved):
I left a bunch of comments to help clarify my understanding but otherwise looks good to me!

docs/source/concept_models/vivarium_ciff_sam/concept_model.rst
alibow: Got it, thanks!

docs/source/concept_models/vivarium_ciff_sam/concept_model.rst
alibow: Yeah... I think we'll have to think through if it makes more sense to correlate stunting with birth weight or stunting with with wasting directly. I can think through this with the rest of the risk risk correlation documentation!

alibow: But either way, the three of those risks will be correlated together somehow, which is implied through the current concept model, so I think its not super critical to add a dashed line between wasting and stunting, although we could to be explicit!

PR #577, opened 2021-08-10
Added details to SQ-LNS documentation 

Ninicorn:


Ninicorn:
are people see this table in this latest commit? 
image

aflaxman (approved):
You have two empty .rst files in this commit, is that intended?

Ninicorn:
> You have two empty .rst files in this commit, is that intended?

That's weird...it's not in my branch....

alibow (approved):

PR #576, opened 2021-08-10
LBWSG 2019 risk exposure stub

NathanielBlairStahn:
No content yet, just copied the risk exposure template to create a page stub.

PR #580, opened 2021-08-11
LBWSG 2019 exposure: Start Vivarium modeling section

NathanielBlairStahn:
Sequel to https://github.com/ihmeuw/vivarium_research/pull/576

Describe strategy for converting LBWSG categorical exposure distribution to a continuous exposure distribution, with links to sample code.

### DIFF
--- a/docs/source/gbd2019_models/risk_exposures/low_birthweight_short_gestation/index.rst
+++ b/docs/source/gbd2019_models/risk_exposures/low_birthweight_short_gestation/index.rst
@@ -73,10 +73,20 @@ continuous joint exposure distribution of birthweight and gestational age by
 assuming a uniform distribution of birthweights and gestational ages within each
 bin/category. In this way, each simulant can be assigned a continuously
 distributed birthweight and gestational age, which can then be easily mapped
-back to the appropriate risk category in GBD. Python code for achieving these
-transformations can be found in `Abie's notebook
-<https://github.com/ihmeuw/vivarium_data_analysis/blob/master/pre_processing/lbwsg/2019_03_19c_lbwsg_cat_to_continuous_abie.ipynb>`_
-in the Vivarium Data Analysis repo.
+back to the appropriate risk category in GBD. Example Python code for achieving
+these transformations can be found here:
+
+* `Abie's LBWSG cat-to-continuous notebook
+  <abie_lbwsg_cat_to_continuous_notebook_>`_ in the ``vivarium_data_analysis``
+  repo has a simple implementation demonstrating what we want.
+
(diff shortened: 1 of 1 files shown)


docs/source/gbd2019_models/risk_exposures/low_birthweight_short_gestation/index.rst
```diff
@@ -61,9 +61,44 @@ Risk Exposures Description in GBD
...
+
+.. _low_birth_weight_and_short_gestation.py: https://github.com/ihmeuw/vivarium_public_health/blob/main/src/vivarium_public_health/risks/implementations/low_birth_weight_and_short_gestation.py
+
+.. note::
 
-  Include here an overview of the Vivarium modeling section
+    The strategy of assuming a uniform distribution on each risk category is
+    likely biasing towards overestimating extreme birthweights or gestational
+    ages. For example, in the 0-500g category, most babies are probably pretty
+    close to 500g, not equally likely to be <1 gram versus 499-500 grams.
```
aflaxman, line 101:
```suggestion
    close to 500g, not equally likely to be <1 gram versus 499-500 grams.
    A limitation of this approach is therefore to overestimate the severity of the risk exposure distribution.  Since these extremely high risk categories are quite rare, we expect that the impact of this will be small.  In future work, we could use a more complex transformation to derive continuous values from the risk categories, but we should not pursue this until we have an application where it is clear that this limitation is a risk to the validity of our results.

```

aflaxman (approved):

docs/source/gbd2019_models/risk_exposures/low_birthweight_short_gestation/index.rst
NathanielBlairStahn: @aflaxman This overestimation of rarer events and underestimation of less rare events will be happening in _every_ category, not just the extremely high risk categories. I think the overall effect will be the same, i.e. we'll be overestimating the severity of the risk distribution, but I the effect is not limited to just the least probable categories.

NathanielBlairStahn:
@rmudambi Here's the first PR describing our strategy for LBWSG exposure.

docs/source/gbd2019_models/risk_exposures/low_birthweight_short_gestation/index.rst
NathanielBlairStahn: Although I guess when you're at interior points rather than near the boundary, the overestimation and underestimation would tend to cancel each other out as you cross over boundaries of categories, depending on how the RR is changing as you move from one category to an adjacent category.

Ninicorn (approved):

PR #587, opened 2021-08-13
CGF risk correlation modeling strategy

alibow:


docs/source/gbd2019_models/risk_risk_correlation/child_growth_risks/index.rst
```diff
@@ -104,21 +104,106 @@ Wasting and Stunting
...
+
+   Link to x-factor documentation for wasting model when it is developed
+
 Correlation
 ^^^^^^^^^^^^
 
+Initialization
+~~~~~~~~~~~~~~~
+
+When simulants are initialized into the early or late neonatal age groups or born into the simulation, they will be assigned a birthweight exposure value. When they transition into the postneonatal age group, they should be initialized into a particular state in the wasting model (e.g. TMREL, mild wasting, moderate wasting, or severe wasting) *based on their birthweight percentile within their overall location- and sex-specific birthweight distribution among the late neonatal age group.*
```
aflaxman, line 139: I thought simulants were initialized into a state of the wasting model at birth, and it has RR 1.0 before post-neonatal age

aflaxman (approved):

docs/source/gbd2019_models/risk_risk_correlation/child_growth_risks/index.rst
alibow: Yes, we were thinking that because we have the coefficient for one month instead of zero months that we would enforce it upon transition to the postneonatal age group since it won't have an effect on morbidity/mortality before that anyway, but now I am realizing that infants 0-6 months are eligible for wasting treatment, so we may want to continue to model wasting exposures in those groups too. @Ninicorn, do you agree?

This means that we would enforce the correlation (and causal impact) of birthweight on wasting *at birth* rather than one month and then track exposures accordingly. This is likely a conservative estimate, because I would expect the correlation to decrease over time (by the time wasting has an impact on outcomes) without enforcing a correlation with the x-factor, but I think I'll want to play around with it in a nano sim too to make sure.

Ninicorn: yes...we want to model wasting exposure from 1 month - 6months. Under 1 months don't have any burden due to wasting so we don't necessarily need to model it in those groups...but it messed up our model when we didn't have 'birth' prevalence. So yes agreed, we can enforce correlation at birth.

docs/source/gbd2019_models/risk_risk_correlation/child_growth_risks/index.rst
```diff
@@ -104,21 +104,106 @@ Wasting and Stunting
...
+
+This should be done according to the methodology described in the :ref:`risk-risk correlation proposal page <2017_risk_models>`.
+
+.. todo::
+
+   Add more detail on the methodology here/as implemented in the BEP project
+
```
Ninicorn, line 146: This is something for us to think about more, perhaps for discussion and how to justify the birthweight and wasting causal relationship for BEP since the trials didnt show an effect on wasting/stunting. It possibly is a 'no evidence for does not mean evidence against'.  We might say something like although the BEP trials (with their limitations and weaknesses etc) did not show an effect of BEP on wasting/stunting at 6 months, we believe there is strong evidence that  birthweight has a causal effect on stunting/wasting. Hence in our model, we modelled this relationship.

docs/source/gbd2019_models/risk_risk_correlation/child_growth_risks/index.rst
```diff
@@ -104,21 +104,106 @@ Wasting and Stunting
...
 Causation
 ^^^^^^^^^^^
 
+For interventions that impact birthweight, there should be a corresponding change in a child's wasting risk exposure attributable *if the impact of the intervention on wasting is not modeled directly*.
+
+Initialization
+~~~~~~~~~~~~~~~~
+
+For each gram increase in a simulant's birthweight due to a simulation intervention, the category 1 (severe wasting) and category 2 (moderate wasting) exposures used to determine the probability of initialization into those states should be reduced proportionately such that the total reduction in moderate and severe wasting is equal to 0.0115 / 200 = 0.00575. The exposure distribution of category 3 (mild wasting) should be increased by 0.0115 / 200 = 0.00575. The figure below demonstrates how to implement this change visually.
+
```
Ninicorn, line 167: Do we apply this reduction in wasting from birth?

Ninicorn (approved):

docs/source/gbd2019_models/risk_risk_correlation/child_growth_risks/index.rst
```diff
@@ -104,21 +104,106 @@ Wasting and Stunting
...
+- :ref:`GBD 2020 Wasting risk exposure <2020_risk_exposure_wasting_state_exposure>`
+
+- :ref:`GBD 2020 Stunting risk exposure <2020_risk_exposure_child_stunting>`
+
+Vivarium Modeling Strategy
+----------------------------
+
+The strategy for modeling risk-risk correlations related to child anthropometry in this document was developed for the needs of the :ref:`acute malnutrition treatment and prevention simulation <2019_concept_model_vivarium_ciff_sam>`. Different strategies may be more appropriate for different project needs and should be reavaluated when necessary.
+
```
plinds, line 123: I like this note!

docs/source/gbd2019_models/risk_risk_correlation/child_growth_risks/index.rst
alibow: Ok! Thanks, I will make this update.

docs/source/gbd2019_models/risk_risk_correlation/child_growth_risks/index.rst
alibow: Yes, added this in the latest commit!

PR #593, opened 2021-08-19
x-factor preliminary concept

Ninicorn:


aflaxman (approved):
Great, thanks for getting this written.  @rmudambi : does this make sense to you?  I think the description of the "risk effect model" is terse, but perhaps sufficient.

plinds (approved):

PR #597, opened 2021-08-25
got rid of x-factor effect on stunting and added equations and detail…

Ninicorn:
…s of the risk effect application

### DIFF
--- a/docs/source/concept_models/vivarium_ciff_sam/concept_model.rst
+++ b/docs/source/concept_models/vivarium_ciff_sam/concept_model.rst
@@ -340,9 +340,7 @@ SAM and MAM treatment baseline and treatment scale-up
 
 **Risk exposure**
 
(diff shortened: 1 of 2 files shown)


docs/source/concept_models/vivarium_ciff_sam/concept_model.rst
```diff
@@ -357,13 +355,25 @@ The proportion of reproductive age women underweight is 0.188 for GBD 2020.
...
 
+The risk effect (relative rate ratio) of incidence would be applied as such (breaking out the exposed vs non-exposed incidence from the exposure weighted overall incidence): 
+
+ - :math:`i_{x1} = i_{wasting} (1-PAF) \times rr_{x_{factor}}`
+ - :math:`i_{x0} = i_{wasting} \times (1-PAF)` 
+
+ where :math:`i_{wasting}` are the wasting transition incidences from the Markov calibration. And the PAF is calculated as
+
+ PAF = :math:`\frac{rr_{x_{factor}}-1}{rr_{x_{factor}}}`
+
```
alibow, line 373: Doesn't this need to be the mean RR among the population rather than just the RR?

alibow (changes requested):
I think we need to add the bar in the PAF calculation to indicate mean RR.

PR #620, opened 2021-09-02
[data] maternal anemia intervention

yongqx:


aflaxman (approved):

alibow:
Generally, the equations written out from that google doc are calculations of how the population mean hemoglobin level will change assuming 100% intervention coverage, but this might not be how we implement it for a vivarium simulation. The equations would be good validation criteria if we asked for population mean hemoglobin stratified by target population as a simulation output. However, it might be good in a subsequent PR to indicate how we would model the intervention effects at the simulant level for a vivarium simulation rather than at the population level as we did for the initial maternal anemia project. I am approving for now, but let me know if you have any questions on what I mean!

alibow (approved):

alibow:
Also, in the assumptions and limitations section, it would be good to note why we are taking an average of those two values for the effect size of IV iron (average of maximum and additive effect) so we don't get confused again.

alibow:
Something else to note for subsequent PRs is that the post-DFS intervention hemoglobin level should be used to determine eligibility for IV iron (hgb < 10 g/dL) rather than the pre-DFS intervention hemoglobin level (again, let me know if this is unclear)

yongqx:
> Generally, the equations written out from that google doc are calculations of how the population mean hemoglobin level will change assuming 100% intervention coverage, but this might not be how we implement it for a vivarium simulation. The equations would be good validation criteria if we asked for population mean hemoglobin stratified by target population as a simulation output. However, it might be good in a subsequent PR to indicate how we would model the intervention effects at the simulant level for a vivarium simulation rather than at the population level as we did for the initial maternal anemia project. I am approving for now, but let me know if you have any questions on what I mean!

Thanks for noting that!

PR #624, opened 2021-09-08
Bed net baseline coverage

alibow:


aflaxman (approved):

docs/source/intervention_models/insecticide_treated_nets/index.rst
```diff
@@ -62,58 +62,26 @@ Insecticide treated bed nets effectively reduce malaria incidence (and therefore
...
+    - 0.44 (SD: 0.021)
+    - Assume normal distribution of uncertainty. SD of uncertainty distribution calculated from :math:`SE = \sqrt{p * (1 - p) / n}`. Data from Ethiopia MIS 2015.
 
-.. todo::
-
-  Determine baseline coverage value to be used for Ethiopia, either by meta-analysis or by selection of large/representative recent study
+.. note::
 
-  NOTE: GBD estimates ITN coverage. Use this value instad after discussion with modelers.
+  Because we do not currently model malarious areas in our simulations (but rather generalize the malaria burden to the entire population), we will assume that the coverage proportion of ITNs in malarious areas is representative of the general population of pregnant women as well.
```
aflaxman, line 84: This will likely lead to a dramatic over-estimate of the impact of ITN scale-up in Ethiopia, so let's get an updated number for the national coverage at some point.  By eye, I would say something 10x smaller, see https://malariaatlas.org/research-project/metrics-of-insecticide-treated-nets-distribution/exceedance-surfaces/

aflaxman (approved):

docs/source/intervention_models/insecticide_treated_nets/index.rst
alibow: Part of the difference between the figure here and the one you linked is because use in the general population is lower than among pregnant women. The MIS survey I cited has use in the general population in malarious areas at 38% vs. 70% of pregnant women. Maybe we could derive a crosswalk value for pregnant women from MIS and then apply it to an updated national value?

alibow: Thinking through it again, though, I wonder if we could derive some fraction for the proportion of the population living in "malarious areas" and then only apply the intervention to that fraction? The effect size of the intervention comes from areas with high malaria burden, so we will need to make some decisions/assumptions about the burden of malaria in Ethiopia. I will follow up about this!

aflaxman: Good strategy.  It is possible that this fraction is already available as an ancillary input for the MAP team's work.

alibow: I made an update using just the MIS data and associated assumptions in this PR, but I will keep looking into the MAP work and finding the best data source for this on my todo list to revisit after finalizing the wasting treatment model first. https://github.com/ihmeuw/vivarium_research/pull/625

NathanielBlairStahn (approved):

PR #627, opened 2021-09-09
Wasting treatment assumptions from slack discussion

alibow:
Updated "how to apply treatment effects to simulants" section that reflects standard vivarium tools to follow.


docs/source/intervention_models/wasting_treatment/index.rst
```diff
@@ -632,9 +632,12 @@ Also note that since wasting and LBWSG are correlated, those with more severe wa
...
 
-- We are not applying a differential death rate to those effectively covered vs not effectively covered
-- We are generalizing across the whole country. There is likely to be a lot of heterogeneity within the country. 
-- We do not have the durations of untreated SAM and MAM for 0-6 age groups hence we are using the durations from 6-59 age groups. 
+#. We are not applying a differential death rate to those effectively covered vs not effectively covered
+#. We are generalizing across the whole country. There is likely to be a lot of heterogeneity within the country. 
+#. We do not have the durations of untreated SAM and MAM for 0-6 age groups hence we are using the durations from 6-59 age groups. 
+#. We assume that MAM treatment coverage is equal to SAM treatment coverage. Given that SAM treatment is more intensive than MAM treatment, we may underestimate MAM treatment coverage as a result of this assumption.
+#. We assume that MAM and SAM treatment effectivenesses are independent from one another.
+#. We assume that individual simulant's propensity to respond to wasting treatment is not auto-correlated.
```
NathanielBlairStahn, line 640: Will this change when we implement the X factor, or will that work by a different mechanism?

NathanielBlairStahn (approved):

docs/source/intervention_models/wasting_treatment/index.rst
```diff
@@ -468,7 +468,7 @@ where t is the period for which transition the is estimated (a year) eg. 365 day
     - 0-59 months old
     - 0.488 (95% CI:0.374-0.604)
     - normal
-    - baseline scenario value. assumed to be the same for SAM and MAM and across age groups.
+    - Baseline scenario value. Assumed to be the same for SAM and MAM and across age groups at the draw level.
```
aflaxman, line 471: The additional value of treating MAM (and not just SAM) is something I have heard interest in, so we might want/need to extend this to afford different levels of baseline/scaled-up coverage of SAM and MAM in the future.

docs/source/intervention_models/wasting_treatment/index.rst
```diff
@@ -587,9 +587,9 @@ Rather than apply an effect size associated with the intervention to a particula
...
 
-#. If simulant is uncovered according to #1: apply untreated transition rates (:math:`r_{SAM,ux}` and :math:`r_{MAM,ux}`). If simulant is covered according to #1: determine if the simulant is "effectively" covered using the wasting state-specific probability :math:`E_{SAM}` or :math:`E_{MAM}` corresponding to the wasting state that they occupy at the current time step.
+#. If simulant is uncovered according to #1: apply untreated transition rates (:math:`r_{SAM,ux}` and :math:`r_{MAM,ux}`). If simulant is covered according to #1: determine if the simulant is "effectively" covered using the wasting state-specific probability :math:`E_{SAM}` or :math:`E_{MAM}` corresponding to the wasting state that they occupy at the current time step. NOTE: distributions of :math:`E_{SAM}` and :math:`E_{MAM}` should not be correlated at the draw level. Simulants should have random propensity values for :math:`E_{SAM}` and :math:`E_{MAM}` such that it is possible for them to be effectively covered for one course of MAM treatment and not effectively covered for the next and vice versa.
```
aflaxman, line 592: Would a different way to say this be, "the probability an individual is cured is independent of past treatment/non-treatment history"?

aflaxman (approved):

docs/source/intervention_models/wasting_treatment/index.rst
NathanielBlairStahn: Are we saying it's independent of the _treatment history_, or independent of whether they've _responded_ to past treatments?

alibow: The probability that an individual responds to treatment is independent of whether they've responded to past treatments.

After talking it over with Rajan, this point has important implications for model structure. If we keep this assumption, we can use standard vivarium tools. If we assume instead that they have a static propensity for treatment response, we will need a custom model for SAM treatment. I am looking into the literature to see which assumption appears more valid right now. Do either of you have an opinion on this already?

docs/source/intervention_models/wasting_treatment/index.rst
alibow: This will not change with the x-factor. The x-factor only affects the incidence rates to worse wasting categories (which are not affected by treatment), not the recovery rates to better wasting categories (which are affected by treatment)

docs/source/intervention_models/wasting_treatment/index.rst
alibow: From this paper, important predictors of treatment response include 1) diarrhea (we can incorporate this into our vicious cycle model later), 2) oedema (which should be associated with SAM/MAM episode in most cases rather than chronic state), 3) antibiotics given with treatment. Other possible factors include distance from treatment center and vitamin A supplementation, although direct evidence was not available in this review.

Given this, I think we should stick with the independence assumption that allows for the use of standard tools. Do you agree? 

https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7339430/pdf/12887_2020_Article_2188.pdf

aflaxman: Agree, let's use an independence assumption.  I don't understand how this is more amenable to our standard tools, but I'll keep my eye out for the PR on the python repo and maybe that will make it clearer to me.

docs/source/intervention_models/wasting_treatment/index.rst
NathanielBlairStahn: Ok, good to know. Thanks for the clarification.

PR #642, opened 2021-09-22
Wasting treatment documentation clean up

alibow:
I added some clarification and limitations about the "duration" parameters currently used in our treatment model and added some additional related assumptions.

This PR does not change our modeling strategy or any data used in the model.

### DIFF
--- a/docs/source/intervention_models/wasting_treatment/index.rst
+++ b/docs/source/intervention_models/wasting_treatment/index.rst
@@ -486,7 +486,7 @@ where t is the period for which transition the is estimated (a year) eg. 365 day
     - 0-59 months old
     - 6.7(95% CI: 5.3-8.4)
     - lognormal
-    - :math:`k = 1 / duration_\text{overall SAM}`. :math:`duration_\text{overall SAM}` is the average duration of disease that is not conditional on survival of SAM (:math:`k` is *smaller* than it would be if it were conditional on those who recovered as is :math:`\text{time to recovery}_\text{treated SAM}`).
+    - See notes on uncertainty distribution and interpretation of this value below
     - [Isanaka_2021]_
   * - :math:`mortality_{SAM|a,s,l,y}`
     - GBD demographic group
@@ -513,6 +513,18 @@ where t is the period for which transition the is estimated (a year) eg. 365 day
   confidence interval on the :ref:`Statistical Distributions of Uncertainty
   <vivarium_best_practices_statistical_distributions>` page (note that the
   median of a lognormal distribution equals its geometric mean).
+
+.. note::
+
+  A note on the incidence correction factor, :math:`k`:
+
+    The incidence correction factor :math:`k` is = :math:`\frac{t}{\text{average duration of disease}}`, where :math:`t` is the period for which incidence is estimated (a year) eg. 365 days using days as the unit. Note that the average duration of disease in this equation is **not** conditional on recovery from SAM, unlike the :math:`\text{time to recovery}` parameters. Therefore, the average duration of disease is *lower* and :math:`k` is greater than it would be if it were conditional on recovery from SAM in the same manner as the :math:`\text{time to recovery}` parameters.
+
+    k = :math:`\frac{\text{number of incident cases}}{\text{number of prevalent cases}}` see [Isanaka_2021]_ for full proof and equations.
+
+    Number of incident cases = :math:`\frac{\text{annual programme admissions}}{\text{treatment coverage}}`
+
+    Note: [Isanaka_2021]_ relies on estimates of SAM treatment coverage to estimate the incidence correction factor/average duration of disease in Ethiopia. SAM treatment coverage is a parameter that is difficult to estimate and measurement may be prone to bias. SAM treatment coverage in the [Isanaka_2021]_ is estimated through standardized surveys [SQUEAC-SLEAC]_
 
 .. todo::
 
@@ -543,19 +555,6 @@ So...
   ...
 
   :math:`r_{SAM,ux} = \frac{k/365 - r_{SAM,tx} * C * E_{SAM} - mortality_{SAM|a,s,l,y}}{1 - C * E_{SAM}}`
-
-.. note::
-
-  A note on the incidence correction factor, :math:`k`:
-
-    The incidence correction factor :math:`k` is = :math:`\frac{t}{\text{average duration of disease}}`, where :math:`t` is the period for which incidence is estimated (a year) eg. 365 days using days as the unit. Note that the average duration of disease in this equation is **not** conditional on recovery from SAM, unlike the :math:`\text{time to recovery}` parameters. Therefore, the average duration of disease is *lower* and :math:`k` is greater than it would be if it were conditional on recovery from SAM in the same manner as the :math:`\text{time to recovery}` parameters.
-
-    k = :math:`\frac{\text{number of incident cases}}{\text{number of prevalent cases}}` see [Isanaka_2021]_ for full proof and equations.
-
-    Number of incident cases = :math:`\frac{\text{annual programme admissions}}{\text{treatment coverage}}`
-
-    Note: [Isanaka_2021]_ relies on estimates of SAM treatment coverage to estimate the incidence correction factor/average duration of disease in Ethiopia. SAM treatment coverage is a parameter that is difficult to estimate and measurement may be prone to bias. SAM treatment coverage in the [Isanaka_2021]_ is estimated through standardized surveys [SQUEAC-SLEAC]_
-
 
 Affected Outcomes
 +++++++++++++++++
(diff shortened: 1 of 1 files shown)


docs/source/intervention_models/wasting_treatment/index.rst
```diff
@@ -434,35 +434,35 @@ where t is the period for which transition the is estimated (a year) eg. 365 day
...
-    -
+    - NOTE: this study reports mean duration of stay in the inpatient therapeutic feeding center rather than time to recovery. It is currently being implemented as time to recovery. Of note, 85% of participants were successfully discharged in this study.
     - [Vygen_2013]_; Niger
-  * - :math:`duration_\text{treated SAM}`
+  * - :math:`\text{time to recovery}_\text{effectively treated SAM}`
     - 6-59 months old
     - 48.3
     - point value
-    -
+    - Studies censored participants who did not recovery in estimation of median time to recovery. 48.3 value is as a weighted average between seven studies reported by [Zw_2020tx]_. Random effects meta analysis would improve this estimate (see todo note below).
```
NathanielBlairStahn, line 447:
```suggestion
    - Studies censored participants who did not recover in estimation of median time to recovery. 48.3 value is a weighted average between seven studies reported by [Zw_2020tx]_. Random effects meta analysis would improve this estimate (see todo note below).
```

docs/source/intervention_models/wasting_treatment/index.rst
```diff
@@ -434,35 +434,35 @@ where t is the period for which transition the is estimated (a year) eg. 365 day
...
     - point value
     - IQR reported, but no uncertainty about median value. Used MUAC definition of malnutrition. Value specific to >6 month population, but assumed to generalize to <6 month population.
     - [James_2016]_; Ethiopia
-  * - :math:`duration_\text{treated MAM}`
+  * - :math:`\text{time to recovery}_\text{effectively treated MAM}`
     - 0-6 months old
     - 20.8
     - point value
-    -
+    - NOTE: this study reports mean duration of therapy without specifying if this duration is conditional on recovery. It is currently being implemented as time to recovery. Of note, 81% of participants were successfully discharged in this study.
```
NathanielBlairStahn, line 459: When you say it is currently being implemented as time to recovery, does that mean we are treating the duration as being conditional on recovery? What happened to the 19% of patients who were not successfully discharged?

docs/source/intervention_models/wasting_treatment/index.rst
```diff
@@ -434,35 +434,35 @@ where t is the period for which transition the is estimated (a year) eg. 365 day
...
     - [Woeltje_2020]_; Malawi
-  * - :math:`duration_\text{treated MAM}`
+  * - :math:`\text{time to recovery}_\text{effectively treated MAM}`
     - 6-59 months old
     - 41.3 (95% CI: 34.4, 49)
     - normal
-    -
+    - Censored participants who did not recovery in estimation of median time to recovery.
```
NathanielBlairStahn, line 465:
```suggestion
    - Censored participants who did not recover in estimation of median time to recovery.
```

docs/source/intervention_models/wasting_treatment/index.rst
alibow: Yes, correct. The other 19% either died, failed to respond to treatment, or were lost to follow-up/status unknown.

docs/source/intervention_models/wasting_treatment/index.rst
```diff
@@ -723,13 +708,13 @@ Validation and Verification Criteria
 
 .. math::
 
-  \frac{prevalence_\text{MAM|covered}}{prevalence_\text{MAM|uncovered}} ~ \frac{duration_\text{treated MAM} * E_\text{MAM} + duration_\text{untreated MAM} * (1 - E_{MAM})}{duration_\text{untreated MAM}}
+  \frac{prevalence_\text{MAM|covered}}{prevalence_\text{MAM|uncovered}} ~ \frac{\text{time to response}_\text{effectively treated MAM} * E_\text{MAM} + \text{time to response}_\text{untreated MAM} * (1 - E_{MAM})}{\text{time to response}_\text{untreated MAM}}
```
NathanielBlairStahn, line 711: In the previous equations you were using "time to recovery" instead of "time to response". Are these two things supposed to be the same?

docs/source/intervention_models/wasting_treatment/index.rst
```diff
@@ -723,13 +708,13 @@ Validation and Verification Criteria
...
 
 and
 
 .. math::
 
-  \frac{prevalence_\text{SAM|covered}}{prevalence_\text{SAM|uncovered}} ~ \frac{duration_\text{treated SAM} * E_{SAM} + duration_\text{untreated SAM} * (1 - E_{SAM})}{duration_\text{untreated SAM}}
+  \frac{prevalence_\text{SAM|covered}}{prevalence_\text{SAM|uncovered}} ~ \frac{\text{time to response}_\text{effectively treated SAM} * E_{SAM} + \text{time to response}_\text{untreated SAM} * (1 - E_{SAM})}{\text{time to response}_\text{untreated SAM}}
```
NathanielBlairStahn, line 717: Same comment about  "time to recovery" vs. "time to response".

docs/source/intervention_models/wasting_treatment/index.rst
```diff
@@ -709,6 +688,12 @@ Assumptions and Limitations
 #. We assume that MAM treatment coverage is equal to SAM treatment coverage. Given that SAM treatment is more intensive than MAM treatment, we may underestimate MAM treatment coverage as a result of this assumption.
 #. We assume that MAM and SAM treatment effectivenesses are independent from one another.
 #. We assume that individual simulant's propensity to respond to wasting treatment is independent of their previous response/non-response to treatment. According to [Zw_2020tx]_, SAM treatment response rates are associated with diarrhea, oedema, and use of antibiotics in the treament course in Ethiopia. Additionally, vitamin A supplementation and distance from the treatment center may be associated with SAM treatment response rates, although direct evidence was not provided [Zw_2020tx]_. We chose to make this assumption given the non-deterministic nature of these factors.
+#. We assume that MAM and SAM treatment non-responders have the same time to recovery as untreated MAM and SAM cases.
```
NathanielBlairStahn, line 691: What does this mean exactly? It sounds self-contradictory at first glance, so I don't think I understand...

docs/source/intervention_models/wasting_treatment/index.rst
NathanielBlairStahn: Ok thanks. What is the implication of our interpretation of the duration as time to recovery, given that 81% of patients were successfully discharged?

NathanielBlairStahn: Perhaps to be more explicit about our assumptions, you could say something like:
> NOTE: this study reports mean duration of therapy without specifying if this duration is conditional on recovery. We are currently interpreting the duration as being conditional on recovery and therefore using it as the time to recovery in our model.

And then say something about the 81% successfully discharged, and perhaps mention in the assumptions and limitations what the implications would be if our assumption about it being time-to-recovery were wrong.

NathanielBlairStahn: I think all the same comments apply to `\text{time to recovery}_\text{effectively treated SAM}` above, as it looks like you are making the same assumption there when you say:
> NOTE: this study reports mean duration of stay in the inpatient therapeutic feeding center rather than time to recovery. It is currently being implemented as time to recovery. Of note, 85% of participants were successfully discharged in this study.

docs/source/intervention_models/wasting_treatment/index.rst
```diff
@@ -434,35 +434,35 @@ where t is the period for which transition the is estimated (a year) eg. 365 day
     - Distribution
     - Note
     - Source
-  * - :math:`duration_\text{treated SAM}`
-    - 0-6 months old
+  * - :math:`\text{time to recovery}_\text{effectively treated SAM}`
+    - 0-6 months olds
```
NathanielBlairStahn, line 438:
```suggestion
    - 0-6 months old
```
The s at the end is inconsistent with the other table entries

aflaxman (approved):

docs/source/intervention_models/wasting_treatment/index.rst
NathanielBlairStahn: Are you saying that the remission rate (SAM -> MAM or MAM -> MILD) of non-responders is the same as the remission rate of untreated cases? I think the wording "time to recovery" was throwing me off because it sounds like they are recovering (e.g. SAM -> MILD) even though they don't respond to treatment.

docs/source/intervention_models/wasting_treatment/index.rst
alibow: Yes, those are meant to be the same. Updating now, thanks for catching

docs/source/intervention_models/wasting_treatment/index.rst
alibow: Thanks! Updating.

docs/source/intervention_models/wasting_treatment/index.rst
alibow: So the studies that investigated time to recovery defined non-responders as treated children who had not recovered (no longer wasted) within two months of treatment. They censored non-responders along with those who died or were lost to follow up in the estimation of time to recovery. Since we don't have information on transition rates among this group, we are treating them as if they were not treated at all (untreated SAM to MAM recovery rate or untreated MAM to mild recovery rate, depending on which state they were in initially), which is likely not actually the case. I will expand upon this in the limitation.

docs/source/intervention_models/wasting_treatment/index.rst
alibow: Yeah, the implications of the 81% successfully discharged is that the majority of patients were successfully discharged (recovered) and the estimate may be skewed by the other 19% who either died or were lost to follow up before recovery, which may decrease the estimate of length of stay. The lower the proportion of patients who were successfully discharged, the greater the potential for bias. 

These studies are in the 0-6 month olds which we are not modeling right now and I plan to address all of these studies in more detail when revisiting the modeling strategy for this age group, but I will add these suggestions now too. Thanks

NathanielBlairStahn (approved):

PR #646, opened 2021-09-24
Exclude infants <6 months from wasting exposure transitions

alibow:


aflaxman (approved):

docs/source/gbd2020_models/risk_attributable_causes/wasting/index.rst
```diff
@@ -532,6 +532,28 @@ Deriving wasting transition probabilities
...
+
+  The magnitude of the bias introduced by this modeling strategy due to the two points above should be investigated upon implementation to determine if different modeling strategies are necessary. This should be done by comparing the wasting exposure and wasting-affected outcomes in the simulation output to the GBD inputs by age group.
+
+  NOTE:
+
+  - The modeling decision not to model wasting transitions among simulants less than six months of age is due to the reliance of the wasting model transition rates on the wasting treatment model and the lack of data to inform treatment-related transition rates among this age group. Note that a sensitivity analysis scenario that includes infants less than six months of age in the treatment model may be performed in the future.
+
+  - The decision to initialize simulants with the birth prevalence of the 6-11 month age group with all wasting transition rates equal to zero before six months of age rather than have a static wasting exposure model among simulants less than six months of age (similar to the stunting exposure model) was chosen due to the ease of engineering implementation and will be revisited if it introduces a large bias.
```
rmudambi, line 555: Shouldn't this be the prevalence in the 1-6 months age group? Since that's the prevalence we'd expect to be reflected in a vivarium sim in which we were modelling wasting in that age group.

docs/source/gbd2020_models/risk_attributable_causes/wasting/index.rst
```diff
@@ -886,7 +907,7 @@ Data Description Tables
    * - TMREL, MOD, MAM, SAM
      - birth prevalence
      - :math:`prevalence_{240_{cat-1-4}}`
-     - Use prevalence of age_group_id = 2 (early neonatal)
+     - Use prevalence of age_group_id = 389 (six to 11 months)
```
rmudambi, line 910: Ditto, should be 1-6 months?

rmudambi (changes requested):

docs/source/gbd2020_models/risk_attributable_causes/wasting/index.rst
alibow: @rmudambi Hmm, I chose the 6-11 month age group because that is the age group where we will be measuring our intervention impact, so we want to prioritize getting exposure correct in this age group. The exposures between these groups are very similar, so I don't think that it will cause an especially large bias in the mortality rate among 1-5 month olds. Notably, exposure is slightly greater among the 6-11 month age group than the 1-5 month age group. Given that we know our strategy will bias exposure lower (upon the BW/wasting correlation implementation), selecting the higher exposure group will offset this bias more than selecting the lower exposure group.

Does that address your concern?

rmudambi (approved):

PR #652, opened 2021-09-27
make page specific to bw and wasting, update to reflect six month age

alibow:


aflaxman (approved):

docs/source/gbd2019_models/risk_risk_correlation/birthweight_wasting/index.rst
```diff
@@ -114,111 +92,88 @@ Links to documentation for relevant risk exposure pages include:
...
 
-.. note::
+2. Assess the percentile of a simulant's birthweight exposure value (after application of all birthweight-affecting variables in the baseline scenario but before the application of *new* intervention coverage in the alternative scenario) within the overall sex, location, and year-specific birthweight exposure distribution *at birth*.
 
-   The correlation coefficient of one month of age should be used for all simulants upon birth as well as those initialized into the early neonatal or late neonatal age groups because we do not have correlation coefficients specific to more detailed age ranges. 
+3. Assign the simulant a wasting propensity score that is correlated to their birthweight percentile according to the spearman correlation coefficient of birthweight and wasting at six months of age (mean: 0.3083092, standard deviation: 0.2225747, normal distribution of uncertainty). This should be done according to the methodology described in the :ref:`risk-risk correlation proposal page <2017_risk_models>`.
```
NathanielBlairStahn, line 118: Just a general comment: It looks like "propensity score" has a technical meaning that I think is different from the way we use the term "propensity" in our models (though it's possible they're related -- I can't really tell on a cursory reading -- see articles below), so we might want to be a bit more careful with our terminology.

https://en.wikipedia.org/wiki/Propensity_score_matching
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3144483/

NathanielBlairStahn (approved):

aflaxman:
Agree--- I was thinking of “propensity scores” when we named the propensity attribute, but there is no formal connection.  It is more like the common meaning of “propensity”, an inclination or tendency towards the risk factor exposure.  Let’s avoid saying “propensity score” and keep using “propensity” for this.

PR #656, opened 2021-09-29
BW wasting correlation doc clean up

alibow:


docs/source/gbd2019_models/risk_risk_correlation/birthweight_wasting/index.rst
```diff
@@ -130,11 +130,7 @@ When simulants are initialized into the early or late neonatal age groups or bor
 Transitions
 ^^^^^^^^^^^^^
 
-.. todo::
-
-   Add detail on how to correlate wasting x-factor propensity to birthweight exposure
-
-   This will be done through the correlation between maternal BMI (proxy measure for x-factor risk exposure) and birthweight
+We will not model a correlation between birthweight exposure and wasting incidence rates directly. However, these measures will be positively correlated in the :ref:`acute malnutrition simulation <2019_concept_model_vivarium_ciff_sam>` as a result of the correlation between birthweight and maternal BMI risk exposures (:ref:`described here <2019_risk_correlation_maternal_bmi_birthweight>`), the perfect correlation between the x-factor and maternal BMI risk expousre, and the impact of the x-factor on wasting incidence rates (note: x-factor documentation can be found in the :ref:`acute malnutrition concept model document <2019_concept_model_vivarium_ciff_sam>`).
```
aflaxman, line 133: Does this imply that a scale up in BEP or ITN will not have any ripple effect reduction in wasting?

aflaxman (approved):

docs/source/gbd2019_models/risk_risk_correlation/birthweight_wasting/index.rst
alibow: That is not what this section means. This section means just that we will not directly correlate birthweight with wasting transition rates like we do for birthweight and wasting exposure initialization, but that they will be correlated through their correlation through maternal BMI/x-factor.

The current strategy for the birthweight ripple effect is to:
- Apply a decrease in the probability of being initialized into SAM or MAM states at 6 months of age associated with an increase in birthweight from BEP and ITN interventions (in the "causation" section of this document).
- Do not change x-factor risk exposure or risk effect magnitude associated with an increase in birthweight.

aflaxman:
Makes sense, thanks for reassuring me!

PR #662, opened 2021-10-01
CIFF risk correlation structure figure and description

alibow:


yongqx (approved):

aflaxman (approved):
In my mind, there is a clear direction for an arrow corresponding to each of the dashed lines.  What do you think of using arrows to make it a causal DAG instead of dashed lines which I am interpreting to mean "correlation" and not "causation".

PR #666, opened 2021-10-12
Use x-factor effect of 1.3 until further notice

alibow:


plinds (approved):

docs/source/concept_models/vivarium_ciff_sam/concept_model.rst
```diff
@@ -424,6 +424,10 @@ We do not have direct evidence or data for the risk effect of the x-factor *prox
    :widths: 30, 10, 10, 10, 10
    :header-rows: 1
 
+.. note::
+
+  For model runs following validation of model #4.5, use x-factor risk effect = 1.3 until consensus is reached on x-factor effect magnitude or the model is ready for simulation results for the full range of sensitivity analysis values.
```
aflaxman, line 429: Strategy for data-driven calibration of x-factor effect size: (1) identify one or more measurable quantities that vary when effect size varies (e.g. average number of SAM treatments per unique children with SAM during a single year period); (2) compare measurements of quantity/quantities to value(s) estimated by simulation for a grid of x-factor effect size values; and choose effect size that makes simulation produce quantity/quantities most similar to empirical measurement(s).

NathanielBlairStahn: Would it be worth also running the model with an X-factor effect of 1.0, to compare with previous models and to be able to tell if when we add model changes, the X-factor is having any effect?

PR #677, opened 2021-11-01
splitting BP ramp into two parts

catherineojohnson:


ndecleene (approved):

mattkappel (approved):
Nice! This makes sense to me. We need to track simulants prescription of each class, and some set dosage (half, full, double, IIRC). We'll need ways of combining effects.

PR #684, opened 2021-11-09
fixing SBP exposure level label

catherineojohnson:


catherineojohnson:
tried to change label to match other risk exposure documentation, but that broke things. Can update intervention documentation to use existing label.

PR #692, opened 2021-11-29
Feature/rf neb exposure model

yongqx:


alibow (approved):

alibow:
I think this looks good, with just a couple of notes. This looks like the risk exposure page, but also has some information on the risk effects. Maybe we could consider moving the information on relative risks/PAFs/etc. to a risk effects page rather than the risk exposure to be consistent with how we have structured other risk factor documentation.

Also, I think writing a note about where all of this information was obtained to the reference section would be good so we don't get confused that it references the 2019 appendix and can pay attention if any of the information changes upon publication of GBD 2020.

yongqx:
> I think this looks good, with just a couple of notes. This looks like the risk exposure page, but also has some information on the risk effects. Maybe we could consider moving the information on relative risks/PAFs/etc. to a risk effects page rather than the risk exposure to be consistent with how we have structured other risk factor documentation.
> 
> Also, I think writing a note about where all of this information was obtained to the reference section would be good so we don't get confused that it references the 2019 appendix and can pay attention if any of the information changes upon publication of GBD 2020.

Thanks for your suggestion. I can create a risk effect folder under GBD 2020 model and move the pieces of RRs/PFAs to desired location. I will add notes to distinguish 2020 methods from published 2019 appendix.

aflaxman (approved):

PR #708, opened 2021-12-15
cause model templates for IV iron project

alibow:


aflaxman (approved):
Should we perhaps go straight to GBD 2020 for these models?

plinds (approved):

alibow:
> Should we perhaps go straight to GBD 2020 for these models?

After looking through the current GBD 2020 results, it looks like there isn't a "best" version of the results yet (I don't think they've gone through codcorrect/como yet) and there aren't results for the income level location IDs. GBD 2019 has results for the income level location IDs that we could start using.

Do you think we should start with the GBD 2020 results even though they're not finalized and wait for the numbers to come through or start with GBD 2019 and then update when 2020 is ready?

aflaxman:
> Do you think we should start with the GBD 2020 results even though they're not finalized and wait for the numbers to come through or start with GBD 2019 and then update when 2020 is ready?

Let's start with 2019 and update when 2020 is ready.  Do you know if there were any big changes between 19 and 20?

alibow:
> > Do you think we should start with the GBD 2020 results even though they're not finalized and wait for the numbers to come through or start with GBD 2019 and then update when 2020 is ready?
> 
> Let's start with 2019 and update when 2020 is ready. Do you know if there were any big changes between 19 and 20?

It looks like the only major update was the systematic review of input data that they do annually.

PR #714, opened 2021-12-16
Maternal hemorrhage incidence vivarium modeling strategy

alibow:


docs/source/gbd2019_models/causes/maternal_hemorrhage_incidence/index.rst
```diff
@@ -109,220 +111,99 @@ on the ages and sexes to which the cause applies.
...
 Assumptions and Limitations
 +++++++++++++++++++++++++++
 
-.. todo::
-
-  Describe the clinical and mathematical assumptions made for this cause model,
-  and the limitations these assumptions impose on the applicability of the
-  model.
+- We will assume that all causes of maternal hemorrhage occur at the time of birth.
+- Our model of maternal hemorrhage incidence will be measured as cases that do not die due to maternal hemorrhage.
```
aflaxman, line 127: wouldn't it also count as an incident cases if it *did* lead to death?

aflaxman (approved):

docs/source/gbd2019_models/causes/maternal_hemorrhage_incidence/index.rst
alibow: Yes, that's why in the measure of incidence for this cause model it's measured as (incidence - mortality). However, it is possible that someone had a case of maternal hemorrhage and then died of another maternal disorder and therefore would not experience the disability of living with the hemoglobin deficit due to hemorrhage because they died of something else.

PR #722, opened 2021-12-20
Anemia threshold clarification

alibow:


docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/anemia_impairment/index.rst
```diff
@@ -78,6 +78,8 @@ Once the hemoglobin concentration distribution is estimated, the anemia envelope
 
 .. note::
 
+  GBD uses the hemoglobin thresholds specific to *Males and Females 5-14 years* for pregnant females in this age group (as opposed to the pregnancy-specific thresholds).
```
aflaxman, line 81: Will this be a problem for us?

aflaxman (approved):

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/anemia_impairment/index.rst
alibow: I proposed that we follow this same protocol to be consistent with GBD, but we could also use the pregnancy-specific rather than age-specific thresholds for pregnant 10-14 year olds instead. The fertility rate among this group is super small, so it shouldn't make a huge difference in the sim at the population level either way. Also, I have left todo notes that we may restrict fertility to 15-50 year olds (WHO definition) instead of 10-55 year olds (GBD definition) pending BMGF input (they asked us to do this for our previous multiplication model).

plinds (approved):

PR #725, opened 2021-12-20
add note about ensemble distribution sampling

alibow:


plinds (approved):

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
```diff
@@ -174,69 +174,16 @@ Data Description Tables
...
-    ##parameters
-    params_gamma = gamma_mv2p(mn, vr)
-    params_mgumbel = mgumbel_mv2p(mn, vr)
-
-    ##weighting
-    prev = sum(
-      w[1] * pgamma(x, params_gamma$shape, params_gamma$rate), 
-      w[2] * pmgumbel(x, params_mgumbel$alpha, params_mgumbel$scale, lower.tail=T))
-    prev}
+  Determine if this code follows the protocol to appropriately sample from an ensemble distribution, as identified by Nathaniel.
```
NathanielBlairStahn, line 181: Yes, Kjell implemented the corrected sampling algorithm I suggested, though the notebook also has the original sampling method, so it probably won't be clear to the reader what they're looking for. I think it would be good to copy the relevant functions into the documentation (and perhaps give them better names):
```Python
import numpy as np
import pandas as pd
import scipy

def _gamma_ppf(propensity, mean, sd):
"""Returns the quantile of a gamma distribution with the specified mean and
standard deviation corresponding to the specified quantile rank (propensity).
"""
    shape = (mean / sd)**2
    scale = sd**2 / mean
    return scipy.stats.gamma(a=shape, scale=scale).ppf(propensity)

def _mirrored_gumbel_ppf(propensity, mean, sd):
"""Returns the quantile of a mirrored Gumbel distribution with the specified mean
and standard deviation corresponding to the specified quantile rank (propensity).
"""
    x_max = 220 # data_values.HEMOGLOBIN_DISTRIBUTION.EXPOSURE_MAX
    _alpha = x_max - mean - (sd * np.euler_gamma * np.sqrt(6) / np.pi)
    scale = sd * np.sqrt(6) / np.pi
    tmp = _alpha + (scale*np.euler_gamma)
    alpha = _alpha + x_max - (2*tmp)
    return scipy.stats.gumbel_r(alpha, scale=scale).ppf(propensity)

# Called `viv_calc_iron_nbs` in Kjell's notebook
def sample_from_hemoglobin_distribution(prop_dist, propensity, exposure_parameters):
"""
Returns a sample from an ensemble distribution with the specified mean and
standard deviation (stored in `exposure_parameters`) that is 40% Gamma and
60% mirrored Gumbel. The sampled value is a function of the two propensities
`prop_dist` (used to choose whether to sample from the Gamma distribution or
the mirrored Gumbel distribution), and `propensity` (used as the quantile rank
for the selected distribution).
"""
    #propensity = clip(propensity)
    exposure_data = exposure_parameters
    mean = exposure_data['mean']
    sd = exposure_data['sd']

    gamma = prop_dist < 0.4
    gumbel = ~gamma
    ret_val = pd.Series(index=prop_dist.index, name='value')
    ret_val.loc[gamma] = _gamma_ppf(propensity.loc[gamma], mean, sd)
    ret_val.loc[gumbel] = _mirrored_gumbel_ppf(propensity.loc[gumbel], mean, sd)
    return ret_val
```
Note that the corrected version of ensemble sampling has already been incorporated into Vivarium's `risk_distribution` module in the [.ppf method of the EnsembleDistribution class](https://github.com/ihmeuw/risk_distributions/blob/main/src/risk_distributions/risk_distributions.py#L495) (which is misleadingly named because this sampling method is not the same thing as the quantile function...).

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
```diff
@@ -174,69 +174,16 @@ Data Description Tables
...
-.. note::
-
-  This code has been translated to python in a notebook hosted `here <https://github.com/ihmeuw/vivarium_gates_lsff/blob/main/tests/lsff_iron_exposure.ipynb>`__.
+**Code to sample from the hemoglobin exposure distribution in python has been developed and hosted in a notebook hosted** `here <https://github.com/ihmeuw/vivarium_gates_lsff/blob/main/tests/lsff_iron_exposure.ipynb>`__.
 
 .. todo::
 
-	Integrate python code from above notebook into this document.
-
-.. code-block:: R
```
NathanielBlairStahn, line 185: I think it would be better to leave the R code in the documentation for reference, since this is the original source we are trying to match, and who knows whether the links to external code will break at some point.

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
```diff
@@ -172,15 +147,13 @@ Data Description Tables
     - 0.6
     - 
 
-Below is R code written to randomly sample hemoglobin concentration values from the hemoglobin distribution parameters and constants defined in the tables above. Additionally, the code block contains functions that will evaluate the proportion of the distribution below a given threshold. This code was adapted from the GBD stash code found `here <https://stash.ihme.washington.edu/projects/MNCH/repos/anemia/browse/model/envelope>`__, specifically the *DistList_mv2p.R* and *fit_ensemblemv2p_parallel.R* files.
+**Code to sample from the hemoglobin exposure distribution in python has been developed and hosted in a notebook hosted** `here <https://github.com/ihmeuw/vivarium_gates_lsff/blob/main/tests/lsff_iron_exposure.ipynb>`__. The functions that should be used for sampling from the hemoglobin distribution include :code:`mirrored_gumbel_ppf` (*not* mirroed_gumbel_ppf_2017, which is an implementation that did not validate to GBD) and :code:`viv_calc_iron_nbs` (*not* viv_calc_iron or viv_calc_iron_2017, which implement the incorrect strategy for :ref:`sampling from an ensemble distribution <vivarium_best_practices_ensemble_distributions>`).
```
NathanielBlairStahn, line 150: Why not also include the code in the documentation (i.e. the three functions I suggested), so the reader doesn't have to click on the notebook and search for the relevant pieces unless they want more details?

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
```diff
@@ -172,15 +147,63 @@ Data Description Tables
...
+      gumbel = ~gamma
+      ret_val = pd.Series(index=prop_dist.index, name='value')
+      ret_val.loc[gamma] = _gamma_ppf(propensity.loc[gamma], mean, sd)
+      ret_val.loc[gumbel] = _mirrored_gumbel_ppf(propensity.loc[gumbel], mean, sd)
+      return ret_val
 
 .. note::
 
-  This code has been translated to python in a notebook hosted `here <https://github.com/ihmeuw/vivarium_gates_lsff/blob/main/tests/lsff_iron_exposure.ipynb>`__.
+	While not explicitly enforced by the code above, all hemoglobin values should be non-zero positive numbers. The probability of sampling a negative value is small, but if it occurs, the value should be resampled until it is a positive number or clipped to a value of 1.
```
aflaxman, line 201: If we clip negatives to 1.0, should we also clip positives that are less than 1.0 to this minimum value?

aflaxman (approved):

docs/source/gbd2019_models/hemoglobin_anemia_and_iron_deficiency/hemoglobin/index.rst
alibow: I just found some citations that claim the lowest ever observed hemoglobin value in a patient that survived blood loss was 6 g/L and the lowest observed value in a stable patient not on critical care support was 14 (!!). I updated the strategy to clip anything less than 6 back to 6.

PR #727, opened 2021-12-21
iv iron additional assumptions and limitations

alibow:


plinds (approved):

docs/source/concept_models/vivarium_iv_iron/concept_model.rst
```diff
@@ -463,6 +463,13 @@ For model version 1:
 6.0 Limitations
 +++++++++++++++
 
+In addition to the assumptions and limitations listed in each of the pages included in the `4.1 Vivarium concept model diagram components`_ section, our simulation is also subject to the following assumptions and limitations:
+
+#. Our simulation is run at the regional level and does not consider sub-regional heterogeneity.
+#. We do not model a distinction between iron-deficiency anemia and other types of anemia. Therefore, we may overestimate the number of individuals whose hemoglobin levels respond to our interventions, which may also vary by modeled location.
```
aflaxman, line 469: Do you have a way to roughly calculate by how much might we overestimate the impact because of this?

aflaxman (approved):

docs/source/concept_models/vivarium_iv_iron/concept_model.rst
alibow: We can calculate the proportion of iron-responsive women of reproductive age from GBD. However, the proportion of iron responsive WRA is really most relevant in the measurement of the effect size of the intervention. The optimistic target profile of IV iron from BMGF is that it will increase hemoglobin concentration by 10 g/dL, on average, but that could mean that it doesn't increase hemoglobin at all in a proportion of the population that is not iron responsive and increases it more than 10 in the iron responsive population. By this logic, the intervention may have a larger average effect in South Asia than in SSA due to fewer hemoglobinopathies in South Asia, but this isn't reflected in the optimistic target profile. I've asked Laura for more information on the effect size data sources for the oral iron intervention on hemoglobin so we can see what specific populations they were measured in (South Asia vs. SSA, excluded subjects with hemoglobinopathies or not, etc.) to inform our assumptions regarding those effects, but we haven't heard back yet.

PR #737, opened 2021-12-30
CIFF V and V table update

alibow:


aflaxman (approved):
Should we note the 5.1.0 and 5.1.1 attempts in this table also?

plinds (approved):

aflaxman (approved):
👍

PR #741, opened 2022-01-18
IV iron model 0 validation tracking

alibow:


albrja (approved):

aflaxman (approved):

mattkappel:
```
population = import_data_frame('population', paths, locations)
population

# might expect to see year stratification here?
# not sure how we got to 500,000 population size... was expecting 100,000
```
I think we could add year stratification here, it's not the default apparently. The population was off because we didn't account for actual population size being `population_size * random_seed_count` in `psimulate`.  We'll need to fix `random_seed_count` and `input_draw_count`, and then can set `population_size` to match expected population.

mattkappel (approved):

alibow:
> ```
> population = import_data_frame('population', paths, locations)
> population
> 
> # might expect to see year stratification here?
> # not sure how we got to 500,000 population size... was expecting 100,000
> ```
> 
> I think we could add year stratification here, it's not the default apparently. The population was off because we didn't account for actual population size being `population_size * random_seed_count` in `psimulate`. We'll need to fix `random_seed_count` and `input_draw_count`, and then can set `population_size` to match expected population.

Updated to clarify/specify :)

PR #752, opened 2022-01-26
update diarrheal diseases remission rate

alibow:


plinds (approved):

docs/source/gbd2019_models/causes/diarrheal_diseases/index.rst
```diff
@@ -224,9 +224,12 @@ Data Description
 	* - r
 	  - I
 	  - S
-	  - remission_rate_m1181
-	  - Already a rate within with-condition population
+	  - (incidence_rate_c302 - duration_c302 / 365 * (ACMR - csmr_c302 + emr_c302)) / prevalence_c302
```
aflaxman, line 227: I don't understand how this can work --- why does incidence rate enter into your calculation of remission rate at all?  How can you add emr to a population-level cause-deleted mortality rate?  Could we simplify this by making the I->S transition duration based?

aflaxman (approved):

docs/source/gbd2019_models/causes/diarrheal_diseases/index.rst
alibow: So it's assuming that all incident cases will remit unless they die while infected. Note that the incidence rate is the "total population" incidence rate, so incident cases / total population person time. The incidence rate is there because it is the rate at which remitted cases would occur if no one died (remitted cases = incident cases, so incidence_rate = remission_rate). But since we expect people to die while they are infected we will subtract the mortality rate of those infected with diarrheal diseases (ACMR - csmr_c302 + emr_c302), scaled to the average duration of the episode (duration_c302 /365). Then we transform back to the relevant population (those infected with diarrheal diseases) by dividing by diarrheal disease prevalence.

We could make this a duration-based transition rate, but I thought it might be preferable to represent the spread of individual variation within the population for the duration of a diarrheal episode rather than assign each individual the population-mean duration.

aflaxman: I'm suspicious of this approach, but since we have a solid V&V process in place, we can give it a try your way, and have the duration-based alternative in our back pocket, just in case.

alibow: Ok. This is how I hacked it to get the calculation of the vicious cycle relative risks to work because having more remitted cases than incident cases (which is the case from GBD) caused negative rates. The hack seemed to work for my vicious cycle calculations, but I didn't have the ability to implement the duration-based strategy for that process. I was hoping to hang on to stochastic variation for diarrhea disease remission rather than update to the duration based approach, but I agree that we should look at this carefully in V&V... I just proposed that we make this update before implementation of the full vicious cycle to Rajan here: https://ihme.slack.com/archives/C018BLX2JKT/p1643227667019800

NathanielBlairStahn: When we say "duration-based I->S transition," does that mean all simulants will have exactly the same duration of diarrheal diseases (4.3 days, unless they die?), or does it mean we define the remission rate as 1/duration (which would automatically incorporate stochastic uncertainty)?

alibow: Good question. I thought the former, but I could be wrong. @rmudambi can you confirm?

rmudambi: Yes all simulants will have the exact same duration of the disease, though the dwell time is a pipeline, so it could be modified by an intervention, risk factor, etc.

PR #770, opened 2022-02-08
update diarrheal diseases on wasting PAF equation

alibow:
Here's what the equations look like formatted: 

image

### DIFF
--- a/docs/source/gbd2019_models/risk_effects/diarrheal_diseases/index.rst
+++ b/docs/source/gbd2019_models/risk_effects/diarrheal_diseases/index.rst
@@ -271,15 +271,15 @@ The age- and sex-specific values for RR_{i3, i2, i1} calculated as described abo
 
 .. math::
 
-   \text{prevalence_diarrheal_diseases{x+1}} = \frac{\text{p_D{x+1}}}{\text{p_D{x+1}}+\text{p_S{x+1}}}
+   \text{p_diarrhea}_\text{{x+1}} = \frac{p_\text{D{x+1}}}{p_\text{D{x+1}}+p_\text{S{x+1}}}
 
 .. math::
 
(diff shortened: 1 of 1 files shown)


aflaxman (approved):
the equation might look cooler if the `{x}` and `{x+1}` values were all subscripts (e.g. `_{x}` and `_{x+1}`

docs/source/gbd2019_models/risk_effects/diarrheal_diseases/index.rst
```diff
@@ -271,15 +271,19 @@ The age- and sex-specific values for RR_{i3, i2, i1} calculated as described abo
 
 .. math::
 
-   PAF_\text{i{x}} = \frac{\text{RR_i{x}} * \text{prevalence_diarrheal_diseases} + (1 - \text{prevalence_diarrheal_diseases}) - 1}{\text{RR_i{x}} * \text{prevalence_diarrheal_diseases} + (1 - \text{prevalence_diarrheal_diseases})}
+   \text{prevalence_diarrheal_diseases{x+1}} = \frac{\text{p_D{x+1}}}{\text{p_D{x+1}}+\text{p_S{x+1}}}
+
+.. math::
+
+   PAF_\text{i{x}} = \frac{\text{RR_i{x}} * \text{prevalence_diarrheal_diseases{x+1}} + (1 - \text{prevalence_diarrheal_diseases{x+1}}) - 1}{\text{RR_i{x}} * \text{prevalence_diarrheal_diseases{x+1}} + (1 - \text{prevalence_diarrheal_diseases{x+1}})}
```
aflaxman, line 278: Is it intended to have  {x} on the left-hand side and {x+1} on the right?

alibow: Yes, the PAF_i{x} is for the impact of diarrheal diseases on the transition from wasting state {x+1} to {x}. Using the diarrheal diseases prevalence in the {x+1} state as the exposure for the PAF calculation will be specific to the exposure in the "at-risk" (source state) population for the i{x} transition, so that we will avoid issues with the PAF estimation that we saw with the x-factor that had different exposures across wasting states when we used the population x-factor exposure rather than the wasting state-specific exposures.

rmudambi (approved):

PR #775, opened 2022-02-10
IV iron timeframe (including burn in period) and intervention scale-up

alibow:
I know we said that we only needed a one year burn-in for the pregnancy model when we discussed it earlier today, but I think we will need to run for two years otherwise we will underestimate YLDs due to maternal disorders for the first year of the simulation, so I upped it to two years. Happy to discuss this PR as a group in our check-in meeting tomorrow.


albrja (approved):

alibow:
> Looks good. Did I miss a meeting today?

Not a formal meeting, I just chatted quickly with Matt, Jim, and James about pregnancy initialization and we decided to run a burn-in period for it rather than try to initialize given that the maternal simulation should run quickly given the long time-steps and we will avoid any sampling inaccuracies given the differences in durations across different pregnancy outcomes that will cause different prevalence to incidence ratios and require a more complicated sampling protocol.

PR #776, opened 2022-02-15
Vicious cycle remission rate equations fix

alibow:
This was an error in the documentation that was inconsistent with the equations I was using in my calculation of the diarrheal diseases risk effects external to this document. The inconsistency was identified by Rajan and he is planning on implementing it as shown in this updated PR.


zmbc (commented):
In the table above it says `remission_diarrheal_diseases` is "scaled to total population person time as the denominator." It looks like this change would be equivalent to undoing that and removing `* prevalence_diarrheal_diseases` from that definition. Is there a reason it's better to do it this way?

NathanielBlairStahn (approved):

zmbc (approved):

PR #786, opened 2022-02-22
custom paf calculation for wasting and diarrhea

alibow:


aflaxman (approved):
How different will the PAF be when computed this way?

alibow:
> How different will the PAF be when computed this way?

@aflaxman the top panel is my custom-calculated PAF and the bottom is GBD's PAF


https://github.com/ihmeuw/vivarium_research/pull/786#pullrequestreview-890495801

plinds (approved):

PR #793, opened 2022-03-02
updates for vicious cycle

alibow:
Here are the updates that I think we should make that I am thinking (hoping!) will fix our validation issues with the vicious cycle:
- update to diarrheal diseases remission rate that should give us the correct observed rate given the timestep issues discussed in this slack thread: https://ihme.slack.com/archives/C018BLX2JKT/p1646183763054739
- I caught that the prevalence ratios were still based on the implied duration of the diarrheal diseases remission rate from GBD (~6 days) rather than the 4.3 days from the Troger paper, so I updated them here and they should also be updated for the simulation. Sorry for not getting this update in earlier!


docs/source/gbd2019_models/causes/diarrheal_diseases/index.rst
```diff
@@ -219,7 +219,7 @@ Data Description
 	* - i
 	  - S
 	  - I
-	  - :math:`\frac{\text{incidence_rate_c302}}{1-\text{incidence_rate_c302}*(\text{duration_c302} / 365)`
+	  - :math:`\frac{\text{incidence_rate_c302}}{1-\text{incidence_rate_c302}*(\text{duration_c302} / 365)}`
```
rmudambi, line 222: Is there any change I need to make related to this?

rmudambi (approved):

docs/source/gbd2019_models/causes/diarrheal_diseases/index.rst
```diff
@@ -303,6 +303,28 @@ Data Description
...
+
+		r' = (-1/dt)*log(1-dt*r)
+
+			= (-1/0.5)*log(1-0.5/4.3)
+
+			= 0.24722791193435328
+
+		1 / r' = 4.044850729740949 days
+
+	We then also solved for the upper and lower bound estimates using the same methodology.
```
aflaxman, line 326: Note for future: This locks us in to a simulation with a 0.5 day timestep.  It would be cooler to have a calculation of this type embedded in vivarium so that it will work without additional updates if we make the timestep different.  Not necessary for this project to succeed, though!

aflaxman (approved):
If this works, let's do it for LRI remission, too.

docs/source/gbd2019_models/causes/diarrheal_diseases/index.rst
alibow: No, this was just a formatting bug that I fixed

docs/source/gbd2019_models/causes/diarrheal_diseases/index.rst
NathanielBlairStahn: > I like this way, but adding one other idea of how this could be embedded in vivarium: using a draw from the continuous exponential distribution of event times for calculating the risk effects.

@zmbc What exactly do you mean by this? I don't think I understand.

NathanielBlairStahn: > Note for future: This locks us in to a simulation with a 0.5 day timestep. It would be cooler to have a calculation of this type embedded in vivarium so that it will work without additional updates if we make the timestep different. Not necessary for this project to succeed, though!

@aflaxman Yeah, if this works, the best way to incorporate it into Vivarium would be to simply set the probability of transitioning on a timestep equal to `r * dt` rather than `1-exp(-r*dt)` as is currently done. That would be more straightforward and would avoid the issue of having to back-calculate `r'` like we're doing here.

Of course, this requires that `r*dt<1`, which is not a constraint of our current approach, but if that inequality is not satisfied, then we're probably using too large of a timestep anyway, and we'd have an even worse validation problem than is showing up here with `r*dt ~= 0.116`.

The difference between choosing `p = 1-exp(-r*dt)` vs. `p = r * dt` for the parameter of the geometric random variable in Vivarium is that with the first approach (Vivarium's current method), the tail probabilities of the geometric random variable decay at the same rate as the corresponding exponential random variable, but the mean will be larger; whereas in the second case (corresponding to this PR), the tail probabilities of the geometric random variable will decay faster than the tail of the corresponding exponential, but the means will be equal.

In the first case, the CDF of the geometric random variable lies entirely below the CDF of the exponential random variable, whereas in the second case, the CDF after the point `t=dt` is shifted up so that sometimes it is above the exponential's CDF and sometimes below. Basically the issue is that the geometric distribution assigns zero probability to the interval [0, dt), but the exponential distribution has its maximum at 0 so assigns significant probability to this interval. Making the geometric random variable decay faster than the exponential random variable places extra mass at the point `x=dt` and shifts the CDF to compensate for the "missed" probability near zero.

NathanielBlairStahn (approved):

docs/source/gbd2019_models/causes/diarrheal_diseases/index.rst
NathanielBlairStahn: Another option that just occurred to me would be to use a geometric random variable supported on {0,1,2,...} rather than {1,2,3,...}. That is, there would have to be a nonzero probability of transitioning out of a state on the same timestep that you enter it. In that case, you could set the parameter `p` of the geometric distribution equal to `r*dt / (1+r*dt)` in order to get the same mean as the exponential distribution, avoiding the constraint that `r*dt < 1`. I haven't thought through this carefully, though.

zmbc: > What exactly do you mean by this? I don't think I understand.

As I understand it, we did this because our risk effects were calibrated to a certain duration, which wasn't a multiple of our time-step, so the overall risk effects we saw did not validate.

In my head the easiest way to think about this is that Vivarium, at each half-day increment, "catches up" to whatever would have been happening (in continuous time) in the interim. That is, the transition probability at `dt` should be `P(X <= dt)`, at `dt * 2` should be `P(X <= 2dt) - P(X <= dt)`, etc. Geometric with `p = 1-exp(-r*dt)` is correct for that, if we believe that r is a continuous hazard, not a risk. (But, I am not clear enough about where r comes from to know if that's true.)

However, in this case, our data source gives us an effect of diarrhea on wasting in terms of units per day of diarrhea. I am not sure I totally follow the way this data value ends up passing through our steady-state calculation, etc. but it seems to me that the issue we are seeing is because the RR actually ends up getting applied in half-day increments -- when a simulant is in the diarrhea state at the end of a timestep (whether they are transitioning out of it or not), we assume they were in that state for the full interval and apply the full RR. What I was proposing is that when we apply the RR, we "interpolate" between 1 and that RR in some way using a non-rounded amount of time in the state, sampled from the section of the exponential distribution between `dt` and `dt * 2`, for example.

I'm not sure if that makes sense, or is at all compatible with the prevalence ratio approach we use for the vicious cycle calibration.

NathanielBlairStahn: > I like this way, but adding one other idea of how this could be embedded in vivarium: using a draw from the continuous exponential distribution of event times for calculating the risk effects.

@zmbc Do you mean we'd keep the parameter `p = 1-exp(-r*dt)` as it is, but then every time a transition happens, instead of recording it as happening exactly at the current time step, say `n*dt`, we'd draw from an exponential distribution and record the transition as occurring at some random time in the interval in `[(n-1)*dt, n*dt]`? That also seems like a good idea that should solve the problem, though I'm not sure whether Vivarium can easily deal with continuous times in that way. And I'm still not sure whether I'm correctly interpreting your comment.

zmbc: @NathanielBlairStahn Yes, that's right; I also don't know whether Vivarium would be able to "record" the event at that continuous time (for purposes of DALYs, etc), but we would at least use that continuous time for applying the RR on wasting.

NathanielBlairStahn: > Note for future: This locks us in to a simulation with a 0.5 day timestep. It would be cooler to have a calculation of this type embedded in vivarium so that it will work without additional updates if we make the timestep different. Not necessary for this project to succeed, though!

@aflaxman @rmudambi @alibow If we want code that we don't have to update if we change the timestep, Rajan could just use the equation`r' = (-1/dt)*log(1-dt*r)` to calculate `r'` from the original value `r=1/4.3` (and the same thing for the confidence interval bounds) and from whatever timetep `dt` the simulation is using. We don't have to hard-code the "transformed" `r'` value, we can just give the SWE's the real value of r=4.3 and tell them to transform it using the above equation before inputting the rate into Vivarium.

alibow: Cool, thanks. Regarding Nathaniel's last point, I'd say let's go with whatever is easiest for @rmudambi to implement on our timeline. Let me know what you end up doing and then I can add notes into the documentation to indicate both what we did and how we might want to improve it in the future, if necessary.

beatrixkh (approved):

PR #798, opened 2022-03-09
ssh required to push to github now, I think

aflaxman:


albrja (approved):

NathanielBlairStahn:
Not if you use a personal access token:

- https://stackoverflow.com/questions/67077837/in-what-ways-is-an-ssh-key-different-from-tokens-for-git-authentication#:~:text=While%20SSH%20keys%20can%20be,recommends%20tokens%20over%20SSH%20keys
- https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token

I have a personal access token for my laptop, but I use SSH from the cluster because I wasn't sure how to get tokens to work on the cluster.

zmbc:
I use HTTPS; for the cluster I have a token that I keep in my home directory with permissions so only I can read it.

aflaxman:
Will it make things easier or harder if we steer people towards this approach?  If it doesn't make things easier, I'll leave the docs as they are.

NathanielBlairStahn:
> Will it make things easier or harder if we steer people towards this approach? If it doesn't make things easier, I'll leave the docs as they are.

I think either way, the current documentation is now incomplete because you can't just type in your password to access github. If we want these instructions to be truly useful, we would need to add instructions for creating and using either a token or an SSH key (or both).

aflaxman:
> I think either way, the current documentation is now incomplete because you can't just type in your password to access github. If we want these instructions to be truly useful, we would need to add instructions for creating and using either a token or an SSH key (or both).

How about you give your approval and I merge this for now, but we also note that we should add additional instructions in the future?

NathanielBlairStahn (approved):

PR #816, opened 2022-03-17
Pregnancy model fix wrong covariate ID

alibow:
I accidentally recorded the covariate ID for live births by sex instead of the still birth ratio. I'd be surprised if this was included in the model since the correct ID for SBR is included in other places and the SBR is less than 1 and live births by sex is in the many thousands, so I think we would have noticed, but just FYI!


aflaxman (approved):

albrja (approved):

mattkappel:
I don't think this is an issue on the SW side. We are using the fully-qualified name `'covariate.stillbirth_to_live_birth_ratio.estimate'` for SBR to get at it, not the covariate ID directly.

PR #818, opened 2022-03-17
Update to maternal disorders ratio equation

alibow:


albrja (approved):

mattkappel (approved):
Note to self: create a ticket for this change. Not a heavily lift.

aflaxman (approved):
I don't expect this change to make a big difference, but it makes logical sense.

PR #821, opened 2022-03-17
update pregnancy document to be consistent with implementation

alibow:
This is my understanding of the update of what the SWEs implemented. Please let me know if I am off on anything!


docs/source/other_models/pregnancy/index.rst
```diff
@@ -118,6 +118,9 @@ We will model pregnancy as a characteristic of women of reproductive age in our
   * - p
     - Pregnant
     - 
+  * - md
+    - Maternal disorders 
+    - Simulated state that lasts for single timestep in order to accrue maternal disorders YLDs - not representative of biological state. Can be with (infected/md_i) or without (susceptible/md_s) maternal disorders while within this state according to the maternal disorders cause model document.
```
aflaxman, line 123: I think infected and susceptible will lead to confusion here. What do you think about md_w and md_wo for with and without maternal disorders?

aflaxman (approved):

mattkappel (approved):
FWIW, in the code we call the "md" state "prepostpartum" but that's implementation detail that's probably not useful and more a simplification for engineering.

docs/source/other_models/pregnancy/index.rst
alibow: Good thinking, I've made the update.

PR #822, opened 2022-03-18
Solution for negative maternal disorders YLDs

alibow:
This is a fix for the temporary solution, but it's a low priority since it will end up making just a slight difference!


alibow:
Here are the rendered versions of the equations 

image

aflaxman (approved):

plinds (approved):

mattkappel (approved):

PR #829, opened 2022-03-28
hemorrhage risk effect update

alibow:
I believe that this has not already been built by the software engineers, so we shouldn't need a new ticket for it, but SWE feedback would be helpful!


albrja (approved):

alibow:
The fraction of each severity is now included in this PR: https://github.com/ihmeuw/vivarium_research/pull/834

mattkappel (approved):

aflaxman (approved):

PR #833, opened 2022-03-28
Hgb effect on birth outcome

alibow:


albrja (approved):

alibow:
> This is a large effect! How strong is the evidence base in that systematic review?

There were 21 studies included in the review for stillbirths and there was a consistent monotonic effect with severity of anemia. We could alternatively model moderate/severe anemia with an RR of 1.88 (1.27-2.47) or any anemia (mild, moderate, and severe) with an RR of 1.49 (1.15-1.92). Now that I write it out though, it might be most impactful to model the moderate anemia threshold, which is the same threshold used for IV iron targeting (10 g/dL).

Here's a plot that shows the relative incidence of stillbirth, FWIW. 
image

PR #838, opened 2022-04-08
add birth prevalence to infectious disease cause models

alibow:
This will be an update that requires an engineering ticket eventually, but let's hold off on making any updates to the model until we settle on a consensus strategy.


aflaxman (approved):
Should we put some uncertainty around these point estimates? I'm thinking about what will sound more acceptable in our methods section (or methods appendix, more likely)

alibow:
> Should we put some uncertainty around these point estimates? I'm thinking about what will sound more acceptable in our methods section (or methods appendix, more likely)

Yes, I also think that would be ideal, but I'm not quite sure how to arrive at a sensible confidence interval. I could spot-check specific draws, but it wouldn't be very systematic. Do you have any ideas for sensible approaches?

rmudambi (approved):

PR #840, opened 2022-04-08
Hgb bmi bw shift values

alibow:


alibow:
Oops, I just realized all of these shifts should be negative rather than positive the way I've structured it. Working on fixing it now in case you notice before I update!!

aflaxman (approved):

PR #844, opened 2022-04-13
document age shifting of pregnancy parameters

alibow:
this is a new feature request that we will need to create a ticket for

Software engineers let me know if the goal here isn't clear! We're hoping that this will be a simple change in the artifact age_start and age_end values that vivarium should be able to handle automatically.

### DIFF
--- a/docs/source/other_models/pregnancy/index.rst
+++ b/docs/source/other_models/pregnancy/index.rst
@@ -185,7 +185,7 @@ We will model pregnancy as a characteristic of women of reproductive age in our 
     - Data ID
     - Source
     - Vivarium age start
-    - Vivarium age start
+    - Vivarium age end
     - Note
   * - ASFR
     - Covariate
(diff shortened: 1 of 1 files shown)


docs/source/other_models/pregnancy/index.rst
```diff
@@ -184,27 +184,45 @@ We will model pregnancy as a characteristic of women of reproductive age in our
     - Data type  
     - Data ID
     - Source
+    - Vivarium age start
+    - Vivarium age start
```
mattkappel, line 188: `Vivarium age end` here?

mattkappel (approved):

docs/source/other_models/pregnancy/index.rst
alibow: Yes, thanks for catching that!

aflaxman (approved):

PR #861, opened 2022-04-27
update child v and v table

alibow:
I had a comment about underestimating the PEM CSMR, but it was really because I was misunderstanding the definition of the severity-specific PEM CSMR in the artifact, which was actually for all PEM not severity-specific. Plots are now updated to reflect that here: https://github.com/ihmeuw/vivarium_research_iv_iron/blob/main/validation/child/model_2/model_2.1.1_cause_model_validation.ipynb


aflaxman (approved):
Yay!

plinds (approved):

PR #867, opened 2022-05-04
cause model updates to fix validation issues

alibow:
I think these updates should address our validation issues for diarrheal diseases and LRI that we have been seeing and would like them to be incorporated into the existing IV iron child sim.

Please let me know if there are any questions about this strategy that I can help clarify.

### DIFF
--- a/docs/source/gbd2019_models/causes/lower_respiratory_infections/index.rst
+++ b/docs/source/gbd2019_models/causes/lower_respiratory_infections/index.rst
@@ -189,12 +189,8 @@ Data Description
      -
    * - I
      - prevalence_calculated
-     - incidence_rate_c322 * duration_mean
-     - Justification included below
-   * - I
-     - duration_mean
-     - 3.5/365 if age_group=='Early Neonatal' else (7.79 days; 95% CI 6.2–9.64; normal distribution of uncertainty)/365
-     - Justification included below
+     - **For early neonatal age group:** (birth_prevalence_I + (incidence_rate_c322 * duration_c322))/2. **For all other age groups:** incidence_rate_c322 * duration_c322
+     - Justification included below. Early neonatal age group exception due to non-steady state dynamics in this age group given birth prevalence of zero causes increasing prevalence within age group and short duration of age group. `Citation on these dynamics and approximations here for reference <https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3465772/>`_.
    * - I
      - excess mortality rate
      - :math:`\frac{\text{deaths_c322}}{\text{population} \,\times\,\text{prevalence_calculated}}`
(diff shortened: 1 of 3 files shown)


docs/source/gbd2019_models/causes/lower_respiratory_infections/index.rst
```diff
@@ -211,12 +207,7 @@ Data Description
...
-7.79 days (95% CI 6.2–9.64) days [GBD-2019-Capstone-Appendix-LRI]_. Note that this duration is longer than the Early Neonatal period (7 
-days). It follows that any incident LRI case in the early neonatal age group 
-will not remit before either dying or aging out. As an early neonate is 
-uniformly likely to become an incident LRI case on any of the 7 days of 'Early 
-Neonatal', it follows that the average duraiton of LRI in the early neonatal 
-age group is approximately 3.5 days (not accounting for deaths).
+7.79 days (95% CI 6.2–9.64) days [GBD-2019-Capstone-Appendix-LRI]_. This assumption is valid under steady state conditions. However, the prevalence of LRI is not in stead state for the early neonatal age group given a birth prevalence of 0 and a short duration of the age group (prevalence will increase as the population ages within the age group). Therefore, we calculate the prevalence in the early neonatal age group as an average of the birth prevalence and the approximated prevalence under a stead state transition (incidence * duration). This is approach is `discussed in this citation for reference <https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3465772/>`_.
```
aflaxman, line 210:
```suggestion
7.79 days (95% CI 6.2–9.64) days [GBD-2019-Capstone-Appendix-LRI]_. This assumption is valid under steady state conditions. However, the prevalence of LRI is not in steady state for the early neonatal age group given a birth prevalence of 0 and a short duration of the age group (prevalence will increase as the population ages within the age group). Therefore, we calculate the prevalence in the early neonatal age group as an average of the birth prevalence and the approximated prevalence under a steady state transition (incidence * duration). This is approach is `discussed in this citation for reference <https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3465772/>`_.
```

aflaxman (approved):
I think this will work.  The main change is to the prevalence (or in LRI, prevalence_calculated), right?  You cleaned up a bunch of other things, but it is the way changing the prevalence will change the derived EMR that will make the CSMR match between simulation and GBD, IIUC.

alibow:
> I think this will work. The main change is to the prevalence (or in LRI, prevalence_calculated), right? You cleaned up a bunch of other things, but it is the way changing the prevalence will change the derived EMR that will make the CSMR match between simulation and GBD, IIUC.

Yes, that's how I am thinking of it as well!

albrja (approved):

PR #877, opened 2022-05-12
child model verification table update

alibow:


hussain-jafari (approved):

aflaxman (approved):
noting the competing hypothesis for LRI ENN EMR being low: step size is too long to accurate approximate the high risk LRI death among neonates with LBWSG

PR #884, opened 2022-05-20
update output table shell

alibow:
- Added a request to stratify deaths and YLLs by pregnancy state in addition to YLDs (realized this will be helpful for results presentation)
- Clarified intervention administration counts as desired output


aflaxman (approved):
Can you screenshot the updated version and highlight the changed rows to make sure the SEs see just what you have changed?  GitHub is not providing a useful diff for me.

plinds (approved):

PR #889, opened 2022-06-04
First draft of documenting treatment prediction

zmbc:


docs/source/intervention_models/multiple_myeloma_treatment/index.rst
```diff
@@ -298,98 +298,316 @@ We plan to use Flatiron to inform baseline coverage of each treatment regimen.
...
+
+  The drug class components represent *exactly one* drug from
+  that class. If there is more than one, the treatment regimen will always fall
+  into one of the Other treatment regimen categories described in the next section.
+
+Treatment Regimen Categories
+++++++++++++++++++++++++++++
+
+Based on conversations
+with the client and with our clinical expert Manoj Menon, we are expanding the
```
plinds, line 366: We've never explicitly named anyone in these public docs before (at least to my knowledge) - does anyone have any thoughts on whether we should call out Manoj here or leave as simply 'our clinical expert'?

plinds (approved):

docs/source/intervention_models/multiple_myeloma_treatment/index.rst
aflaxman: For some reason, my gut reaction is to avoid naming the CSU clients, but no problem with naming our expert advisors.  Let's discuss briefly and decide about our approach together.  If Zeb wants to merge before we talk, please do not include the name for now.

PR #893, opened 2022-06-09
Update MM cause model without updating hazard calculation methods

zmbc:
This should be all cause model changes that impact the software engineers, except for the actual numbers (I have yet to make even placeholder files for the survival curves).

On the research side, I will still need to update these methods to reflect how we actually inform the curves.


docs/source/gbd2019_models/causes/neoplasms/multiple_myeloma/cancer_model.rst
```diff
@@ -125,55 +143,34 @@ This cause's restrictions have not changed since GBD 2017.
...
+is receiving Line 2 treatment, etc. New treatment assignment is therefore performed only at the time of
+relapse.
+
+Notably, the OS and TTD survival analyses supported by Flatiron data provide
+data on death *from any cause* among multiple myeloma patients and
+not death due to multiple myeloma specifically.
+Therefore, the excess mortality defined in this document will be the only source
+of mortality among simulants with multiple myeloma and excess mortality among
+simulants with multiple myeloma for "other causes," as typically defined by the
+cause-deleted all-cause mortality rate, should be zero.
```
aflaxman, line 172: I tried to explain previously why I want additional background mortality here, but in the spirit of keeping our changes minimal, I will not push more on it now.

aflaxman (approved):

docs/source/gbd2019_models/causes/neoplasms/multiple_myeloma/cancer_model.rst
zmbc: To briefly summarize the approach I would prefer, though it is a stretch goal and I don't want to distract from more important things:

Our curves, instead of depending only on line number and ASCT, should also depend on other covariates that we think affect hazard of death, primarily age. That way, we can account for the variation in death patterns that we would normally be getting from GBD (by age and sex) without double-counting any deaths (our "EMR" for this cause is derived from data on all-cause mortality), **and** allow MM to have age-varying EMR.

If we added the GBD other-causes mortality on top of our age-invariant EMR, we would be double-counting all non-MM deaths that occur among people with MM, and still not allowing MM to be more fatal in older people.

Perhaps the ideal would be to use GBD EMR for this cause and essentially rake the EMR of all our sub-MM states to that envelope, but that would be complicated.

PR #894, opened 2022-06-09
Document naive MM treatment prediction models

zmbc:
Based on discussion of starting with the naive prediction models, I have supplied them both in CSV (easier to load) and pickle (more similar to how the later models will be loaded, if we stick with the current plan) formats.

### DIFF
--- a/docs/source/intervention_models/multiple_myeloma_treatment/index.rst
+++ b/docs/source/intervention_models/multiple_myeloma_treatment/index.rst
@@ -431,19 +431,54 @@ Treatment Assignment
 ++++++++++++++++++++
 
 The Phase 1 simulation had fixed coverage percentages for each treatment regimen category
-in each line. In Phase 2, we will implement a more sophisticated model that takes
-into account covariates and prior treatment in a patient's history, informed by Flatiron
-data.
+in each line, which varied between scenarios. In Phase 2, we will assign treatment regimen categories
+according to these three schemes:
 
-Because the model needs to be trained within Foundry, it has to be passed from
+#. In the naive-model scenario (Model 1), we use naive models that always predict the same probabilities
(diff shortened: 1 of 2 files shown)


docs/source/intervention_models/multiple_myeloma_treatment/index.rst
```diff
@@ -431,19 +431,54 @@ Treatment Assignment
...
+
+Sophisticated Models
+~~~~~~~~~~~~~~~~~~~~
+
+The sophisticated models require a more complex approach. Because they need to be trained within Foundry, they have to be passed from
 Foundry to Vivarium in a serialized format. While a human-readable and interoperable
-format would be ideal, due to Foundry constraints we will use :code:`joblib.dump` to output a :code:`.pkl` file. The pickled
-form of the model may depend on the version of Python, :code:`sklearn`,
-and possibly sklearn subdependencies it was trained with. We will need to align
+format would be ideal, due to Foundry constraints we will use :code:`joblib.dump` to output :code:`.pkl` files. The pickled
```
aflaxman, line 475: I guess that on the research side @zmbc could load the .pkl file outside of foundry and resave the model in a more interoperable format.  But if this approach is working, let's go with it.

aflaxman (approved):

docs/source/intervention_models/multiple_myeloma_treatment/index.rst
zmbc: Good point and good idea. Unless we have problems matching package versions, I think that the `.pkl` file will actually be _easier_ to load -- but e.g. a [PMML](https://en.wikipedia.org/wiki/Predictive_Model_Markup_Language) file would be much better for long-term storage. @rmudambi may have thoughts about this.

plinds (approved):

PR #895, opened 2022-06-10
Document treatment and survival output format

zmbc:
I haven't documented the standard outputs -- YLDs, YLLs, etc.

My assumption is that we are no longer interested in the registry observer/output.


aflaxman (approved):
Let's think through what outputs we are going to produce with the treatment and survival output tables, to make sure we have what we need here.

plinds (approved):
"My assumption is that we are no longer interested in the registry observer/output." - confirming this point, we are NOT interested in this output in phase 2!

Also agreeing with Abie's comment - let's make sure we thoroughly think through all outputs needed for V&V and also what we want to keep for final outputs

PR #897, opened 2022-06-16
Document V&V strategy for model 1

zmbc:


zmbc:
It's not obvious how to verify once we are using sophisticated treatment assignment (Model 3). In principle, we could stratify our survival output by everything that has a risk effect on survival, and stratify our treatment output by everything that has an effect on assignment. But in practice I think that is infeasible due to runtime and also due to sample size. Any ideas?

I think it's okay to move forward with the already-documented outputs since I am pretty confident I can verify correctness of Model 1 with them. But we may have to update our observers/V&V strategy for Model 3.

docs/source/intervention_models/multiple_myeloma_treatment/index.rst
```diff
@@ -722,3 +722,39 @@ Assumptions and Limitations
...
+The death risk is :code:`died_by_end` divided by :code:`alive_at_start`, while the relapse risk is :code:`progressed_by_end` divided by :code:`alive_at_start`.
+All of these values are recorded by the survival observer.
+
+When the simulation is using a naive treatment assignment scheme, the relative risk of death or relapse between treatment :math:`t1` and treatment :math:`t2` in a timestep, :math:`risk_{t1} / risk_{t2}`,
+should approximately equal the ratio of the corresponding HRs, :math:`HR_{t1} / HR_{t2}`. In practice,
+we will bin groups of adjacent timesteps in order to have sufficient sample size to verify this.
+
+.. todo::
+
+  How do we verify treatment effects once we are using a sophisticated treatment assignment scheme?
```
aflaxman, line 760: Can we confirm this at the individual level, e.g. in an interactive sim?

aflaxman (approved):

docs/source/intervention_models/multiple_myeloma_treatment/index.rst
zmbc: I think I could use some guidance on when it is more appropriate to test with an interactive sim vs with outputs.

docs/source/intervention_models/multiple_myeloma_treatment/index.rst
```diff
@@ -722,3 +722,39 @@ Assumptions and Limitations
...
+
+ #. Treatments with a positive time trend according to our model outputs in Foundry have higher proportions in the
+    first year. They should have even higher proportions in later simulation years. Both of these effects will be
+    reversed for treatments with negative time trends.
+ #. Treatments with a positive age trend according to our model outputs in Foundry have higher proportions in older
+    age groups, and vice versa.
+
+.. todo::
+
+  How do we verify RRMM treatment assignment once we are using a sophisticated treatment assignment scheme?
```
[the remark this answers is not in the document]
zmbc: I think the model transfer verification script I included in the documentation, which verifies that the assignment probabilities match for a specific set of covariates, goes a long way toward establishing that the probabilities are as intended. The only part it doesn't test is that they are correctly sampled from. Is there a way to unit test that part?

docs/source/intervention_models/multiple_myeloma_treatment/index.rst
aflaxman: > I think I could use some guidance on when it is more appropriate to verify with an interactive sim vs with outputs.

Let's brainstorm on it in real-time during a meeting next week.

docs/source/intervention_models/multiple_myeloma_treatment/index.rst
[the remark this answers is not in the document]
aflaxman: > I think the model transfer verification script I included in the documentation, which verifies that the assignment probabilities match for a specific set of covariates, goes a long way toward establishing that the probabilities are as intended. The only part it doesn't test is that they are correctly sampled from. Is there a way to unit test that part?

I think that there is, let's discuss.

PR #904, opened 2022-06-27
Update parts of Phase 2 concept model that differ from Phase 1

zmbc:
We aren't fully fleshing out the Phase 2 concept model, but at least filling in the spots that will be different from Phase 1.


docs/source/concept_models/vivarium_csu_multiple_myeloma_2/concept_model.rst
```diff
@@ -147,15 +141,35 @@ set of treatment categories based on treatment guidelines from the `NCCN
...
 
+United States.
+
 6.2 Population and Randomness
 -----------------------------
 
+Population description:
+
+* Cohort type: Prospective closed cohort of individuals aged 15 years and older (open cohort if we only examine on age 50+ population)
+* Size of largest starting population: 100,000 simulants
```
zmbc, line 169: Note: Phase 1 concept model says 200k, but it looks like we were already using 100k by the time of our last Model 9 run.

docs/source/concept_models/vivarium_csu_multiple_myeloma_2/concept_model.rst
```diff
@@ -147,15 +141,35 @@ set of treatment categories based on treatment guidelines from the `NCCN
...
+* Time span: Jan 1, 2013 to Dec 31, 2027 (Jan 1, 2013 to Jan 1, 2023 is a 10-year long burn-in period)
+* Time step: 90 days
```
zmbc, line 171: Note: Phase 1 was 28 days, I don't _think_ anything heavily relied on the time step duration, and we have talked about changing to 90 to help with runtime.

aflaxman (approved):

docs/source/concept_models/vivarium_csu_multiple_myeloma_2/concept_model.rst
```diff
@@ -147,15 +141,37 @@ set of treatment categories based on treatment guidelines from the `NCCN
...
 -------------
 
+United States.
+
 6.2 Population and Randomness
 -----------------------------
 
+Population description:
+
+* Cohort type: Prospective closed cohort of individuals aged 15 years and older (open cohort if we only examine on age 50+ population)
```
alibow, line 170: I think the cohort type could use a bit more clarity - this ended up being a bit of a pain point in the IV iron sim when we had to switch from a closed to open cohort halfway through sim development. Basically, I think just stating the rationale behind each cohort type so it's more clear what to prioritize would be valuable.

alibow (approved):

docs/source/concept_models/vivarium_csu_multiple_myeloma_2/concept_model.rst
zmbc: We are definitely going to use a closed cohort at the Vivarium level. I think that the parenthetical is supposed to describe that it can be _considered_ an open cohort if we only look at results for 50+ (or 30+ for that matter, since the sim duration is 15 years). I will clarify.

PR #908, opened 2022-06-30
Document MM multiplier

zmbc:
This is an idea for how to speed up the Phase 2 model.

I could especially use help checking the math here. I considered getting the incidence rate with $$\frac{incidence_\text{c486} * 50}{1 - (prev_\text{c486} * 50)}$$ which would be more accurate where the prevalence in our sim is close to the GBD prevalence, but it's hard to say how different they are so I had trouble estimating the bias to justify the choice of multiplier.


docs/source/concept_models/vivarium_csu_multiple_myeloma_2/concept_model.rst
```diff
@@ -168,7 +168,7 @@ United States.
 Population description:
 
 * Cohort type: Prospective closed cohort of individuals aged 15 years and older. The sim duration is 15 years (see below), so results above age 30 will not be impacted by the open/closed distinction; essentially all multiple myeloma occurs at age 30+.
-* Size of largest starting population: 100,000 simulants
+* Size of largest starting population: 2,000 simulants
```
aflaxman, line 171: I think there is some sweet-spot trade off the the number of simulants and the overhead time for loading the data needed to set up the simulation, so maybe 2,000 simulants is too small.  Engineers might be able to advise.

docs/source/gbd2019_models/causes/neoplasms/multiple_myeloma_2/cancer_model.rst
```diff
@@ -271,7 +290,7 @@ data_dir = J:/Project/simulation_science/multiple_myeloma/data/cause_model_input
    * - incidence_MM
      - S
      - NDMM
-     - :math:`\frac{\text{incidence_c486}}{1-\text{prev_c486}}`
+     - :math:`\frac{\text{incidence_c486} * 50}{1 - \text{prev_c486}}`
```
aflaxman, line 293:
```suggestion
     - :math:`\frac{\text{incidence_c486} * 50}{1 - \text{prev_c486}*50}`
```

I think

aflaxman (approved):
It would be great to get @alibow and @NathanielBlairStahn to think this through with us, too, before sending the software engineers out to try it and see if it still validates.  Are we missing anything?

docs/source/gbd2019_models/causes/neoplasms/multiple_myeloma_2/cancer_model.rst
zmbc: See my note above -- I considered this. If my math is right (it's probably not), this makes our bias depend on the difference between sim prevalence and GBD prevalence, so it's hard to say definitively that it decreases the bias.

alibow: Is sim prevalence off from GBD by a factor of 50? We might be able to make a decision about what's "better" here by magnitude of differences

alibow:
> It would be great to get @alibow and @NathanielBlairStahn to think this through with us, too, before sending the software engineers out to try it and see if it still validates. Are we missing anything?

I think this approach makes sense, but results processing for V&V seems like it will be key... It seems like this is the plan, I just don't see it explicitly written: we won't expect ACMR to validate among the entire sim population without weighting the MM and non-MM groups - although it could end up being similar anyway because I imagine ACMR ~ ACMR - CSMR_MM

docs/source/gbd2019_models/causes/neoplasms/multiple_myeloma_2/cancer_model.rst
zmbc: Sorry -- when I say "difference between sim prevalence and GBD prevalence" I mean the difference _after_ accounting for a 50x multiplier.

In our Phase 1 results (no multiplier), we substantially overestimate MM prevalence relative to GBD in some age groups. That would create a downward bias in incidence per prevalent case, whereas Abie's proposed change could only increase incidence. But, if we ever had a prevalence _underestimate_ in our sim, the biases could compound. Without knowing in advance how big of an underestimate would be possible, I don't know how to tell whether this change would be better or worse.

Also, if we do this I will need some help with the math on what multiplier to choose, I tried working it out and got stuck 😅

aflaxman: IIUC, the Phase I prevalence in the sim was different than the GBD prevalence because we did not use the GBD mortality rate in Phase I.  So transforming the population incidence rate to the without-condition incidence rate has always been a little off.  If we multiply everything by 50, this discrepancy might matter more than it did in Phase I, since the GBD prevalence is so low that the population and without-condition incidence rates are almost identical.

So maybe the compromise solution is to multiply everything by 5 or 10 instead of 50.

NathanielBlairStahn: I'm having difficulty understanding everything well enough to give useful feedback, but I'd be happy to chat with you, @zmbc , so that I can ask questions to help me figure it out.

One question I have now: How did you derive the above formula for the "bias relative to the true GBD incidence count per prevalent case"?

alibow (approved):

PR #910, opened 2022-06-30
Document MM Phase 2 scenarios

zmbc:
~~We are going with using Flatiron (not IQVIA) to extrapolate Dara into the future with a linear (in logistic space) trend. The trend is very extreme, so we may modify our treatment assignment model to bring it lower.~~


plinds (approved):

docs/source/intervention_models/multiple_myeloma_treatment/isa_coverage.csv
```diff
@@ -0,0 +1,17 @@
+Scenario,Date,Line,Coverage
+All,2020/01/01,1,0%
```
mattkappel, line 2: That `%`...

docs/source/intervention_models/multiple_myeloma_treatment/dara_coverage_from_flatiron.csv
```diff
@@ -0,0 +1,66 @@
+Date,Line,Dara % (all),Dara % (Isa corresponding)
+2016/01/01,1.0,0.44%,0.29%
+2016/01/01,2.0,6.64%,6.53%
```
mattkappel, line 3: Why have this as `%` instead of max of 1?

hussain-jafari (approved):

docs/source/intervention_models/multiple_myeloma_treatment/dara_coverage_from_flatiron.csv
zmbc: Because I was trying to find a middle ground between machine- and human-readability. But given Abie's ask for graphs above, I think I will make this a decimal proportion and not display the whole table in the docs.

PR #912, opened 2022-07-06
Adding more todo's and minor updates

SylLutze:
Very minor updates - primarily I am practicing making edits and commits and working on figuring out git

### DIFF
--- a/docs/source/gbd2020_models/causes/measles/index.rst
+++ b/docs/source/gbd2020_models/causes/measles/index.rst
@@ -122,7 +122,7 @@ outbreaks due to lack of vaccination in small communities.
 
 Some of the **assumptions** made in this model are:
 
-1. There is no data available for population in recovered state in GBD. Since the early neonatal, late neonatal, and 1-5 month age groups are not modeled in GBD, we made the assumption that there are no individuals in the recovered state at the start of the 6-11 month age group (NOTE: prior GBD estimates started modeling at 28 days). We then calculated the proportion of the population in the recovered state in the 1-4 year age group using GBD measles incidence and mortality rates. Note that we performed this calculation only for the 1-4 age group because measles is most often of interest among the children under five population, but this assumption could be expanded to other older ages as well.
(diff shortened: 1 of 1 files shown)


docs/source/gbd2020_models/causes/measles/index.rst
```diff
@@ -120,7 +122,12 @@ outbreaks due to lack of vaccination in small communities.
...
+1. There is no data available for population in recovered state in GBD. Since the early neonatal, late neonatal, and 1-5 month age groups are not modeled in GBD, we made the assumption that there are no individuals in the recovered state at the start of the 6-11 month age group (NOTE: prior GBD estimates started modeling at 28 days). We then calculated the proportion of the population in the recovered state in the 1-4 year age group using GBD measles incidence and mortality rates. Note that we performed this calculation only for the 1-4 age group because measles is most often of interest among the children under five population, but this assumption could be expanded to other older ages as well.
+
+.. todo::
+
+   Edit above for age group changes once modeling approach has been decided 
+
```
alibow, line 130: So as described in the note on line 169, the previous assumption that prevalence of the R state = 0 at the start of the simulation caused us not to validate well to GBD burden estimates. But what it doesn't mention is that we didn't validate well for Uganda, which as a relatively high measles burden, but validated pretty well for Ethiopia, which has a lower measles burden. Since we'll be using this model for Ethiopia in the CIFF project, I'd say let's stick with the current assumption that the prevalence of the R state = 0 for the 6-11 month age group at initialization and see how it affects our model validation -- it may not be influential for lower measles burden locations. We can write a note/warning/todo to revisit the assumption if we're not validating well to GBD when we implement as a reminder!

alibow (approved):

alibow:
I also added Nathaniel and Abie as reviewers to keep them in the loop! If I recall correctly, I believe Nathaniel worked on earlier versions of the measles model, so he may have thoughts and input too.

aflaxman (approved):

PR #913, opened 2022-07-07
Updating risk effect table and references

SylLutze:
Updating the data in the risk effect table and adding in the needed references.

### DIFF
--- a/docs/source/gbd2019_models/risk_effects/mm_risk_effects/index.rst
+++ b/docs/source/gbd2019_models/risk_effects/mm_risk_effects/index.rst
@@ -51,12 +51,8 @@ In Phase 2, some of these risk factors will become correlated with treatment in 
 
   Flatiron allows us to estimate the correlation of risk factors, but we would need effect sizes adjusted for treatment and the other risk effects (and nothing else). Since we are not using Flatiron for the risk effects, there is no good way to obtain these. Due to the lack of risk effects, we do not model the correlations, except for those between each risk factor and treatment as described above, and between age and sex.
 
-.. todo::
-
-  Fill in this table.
-
 .. list-table:: Risk Effects Table
-  :header-rows: 1
+  :header-rows: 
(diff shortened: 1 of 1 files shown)


docs/source/gbd2019_models/risk_effects/mm_risk_effects/index.rst
```diff
@@ -68,40 +64,40 @@ In Phase 2, some of these risk factors will become correlated with treatment in
...
-    - Male
+    - 0.86 (0.68, 1.1)
+    - 0.83 (0.67, 1.03)
     - Female
+    - Male
     - Age, treatment(?)
-    -
-    -
-    -
+    - None
```
zmbc: I'm not sure which of these this applies to, but where the source is RCT and treatment was not assigned based on the risk factor, I think this column should say "Treatment (randomized)".

docs/source/gbd2019_models/risk_effects/mm_risk_effects/index.rst
```diff
@@ -68,40 +64,40 @@ In Phase 2, some of these risk factors will become correlated with treatment in
...
+    - 1 (1, 1)
+    - 1 (1, 1)
     - Impaired
     - Not impaired
     - Treatment
-    -
-    -
+    - 
+    - [Dimopoulos_2010]_
     - Impairment defined as eGFR less than 60
```
zmbc, line 92: Can you add to this note to explain how there is low data availability at our cutoff, but that existing data suggests no effect?

zmbc (approved):

docs/source/gbd2019_models/risk_effects/mm_risk_effects/index.rst
zmbc: "This column" = "Adjustment variables"

plinds (approved):

PR #916, opened 2022-07-08
Fix retreatment issues in Model 3 scenarios

zmbc:
~~We do not want to allow Isa->Isa or Isa->Dara.~~

We want Isa to affect future treatment in the same way that Dara does.

We want Isa to selectively replace Dara.

This introduces the versioning of treatment model inputs, which had been overwritten in previous updates.


rmudambi (approved):

hussain-jafari (approved):

aflaxman (approved):
This is what Zeb will be changing again right?  I'm approving to prevent any merge conflicts.

plinds (approved):

PR #918, opened 2022-07-10
Document survival curves

zmbc:
- Describes hazard calculation methods
- Updates hazard files

### DIFF
--- a/docs/source/gbd2019_models/causes/neoplasms/multiple_myeloma_2/cancer_model.rst
+++ b/docs/source/gbd2019_models/causes/neoplasms/multiple_myeloma_2/cancer_model.rst
@@ -174,7 +174,7 @@ cause-deleted all-cause mortality rate, should be zero.
 State and Transition Data Tables
 ++++++++++++++++++++++++++++++++
 
-In the tables below, data_dir = J:/Project/simulation_science/multiple_myeloma/data/cause_model_input/phase_2. The mortality and relapse inputs depend on the timestep size; input files are provided for 90-day and 28-day timesteps.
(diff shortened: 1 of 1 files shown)


docs/source/gbd2019_models/causes/neoplasms/multiple_myeloma_2/cancer_model.rst
```diff
@@ -174,7 +174,7 @@ cause-deleted all-cause mortality rate, should be zero.
 State and Transition Data Tables
 ++++++++++++++++++++++++++++++++
 
-In the tables below, data_dir = J:/Project/simulation_science/multiple_myeloma/data/cause_model_input/phase_2. The mortality and relapse inputs depend on the timestep size; input files are provided for 90-day and 28-day timesteps.
+In the tables below, data_dir = J:/Project/simulation_science/multiple_myeloma/data/cause_model_input/phase_2/2022_07_09. The mortality and relapse inputs depend on the timestep size; input files are provided for 90-day and 28-day timesteps.
```
aflaxman, line 177: I would like to get @zmbc to smooth these if we have time; they seem implausibly wiggly currently:
image

aflaxman (approved):

docs/source/gbd2019_models/causes/neoplasms/multiple_myeloma_2/cancer_model.rst
zmbc: Hmm... isn't that sort of the idea of using a non-parametric curve? You get some noise, without imposing any assumptions. If we wanted them to be smooth, we should probably use a spline or something similar in the model instead of doing postprocessing.

zmbc:
Note: There is a bug here! I calculated hazards as a proportion of the total initial population, instead of the population who would still be at risk at the beginning of that timestep.

plinds (approved):

PR #919, opened 2022-07-10
Update treatment HRs, document calculation method

zmbc:
Code PR: https://github.com/ihmeuw/vivarium_research_multiple_myeloma/pull/4

**Note:** We are expecting feedback from our clinical expert on some aspects of the analysis that lead to these values. If that feedback comes on Monday, I will update these; any later, and whatever the critique is will probably have to be a limitation.

As was noted before, these HR data files do not include every possible combination with/without ASCT. The ones that are not in these files should never occur in the sim -- I would rather fail fast if this ever occurs than fill in these files with placeholders for the impossible combinations.


hussain-jafari (approved):

aflaxman (approved):
We could eventually have @zmbc include the network diagrams and forest plots in these docs, to have a central place for copying them into reports on this work.

PR #940, opened 2022-08-05
China treatment changes

zmbc:
There are a few hacks in here that I don't love, but I think this is "good enough."

Code PR: https://github.com/ihmeuw/vivarium_csu_sanofi_multiple_myeloma_phase_2/pull/27


aflaxman (approved):

docs/source/intervention_models/multiple_myeloma_treatment/index.rst
```diff
@@ -607,33 +607,115 @@ match those generated within Foundry. It requires access to all the files in J:\
...
+    * Other triplets: PI+IMID+Dex and Chemo+PI+Dex, with and without ASCT
+    * Doublets: PI+Dex and IMID+Dex, with and without ASCT
+    * Not seen in China (probabilities set to 0): Chemo+IMID+Dex, with and without ASCT 
+* We approximately match our population-level RRMM induction coverage targets, which are for the same groups of regimen categories, but not separated by cytogenetic risk.
+* We approximately match our population-level NDMM targets for the probability of treatment containing ASCT or not.
+* As much as possible, we do not change the probability ratios between simulants for each regimen category. In other words, if Simulant 1 was twice as likely as Simulant 2 to receive PI+IMID+Dex before these rules, that should still be the case after these steps. However, we are forced to compromise on this requirement when we normalize each simulant's probabilities to sum to 1.
+* Within any of the above groups of regimen categories, for a single ASCT value (e.g. Dara triplets with ASCT), we do not change the ratios of probabilities between the regimen categories within it. This is true not only for the population but also at the simulant level.
```
alibow, line 660: So no correlation between ASCT probability and treatment category? This wasn't immediately obvious to me and seems like something we should flag!

alibow (approved):

docs/source/intervention_models/multiple_myeloma_treatment/index.rst
zmbc: There is a correlation between ASCT and induction, informed by Flatiron (since our China coverage targets do not tell us about the joint distribution).

What I was trying to express with this bullet point is that the "sub-category" breakdowns are retained. From the perspective of our China targets, all Dara triplets with ASCT are the same. So how that group is split between Dara+PI+Dex+ASCT and Dara+IMID+Dex+ASCT is retained from our Flatiron-trained model.

PR #942, opened 2022-08-10
Isa and Dara projections for China

zmbc:
Note the interpretation of the scenarios. The projections we were given assume scale up of Isa in both NDMM and RRMM. For the baseline scenario, I have set NDMM Isa to 0, giving its market share to Dara. Then in the Isa frontline scenario, we scale up Isa in NDMM according to the projection.


aflaxman (approved):
Thanks for pulling all of this together!  Can you add a figure that shows the scale-up curves when you have a chance?

zmbc:
Done!

plinds (approved):

PR #955, opened 2022-08-18
Nigeria and India to country VV

alibow:


aflaxman (approved):
Anything surprising?  The only thing I saw that looks more off than I would like is the `Maternal disorders cases per 100,000 pregnancies` plots in cell 28 of https://github.com/ihmeuw/vivarium_research_iv_iron/blob/main/validation/maternal/model9.3_national_locations/maternal%20disorders%20cause%20model.ipynb

e.g. 
image

albrja (approved):

alibow:
@aflaxman thanks for pointing this out, but from my perspective it's not surprising and is consistent with what we saw for the regional locations as well. It's essentially a different view of this plot here that shows underestimation in the younger ages and overestimation in the older ages. It looks better in this plot though because it's anchored at zero. 

image

The application of birth rates as pregnancy rates in our simulation, which causes us to shift the distribution by average gestational age. We identified this issue but didn't correct it because we match pretty well at the population level and it isn't off by much for mortality (shown here), so only is impacting YLDs.

image

PR #956, opened 2022-08-19
Treatment pathway

SylLutze:
Making some adjustments to the treatment pathway to include the new information from Greg. I also changed how to diagrams are organized to not repeat information and fix some issues. For example, the follow-up assumed patients are on both types of medication, which is not necessarily true. 

This separates it into the two types of treatment and makes the type of visit less important.


docs/source/concept_models/vivarium_us_cvd/concept_model.rst
```diff
@@ -610,68 +510,79 @@ adherent, their adherence score is 1.
 Treatment Assignments
 ~~~~~~~~~~~~~~~~~~~~~
 
-**LDL-C Treatments** 
+**Blood Pressure Treatments** 
 
-The decision to assign a simulant treatment is completed in the healthcare visits above. All LDL-C treatments are 
-statins for simplicity in this model. The choice of intensity is determined by the simulants ASCVD score and LDL-C. 
+In general, blood pressure medication is prescribed "low and slow" where medication is started at a low level
```
aflaxman, line 515:
```suggestion
In general, blood pressure medication is prescribed "start low and go slow" where medication is started at a low level 
```

aflaxman (approved):

plinds (approved):

PR #958, opened 2022-08-25
Creation of a max and min of SBP

SylLutze:


alibow:
I think adding a sentence here about what we do with values outside of the min/max would be worthwhile! IIRC we discussed assigning them the min or max?

alibow (approved):

aflaxman (approved):

plinds (approved):

PR #968, opened 2022-09-01
How to model correlated risks ppt

alibow:
I realized that this was never added to vivarium research, so adding it now.

Also, feedback welcome!


aflaxman (approved):
Thanks for remembering to get this into the repo!  I'm tagging @SylLutze here, too, because there is a possibility that we will step into GBD risk mediation in her CVD sim, if it turns out that we need to include BMI as a risk factor.

SylLutze (approved):

zmbc:
@alibow I think it would be great to revisit this in an RT meeting at some point. It seems to be relevant to many simulations.

I don't yet fully understand this PowerPoint because I ended up approaching this problem in a different way (not PAFs) that was more intuitive for me, and haven't found time to figure out the correspondence yet.

aflaxman:
As luck would have it, we have some funding to further develop our approach to correlated risks, as part of our MNCH work for BMGF.  Let's use some time in a future RT meeting to see how Zeb and Ali's approach compare and contrast, and brainstorm about what extensions to pursue next.

PR #970, opened 2022-09-07
Adding in results table, model 3 V&V and some new limitation

SylLutze:


docs/source/concept_models/vivarium_us_cvd/concept_model.rst
```diff
@@ -724,7 +724,7 @@ Baseline coverage of treatment for elevated SBP and elevated LDL-c is substantia
     - `Validation workbook Model 2 <https://github.com/ihmeuw/vivarium_research_nih_us_cvd/blob/main/Model2_VV_SBP.ipynb>`_ `And interactive sim <https://github.com/ihmeuw/vivarium_research_nih_us_cvd/blob/main/Interactive_Model2_VV.ipynb>`_ Cause model is identical to Model 1 with same pieces correct and the same discrepancies. Risk factors match for exposure, standard deviation and relative risk. Outstanding issue with individual simulant outliers in SBP and incidence. 
   * - 3.0
     - Adding angina as a cause    
-    - 
+    - `Validation workbook Model 3 <https://github.com/ihmeuw/vivarium_research_nih_us_cvd>`_ Cause model is identical to prior models with same pieces correct and the same discrepancies. Risk factors match for exposure, standard deviation and relative risk. Outstanding issue with individual simulant outliers in SBP and incidence. Seems that angina relative risk is highly susceptible to low n-size and leads to high variation.
```
aflaxman, line 727: Can you include the visual on the angina relative risk here?  There is so much stuff in the V&V notebooks that I think I missed this one.

aflaxman (approved):

docs/source/concept_models/vivarium_us_cvd/concept_model.rst
SylLutze: I added it! Good idea to paste it here

docs/source/concept_models/vivarium_us_cvd/concept_model.rst
```diff
@@ -734,54 +734,130 @@ Baseline coverage of treatment for elevated SBP and elevated LDL-c is substantia
...
 
-#. DALYs (YLLs and YLDs) due to cause specific mortality 
-#. Average values for risk factors by state and year (SBP, LDL-C, FPG, and BMI)
-#. Numbers of interventions administered per a) 100,000 population, and b) 100,000 person years 
+#. Total population 
+#. Person-time 
+#. YLLs and YLDs
+#. Deaths 
+#. Transitions for each cause 
+#. Exposure by person time for all risk factors
```
alibow, line 754: This would be for categorical risk factors only, right? Otherwise we're tracking mean exposure values? I think we should try to be really specific here!

alibow: Or oops, maybe I just misinterpreted "exposure by person time" and you do mean sum of all exposure values multiplied by person-time spent with that exposure value summed across all simulants to calculate the mean exposure value. Either way, I think a bit more detail to clarify would be helpful :)

docs/source/concept_models/vivarium_us_cvd/concept_model.rst
```diff
@@ -734,54 +734,130 @@ Baseline coverage of treatment for elevated SBP and elevated LDL-c is substantia
...
+#. Total population 
+#. Person-time 
+#. YLLs and YLDs
+#. Deaths 
+#. Transitions for each cause 
+#. Exposure by person time for all risk factors 
+#. Person time at or below target values for SBP and LDL-C 
+#. Healthcare appointments 
+#. Person time on medication 
+#. Medication effect
```
alibow, line 758: I think more specificity about what is meant by medication effect would also be helpful!

docs/source/concept_models/vivarium_us_cvd/concept_model.rst
SylLutze: Makes sense! Added

docs/source/concept_models/vivarium_us_cvd/concept_model.rst
SylLutze: Added

docs/source/concept_models/vivarium_us_cvd/concept_model.rst
```diff
@@ -734,54 +734,130 @@ Baseline coverage of treatment for elevated SBP and elevated LDL-c is substantia
...
+    - Split by primary non-adherent, secondary non-adherent, and adherent; and split by medication category 
+  * - Population on LDL-C medication 
+    - sum of person time on LDL-C medication 
+    - Split by primary non-adherent, secondary non-adherent, and adherent; and split by medication category 
+  * - SBP medication effect 
+    - sum of change in SBP exposure * person time  
+    - Split by medication category 
+  * - LDL-C medication effect 
+    - sum of change in LDL-C exposure * person time  
+    - Split by medication category
```
alibow, line 859: Hmm, maybe I don't understand enough about the treatment model here, but my instinct would be to split the mean SBP/LCL outputs by medication status and then calculate the difference from there

SylLutze: That makes sense and is aligned to what the engineers said - I adjusted the docs to reflect this change

alibow: Cool, I think we could actually delete these rows and then add medication to the list of stratifications in the "mean SBP/LCL-C" rows above.

SylLutze: I see your point - I updated the docs

alibow (approved):

docs/source/concept_models/vivarium_us_cvd/concept_model.rst
alibow: Cool looks good!

PR #971, opened 2022-09-07
Heart failure modeling updates

SylLutze:


docs/source/gbd2019_models/causes/heart_failure/index.rst
```diff
@@ -155,12 +142,63 @@ Vivarium Modeling Strategy
...
+     - 
+   * - 130-139 
+     - 1.5 (1.3, 1.73) 
+     - 
+   * - 140+ 
+     - 1.76 (1.43, 2.17) 
+     - 
+
+
+.. list-table:: Relative risk of heart failure for LDL-C
```
aflaxman, line 177: Did Greg say today that we should not include a risk-outcome pair for LDL-C to HF?

aflaxman (approved):
Was the concept_model.rst intended to be in this PR?  Maybe it is a hold-over from a previous PR.

docs/source/gbd2019_models/causes/heart_failure/index.rst
SylLutze: Yeah - just updated this to reflect his latest email.

alibow (approved):
I just want to confirm that we won't be double-counting heart failure due to IHD with this modeling strategy and the modeling strategy for IHD in the same model - otherwise looks good to me

PR #978, opened 2022-09-12
Vivarium best practices: timestep considerations

alibow:


aflaxman (approved):

docs/source/model_design/concept_model/choosing_appropriate_time_step/index.rst
```diff
@@ -29,10 +29,69 @@ Choosing an Appropriate Time Step
...
 
-  How to choose an appropriate time-step
+Generally, a timestep should be as large as possible while meeting model verification criteria and producing all desired outputs. Larger timesteps are more desireable because they decrease computational time and resources required to run the model.
 
-    Minimum disease duration, other considerations?
\ No newline at end of file
+Some modeling features that may impact timestep duration decisions include:
+
+- Modeled events between specified time intervals. For example, a monthly timestep may be desired if simulants attend monthly medication sessions
+- Duration-based choices: if the average duration of a given condition is 10 days, a timestep longer than 10 days may not be appropriate
```
zmbc, line 42: While we don't need to go into as much mathematical detail as we do for the third bullet, I think it would be useful to add a bit more detail on why these first two would cause issues. My understanding of the problems:
- Simulants can only do one transition in each cause/intervention model per timestep. So if the timestep is very long, it's unrealistic that only one thing happened to them during that long timestep, because in all likelihood something else would have happened to them in between that first event and the end of the timestep.
- Person-time observers always work on the assumption that someone was in the same state for the entire timestep duration. If people in real life would be quickly moving between states, this assumption becomes highly inaccurate.
- ... is there anything else I'm not thinking of?

I'd add a TODO here to more formally work this out, similar to what you've done for the rate issue below.

zmbc (approved):

docs/source/model_design/concept_model/choosing_appropriate_time_step/index.rst
alibow: Great thinking, thanks! I added all of these points.

PR #979, opened 2022-09-13
Vivarium best practices: results processing and v&v tips

alibow:
Please feel free to add to this page!!

### DIFF
--- a/docs/source/model_design/concept_model/results_processing_steps/index.rst
+++ b/docs/source/model_design/concept_model/results_processing_steps/index.rst
@@ -46,6 +46,10 @@ General things to check/keep in mind
 
 - If something is not verifying, but is close to the target, it could be due to stochastic variation. Check and see if it gets closer when you pool across all scenarios, all ages, etc. If so, stoachstic variation could be to blame for the deviation from the target value. In this case, consider increasing the simulated population size and/or simulated number of draws. Also, try looking at the median value, which may be closer to the target value than the mean is when there are small counts.
 
+- Remember the distinction between model verification (Do model outputs accurately reflect model inputs? Did we make a mistake in *building* the model?) and model validation (Do model outputs accurately represent reality? Did we make a mistake in *designing* the model?). Be sure to validate model results, including to your own back-of-the-envelope calculation (:ref:`see this page for details <vivarium_best_practices_boe>`) as well as external data sources that may have evaluated similar research questions.
+
+- Vivarium has an "untracked" population that can cause confusing issues if it is set to something unexpected. This is something that may be investigated in the `Interactive Simulation`_
+
 General list of things to verify
 -----------------------------------
 
@@ -75,7 +79,7 @@ Verification of these parameters compared to artifact and GBD values is generall
 
 Additionally, especially if simulating across several years, we should check not only that the parameters meet verification criteria across all simulated years, but also that there are no unacceptable/obvious trends in simulated outputs across simulated years (ex: some mortality rate increasing with time, etc.).
 
-Interactive simulations
+Interactive simulation
 ------------------------
 
 Some things may be easier to verify in interactive simulations rather than from count data outputs. Such parameters may include:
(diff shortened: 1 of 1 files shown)


aflaxman (approved):
This is great, thanks for documenting!  @SylLutze , since you have just been learning all of this, can you have a careful read through?  In particular, is there anything else you have been doing that we should add here?  And is there anything here that surprises you that we should elaborate on?

docs/source/model_design/concept_model/results_processing_steps/index.rst
```diff
@@ -24,15 +24,105 @@
...
+
+Some things may be easier to verify in interactive simulations rather than from count data outputs. Such parameters may include:
+
+- Risk factors with many categories (ex: LBWSG) because stratifying simulation outcomes by many categories may be too much of a drain on computation time
+
+- Continuous risk factors. Mean exposure values and/or proportions below a given threshold may be included as simulation outputs (TODO: provide link/info), but otherwise interactive simulations may be helpful to verify risk exposure standard deviation or other measures.
+
+- Continuous risk effects. TODO: provide details or examples
+
+- Continuity at the simulant level (to ensure that parameters that should not change over time *do* not change over time at the individual level)
```
zmbc, line 89: I'd add to this:

```markdown
- Many risk effects on the same event. If it is not computationally feasible to stratify the event rate by all risk factors that affect it, the best way to verify these risk factors may be to re-calculate and verify the event *rate* at the simulant level.
```

zmbc (approved):

docs/source/model_design/concept_model/results_processing_steps/index.rst
```diff
@@ -24,15 +24,105 @@
...
+- Intervention coverage
+
+- Intervention effects
+
+- Any other modeled parameters!
+
+Verification of these parameters compared to artifact and GBD values is generally done for the baseline scenario and alternative scenarios as necessary (especially for intervention coverage and effect verification). However, they should also be evalutated in alternative scenarios to ensure that there are no unexpected changes.
+
+Additionally, especially if simulating across several years, we should check not only that the parameters meet verification criteria across all simulated years, but also that there are no unacceptable/obvious trends in simulated outputs across simulated years (ex: some mortality rate increasing with time, etc.).
+
```
SylLutze, line 77: I would add something about when outside data might be needed to validate against. What is an example of a situation when the simulation results are non-trivial and so comparing against the artifact might not make sense? Especially for alternative scenarios, I still struggle with what might be a bug and what might be a result.

docs/source/model_design/concept_model/results_processing_steps/index.rst
```diff
@@ -24,15 +24,105 @@
...
+- Remember that simulation input data are generally age and sex specific, so it's generally easiest to perform verification of simulation inputs and outputs at the age and sex specific level. However, especially if simulating across many years (which may result in slight changes in the population structure over time), be sure to also **evaluate total population outcome rates as well as age- and sex-specific rates**.
+
+- If something is not verifying, but is close to the target, it could be due to stochastic variation. Check and see if it gets closer when you pool across all scenarios, all ages, etc. If so, stoachstic variation could be to blame for the deviation from the target value. In this case, consider increasing the simulated population size and/or simulated number of draws. Also, try looking at the median value, which may be closer to the target value than the mean is when there are small counts.
+
```
SylLutze, line 48: I know I encountered having untracked population - that might be a helpful starting check to add

SylLutze (approved):

docs/source/model_design/concept_model/results_processing_steps/index.rst
alibow: Great point, thanks!

docs/source/model_design/concept_model/results_processing_steps/index.rst
alibow: I just added a new bullet in the latest commit - let me know if you think it addresses your point or not!

SylLutze: I think that works yeah!

PR #981, opened 2022-09-13
Assorted updates including V&V notes, SE requests, and additional limitations

SylLutze:


docs/source/concept_models/vivarium_us_cvd/concept_model.rst
```diff
@@ -324,8 +324,37 @@ Individual intervention pages:
...
+
+**Final Model Run:**
+
+.. list-table:: CVD simulation model population parameters
+   :header-rows: 1
+
+   * - Parameter
+     - Value
+     - Note
+   * - Population size
```
alibow, line 336: This is not how it's structured in the template, but I think @aflaxman prefers to have "population size per draw" and "number of draws" defined by the researcher and "number of random seeds per draw" and "population size per random seed" defined by the engineers (so that the total population size matches what you've specified). Maybe we can edit this table so that it doesn't specify the number of random seeds to be in accordance with that preference and then eventually update the template too so we don't run into this again!

alibow (approved):

docs/source/concept_models/vivarium_us_cvd/concept_model.rst
SylLutze: Okay, that makes sense from our chat yesterday. Thanks!

SylLutze:
This is failing because edits are on different branches and the references are off. I will fix it once the "outreach intervention" is approved.

docs/source/concept_models/vivarium_us_cvd/concept_model.rst
```diff
@@ -752,11 +779,11 @@ Code is below for reference
...
   * - 4.0
-    - Adding heart failure as a cause  
-    - 
+    - Adding in healthcare system visits 
+    - Planned V&V: stable rate of appointments per CVD case in the population; percent of simulants with a follow-up scheduled is reasonably stable; percent of appointments that are follow-up visits is stable. Source: [Rodgers_2009]_
   * - 5.0
-    - Adding in healthcare system and medications for SBP and LDL-C  
-    - Note: need to confirm if we are overmedicating the population 
+    - Adding medications for SBP and LDL-C  
+    - Planned V&V: rate of medication per simulant with risk factor might increase but should be in line with publiched data [Gu_2012]_; total percent of population that is medicated; types of medication used over time (combo vs mono) should be stable [Derington_2020]_
```
plinds, line 786: Typo: *published

plinds (approved):

docs/source/concept_models/vivarium_us_cvd/concept_model.rst
SylLutze: Thanks!

PR #982, opened 2022-09-13
vivarium best practices: monte carlo uncertainty

alibow:
@SylLutze and I just met to talk through this, so I thought it would be worth writing up to have as a resource. I left to-do notes to link to other resources so if anyone has any in mind that may be good fits, please let me know!

### DIFF
--- a/docs/source/model_design/concept_model/monte_carlo_uncertainty/index.rst
+++ b/docs/source/model_design/concept_model/monte_carlo_uncertainty/index.rst
@@ -87,7 +87,7 @@ Simulated population size per draw will directly affect the impact of stochastic
 Interaction between random seeds and simulated population sizes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Generally, a researcher should communicate to the engineers the desired simulated population size per draw for a given simulation (see below for how to select an appropriate value for this parameter). Then, the engineers (perhaps with input from the researchers!) will determine an appropriate number of subgroups to divide this population across to optimize the balancer between the amount of cluster nodes/memory as well as the run time duration that the simulation will take. 
+Generally, a researcher should communicate to the engineers the desired simulated population size per draw for a given simulation (see below for how to select an appropriate value for this parameter). Then, the engineers (perhaps with input from the researchers!) will determine an appropriate number of subgroups to divide this population across to optimize the balance between the amount of cluster nodes/memory as well as the simulation's run time. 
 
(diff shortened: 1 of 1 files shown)


docs/source/model_design/concept_model/monte_carlo_uncertainty/index.rst
```diff
@@ -29,8 +29,97 @@ Monte Carlo Uncertainty in Vivarium
...
+
+  Link to some good existing resources. Anybody know of any good ones?
+
+Uncertainty Parameterization in Vivarium
+-----------------------------------------
+
+What is a draw?
+++++++++++++++++
+
+We can think of a draw as a single iteration of some calculation or model run within a series of calculations or model runs that make up the summary estimates.
```
zmbc, line 47: I'm not sure I understand what is meant by "iteration" here. I don't know the clearest way to explain what a draw is, but I'd start with (or at least include) that a draw is a sample from a probability distribution and that we use the 1,000 draws/samples to represent the distribution, instead of using a continuous density function.

docs/source/model_design/concept_model/monte_carlo_uncertainty/index.rst
```diff
@@ -29,8 +29,97 @@ Monte Carlo Uncertainty in Vivarium
...
+
+Interaction between random seeds and simulated population sizes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Generally, a researcher should communicate to the engineers the desired simulated population size per draw for a given simulation (see below for how to select an appropriate value for this parameter). Then, the engineers (perhaps with input from the researchers!) will determine an appropriate number of subgroups to divide this population across to optimize the balancer between the amount of cluster nodes/memory as well as the run time duration that the simulation will take.
```
zmbc, line 90:
```suggestion
Generally, a researcher should communicate to the engineers the desired simulated population size per draw for a given simulation (see below for how to select an appropriate value for this parameter). Then, the engineers (perhaps with input from the researchers!) will determine an appropriate number of subgroups to divide this population across to optimize the balance between the amount of cluster nodes/memory as well as the simulation's run time. 
```

zmbc (approved):

SylLutze (approved):

docs/source/model_design/concept_model/monte_carlo_uncertainty/index.rst
alibow: Yes, this is much more clear. I updated to include this language and also added in a link to a monte carlo overview resource which will hopefully also provide some more context. Thanks!

PR #983, opened 2022-09-14
Outreach intervention page revamping 

SylLutze:


alibow:
FYI, I updated the base branch from main to assorted so that we could more easily see what the changes specific to this PR were!

docs/source/intervention_models/crm_mgmt/vivarium_outreach.rst
```diff
@@ -1,54 +1,140 @@
...
 
 
-.. list-table:: Affected Outcome #1 Effect Size
-  :widths: 15 15 15 
+This intervention affects primary adherence to medications, which in turn will affect the simulants exposure to SBP and LDL-C. 
+
+
+This odds ratio above will approximately halve the number of simulants that are primary non-adherent. All simulants that 
+are no longer primary non-adherent, are assumed to be adherent. This means that the new adherence rates for simulants 
+with the intervention are:
```
alibow, line 79: I think it would be useful to provide a bit more context here regarding the update (date that the change was made and maybe a link to the PR where the changes were made so that we can have a record of it). There's an example of this kind of note at the top of the SQ-LNS intervention page https://vivarium-research.readthedocs.io/en/latest/intervention_models/lipid_based_nutrient_supplements/index.html

Maybe it's not as important for this application though - so really whatever you think is best!

alibow (approved):

PR #992, opened 2022-09-23
Make minor edits to PRL concept model

zmbc:
Just things I noticed as I read through the concept model.

We need to make an engineering ticket for the last name assignment by relationship!


aflaxman (approved):
Thank you!

SylLutze (approved):

PR #995, opened 2022-09-28
International immigration

zmbc:
I have some doubts about this:

- Is it detailed enough? In the current plan, this documentation is all that would be provided to the engineers and they would implement everything on their side. **I believe we won't want to implement this until the emigration component is also defined, since it will require additional new ACS columns.**
- Is the sampling approach acceptable? I have some concerns; see "sample size considerations" in [this notebook](https://github.com/ihmeuw/vivarium_research_prl/blob/immigration/migration/exploratory/migration_characteristics.ipynb).


docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -119,26 +119,25 @@ include an attribute for their relationship to a reference person
...
-observations with realistic noise (13).
+actually unique for others (14); and periodic survey, census, and registry
+observations with realistic noise (15).
 
 Additional components we might want: time-dependent changes to
-observers of sex, based on gender assigned at birth (15); multiple
-households for individuals, leading to double counting in census (16);
-twins and multiparous births in fertility model (14).  To capture an
+observers of sex, based on gender assigned at birth (17); multiple
+households for individuals, leading to double counting in census (18);
```
albrja, line 136: We want an observer that double counts individuals?

zmbc: Yes, @SylLutze is working on documenting this. Here I've just changed the number of the component.

albrja (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
aflaxman: Even triple counts sometimes!

aflaxman (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: @zmbc I think we should connect on double counting. I think ideally the double (or triple) counting happening in the simulation, not in the observer. 

An example: someone should move and then exist in the simulation in two households. Rather than at the point of sampling point having them be double counted.

SylLutze (approved):

SylLutze:
For the PUMA sampling - could we look at this spatially? I would imagine that PUMAs that contain or are near cities are more likely to receive immigrants (and are next to each other). We could have some kind of smoothing so that a PUMA's immigrant rate and population are influenced by its nearest neighbors. This might help with noisy data and small sample sizes.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: While I think we will stick with the simpler approach where they are double-counted at enumeration time, it would be interesting to think through _where_ they are double-counted. I'd imagine a good amount of double-counting (>50%?) happens with the duplicates in _different_ households, so double-counting them with the same household ID would make PRL look too easy.

zmbc:
@SylLutze

> For the PUMA sampling - could we look at this spatially? I would imagine that PUMAs that contain or are near cities are more likely to receive immigrants (and are next to each other). We could have some kind of smoothing so that a PUMA's immigrant rate and population are influenced by its nearest neighbors. This might help with noisy data and small sample sizes.

Great question. I actually see two potential approaches in your question: one would be a spatial small-area-estimation model, and another would be adding covariates e.g. for urban-rural status.

I think that there are some theoretical problems with SAE models, but they are almost certainly better than what we plan to do now. The main reason why we hadn't discussed using them is because I'm not familiar with the methods, and we don't use them in other places they could also apply (e.g. VEHSS county modeling) for reasons that @aflaxman may be able to shed more light on.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -119,26 +119,25 @@ include an attribute for their relationship to a reference person
...
-actually unique for others (12); and periodic survey, census, and registry
-observations with realistic noise (13).
+actually unique for others (14); and periodic survey, census, and registry
+observations with realistic noise (15).
 
 Additional components we might want: time-dependent changes to
-observers of sex, based on gender assigned at birth (15); multiple
-households for individuals, leading to double counting in census (16);
-twins and multiparous births in fertility model (14).  To capture an
+observers of sex, based on gender assigned at birth (17); multiple
```
beatrixkh, line 135: i'm reading this as observers getting a simulants sex wrong, as opposed to a simulant's sex changing --- is that accurate?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Agreed. It sounds like for now that we will basically not be including this for now in the observer in any function - same household ID or not. I will list that in the limitations but I think its important to include somehow and have some folks be listed at multiple addresses and some duplicated at the same address.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -556,45 +555,15 @@ and geography, we can calculate the probability of moving from ACS, as
 the weighted average of MIGPUM.isnull(); could also determine if they
 moved within the PUMAs represented in the sim or from outside those
 PUMAs.
+For now, we only model migration within the sim catchment area (this component)
+and to/from other countries (next two components).
+When the simulation only includes part of the US, there is no domestic
+migration into or out of this region.
```
beatrixkh, line 561: just to make sure i'm reading this correctly:

we're saying that when we are only modeling a subset of the US, there is no migration from the subset we are modeling to another part of the US, or migration from another part of the US to the subset we are modeling -- yes?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -627,9 +596,78 @@ simplifying assumptions that we have included:
...
+We simulate two kinds of immigration: household moves and non-reference-person moves.
+
+#. A **household move** is when an entire household (which may be a single person) enters from outside the country as a unit,
+   preserving relationships within the unit.
+#. A **non-reference-person move** is when a person enters from outside the country and joins an existing household.
+   Non-reference-person moves are independent, single-person events that do not preserve relationship structure.
+
+For the purposes of immigration, a group quarters person should be treated identically to a one-person household,
+with the group quarters person as the reference person.
+This means that group quarters people should only enter the US in a household move.
```
beatrixkh, line 614: could you clarify this a bit? are you referring to group quarters individuals or "households" moving into the US?
if individuals, is this group quarters individuals moving into group quarters, into a non-group quarters existing household, or just into their own household (then they become household head?)
if it's an entire group quarters "household", i assume this would mean they move into the correct group quarters type?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -627,9 +596,78 @@ simplifying assumptions that we have included:
...
+We also assume that the proportions of recent immigrants by move type (household or non-reference-person) will remain constant.
+Though in reality not all moves into the US follow one of these two patterns, we assume that any new immigrant in a household
+where the reference person is also a new immigrant was part of a household move, while any new immigrant in a household where the
+reference person is not a new immigrant moved to the US in a non-reference-person move.
+
+Specifically, the yearly rate at which simulants are added to the population by each move type is given by
+the (weighted) proportion of ACS PUMS persons that are recent immigrants consistent with that move type.
+At each time step:
+
+#. PUMS households with reference people who are recent immigrants, after removing household members who are not recent immigrants,
```
beatrixkh, line 632: i assume this is per-state? i.e., we look sample from all 2019 **florida** residents who are recent immigrants?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -627,9 +596,78 @@ simplifying assumptions that we have included:
...
+   are sampled with replacement until the desired number of simulants added in household moves is reached.
```
beatrixkh, line 633: if we aim for the desired number of simulants added in household moves, do we have concerns about the number of recent immigrant households being quite far off from the data?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -627,9 +596,78 @@ simplifying assumptions that we have included:
...
+#. PUMS recent immigrants not living in such households are sampled with replacement until the desired
+   number of simulants added in non-reference-person moves is reached.
+
+Households added in a household move **that are not GQ people** are assigned a new address and household ID so as not
+to conflict with the existing household that was sampled to create them.
+GQ people added as a single-person household should retain the GQ type from sampling.
+
+Simulants added by a non-reference-person move join a randomly-selected existing household matching their PUMA.
```
beatrixkh, line 641: currently the PUMA never changes after initialization. are we concerned about the likelihood that this causes us to add PERSON A in the data to the household they are already in? (but i guess... this would be a younger version of themselves, because the PERSON A from initialization would have aged a bit?)

i think this is probably quite unlikely, but am not sure how large PUMAs tend to be

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
beatrixkh: also, are we sampling with the household weight or the person weight of the household head, here? on initialization we use the household weight to sample households, so i assume we'd do that again here -- but that may(?) create strange behavior in the case that there is some non-reference person with a high person weight, who is a recent immigrant, living in a household with a recent immigrant household head, with a low household weight. << not sure if that scenario exists in the data, but this would result in that non-reference person being unlikely to enter the simulation, despite having a high person weight

this is almost certainly not a concern, but wanted to note it might contrast with the way we sample the non-reference recent immigrants living in houses with non-recent immigrant household heads

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: I'm not sure; @aflaxman wrote this part and it is not substantively updated here.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Yes, that's correct.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: GQ people move as a "household" of one person. They would move to the correct (sampled) group quarters type and address. They would never join an existing household.

I will attempt to reword this to make it more clear.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: As currently written, when we sample recent immigrants, we preserve their PUMA and state. So the sampling happens at the national level, but the immigrants within each state (and indeed PUMA) will all be based on recent immigrants to that state/PUMA.

Re: weights, this was an oversight on my part. I believe it will be correct to sample households using household weights and individuals using person weights, even though it is a bit odd. I will clarify this in the documentation.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: I would not expect this to happen. It would require the household size of the sampled households to be systematically different from the households in ACS with a recent immigrant reference person. Since that is precisely the group we will be sampling from, the household size should be the same, modulo some sampling error. Or am I missing something?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: PUMAs are 100,000+ population, which means ~5,000+ ACS respondents in a 5-year file. That would make this pretty unlikely.

That said, the fact that PUMAs never change after initialization (i.e. at the time of domestic migration) is a problem. It is not correct to be doing this matching on PUMA when it may no longer represent where a household "should" be living. We will need to update the domestic migration component to correct this, and are working out the best way to do so.

zmbc: I did not note this in the documentation (I will add it), but of course, simulants added via migration should get a new unique simulant ID for PRL tracking purposes (they wouldn't literally be the same people, even if added to the same household).

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Actually, I have a question for you: is there any correlation between group quarters type and address? Does each simulant in a group quarters type (e.g. carceral) have a unique address?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Ah, I think I may have misunderstood what you meant by your first question. When we are only _simulating_ Florida and not the entire US, all ACS data used should be filtered to Florida.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
aflaxman: I have a pretty straightforward way to do this that I think is respectful to transgender people.  I haven't had time to include it here, though, and I don't think we need it to get to done on our minimal model.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -614,33 +614,64 @@ with the group quarters person as the reference person, and that GQ "household"
...
 Added GQ people (who all enter in "household" moves) should be assigned a household ID for a randomly-selected GQ type matching
-their institutional/non-institutional status, as is done at population initialization.
+their institutional/non-institutional status, as well as the corresponding shared address, as is done at population initialization.
+
+Additionally, we perturb the PUMA attribute of the sampled person (in the case of a non-reference-person move) or household (in the case of a household move).
+This is intended to ensure a non-zero probability of immigration into all PUMAs,
+and to lower the chances of very similar simulants being sampled from the same ACS person/household.
+The perturbation process is as follows:
+**10%** of the time, replace the sampled person/household's PUMA with a PUMA selected at random according to the
+probabilities in the appropriate file below.
```
beatrixkh, line 655: following these probabilities, do we expect there is any chance we choose a PUMA with no other households? In the case that we do that for a non-household head move, is that okay, and we make them a household head? or that shouldn't have been possible?

PR #1002, opened 2022-10-03
Note on wasting transition truncated normal distribution

alibow:
As determined and implemented by @hussain-jafari!


hussain-jafari (approved):

aflaxman (approved):
Weird that this upper limit is required!  What python method are we using?

alibow:
> Weird that this upper limit is required! What python method are we using?

I think this is the one: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.truncnorm.html

I believe it is because the probability of the previous upper limit we selected was too low to compute, but perhaps @hussain-jafari has a better understanding.

PR #1003, opened 2022-10-04
Handle edge case for PUMA matching

zmbc:


aflaxman (approved):

plinds (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -672,7 +672,8 @@ the 2.5th percentile value, and the proportions were re-normalized.
     before it is appropriate to interpret them as probabilities.
 
 Simulants added by a non-reference-person move join a randomly-selected existing non-GQ household matching their PUMA.
-Their relationship attribute is unchanged from sampling, except that "Father or mother" becomes "Other relative" and
+If there is no such household, the PUMA perturbation process described above is used to repeatedly assign a new PUMA until this is no longer true.
```
beatrixkh, line 675: thank you for updating to catch the edge case! In order to leave us flexibility for how we implement this, could you tweak the docs to describe the desired outcome instead of how we achieve it? i.e., that we should not select a PUMA for an individual to move into, if that PUMA does not contain any households already in the sim? (I think we will probably end up subsetting down the set of PUMAs that we sample from to the inhabited PUMAs, in order to only have to sample once)

PR #1004, opened 2022-10-04
Adding an observer for household surveys

SylLutze:
Adding a generalized observer for household surveys and specifics for the ACS


docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1079,6 +1079,157 @@ These are not currently modeled.
...
+non-response bias matches the census, using the data in the section above. 
+
+.. list-table:: Calculating Simulant Non-Response 
+  :widths: 10 10 10 20
+  :header-rows: 1
+
+  * - Simulant 
+    - Race/Ethnicity 
+    - Survey Modes USed 
+    - Probability of Omission (%)
```
zmbc, line 1188: Is there a reason to use "Omission" here instead of "Non-Response"?

zmbc (changes requested):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1079,6 +1079,157 @@ These are not currently modeled.
...
+
+.. list-table:: Simulant Attribute to Sample 
+  :widths: 20
+  :header-rows: 0
+
+  * - Unique simulant ID (for PRL tracking)
+  * - First name
+  * - Middle initial 
+  * - Last name
+  * - Age
```
aflaxman, line 1117: Might be full DoB sometimes?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1079,6 +1079,157 @@ These are not currently modeled.
...
+.. list-table:: Simulant Attribute to Sample 
+  :widths: 20
+  :header-rows: 0
+
+  * - Unique simulant ID (for PRL tracking)
+  * - First name
+  * - Middle initial 
+  * - Last name
+  * - Age 
+  * - Home Address
```
aflaxman, line 1118: Might just be zip code sometimes, can anticipate by including zip code as a separate field.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
aflaxman: Can anticipate by including month and day as separate fields

aflaxman (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1079,6 +1079,157 @@ These are not currently modeled.
...
+**Step 1:** Households will be randomly selected for participation at a rate 
+predetermined by the researcher. The selection should be stratified by state, 
+but no other variables. This will be a random sample. 
+
+**Step 2:** Households will be chosen to be non-responsonders and removed from 
+the sample. This step will vary significantly based on the mode of the survey. 
+There are three possible modes: mail/online (assumed to be the same for this 
+model), telephone, and personal visits. 
+
+The below table includes the percent of responses for each mode of survey by
```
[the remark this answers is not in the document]
plinds: Echoing Zeb's question (let's make sure to state where the data in this table is coming from) and comment (I think in general it's nice to see the table as it is now, but also have a downloadable .csv file for ease of implementation).

plinds (approved):

zmbc:
Re: the CSV, there is also the ability to display a CSV file inline, which may or may not be useful here: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/tables.html#csv-files

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1079,6 +1079,157 @@ These are not currently modeled.
...
+#. Sampling on a single time step is not representative of most surveys. People might move houses, change names, have babies, or have loved ones die during the census leading to additional noise in the census not modeled here 
+#. Our model does not include an option for double counting or duplicating people 
+#. There are multiple other factors that contribute to omission rate including: tenure in a home, state/geography, age, and having a SSN (as a proxy for citizenship) [Jackson_2007]_. These are not currently included in our model 
+#. Simulants who do not respond to one time point in a longitudinal survey are probably more likely to not respond moving forward. We assume independence here. 
+
+**Initial Survey - American Community Survey** 
+
+The ACS will be used for V&V testing. It is deinfed as: 
+
+- Sample rate of (12,000 / 350 million)%
```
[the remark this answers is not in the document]
SylLutze: Added that this is also stratified by state for clarity

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: No, just a copy paste issue

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Added DOB

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1079,6 +1079,157 @@ These are not currently modeled.
...
+
+- The sample will be taken at set time intervals (monthly, annually) that the research team will communicate in terms of time steps 
+- The total duration of the survey will be specified 
+- Each subsequent sample is assumed to be independent of the prior samples 
+
+2. A longitudinal sample will sample the SAME population on defined time steps (e.g., sample the same 5% of households each month for 6 months) 
+
+- The sample will be taken at set time intervals (monthly, annually) that the research team will communicate in terms of time steps 
+- The total duration of the survey will be specified 
+- Once there are stable addresses, the sample will be taken and kept the same at the address level. For now, when addresses are not stable, we can keep the same households instead.
```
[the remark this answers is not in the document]
SylLutze: Yes, but upon further review I changed this to include the same people each time.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1079,6 +1079,157 @@ These are not currently modeled.
...
+duplicate people or swapped people. Others may be introduced 
+at the time of sampling, such as name and address misspellings. 
+
+These are not currently modeled. 
+
+**Limitations and Possible Future Adds** 
+
+#. Sampling on a single time step is not representative of most surveys. People might move houses, change names, have babies, or have loved ones die during the census leading to additional noise in the census not modeled here 
+#. Our model does not include an option for double counting or duplicating people 
+#. There are multiple other factors that contribute to omission rate including: tenure in a home, state/geography, age, and having a SSN (as a proxy for citizenship) [Jackson_2007]_. These are not currently included in our model
```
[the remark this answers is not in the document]
SylLutze: Added clarification of the data source and changed to non-response

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1079,6 +1079,157 @@ These are not currently modeled.
...
+    - Survey Modes USed 
+    - Probability of Omission (%)
+  * - 1
+    - White 
+    - Mail/Online Only 
+    - 100% - 62.5% = 37.5% non-response 
+  * - 2
+    - Black  
+    - Mail/Online and Telephone  
+    - 100% - (29.7% + 15.1%) = 55.2% non-response
```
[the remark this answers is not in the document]
SylLutze: I think this and the below question have the same miscommunication on my part - these percents add to 100%. So I think it would be better to interpret them not as a probability of non-response, but as an additive percent of respondents reached. I added some clarification above that this data should NOT be taken for only telephone, but rather a look at how many of these methods you completed.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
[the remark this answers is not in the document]
SylLutze: Makes sense! I added a .csv and where to find it. The lack of citing the data was also definitely an oversight - added in here

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1079,6 +1079,168 @@ These are not currently modeled.
...
+
+The below table includes the percent of responses for each mode of survey by 
+race/ethnicity. These will be used to find the non-response based on the mode 
+of the survey. 
+
+This data is based on the ACS which uses all of these methods and tracks 
+the percent of respondents to each. [Jackson_2007]_ Since the ACS has an order to 
+their survey modes (mail -> telephone -> personal visit), this data should 
+not be used out of that order (e.g., we can **not** assume that a telephone only 
+interview would have an 11.9% responce rate for white people).
```
zmbc, line 1145:
```suggestion
interview would have an 11.9% response rate for white people). 
```

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1079,6 +1079,168 @@ These are not currently modeled.
...
+
+This data is based on the ACS which uses all of these methods and tracks 
+the percent of respondents to each. [Jackson_2007]_ Since the ACS has an order to 
+their survey modes (mail -> telephone -> personal visit), this data should 
+not be used out of that order (e.g., we can **not** assume that a telephone only 
+interview would have an 11.9% responce rate for white people). 
+
+For surveys with mail, telephone, and personal visits, the data below 
+will add to 100%. Since this is unrealistic, we assume that non-response 
+bias for this case matches the census, using the data in the section above.
```
zmbc, line 1149: - On my first read, I didn't understand which section you were referring to. I'd be more specific, maybe "using the rates in the Simulant Omission tables in the decennial census section above."
- I don't think it will matter very much because the census omission rate is so low, but wouldn't it make more sense to include the omission rate all the time (not only when a survey has all three modes)? After all, these are percentages of respondents.
- Using the Census omission rates seems very optimistic to me. The [ACS response rate](https://www.census.gov/acs/www/methodology/sample-size-and-data-quality/response-rates/), even before the pandemic, was ~95%, while the Census omission rate would give 99.76% at the population level. Given that the percentages of respondents are from ACS, it might be better to use an ACS number. [This paper](https://www.census.gov/content/dam/Census/library/working-papers/2002/acs/2002_Griffin_01.pdf) has ACS response rate by race/ethnicity, and the first link has it by state.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1079,6 +1079,157 @@ These are not currently modeled.
...
+    - Mail/Online Only 
+    - 100% - 62.5% = 37.5% non-response 
+  * - 2
+    - Black  
+    - Mail/Online and Telephone  
+    - 100% - (29.7% + 15.1%) = 55.2% non-response 
+  * - 3
+    - Asian 
+    - Mail/Online, Telephone, and Personal Visits 
+    - Use census data above, dependent on age/sex
```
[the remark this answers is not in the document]
zmbc: I also think there is potential confusion here because you say "dependent on age/sex" but race/ethnicity is already specified to be Asian so I assume that effect is included as well. Even if you have to specify a particular age and sex, having a full example for this would be helpful.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Yeah, I think for surveys that don't follow the methodology ordering of mail -> telephone -> personal interview (such as the GSS which does face-to-face) we would need new percentages to show only telephone or only face-to-face. There are a lot of surveys and types of surveys that depending on what we would like might require some additional information to complete. 

For using the ACS vs census response rate. The census does the post-enumeration survey and so has better data on who is missed. I wasn't able to find data on who is missed from the ACS broken down by age/sex/race. If you have data (like what you sent but broken down more) I can swap it.

zmbc: Yes, ACS response rates are only available by state and year.

I understand why we would prefer age/sex/race stratified data, if we think that age/sex/race have significant effects on difficulty of PRL in a way that state does not. That said, it's quite confusing to model survey nonresponse while knowingly underestimating it by 20-80x (e.g. [CPS has an 80% response rate](https://www.bls.gov/osmr/response-rates/home.htm)). Maybe it doesn't matter much for PRL, but it would be good to at least add a limitation/note explaining why we did this and why it (counter-intuitively) doesn't matter too much. For example, applying an un-stratified ACS response rate adjustment would be equivalent to a different sample rate -- but then again, it's possible that the sample rate (and therefore data size) do matter for PRL applications.

Also, maybe I missed it, but did you respond to the second bullet point above?

SylLutze: My reasoning for wanting age/sex/race specific data is that in real life, these things all overlap with stuff that makes PRL harder: 
- Having a higher probability for the same people to be missed (by age/sex/race attributes across observers) will provide an additional challenge since they might appear rarely and not have a match 
- Hispanic names can be longer and a combination of other names in the household 
- Larger family sizes or different family structures that are more difficult to match on taxes 
- Younger people are more likely to move, be in group quarters, or be missed in the census 
- Elderly people are also more likely to be in group quarters, and are less likely to be on taxes as they are not employed 

While I know not all of the above are currently in the sim, I was hoping to design something that would grow more complex with a more complex model. I think a state based missing rate or random missing rate would miss some of this. 

Yes, I am aware that ACS has a very high response rate. I think for adding additional specific survey observers, we might have to provide new data for increased missing folks. My hypothesis in creating this was that a lot of surveys would follow the order of the modes includes in ACS. So we could use this data for a mail/online only survey or for a mail and telephone only survey. This obviously does not apply to all surveys. 

The CPS, as with the GSS, does not follow this mode of survey pattern and so the data is not really applicable. The CPS does telephone and personal visits. Adding this specific survey would require some additional documentation. If there is a specific survey that should be added, we can do that. I will add to the limitations that the ACS has a high response rate. I have tried to clearly state that the data should be used when the modes match only. 

I will apply the census omission rate to all modes.

zmbc (approved):

PR #1015, opened 2022-10-11
Adding CPS observer

SylLutze:
Hi all, I have tried to do this in the simplest way possible to see what adding a new survey would be like at the easiest based on what we have. My hope is future surveys would fit more nicely into our "by mode" framework, but maybe that is not realistic. 

Let me know if this seems reasonable to you with the assumptions or not.

### DIFF
--- a/docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
+++ b/docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
@@ -1321,9 +1321,9 @@ These are not currently modeled.
 
 The ACS will be used for V&V testing. It is defined as: 
 
-- Sample rate of (12,000 / 350 million)% 
+- Sample rate of 12,000 households nationwide  
 - Sample will be stratified by state to ensure an even distribution 
-- Sample at each time step (approximates monthly) for 12 time steps 
+- Sample at each time step (approximates monthly)
 - **Not** longitudinal (independent samples) 
 - Includes mail/online, telephone, and personal visits 
 
@@ -1336,9 +1336,9 @@ added here.
 
 To create this survey: 
 
-- Sample rate of (60,000 / 350 million)% 
+- Sample rate of 60,000 households nationwide 
 - Sample will be stratified by state to ensure an even distribution 
-- Sample at each time step (approximates monthly) for 12 time steps 
+- Sample at each time step (approximates monthly)  
 - **Not** longitudinal (independent samples) 
 - This survey utilizes personal visits and phone calls. As this does not fit into the framework above, we will use the values for mail/online, telephone, and personal visits and then apply an overall non-response rate of 27.6%. This additional risk of non-response will be added to all simulants regardless of race/ethnicity, age, or sex 
 
--- a/docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
+++ b/docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
@@ -1317,7 +1317,7 @@ These are not currently modeled.
 #. The ACS data is for a survey that has an unusually high response rate. This data was used as it could give an estimate for mail only or mail and telephone only data. However, this has limitations. Other surveys might have lower response rates and should be handled separately. 
 #. By replicating the census omission rate for the ACS observer, we are limiting the non-response rate below what we might expect. This will lead to overcounting in the ACS. 
 
-**Initial Survey - American Community Survey** 
+**Initial Survey - American Community Survey (ACS)** 
 
 The ACS will be used for V&V testing. It is defined as: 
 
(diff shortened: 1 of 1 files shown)


zmbc (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1327,6 +1327,27 @@ The ACS will be used for V&V testing. It is defined as:
...
+The CPS is a survey run by the Census Bureau and gathers data about the 
+labor force, employment and unemployment, demographics, earnings, and 
+more information. It is an important survey and therefore is being 
+added here. 
+
+To create this survey: 
+
+- Sample rate of (60,000 / 350 million)% 
+- Sample will be stratified by state to ensure an even distribution 
+- Sample at each time step (approximates monthly) for 12 time steps
```
beatrixkh, line 1341: if the simulation is to run longer than 12 timesteps, which 12 timesteps should be sampled from?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1327,6 +1327,27 @@ The ACS will be used for V&V testing. It is defined as:
...
 - Includes mail/online, telephone, and personal visits 
 
+**Survey - Current Population Survey (CPS)** 
+
+The CPS is a survey run by the Census Bureau and gathers data about the 
+labor force, employment and unemployment, demographics, earnings, and 
+more information. It is an important survey and therefore is being 
+added here. 
+
+To create this survey:
```
beatrixkh, line 1337: which columns / what data should be included in this survey?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1327,6 +1327,27 @@ The ACS will be used for V&V testing. It is defined as:
...
+more information. It is an important survey and therefore is being 
+added here. 
+
+To create this survey: 
+
+- Sample rate of (60,000 / 350 million)% 
+- Sample will be stratified by state to ensure an even distribution 
+- Sample at each time step (approximates monthly) for 12 time steps 
+- **Not** longitudinal (independent samples) 
+- This survey utilizes personal visits and phone calls. As this does not fit into the framework above, we will use the values for mail/online, telephone, and personal visits and then apply an overall non-response rate of 27.6%. This additional risk of non-response will be added to all simulants regardless of race/ethnicity, age, or sex
```
beatrixkh, line 1343: could you explain what "use the values for mail/online, telephone, and personal visits" means? i.e., where do those values come from, and how are we using them? does that mean we will sample people at rate X, where X is the rate of any response, regardless of if it was mail/online, telephone, personal visit?

rmudambi, line 1343: There has been discussion about having the simulation's observers output the "true" state of the world and then having processes that operate outside the simulation itself use that data to generate realistic census data, household surveys, etc. enabling all the noising and sampling rates to be configurable. If that is the approach we are moving in, would we want response rate and/or sample rate to be configurable in that step? If so, we'd need to record more data than we'd expect the survey to actually record.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1327,6 +1327,27 @@ The ACS will be used for V&V testing. It is defined as:
...
+**Survey - Current Population Survey (CPS)** 
+
+The CPS is a survey run by the Census Bureau and gathers data about the 
+labor force, employment and unemployment, demographics, earnings, and 
+more information. It is an important survey and therefore is being 
+added here. 
+
+To create this survey: 
+
+- Sample rate of (60,000 / 350 million)%
```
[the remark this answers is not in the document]
SylLutze: Great callout, it is households, but that is not written here

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Good question - I have made this indefinite (through the length of the simulation) rather than for a defined year. I think this could be something we adjust in scenarios if we don't need all of the information though.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: This is a good question and @aflaxman would like your thoughts. 

My plan was to have response rate/sample rate in the observer and then primarily add noise afterwards. Noise would be things like misspelling names/addresses, etc. 

But I think we in general need to better define what should happen in the simulation, in the observer and outside of both. For example duplicating people - is that in the simulation or the observer or outside of it?

SylLutze:
> Because a lot of the implementation details are missing, as this PR stands it's not something we would be able to implement --- but perhaps you want to get this minimal idea into the docs, and update it later once you are ready for us to implement it? it might be good to add some sort of tag to each feature we start to document but that is not yet expected to be reflected in the model
> 
> if you're able to answer all of the questions i left, though, i think we would be able to implement --- whatever your priority is! we still have plenty of things to work on on our end, as of now.

Hey, I think this is confusing because it was spread across 2 PRs. This is the [PR](https://github.com/ihmeuw/vivarium_research/pull/1004) that includes the other information you asked for. There was a "general survey" set of information and then this is for a specific survey. Let me know if there is still information that is missing!

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
aflaxman: Agree. Some things will need to be in the simulation, and some things will be easier if they are in the simulation.  Everything else (e.g. OCR errors), let's do as post-processing after the sim has been run.

aflaxman (approved):

PR #1016, opened 2022-10-11
Generalize ACS sampling perturbation

zmbc:


docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1497,6 +1479,63 @@ simulants living in military group quarters we will set their employer
...
+likely to be a significant concern even at the substantial sample size (>15 million)
+of the ACS PUMS.
+
+In different components of the simulation, we sample different entity types from the PUMS:
+households, group quarters people, or non-GQ people (individuals living in households).
+The perturbation process is similar no matter the entity type being sampled.
+
+To perturb PUMA, **50% of the time,** we replace the PUMA with a sample from
+an unconditional PUMA distribution; in other words, we use the PUMA value of
+a second independently sampled entity.
```
beatrixkh, line 1513: can you go into a little more detail for this sampling process?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1497,6 +1479,63 @@ simulants living in military group quarters we will set their employer
...
+If we are sampling from a subset of the ACS PUMS, for example recent immigrants
+consistent with a non-reference-person move, **80% of the perturbations**
+(40% of all entities sampled, including those without perturbed PUMA)
+sample a replacement PUMA value from the same subset.
```
beatrixkh, line 1517: i think you mentioned these docs aren't yet ready for implementation, but when you end up going over them again for implementation, can you specify which subsets (or what partition?) or ACS PUMS you will want to use?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1497,6 +1479,63 @@ simulants living in military group quarters we will set their employer
...
+The remaining **20%** (10% of all entities sampled, including those without perturbed PUMA)
+receive a replacement PUMA value sampled from the entire ACS PUMS data.
+(If the primary sampling is already using the entire ACS PUMS, all
+replacement PUMA values are also sampled from the full dataset.)
+This PUMA sampling uses the same entity type, and the same weights, as the primary sampling.
+All simulants in a single sampled household should receive the same replacement PUMA value.
+
+To perturb age, we generate an age shift for each entity sampled.
+This is a random draw from a normal distribution with mean 0 and standard deviation
+10 years.
```
beatrixkh, line 1527: do you want this applied to everyone in group quarters, too? i'm assuming you wouldn't want this for people in college dorms?

beatrixkh, line 1527: should this same approach be used for everyone in group quarters? i assume you don't want to perturb the ages of those in college dorms (or the military? maybe nursing homes?) too much

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1497,6 +1479,63 @@ simulants living in military group quarters we will set their employer
...
+This age shift is added to the simulant's age (if sampling individuals) or to the
+ages of all simulants in the household (if sampling households).
+Using a single age shift for an entire household helps prevent illogical combinations of
+age and relationship.
+Finally, all age values are clipped so that they are between 0 and 125 years, inclusive.
```
beatrixkh, line 1532: i haven't thought too much about this --- but might it result in a pile-up at age 0? (probably not one at age 25?)

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1497,6 +1479,63 @@ simulants living in military group quarters we will set their employer
...
+
+To perturb PUMA, **50% of the time,** we replace the PUMA with a sample from
+an unconditional PUMA distribution; in other words, we use the PUMA value of
+a second independently sampled entity.
+If we are sampling from a subset of the ACS PUMS, for example recent immigrants
+consistent with a non-reference-person move, **80% of the perturbations**
+(40% of all entities sampled, including those without perturbed PUMA)
+sample a replacement PUMA value from the same subset.
+The remaining **20%** (10% of all entities sampled, including those without perturbed PUMA)
+receive a replacement PUMA value sampled from the entire ACS PUMS data.
```
beatrixkh, line 1519: should these be state-specific?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
beatrixkh: oh wait --- do you mean that:

on sim initialization, 100% of the perturbations sample from the entire state
on immigration, 80% will go to PUMAs that have recent immigrants in the data, and 20% will go to a random PUMA in the state?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1497,6 +1479,63 @@ simulants living in military group quarters we will set their employer
...
+
+In different components of the simulation, we sample different entity types from the PUMS:
+households, group quarters people, or non-GQ people (individuals living in households).
+The perturbation process is similar no matter the entity type being sampled.
+
+To perturb PUMA, **50% of the time,** we replace the PUMA with a sample from
+an unconditional PUMA distribution; in other words, we use the PUMA value of
+a second independently sampled entity.
+If we are sampling from a subset of the ACS PUMS, for example recent immigrants
+consistent with a non-reference-person move, **80% of the perturbations**
```
beatrixkh, line 1515: should this 80% be chosen uniformly at random?

also --- will the PUMA data that we sample from be the same data that we used for population initialization? are there any other sampling details that are important? do you expect all PUMAs to be represented in that data, for example? Or do you expect us to have to do any filtering to remove PUMAs that aren't residential? (im not sure if such a thing exists)

beatrixkh (commented):
I can't remember if you said in RT that this was not yet a piece of documentation that you want implemented --- if that is the case, could you add some sort of flag to this section of the docs that it's not yet ready for implementation?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Yes, uniform probability, and yes, sampling from the same data as initialization. I expect all PUMAs to be represented in that data.

There may (in real life) be PUMAs that skew heavily toward or away from GQ -- we'll capture this a bit through this "subset" concept that I clearly need to explain better, but otherwise we will not worry about it.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: No -- but this comment made me realize for the first time that the PUMA code itself is not unique between states. Everything I wrote here assumed that "PUMA" meant the globally-unique code. I will clarify this.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Yes -- I thought this was an inevitable limitation but now I think there may be an easy way to avoid it, so I may update this.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: For now, yes. Specific GQ type (more specific than institutional/non-institutional) is already assigned randomly without respect to age, which means there may already be some weirdness here.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: It's almost that. 80% will go to PUMAs sampled (with weighting) from the PUMA values of recent immigrants **of that move type (e.g. non-reference-person move)**. I will add examples here (and possibly in the initialization and immigration sections above) to make this clearer.

zmbc:
@beatrixkh Thank you for your review. The intention here is to make this implementable, but I clearly need to make improvements to get there! I will work on responding to your questions.

aflaxman (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1529,6 +1512,151 @@ simulants living in military group quarters we will set their employer
...
+''''''''''
+
+After sampling a household to add to the simulation, whether at population initialization or
+from international immigration in a household move, the following steps are **always** performed.
+
+#. An age shift is generated by taking a random draw from a normal distribution with mean 0
+   and standard deviation 10 years.
+#. The age shift is added to the age values of all individuals in the household.
+#. If any age values in the household exceed 125 years, they are set to 125 years.
+#. Any individuals with negative age values are removed from the household.
```
beatrixkh, line 1633: won't this result in losing babies a high percentage of the time (and young children a lower percentage of the time, etc)?

aflaxman:
Yeah, I was expecting Zeb to round up the negative ages to zero.

beatrixkh (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Yes, and @aflaxman I think this is better than rounding to 0, which creates a spike at 0 in the age distribution.

I could see an argument for using truncation as I did in the individual case, based on the minimum age in the household, but I don't think it's obviously better.

One way to interpret this is as "what the household looked like X years ago" (assuming no moves, deaths, etc since then).

zmbc: @beatrixkh After discussion with Abie, we have switched to clipping the age to 0 when it is negative. I've updated the docs.

PR #1018, opened 2022-10-12
Creating BMI risk effects section

SylLutze:
It is my first time making a risk effects section so please let me know what I am missing.


docs/source/gbd2019_models/risk_effects/bmi/index.rst
```diff
@@ -0,0 +1,312 @@
...
+below. For the CVD model, BMI will affect ischemic heart disease, 
+ischemic stroke, and heart failure. There were no updates to GBD 
+modeling of BMI for these causes. 
+
+RRs were reported per 5 kg/m\ :sup:`2`\  increase in BMI above the TMREL value 
+(20–25 kg/m\ :sup:`2`\), calculated as in the equation below: 
+
+:math:`\text{RR(x)} = {\text{RR}_0}^{\frac{\max\left((x-\text{TMREL}), 0\right)}{\text{5 kg/m2}}}`
+
+Where RR(x) is the RR at exposure level x and RR\ :sub:`0`\  is the increase in RR for each 5 kg/m\ :sup:`2`\ above the TMREL. We used DisMod-MR 2.1 to pool effect sizes from included studies and generate a dose-response curve for each of the outcomes associated with high SBP. The tool enabled us to incorporate random effects across studies and include data with different age ranges. RRs were used universally for all countries and the meta-regression only helped to pool the three major sources and produce RRs with uncertainty and covariance across ages taking into account the uncertainty of the data points.
```
alibow, line 75: This is a small point, but I think "GBD" may be slightly more clear than "we" because I always interpret "we" to mean simulation science in these docs!

SylLutze: Hahah great catch! I think Catherine wrote this actually and meant "we" to be GBD because that is "we" for her. But yes, I agree this could be more clear as "GBD"

SylLutze:
> How are we accounting for mediation between BMI -> SBP/FPG/cholesterol -> IHD/other outcomes? The GBD mediation matrix for 2019 is linked on this HUB page https://hub.ihme.washington.edu/display/GBD2020/GBD+2021+Risk+factors+mediation

Great call out and I should have added this. We were going to not include mediation because the scenarios we have don't change BMI and SBP/LDL-C simultaneously. So one intervention changes only SBP and LDL-C and another intervention changes only BMI. The idea was that then we could avoid having mediation. Does that make sense to you as well? 

I have added some information including this rationale and why we are skipping it here.

alibow:
> > How are we accounting for mediation between BMI -> SBP/FPG/cholesterol -> IHD/other outcomes? The GBD mediation matrix for 2019 is linked on this HUB page https://hub.ihme.washington.edu/display/GBD2020/GBD+2021+Risk+factors+mediation
> 
> Great call out and I should have added this. We were going to not include mediation because the scenarios we have don't change BMI and SBP/LDL-C simultaneously. So one intervention changes only SBP and LDL-C and another intervention changes only BMI. The idea was that then we could avoid having mediation. Does that make sense to you as well?
> 
> I have added some information including this rationale and why we are skipping it here.

Ok, great, sounds good! Is there a separate plan for FPG, then? It looks like there may be interventions that simultaneously affect BMI and FPG in the concept model so just want to check!

SylLutze:
> > > How are we accounting for mediation between BMI -> SBP/FPG/cholesterol -> IHD/other outcomes? The GBD mediation matrix for 2019 is linked on this HUB page https://hub.ihme.washington.edu/display/GBD2020/GBD+2021+Risk+factors+mediation
> > 
> > 
> > Great call out and I should have added this. We were going to not include mediation because the scenarios we have don't change BMI and SBP/LDL-C simultaneously. So one intervention changes only SBP and LDL-C and another intervention changes only BMI. The idea was that then we could avoid having mediation. Does that make sense to you as well?
> > I have added some information including this rationale and why we are skipping it here.
> 
> Ok, great, sounds good! Is there a separate plan for FPG, then? It looks like there may be interventions that simultaneously affect BMI and FPG in the concept model so just want to check!

Good point, @aflaxman I thought we had removed the need for mediation? But the current lifestyle intervention is set up to affect both BMI and FPG. Thoughts?

alibow:
> > > > How are we accounting for mediation between BMI -> SBP/FPG/cholesterol -> IHD/other outcomes? The GBD mediation matrix for 2019 is linked on this HUB page https://hub.ihme.washington.edu/display/GBD2020/GBD+2021+Risk+factors+mediation
> > > 
> > > 
> > > Great call out and I should have added this. We were going to not include mediation because the scenarios we have don't change BMI and SBP/LDL-C simultaneously. So one intervention changes only SBP and LDL-C and another intervention changes only BMI. The idea was that then we could avoid having mediation. Does that make sense to you as well?
> > > I have added some information including this rationale and why we are skipping it here.
> > 
> > 
> > Ok, great, sounds good! Is there a separate plan for FPG, then? It looks like there may be interventions that simultaneously affect BMI and FPG in the concept model so just want to check!
> 
> Good point, @aflaxman I thought we had removed the need for mediation? But the current lifestyle intervention is set up to affect both BMI and FPG. Thoughts?

Does the intervention have a direct effect on FPG? Or is it just an effect mediated by BMI? If it's the latter, we could probably just remove the intervention effect on FPG and use total effects of BMI on all of the outcomes and avoid mediation that way. But it looks like then we might lose the intervention impacts on outcomes that are affected by FPG and not BMI, which seem to include PAD and HF from my review of the concept model. Otherwise though it seems to me like we'd need to account for mediation of BMI on modeled outcomes by FPG.

alibow:
And for what it's worth I think we should be able to relatively easily adjust for mediation by FPG! It would just involve utilizing the appropriate GBD mediation factor (linked in this thread) and then recalculating the RRs according to this equation `RR_adjusted = (1 - MF) * (RR_unadjusted - 1) + 1`

alibow:
In other words, it should just take some careful attention to which effect sizes to use on the research side (which may depend on how the intervention effects are structured!), but mediation implementation should only involve data changes for the engineers, not structural changes. 

image

SylLutze:
Maybe it would be helpful to chat about this...  I see two options that would work, let me know if this makes sense to you: 

1. Have the intervention only impact BMI, not FPG. Then I think we could actually remove FPG entirely from the model since no intervention would affect it directly. This would be a simplification over what was done previously, but might be okay. We are already removing the 1:1 PAF of FPG and diabetes (as diabetes is not in the model). This would also change how we enrolled people in the intervention, since it was based on FPG. 

2. Have the intervention affect FPG and BMI, add in the mediation effect between these two only.

SylLutze:
One other note/complication. We are creating RRs for heart failure from literature, not GBD. I doubt that we would be able to find readily applicable MF values for this? Are those something that is reported outside of GBD land? How would we approach this?

alibow:
> Maybe it would be helpful to chat about this... I see two options that would work, let me know if this makes sense to you:
> 
> 1. Have the intervention only impact BMI, not FPG. Then I think we could actually remove FPG entirely from the model since no intervention would affect it directly. This would be a simplification over what was done previously, but might be okay. We are already removing the 1:1 PAF of FPG and diabetes (as diabetes is not in the model). This would also change how we enrolled people in the intervention, since it was based on FPG.
> 2. Have the intervention affect FPG and BMI, add in the mediation effect between these two only.
> 
> I think 2 is the better approach. Also, the only causes left in the model (it is the smaller one on vivarium) are affected by both BMI and FPG.



> Maybe it would be helpful to chat about this... I see two options that would work, let me know if this makes sense to you:
> 
> 1. Have the intervention only impact BMI, not FPG. Then I think we could actually remove FPG entirely from the model since no intervention would affect it directly. This would be a simplification over what was done previously, but might be okay. We are already removing the 1:1 PAF of FPG and diabetes (as diabetes is not in the model). This would also change how we enrolled people in the intervention, since it was based on FPG.
> 2. Have the intervention affect FPG and BMI, add in the mediation effect between these two only.
> 
> I think 2 is the better approach. Also, the only causes left in the model (it is the smaller one on vivarium) are affected by both BMI and FPG.

Yeah, those are the two approaches I see too! Looking at the smaller concept model instead makes it seem like we'd lose less by excluding FPG. Maybe how the intervention effects are reported will be informative. 

Can we get a sense of if the magnitude of the total effect of the intervention on FPG is much greater than the expected mediated effect through BMI? If so, then we might want to include it, but if not we could likely drop it.

The effect of BMI on FPG isn't directly saved as a GBD outcome, but the CVD modelers may have a reference they use for it. Otherwise, it can be back-calculated from the RR and MF in GBD (see the figure here). 

image

Then we can try to use this as a sort of order-of-magnitude comparison for the total versus direct intervention effect on FPG. Essentially, we'd compare the observed (total) effect of the intervention on FPG compared to the expected mediated effect of the intervention on FPG through BMI, as the intervention effect on BMI * the expected effect of BMI on FPG (delta from above).

alibow:
> One other note/complication. We are creating RRs for heart failure from literature, not GBD. I doubt that we would be able to find readily applicable MF values for this? Are those something that is reported outside of GBD land? How would we approach this?

We could calculate our own mediation factors with our own data sources! It's essentially just the proportion of total effect attributable to the mediated pathway, so we just need effect estimates for risk->mediator, risk->outcome, and mediator->outcome to calculate it. We could probably use some combination of GBD and literature sources here if needed.

SylLutze:
Okay, I have done some digging this morning. It might be nice to chat today about this when you have a min? Just to make sure I am understanding everything and not over simplifying.  

Here is what I am thinking: 
- I looked more in depth at the intervention trial results and FPG increases over the trial (higher = closer to diabetes) so I don't think it makes sense to model it as an effect of the intervention 
- However, I still want to include it in the model since it is a qualification to get enrolled in the intervention (designed for pre-diabetic people) 
- Since it won't be directly affected by the intervention, I am thinking I don't need to do risk mediation?

alibow:
Commenting this figure that Syl and I talked through just for documentation/reference in the above thread.
mediation drawio

docs/source/gbd2019_models/risk_effects/bmi/index.rst
```diff
@@ -0,0 +1,328 @@
...
+Risk Outcome Pair #1: Ischemic heart disease
+++++++++++++++++++++++++++++++++++++++++++++
+
+:ref:`See ischemic heart disease documentation <2019_cause_ihd>`
+
+The relative risks apply to the incidence rates of acute 
+myocardial infarction and stable angina. They should be 
+applied using the formula: 
+
+incidence(i) = incidence*(1-PAF\ :sub:`r108`\)*RR^{max((BMI_i - TMREL),0)/5}
```
aflaxman, line 235: I think this RR should be applied to arrow 1 (from S1 to AMI), arrow 3 (from P to AMI), and arrow 4 (from S2 to SA).  If I'm wrong about that, I won't be too surprised, but whatever we decide, we should be precise about which incidence we are referring to here.

docs/source/gbd2019_models/risk_effects/bmi/index.rst
```diff
@@ -0,0 +1,328 @@
...
+incidence(i) = incidence*(1-PAF\ :sub:`r108`\)*RR^{max((BMI_i - TMREL),0)/5} 
+
+The association was evaluated at the cause level, but the 
+associations should be applied to the incidence rates for 
+both nonfatal components of ischemic heart disease. The 
+relative risk for GBD 2019 is for a 5-unit increase in BMI.
+
+PAFs and relative risks can be pulled using the following code: 
+
+rrs = get_draws(gbd_id_type='rei_id', gbd_id=108, source='rr', year_id=2019, gbd_round_id=6, status='best', decomp_step='step4')
```
aflaxman, line 244: this will look better as a code block

docs/source/gbd2019_models/risk_effects/bmi/index.rst
```diff
@@ -0,0 +1,328 @@
...
+The association was evaluated at the cause level, but the 
+associations should be applied to the incidence rates for 
+both nonfatal components of ischemic heart disease. The 
+relative risk for GBD 2019 is for a 5-unit increase in BMI.
+
+PAFs and relative risks can be pulled using the following code: 
+
+rrs = get_draws(gbd_id_type='rei_id', gbd_id=108, source='rr', year_id=2019, gbd_round_id=6, status='best', decomp_step='step4') 
+
+pafs = get_draws(gbd_id_type=['rei_id', 'cause_id'], gbd_id=[108, 493], source='burdenator', measure_id=2, metric_id=2, year_id=2019, gbd_round_id=6, status='best', decomp_step='step5')
```
aflaxman, line 246: this will look better as a code block

docs/source/gbd2019_models/risk_effects/bmi/index.rst
```diff
@@ -0,0 +1,328 @@
...
+
+Risk Outcome Pair #2: Ischemic stroke
++++++++++++++++++++++++++++++++++++++
+
+:ref:`See ischemic stroke documentation <2019_cause_ischemic_stroke>`
+
+The relative risks apply to the incidence rates of acute 
+ischemic stroke. They should be applied using the formula: 
+
+incidence(i) = incidence*(1-PAF\ :sub:`r108`\)*RR^{max((BMI_i - TMREL),0)/5}
```
aflaxman, line 256: I think this should apply to arrow 1 and 3 in the IS cause model diagram https://vivarium-research.readthedocs.io/en/latest/gbd2019_models/causes/ischemic_stroke/index.html#cause-model-diagram

aflaxman (approved):

docs/source/gbd2019_models/risk_effects/bmi/index.rst
SylLutze: Thanks! Added

docs/source/gbd2019_models/risk_effects/bmi/index.rst
SylLutze: Thanks! Added

alibow:
> @alibow including the updates with mediation factors - does this look right?

Yes, these updates all look good to me! Just one note that I think we should indicate where the mediation factors came from! Maybe link to the GBD 2019 mediation matrix for those two values and then indicate how we calculated the HF one and what sources we used. Otherwise I think it's ready to merge!

PR #1019, opened 2022-10-13
Adding matching within names

SylLutze:


docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -903,6 +905,15 @@ sources described above to create conditional samplers for first name
 and last name based on soundex.  Perhaps measure of success is to look
 at entropy of character n-gram distribution.
 
+To simulate naming after a parent or family member, have 5% of children 
+born in the simulation be named after a parent that matches their 
+gender. If there is not a parent that matches their gender, or that 
+parent is not tracked, they will have the name from another relative
```
zmbc, line 911: We may not know who in the household is a relative of this simulant, if the person who gives birth to them is not a relative of the reference person. I don't think what you've written here is incorrect, but you might want to spell this out more explicitly.

My understanding: 5% chance -> if tracked parent of same sex, assign that first name -> if relative of ref person and there is another relative of ref person of the same sex, assign that first name -> assign a random first name.

zmbc (approved):

aflaxman (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Yeah, this looks cleaner. I made the text follow this flow

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -903,6 +905,15 @@ sources described above to create conditional samplers for first name
 and last name based on soundex.  Perhaps measure of success is to look
 at entropy of character n-gram distribution.
 
+To simulate naming after a parent or family member, have 5% of children 
+born in the simulation be named after a parent that matches their 
+gender. If there is not a parent that matches their gender, or that
```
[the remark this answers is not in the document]
SylLutze: Good callout

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
beatrixkh: I agree with Zeb

At the moment, the only defined "parents" that exist in the sim are the mothers of childs born in the simulation.

If we want 5% of children to take on a parent's name, and you guys are also planning on coming up with a way for us to assign parents, we should hold off on assigning 5% of children their parents names until you've come up with a method for assigning parents to children.

however, if you guys end up deciding you don't want to build in structure that links parents and children, we'll want a more specific set of rules for this. e.g.:
- for anyone in a household with the relationship to head of household "biological child" or "adopted child", if they share the sex of the "Reference person", then have a 5% chance of getting the same first name as the reference person
^ though this will miss a lot of the parent-child dyads you were hoping to catch

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -903,6 +905,15 @@ sources described above to create conditional samplers for first name
 and last name based on soundex.  Perhaps measure of success is to look
 at entropy of character n-gram distribution.
 
+To simulate naming after a parent or family member, have 5% of children 
+born in the simulation be named after a relative that matches their 
+sex. To do this: 
+
+#. If there is a tracked parent of the same sex, assign the same first name 
+#. If there is a relative of the same sex, assign the same first name. Note that this might be a relative of the reference person (i.e., a male child has a tracked mother, but that tracked mother has a tracked father) 
+#. If none of these are available, assign a random name
```
beatrixkh, line 914: Could you define "tracked parent" here?

beatrixkh: It might be helpful to separate two concepts:

We have a "relation to household head" attribute for every simulant, and in specific cases (such as ["reference person" and "biological child"] pairs) we can derive information about the relation between two sims, but these are sort of implied relationships -- and they get murky, for example, if we have ["reference person", "parent", "parent"] or other such collections of people we might assume to be married, children of, etc.

We also "track" mothers of children born in the sim. this means that the child sim has an attribute called "parent_id", and this attribute links directly to the unique identifier associated with the mother. I 

I think you guys are considering creating more of these direct links for simulants on initialization, and if you do that, it will be much more straightforward to establish who a "tracked parent" is. however, in that case, i think we should hold off on implementing this until you've decided how to implement "tracked parent"

in the case that you _dont_ end up creating maps from children to parents on initialization, then we'll need a more specific set of rules for who parents are in order to implement this. for example:
- two people who live in the same household and have ["reference person", "biological child"], ["reference person", "adopted child"],
- two people who live in the same household and have ["reference person", "parent"] (if there are multiple parents of the same sex, we would want a way to choose which parent
- grandparent/parent
- grandchild/child
- etc

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -848,7 +848,9 @@ North Carolina, filename VR_Snapshot_20220101.txt; see
 2022_06_02b_prl_code_for_probs_of_spaces_and_hyphens_in_last_and_first_names.ipynb
 for computation details.)
 
-
+For now, assign all simulants within the same household the same last name. 
+This is an oversimplification as some might not be related, but will work 
+for the simple model.
```
beatrixkh, line 853: it would not be too complicated for us to exclude anyone with "relation to head of household" "roommate" or "other nonrelative".

i also assume (and if so can you include) that you don't want this to be done to anyone in group quarters?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: @SylLutze correct me if I'm wrong, but I read "tracked parent" here to refer exclusively to the existing mother ID and a (possibly implemented in future) father ID. So for now, the "tracked parent" line in this logic would only ever happen when a female child is born during the simulation.

SylLutze: Yes, I wrote this meaning it to only apply to new people born in the simulation. I intended "tracked parent" as the mother we are tracking right now

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Yeah, I will add GQ for sure! Good catch. We were discussing roommate/other non-relative. If it is relatively easy I can add it in

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
beatrixkh: ohh i see. that makes sense.

could you define for case #2 how the relative is determined and chosen? this is (unfortunately) i think rather complicated. if the mother has a male child, and the household structure is:

- reference person
- sister
- brother
- brother
- parent-in-law
- grandparent

^ then if she is the sister, we need to know how to choose which male relative's name to use
but in the case she is, say, "parent", we would need a different set of rules for which male relative's name to use

PR #1021, opened 2022-10-14
Emigration

zmbc:
**Note to self: Do not merge this before #1016!**


docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -678,9 +689,120 @@ This could be done by assigning unique (or practically unique, with very low pro
...
+use the published NIM values instead of directly estimating emigration.
+
+The NIM values are not published fully stratified.
+Out of the available stratifications, we chose to use the values stratified
+by (broad categories of) race/ethnicity, because these are most likely to have
+PRL implications.
+
+Inspired by the methodology of the PopEst team at the Census Bureau,
+we further stratify emigration by assuming that **emigrants** have the same
+characteristics as **immigrants**.
```
SylLutze, line 727: This is interesting to me and not intuitive. I would imagine that different folks might want to enter and exit the US, but I assume the Census did this for a review!

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -678,9 +689,120 @@ This could be done by assigning unique (or practically unique, with very low pro
...
+''''''''''''''''''''''''''
+
+The at-risk population for GQ person moves is all simulants living in non-GQ households in the US, except for those who are a household reference person.
+This at-risk population should be stratified by age group, sex, race/ethnicity, nativity (born in or outside the US),
+and US state.
+On each time step, within each stratum, the corresponding non-reference-person move emigration rate **per year of person-time**
+should be applied to sample simulants to emigrate.
+The simulant is removed from the household (they may be given a blank or placeholder household ID) and the
+rest of the household is unaffected by this event.
+
```
SylLutze, line 792: Do we want to consider any kind of relationships in this? For example, it might be less likely for a spouse to emigrate alone than another person. But this is likely unneeded complication.

SylLutze (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Yes, it's not necessarily intuitive. The PopEst team does this in their own modeling with essentially no justification -- they just state that it is an assumption. (In fact, it is a bit more complicated than even that, because for people born in the US they use the set of immigrants *or* between-state movers which seems like an even bigger leap to me.)

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Good point. We will get some of this by age stratification (age is likely correlated with relationship) but the specific example you mention is likely to happen too often.

As I've currently implemented this, we already have serious computational challenges calculating the rates, so adding stratification by relationship would be difficult.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -678,9 +689,120 @@ This could be done by assigning unique (or practically unique, with very low pro
...
+Emigration events are modeled as happening to an at-risk population at a certain rate.
+They are constant across time in the simulation.
+
+Households and individuals selected to have emigration events should remain in the simulation, but their
+state and PUMA attributes should be set to a placeholder value that signifies they are
+no longer in the US.
+Emigrating simulants should also have their address attribute removed, and they should terminate
+employment.
+In the future, we may want to sample some of these simulants for re-entry in the immigration component, but for now
```
aflaxman, line 760:
```suggestion
In the future, we may want to sample some of these simulants for re-entry in the immigration component (or for continued employment), but for now
```

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -678,9 +689,120 @@ This could be done by assigning unique (or practically unique, with very low pro
...
+Then, households within the stratum should be sampled at random for emigration until the desired number of simulants is reached.
+
+GQ person moves
+'''''''''''''''
+
+The at-risk population for GQ person moves is all simulants living in GQ in the US.
+This at-risk population should be stratified by age group, sex, race/ethnicity, nativity (born in or outside the US),
+and US state.
+On each time step, within each stratum, the corresponding GQ person move emigration rate **per year of person-time**
+should be applied to sample simulants to emigrate.
```
aflaxman, line 780: Are you proposing to have the same rate for GQ people, regardless of GQ type?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -678,9 +689,120 @@ This could be done by assigning unique (or practically unique, with very low pro
...
+The at-risk population for GQ person moves is all simulants living in GQ in the US.
+This at-risk population should be stratified by age group, sex, race/ethnicity, nativity (born in or outside the US),
+and US state.
+On each time step, within each stratum, the corresponding GQ person move emigration rate **per year of person-time**
+should be applied to sample simulants to emigrate. 
+
+Non-reference-person moves
+''''''''''''''''''''''''''
+
+The at-risk population for GQ person moves is all simulants living in non-GQ households in the US, except for those who are a household reference person.
```
aflaxman, line 785:
```suggestion
The at-risk population for non-reference-person moves is all simulants living in non-GQ households in the US, except for those who are a household reference person.
```

aflaxman (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Yes, after stratifying by demographics. Using the data, we could split this by institutional/non-institutional.

Should I split it, or add this as a limitation? It seems likely to me that the GQ type (institutional vs non) has a big effect on emigration rates, but it's not clear that this would have substantial PRL implications.

aflaxman:
Add as limitation, because we should not burden the engineers with additional feature requests if we can avoid it.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -678,9 +689,136 @@ This could be done by assigning unique (or practically unique, with very low pro
...
+we calculate the denominator for the rate from a resample of the "at risk" population, with perturbation.
+
+Simulation strategy
+^^^^^^^^^^^^^^^^^^^
+
+Emigration events are modeled as happening to an at-risk population at a certain rate.
+They are constant across time in the simulation.
+
+Households and individuals selected to have emigration events should remain in the simulation, but their
+state and PUMA attributes should be set to a placeholder value that signifies they are
```
beatrixkh, line 761: by "state" do you mean US state, or simulation-state? 

could you specify which columns become placeholders vs. which ones continue to be "live"?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -678,9 +689,136 @@ This could be done by assigning unique (or practically unique, with very low pro
...
+
+Household moves
+'''''''''''''''
+
+The at-risk population for household moves is all simulants living in non-GQ households in the US.
+This at-risk population should be stratified by age group, sex, race/ethnicity, and nativity (born in or outside the US)
+**of the simulant's household's reference person**, as well as US state.
+On each time step, within each stratum, the corresponding household move emigration rate **per year of person-time** should be applied to determine
+a number of **simulants** to emigrate as part of household moves.
+Then, households within the stratum should be sampled at random for emigration until the desired number of simulants is reached.
```
beatrixkh, line 777: should we hit this number exactly?

i think there are three reasonable choices if the desired number of simulants to move = X
- sample at random within the stratum until we're very close to X, then fill in the remaining, say, X-2 by sampling only from 1-2 person households.
- sample at random until we're at Y, where X-Y < size(largest household)
- sample at random until we're at Y, where Y >= X

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -678,9 +689,136 @@ This could be done by assigning unique (or practically unique, with very low pro
...
+rest of the household is unaffected by this event.
+
+Simulation inputs
+^^^^^^^^^^^^^^^^^
+
 .. todo::
 
-    Describe this component.
+  These are placeholder files that I will finalize soon!
+  For now, these are an example of the format/structure.
```
beatrixkh, line 805: thank you for making these!

beatrixkh (approved):
just a couple questions

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: I think option 3 is the simplest to reason about and I don't expect this to matter much. I will add more detail about this.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Oops, I actually meant to remove this note! The files in this PR are the real values now.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: US state; will revise to clarify.

My intention here is that columns besides the ones explicitly mentioned continue to have non-placeholder values. Then, separately, in many of the other components we will need to specify that they only act on simulants in the US. I have done that in this PR. This is a bit tricky and error-prone, but a) I *think* you will need to implement it this way, if it works anything like how death works in Vivarium, and b) we would want certain events to continue to happen, e.g. death, if we ever add the ability to bring simulants back from abroad.

I will spell this out more.

PR #1023, opened 2022-10-18
Adding parents/guardians on initialization

SylLutze:
Hello, some questions I am struggling with: 

- I assign foster children to the person they are living with at the moment. I think the rules here on taxes and moving are rather complicated so this was the easiest method 
- Is there a way to "backwalk" the fertility model to give a more comprehensive picture of the distribution of the ages of mothers? Is this worth it? 
- I am only assigning GQ "college" folks to be in "college". This is wrong, but we don't track who else is a student - any thoughts? 
- Some college students do not receive support from their parents and shouldn't be claimed as a dependent, I think this makes more sense in the tax section though. Maybe parents who aren't filing taxes then won't end up claiming them? 
- We are assigning GQ college students to "parents" not "parent/guardians". Is it worth assigning some to people we know are not likely "parents"? (e.g., some grandparent age folks)


zmbc (commented):
Re: your questions, I think most are fine to keep as-is, but should be listed as limitations.

> Some college students do not receive support from their parents and shouldn't be claimed as a dependent, I think this makes more sense in the tax section though. Maybe parents who aren't filing taxes then won't end up claiming them?

I agree this should be in the tax observer, and I'd propose using an income threshold on the student (assuming that we will eventually add complexity to income). In order to be a dependent, they need to earn <50% of their expenses, and we could make some reasonable assumption about how much expenses are for a college student.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -379,6 +379,72 @@ I will also verify that the household
...
+simulant (reference person and spouse ONLY), then include both as 
+parents/guardians. Otherwise only include the one as a parent/guardian. 
+
+**For a simulant who is 18-24 and in GQ at college:**
+
+Simulant will be randomly assigned to a parent/guardian based on the below rules: 
+
+- 78.5% will be assigned to a parent/guardian within their state. The remainder will be assigned out of state source1_  
+- Match to a person 20 to 45 years older than the child 
+- If child is not mixed/other race, match parent's race. If child is mixed/other race, then assign to a parent of any race
```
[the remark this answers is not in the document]
SylLutze: Good point! Honestly I was thinking all college kids are 18+ but that is definitely not true.

SylLutze:
> I agree this should be in the tax observer, and I'd propose using an income threshold on the student (assuming that we will eventually add complexity to income). In order to be a dependent, they need to earn <50% of their expenses, and we could make some reasonable assumption about how much expenses are for a college student.

I checked, and we do have a limit on income for a dependent already! So I think this will select people out well enough.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -379,6 +379,79 @@ I will also verify that the household
...
+.. _source1: https://www.statista.com/statistics/236069/share-of-us-students-who-enrolled-in-a-college-in-their-own-state/ 
+
+.. _source2: https://nces.ed.gov/programs/coe/indicator/cce/family-characteristics 
+
+
+**Limitations**
+
+#. The foster care system is complex. We have the foster kid assigned within the house they are currently living. If we model the foster care system in more detail, we might improve this at some point. 
+#. We have "parents" fall between 20-45 older than the child. This is an oversimplification. Some parents (especially men) fall outside of this range. Also some age gaps are more common than others. 
+#. The only people who are seen as "in college" are in GQ in college. Plenty of people attend college from home, but we do not track occupation so are not accounting for this.
```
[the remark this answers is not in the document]
SylLutze: We do track employer, but I think it might get complicated. For example, how do we separate a U of Washington employee from a U of Washington student? We would probably need some kind of special "student" tag separate from employer. We could consider adding this to the model though if it is important.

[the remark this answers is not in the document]
zmbc: And, people would work and attend school at the same time. I'd change this sentence to say "education" instead of "occupation" to reflect that we would model it separately.

aflaxman:
I would say “not important”, at least for now.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -379,6 +379,81 @@ I will also verify that the household
...
+- Child is the reference person 
+    * Assign a parent, if available 
+    * Otherwise, assign another relative who is between 20 and 45 years older than the child. If there are multiple, select one of the same race/ethnicity. If there are multiple of the same race/ethnicity, assign at random.
+    * If there are no other relatives in the house, make "N/A"
+
+Once the parent/guardian is assigned, if there is a spouse or unmarried partner 
+for that simulant (reference person and spouse/unmarried partner ONLY), then 
+include both as parents/guardians. Otherwise only include the one as a parent/guardian. 
+
+(note to software engineers: if any of these rules turn out to be surprisingly hard to implement, please be in touch with research --- we have some flexibility in just how we do this!)
```
beatrixkh, line 430: thank you!

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -379,6 +379,81 @@ I will also verify that the household
...
+
+**For a simulant who is below 24 and in GQ at college:**
```
beatrixkh, line 432: do you care about below 24 and in the military?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -379,6 +379,81 @@ I will also verify that the household
...
+
+Simulant will be randomly assigned to a parent/guardian based on the below rules: 
+
+- 78.5% will be assigned to a parent/guardian within their state. The remainder will be assigned out of state source1_
```
beatrixkh, line 436: for now, we have only ever run for Florida. Until we run for multiple states, would it be appropriate to ignore those remaining 21.5%?

beatrixkh (approved):
It would be helpful to choose definite relationships that you want to establish. for example, do you want to track an entity called ["parent"], or both ["mother", "father"], vs ["guardian"] vs ["sim_who_will_claim"]?

^ i think the answer probably depends on if you will want to use this relationship for something other than just tax returns, in which case more specificity might be helpful, or if you want this to strictly be for tax returns, in which case simply tracking something like "sim_who_will_claim" might be better, and then you would be able to include, for example, grandparents

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: I am pretty sure the answer is no, because in order to be claimed as a dependent on taxes, you must be a student.

SylLutze: Yes, as Zeb mentioned its not relevant for taxes since they can't be a dependent, so I didn't include them

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Good point! I will add something about this. I think we can just ignore for now, yes.

SylLutze:
> It would be helpful to choose definite relationships that you want to establish. for example, do you want to track an entity called ["parent"], or both ["mother", "father"], vs ["guardian"] vs ["sim_who_will_claim"]?
> 
> ^ i think the answer probably depends on if you will want to use this relationship for something other than just tax returns, in which case more specificity might be helpful, or if you want this to strictly be for tax returns, in which case simply tracking something like "sim_who_will_claim" might be better, and then you would be able to include, for example, grandparents

This is a good point. I am going to add the name ["guardian"]. Right now this is mainly used for taxes, but I do think there is potential elsewhere. Plus, ["guardian"] is still loose enough to include a grandparent, etc.

PR #1025, opened 2022-10-19
Clarification on missed appointments and screening

SylLutze:


stevebachmeier (approved):

docs/source/concept_models/vivarium_us_cvd/concept_model.rst
```diff
@@ -833,7 +833,7 @@ These covariate values are calculated for each simulant and are then plugged int
  :math:`Both_{i} = exp((-6.26) + (0.018 * SBP_{level}) + (-0.014 * LDL_{level}) + (0.069 * age_{(yrs)}) + (0.13 * sex))` 
 
 Where sex = 1 for men and 2 for women 
-
+and SBP and LDL level refer to the raw values from GBD
```
aflaxman, line 836: are you sure you want the _raw_ value here, and not their current value (which might include treatment)?

aflaxman (approved):

docs/source/concept_models/vivarium_us_cvd/concept_model.rst
SylLutze: This is for initializing who is on treatments, so we can't use the treatment effected values. It is a chicken/egg issue the SE team found.

PR #1027, opened 2022-10-20
ENH: add detail about unique names for employers

aflaxman:


docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1529,6 +1529,17 @@ simulants living in military group quarters we will set their employer
...
 employer_address and employer_zipcode.
 
+Speaking of the names for employers, we will generate names for
+employers based on a conditional random model that Abie developed and
+Jim refactored into the vivarium model.  It should have more details
+documented at some point.  Jim will ensure that all employer names are
+unique for now, although it would be interesting one day to add some
+duplicates.  Businesses can have the same name in real life but
+if/when we do have duplicate names we need to prevent duplicates for
+certain most common names.
```
zmbc, line 1539: What do you mean by this? No duplicates for the biggest employers?

zmbc (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
aflaxman: to be determined, but for example, there should perhaps only be one Postal Service (https://en.wikipedia.org/wiki/The_Postal_Service#United_States_Postal_Service_controversy)

zmbc: Makes sense, but I don't clearly understand that meaning from what you have written here -- how can it be "common" if there is only 1 business with that name?

That's why I was proposing that it should be "... we need to prevent duplicates of the names of the largest employers." Your USPS example makes me think I got that right, but I'm not sure.

SylLutze: I agree with Zeb on this phrasing ^^

aflaxman:
```suggestion
the names of the largest employers
```

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3890,6 +3890,15 @@ Limitations
...
    This likely makes business linking easier than it is in reality.
 
+Speaking of the names for employers, we will generate names for
+employers based on a conditional random model that Abie developed and
+Jim refactored into the vivarium model.  It should have more details
+documented at some point.  Jim will ensure that all employer names are
+unique for now, although it would be interesting one day to add some
+duplicates.  Businesses can have the same name in real life but
+if/when we do have duplicate names we need to prevent duplicates for
+certain most common names.
```
aflaxman, line 3900:
```suggestion
the names of the largest employers.
```

PR #1028, opened 2022-10-25
Standard vivarium results formatting guide

alibow:
Proposal as discussed in the 10/20 research team meeting


docs/source/model_design/concept_model/standard_results_formats/index.rst
```diff
@@ -0,0 +1,197 @@
...
+      * uncovered
+  * - :code:`measure`
+    - string
+    - Description of what is being measured about the entity
+    - * ylds
+      * ylls
+      * deaths
+      * person_time
+      * transition_count
+      * first_moment
```
zmbc, line 120: Are these really measures? They feel like metrics. But, maybe it's fine to be inexact as long as we never want to use them at the same time as the other measures.

For the MM results CSVs, I had a column for the (equivalent GBD) metric. Seeing this makes me wonder if that could be more broadly useful/whether it should be a standard column here. 🤔

docs/source/model_design/concept_model/standard_results_formats/index.rst
```diff
@@ -0,0 +1,197 @@
...
+  * - :code:`entity_type`
+    - string
+    - Type of entity being measured
+    - * cause
+      * rei
+      * {other}
+  * - :code:`entity`
+    - string
+    - Name of entity being measured
+    - :code:`gbd_{entity_type}_name.replace(" ","_").lower()`
```
zmbc, line 131:
```suggestion
    - :code:`gbd_{entity_type}_name.replace(" ","_").lower()` if there is a corresponding GBD entity, otherwise agreed upon between research and engineering teams
```

docs/source/model_design/concept_model/standard_results_formats/index.rst
```diff
@@ -0,0 +1,197 @@
...
+      
+      * measles
+      * child_stunting
+  * - :code:`sub_entity`
+    - string
+    - Name of entity state or category
+    - * measles/susceptible_to_measles
```
zmbc, line 138: Is there ever a matching GBD entity here? e.g. a sequela?

docs/source/model_design/concept_model/standard_results_formats/index.rst
```diff
@@ -0,0 +1,197 @@
...
+
+.. note::
+
+  As a reminder, the definitions of the first and second moments for a continuous exposure measure, :math:`Y`, for our purposes are defined below. 
+
+  .. math::
+
+    \text{First moment} = \sum_{i}Y_i \times \text{person time}_i
+
+    \text{Second moment} = \sum_{i}Y_i \times \text{person time}_i
```
zmbc, line 153: This probably isn't supposed to be the same as the first?

docs/source/model_design/concept_model/standard_results_formats/index.rst
```diff
@@ -0,0 +1,197 @@
...
+  * - :code:`scenario`
+    - string
+    - Simulated scenario name
+    - * baseline
+      * alternative
+      * etc.
+  * - :code:`age`
+    - string
+    - Age group of stratum
+    - :code:`gbd_age_group_name.replace(" ","_").lower()`
```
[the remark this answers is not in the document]
alibow: Great point! I updated the age and year columns to include instructions here. For sex I just added another option of "both", which is the same as the GBD name for sex_id=3.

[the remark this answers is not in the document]
alibow: I figured it was better to actually define age/year start/end instead of "all" just to make sure we know what it represents.

docs/source/model_design/concept_model/standard_results_formats/index.rst
alibow: Good catch! Yes, it's updated now.

docs/source/model_design/concept_model/standard_results_formats/index.rst
alibow: GBD defines measures as deaths/dalys/ylds/etc. and metrics as rates/counts/etc., so I'm thinking that we're consistent in the terminology here? Otherwise I'm not sure I'm following this comment, maybe it's because I'm confused about measure/metric terms. Could you please provide an example of the column from the MM result CSVs?

docs/source/model_design/concept_model/standard_results_formats/index.rst
alibow: Hmm, I think it's possible but not common. One issue is that we'll pretty much always have some sort of susceptible state that won't be included in the GBD sequelae. For many causes I think sequelae won't really be relevant and applies more to symptomology than stage of disease. 

There are cases were we might model sequelae as cause model states (maybe for cancers like shown in the diagram below?), but I think it will probably be rare and the names are typically wordy enough that we might want to rename them anyway. 
image

One thing I did think through though was whether it would be preferable to have either...

A: entity=measles, sub_entities=measles, susceptible_to_measles, recovered_from_measles 

or 

B: entity=measles, subentities=infected, susceptible, recovered

Our current results are structured like A, but that's really because we don't break apart entities and sub-entities. I might update it to B unless you think otherwise!

docs/source/model_design/concept_model/standard_results_formats/index.rst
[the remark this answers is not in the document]
zmbc: That makes sense! But, is it always/usually the case that we have "simulation minimum and **maximum** age values"? I was under the impression that for most of our open cohort sims, we had minimum 0 with no set maximum. In that case, I think it would make sense to consider "All Ages" -> "all_ages" (age_group ID 22) the matching GBD age group, since that is the GBD result we would want to compare with if results were not age-stratified.

docs/source/model_design/concept_model/standard_results_formats/index.rst
zmbc: Makes sense. In MM there was kind of a mapping between sequelae and cause states but it wasn't 1:1. And, great point that we wouldn't want to use the sequela names anyway, because they are so cumbersome.

I like B!

docs/source/model_design/concept_model/standard_results_formats/index.rst
zmbc: It wasn't that interesting (didn't vary), but here is an example of how I linked results to the Rate metric from GBD: https://github.com/ihmeuw/vivarium_research_multiple_myeloma/blob/main/results/china/Multiple%20Myeloma%20Output%20Shell%20Table%20-%20China.csv. The context was that they wanted it to be as similar to GBD outputs as possible.

I don't have much experience with the continuous exposures. My expectation was that the measure would be something like "continuous" and things like mean and SD would be metrics. That's just based on the fact that mean/SD/first-moment/second-moment/etc seem more like rate/count/etc than deaths/dalys/ylds.

I see GBD measure ID 19 is "Continuous" but there are no metrics that bear out my intuition.

All of this only really matters when we are merging with inputs. When you get e.g. mean and SD of a hemoglobin distribution, does it come with a measure and/or metric column? Or is it through an interface sufficiently different from `get_outputs` that there is no such concept?

alibow: Ah, I see what you're saying. Continuous exposure mean and SDs typically come from get_draws as modelable entity IDs that don't really follow these naming conventions. So I think that GBD consistency is less critical here because of that, but still, maybe this is more intuitive? 

measure = continuous_exposure
entity = hemoglobin
sub_entities = first_moment, second_moment

Does that seem like a better fit?

docs/source/model_design/concept_model/standard_results_formats/index.rst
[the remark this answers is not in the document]
alibow: Another good point. I imagine we could have a sim with "adults" that would be something like "20_to_95_plus", so I'll add a note about what to do if there is no age max and then also add a note about all_ages being preferred over 0_to_95_plus. Let me know if you think this doesn't cover it, though!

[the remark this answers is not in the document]
alibow: Oops, coming back to say that we should do "{age_min}_plus" when there is no age max, because that is more consistent with GBD.

docs/source/model_design/concept_model/standard_results_formats/index.rst
zmbc: Yes, and I think it is more consistent with get_draws.

When you get_draws from the modelable entity for a continuous risk exposure, e.g. `get_draws(gbd_id_type="modelable_entity_id", gbd_id=10487, source='stgpr', location_id=6, release_id=6, year_id=2019)`, it comes back with measure_id 19 ("continuous") and metric_id 3 (for some reason, it is considered a rate?). This is also true when you use rei_id with source = "exposure" or "exposure_sd." So, there is some GBD precedent for treating different summary values like "sub-entities" as opposed to measures (or metrics, as it turns out).

zmbc (approved):

SylLutze (approved):

NathanielBlairStahn:
@alibow Looks great, thanks for putting this together! I have a couple comments:

1. For a continuous measure like hemoglobin, I think "first_moment" or "second_moment" should go in the _measure_ column, not the sub-entity column. Why? We should think of the measure column as the "units" of the numerator or denominator of the thing we're computing. In my standardized output processing code, I actually record the measure of the numerator and denominator dataframes whenever I compute a ratio so that I can easily see what the numbers in the new ratio dataframe represent. If it just said "numerator measure = continuous exposure" and "denominator measure = person time," that's not very useful because "continuous exposure" isn't specific enough. But if it said "numerator measure = first moment" and "denominator measure = person time," then I know that represents a mean.

2. It might be worth adding a few more examples of somewhat non-standard things we have dealt with, but probably that should go in a separate PR. For example, how should we label the columns if we want to compute anemia prevalence (anemia is a bit weird because it's a GBD impairment, and we typically compute it from hemoglobin values rather than GBD prevalence)? Or what about things that happen at birth, such as neural tube defects or stillbirths? In that case we'd be using a population count (number of births) for the "Birth" age group rather than person-time like we usually do.

zmbc:
@NathanielBlairStahn point number 1 is [my fault](https://github.com/ihmeuw/vivarium_research/pull/1028#discussion_r1004919169). To me, it doesn't make sense as a measure, and there is some consistency-with-GBD argument for it not being a measure, but if it seems like a measure to both of you, I'm happy to be overruled.

docs/source/model_design/concept_model/standard_results_formats/index.rst
alibow: @NathanielBlairStahn here's some context for the first_moment classification, in case you hadn't seen it because it was hidden.

alibow:
@NathanielBlairStahn I see your point! I had this initially and updated it in response to the conversation with Zeb, but I think it make sense in light of the output processing code you referenced. Let me know if you think otherwise after looking through the existing thread. 

And good point about births - I will work that into this PR because I think it's a pretty important example.

NathanielBlairStahn:
> @NathanielBlairStahn I see your point! I had this initially and updated it in response to the conversation with Zeb, but I think it make sense in light of the output processing code you referenced. Let me know if you think otherwise after looking through the existing thread.
> 
> And good point about births - I will work that into this PR because I think it's a pretty important example.

@alibow @zmbc Thanks for pointing out the existing thread -- I hadn't seen it before posting my comment. I think I still prefer my way over the current version, though I'm open to suggestions if people have better names than "first moment" and "second moment" for the measures, as long as they're sufficiently specific. I'm about to write a reply in the thread with more details about my rationale.

docs/source/model_design/concept_model/standard_results_formats/index.rst
NathanielBlairStahn: > @NathanielBlairStahn point number 1 is [my fault](https://github.com/ihmeuw/vivarium_research/pull/1028#discussion_r1004919169). To me, "first moment" doesn't feel like a measure, and there is some consistency-with-GBD argument for it not being a measure, but if it seems like a measure to both of you, I'm happy to be overruled.

@zmbc @alibow I think it is less critical for Vivarium outputs to be consistent with GBD in the `measure` column than, for example, with cause names, risk names, age groups, or locations, as it is much easier to find the correct match for a measure (which has relatively few possible values in GBD) than for a cause, age group, etc. And as Ali said, GBD is not necessarily consistent about how it stores things that we might want to call a "measure," depending on where we pull the data from.

Also, I'm a bit confused by GBD's distinction between "measure" and "metric"; I think their goals with this distinction are different from our use cases in Vivarium outputs. For example, our outputs from Vivarium are (almost?) always "count space" values, so we should rarely if ever need to specify that an output is a "rate" rather than a "count" since our preferred methodology is to output the numerator and denominator separately. When I was originally thinking through which columns to put in a standardized output table, I considered having three separate columns, `measure`, `metric`, and `units`. After thinking through some examples, I was finding little utility for the `metric` column. And while `units` would be nice, oftentimes it is unnecessary (e.g., when the measure is DALYs, YLLs, or YLDs, which are already units), and otherwise we usually have conventions such as always measuring person-time in person-years. The exception would be something like hemoglobin concentration, which could be measured in g/L, g/dL, mmol/L, etc., so in these cases we'll just have to specify the desired units (consistently with GBD if possible) and remember what they are.

(to be continued...)

docs/source/model_design/concept_model/standard_results_formats/index.rst
```diff
@@ -0,0 +1,242 @@
...
+  * - Person time in the susceptible to measles state
+    - person_time
+    - cause
+    - measles
+    - susceptible
+  * - Transitions from measles to recovered from measles
+    - transition_count
+    - cause
+    - measles
+    - infected_to_recovered
```
NathanielBlairStahn, line 232:
```suggestion
    - infected__to__recovered
```
I suggest we request that the _from_ and _to_ states of a transition be separated by a unique string such as `'__to__'`, so that the two states can be easily extracted using `.split()`, e.g., `states = sub_entity.split('__to__')`. This is not possible with the current Vivarium naming scheme, which has transitions with names like `'diarrheal_diseases_to_susceptible_to_diarrheal_diseases'`. Even if we replace these annoying names with something simpler like `'infected_to_susceptible'`, there's no guarantee that we'll never have a complicated cause model with many cause states (such as the LTBI model), some of which could potentially include the word "to." So my proposal is to use double underscores for the separator, and to never use double underscores in the name of a state.

docs/source/model_design/concept_model/standard_results_formats/index.rst
NathanielBlairStahn: So eventually I decided that the simplest thing to do would be to have a single `measure` column that would describe, as precisely as possible, the quantity being measured by the number we get in the output. The name of the measure should be some sort of physical quantity that can be measured concretely, and it can encompass both the "measure" and the "metric" in GBD's terminology (whatever those are supposed to mean). For example, if we wanted the count of deaths, the measure can just be called "deaths". If for some reason we wanted the simulation to directly output a mortality rate rather than a count, we could call the measure "death_rate". In my output processing code for the [CIFF SAM project](https://github.com/ihmeuw/vivarium_research_ciff_sam/tree/main/model_validation), I assigned new measures such as "prevalence", "transition_rate", or "rate_ratio" to derived quantities I computed from the raw count space outputs.

For a continuous quantity like hemoglobin concentration, it's a little more complicated; there are many possible aspects of the distribution of hemoglobin concentration in the population that we may want to measure. In previous simulations Vivarium has directly output the mean and standard deviation of the Hb concentration, in which case the measures could be called simply "mean" and "standard_deviation" (since those are the quantities that the numbers in the output are measuring). However, for consistency with other Vivarium outputs, and to make it easier to aggregate values over different population strata, I think it's better to output the numerators and denominators we need to compute the mean and standard deviation (if anyone disagrees, please chime in). In trying to describe what numerators we need to compute these (with person-time as the denominator), I came up with "first_moment" and "second_moment", based on the concept of "moment" from [physics](https://en.wikipedia.org/wiki/Moment_(physics)) and [mathematics](https://en.wikipedia.org/wiki/Moment_(mathematics)), so that's what I put in the `measure` column. If anyone has a better description for the quantities/measures defined by [Ali's equations](https://github.com/ihmeuw/vivarium_research/pull/1028/files#diff-27a94db806f9b410773a575674886b27c967e01728d7a689e49e74275fa34395R194-R196), I'm open to other suggestions. (Note the second moment can't be used directly to compute the variance or standard deviation as a ratio, but can be combined with the mean to compute the variance via the formula Var(Y) = E(Y^2) - (EY)^2.)

zmbc: We are definitely bikeshedding here; I'm fine with changing it back. I don't really _love_ either option, and only suggested the other one for more consistency with GBD. But, as we have discussed, with continuous exposures it requires a few steps to link with GBD anyway.

docs/source/model_design/concept_model/standard_results_formats/index.rst
zmbc: This is an example of the downside of having every observer output a file with the same columns -- otherwise, the transition count observer could have two separate columns for this. But I agree that double underscore is better than single.

NathanielBlairStahn: Well I guess we could just have separate columns for the states of a transition, though with this change it will be easy enough to do ourselves. In the CIFF SAM project, since nothing really looked how I wanted, I [preprocessed the data](https://github.com/ihmeuw/vivarium_research_ciff_sam/blob/main/model_validation/ciff_sam_results.py#L159) into the form I wanted, including putting the states of the transitions into separate columns. In the current Vivarium output, the entire transition description is stored in the "measure" column.

NathanielBlairStahn: I think the point of conforming the output to the same format as much as possible is to avoid having to specify lots of special cases to the engineers. Ali and I added the `sub_entity` column as a sort of catch-all that could handle several types of entities such as causes, risks, and transitions, and hopefully it will provide a structure that can also capture new types of outputs that we haven't thought of yet, without us having to think to carefully about what we want. Having a more standardized formate reduces the burden of communication between researchers and engineers, which is often a sticking point. And as long as we understand what we're getting in the output and it's well-structured, it should be easy for us to write preprocessing code to reformat the Vivarium output into a form more suitable for answering research questions.

docs/source/model_design/concept_model/standard_results_formats/index.rst
NathanielBlairStahn: Haha, yeah, all this discussion isn't really necessary to make this one decision, but I realized I had done a lot of thinking about the table design awhile ago, but it was all in my head, so I thought it might be good to document my reasoning and open it up for discussion in case anyone has ideas for improvement.

PR #1031, opened 2022-10-25
Handle deaths of reference people

zmbc:
I decided to do this a more complicated way, in part because I want to allow reference people to migrate out of their household in a future PR.


aflaxman (approved):
Fun!

docs/source/concept_models/vivarium_census_synthdata/reference_person_update_relationship_mapping.csv
```diff
@@ -0,0 +1,241 @@
...
+Other nonrelative,Same-sex partner,Other nonrelative,
+Opp-sex spouse,Biological child,Parent,
+Opp-sex partner,Biological child,Parent,
+Same-sex spouse,Biological child,Parent,
+Same-sex partner,Biological child,Parent,
+Biological child,Biological child,Sibling,"Could be half-sibling, which might not be reported as sibling"
+Adopted child,Biological child,Sibling,Might not be reported as sibling
+Stepchild,Biological child,Other relative,"Assuming step-siblings are not reported as ""brother or sister"""
+Sibling,Biological child,Other relative,
+Parent,Biological child,Other relative,
```
SylLutze, line 58: Why is this other relative but row 60 is grandchild? I think this should also be grandchild. Same comment for adopted child below. Granted if we assign to the oldest this should never happen anyways

SylLutze (approved):

docs/source/concept_models/vivarium_census_synthdata/reference_person_update_relationship_mapping.csv
zmbc: Excellent catch -- row 60 is definitely wrong because it turns the parent-in-law _into_ the grandchild. I'm going to go back through these one more time and make sure I didn't do any more of them backward 😬

zmbc: Re: "assign to the oldest," yes good point this file is a little overkill. But, now we have it if we ever want to change that rule ¯\_(ツ)_/¯

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -577,6 +577,40 @@ GBD has state-level all-cause mortality, does FBD forecast at the US
...
+   attribute and the :code:`new_reference_person_relationship_to_old_reference_person` column
+   matches the previous relationship attribute of the new reference person.
+#. If there is no such row in the file (which would only happen with very strange combinations,
+   e.g. a person having two spouses), the simulant is assigned 'Other nonrelative.'
+
+:download:`reference_person_update_relationship_mapping.csv`
+
+This file is not exact:
+
+* There is not always sufficient information to uniquely determine a new relationship. We err
```
beatrixkh, line 598: when you say not enough to "uniquely determine", does that mean that sometimes the csv would suggest 2+ different new relationships, or sometimes that it would not suggest a new relationship?

(the latter wouldnt be a problem if paired with line 592 / your 'other nonrelative' rule, so i think that's probably what you mean?)

beatrixkh (approved):
i had one comment. we would need more info if the csv + the "other nonrelative" rule does not uniquely determine an assignment, but this looks implementation ready if those two things do uniquely determine.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Apologies for the confusion -- I did not mean either of these things.

The CSV will never suggest more than 1 relationship. However, it will not always suggest a relationship. This is the reason for the 'other nonrelative' rule.

However, the "uniquely determine" language and indeed this entire list of bullets were intended to be assumptions/limitations regarding the _construction_ of the file, not its use. (Only important from a research perspective, not an engineering one.) I will change the text above the list to make this more clear.

PR #1032, opened 2022-10-25
wasting paper model 2.0 v and v

alibow:
I did not make a ticket for the bugfix because Hussain already has.


hussain-jafari (approved):

aflaxman (approved):
I misunderstood those docs, too. Thanks to Ali and Hussain for the quick fix, and I'm glad to see that V&V is catching things!

PR #1033, opened 2022-10-26
Adding SSA observer

SylLutze:
Hello - one question: do we need this defined for initialization for everyone who starts with an SSN? Otherwise this will be incomplete to only events during the simulation.


aflaxman:
> Hello - one question: do we need this defined for initialization for everyone who starts with an SSN? Otherwise this will be incomplete to only events during the simulation.

Please include birth for everyone who starts in the sim and starts with an SSN.

aflaxman (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1900,6 +1900,86 @@ These are not currently modeled.
...
+The date of event is either the date of birth or the date of immigration. 
+If these values are unknown, any date within the time step can be recorded. 
+
+The date of death is a recording of anyone who has died and has a SSN. The 
+date of the event is the date of death. This will be listed as "Date 
+of Death". 
+
+Here is an example: 
+
+.. image:: SSA_example.png
```
zmbc, line 1947: In general/for the future, it would be nice to have these as actual tables instead of images. But, no big deal if that is a pain to do here.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1900,6 +1900,86 @@ These are not currently modeled.
...
+Currently, we will only track 2 types of events: 
+
+#. Creation of a SSN 
+#. Date of death recorded 
+
+The creation of a SSN can be triggered by a birth or by immigration into 
+the US. Both will be listed as "Creation" and are not differentiated. This 
+event is only recorded for simulates that receive a SSN. 
+The date of event is either the date of birth or the date of immigration. 
+If these values are unknown, any date within the time step can be recorded.
```
zmbc, line 1939: Fine to have this here, but my understanding is that these dates will always be known.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1900,6 +1900,86 @@ These are not currently modeled.
...
+
+100% of simulants with a SSN and a qualifying event in that time step 
+will be recorded. 
+
+**Data Errors/Noise** 
+
+- The omission rate (currently 0%) should be parameterizable. 
+- Additional noise in names and dates will be added in a separate function, at a later date. 
+
+**Limitations and Possible Future Adds**
```
zmbc, line 1963: I think this could be a bit confusing -- because "possible future adds" doesn't sound so different from saying that something is parameterizable.

I'd just say "Limitations," or if you want to keep the second part, somehow make it clear that this means in a future model, not the minimum viable model we will be using for our first US-scale run.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1900,6 +1900,86 @@ These are not currently modeled.
...
+**Limitations and Possible Future Adds** 
+
+There are a significant number of other possible "events" that 
+could be included in the observer. These are either not implemented 
+in the larger simulation at this time, or more difficult and so 
+have not been included. These are: 
+
+#. Name changes - either first names (trans folks or others) or last names (commonly marriage or divorce). Not included in larger simulation 
+#. Sex-coding changes - not included in larger simulation 
+#. Correction of incorrect information - not included in larger simulation
```
zmbc, line 1972: "Not included in larger simulation" doesn't seem like the right way to say this, since this event doesn't mean much outside the context of this observer.

By the way, this list of other events is a good example of something that I think would be infeasible or very difficult to do outside the sim (cannot be parameterizable), which is why it's especially important that we are clear that these are not happening in our Model 1 (see comment above).

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1900,6 +1900,86 @@ These are not currently modeled.
...
+
+There are a significant number of other possible "events" that 
+could be included in the observer. These are either not implemented 
+in the larger simulation at this time, or more difficult and so 
+have not been included. These are: 
+
+#. Name changes - either first names (trans folks or others) or last names (commonly marriage or divorce). Not included in larger simulation 
+#. Sex-coding changes - not included in larger simulation 
+#. Correction of incorrect information - not included in larger simulation 
+#. Social security benefits - this could be an easy add here if we add it in employment
```
zmbc, line 1973:
```suggestion
#. Receipt of social security benefits - this could be an easy add here if we add it in employment 
```

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1900,6 +1900,86 @@ These are not currently modeled.
...
+There are a significant number of other possible "events" that 
+could be included in the observer. These are either not implemented 
+in the larger simulation at this time, or more difficult and so 
+have not been included. These are: 
+
+#. Name changes - either first names (trans folks or others) or last names (commonly marriage or divorce). Not included in larger simulation 
+#. Sex-coding changes - not included in larger simulation 
+#. Correction of incorrect information - not included in larger simulation 
+#. Social security benefits - this could be an easy add here if we add it in employment 
+#. Diability benefits - similarly, could be an easy add if it is added in employment
```
zmbc, line 1974:
```suggestion
#. Receipt of disability benefits - similarly, could be an easy add if it is added in employment 
```

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1900,6 +1900,86 @@ These are not currently modeled.
...
+could be included in the observer. These are either not implemented 
+in the larger simulation at this time, or more difficult and so 
+have not been included. These are: 
+
+#. Name changes - either first names (trans folks or others) or last names (commonly marriage or divorce). Not included in larger simulation 
+#. Sex-coding changes - not included in larger simulation 
+#. Correction of incorrect information - not included in larger simulation 
+#. Social security benefits - this could be an easy add here if we add it in employment 
+#. Diability benefits - similarly, could be an easy add if it is added in employment 
+#. New employment - it's unclear what percent of new jobs are actually recorded. Also this leads to complications if someone shares a SSN.
```
zmbc, line 1975: Right now in the sim, nobody shares an SSN.
```suggestion
#. New employment - it's unclear what percent of new jobs are actually recorded. Also this could be harder in a future model that included shared SSNs. 
```

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1900,6 +1900,86 @@ These are not currently modeled.
...
+
+**Who to Sample** 
+
+100% of simulants with a SSN and a qualifying event in that time step 
+will be recorded. 
+
+**Data Errors/Noise** 
+
+- The omission rate (currently 0%) should be parameterizable. 
+- Additional noise in names and dates will be added in a separate function, at a later date.
```
zmbc, line 1961:
```suggestion
- Additional noise functions on names and dates should be parameterizable.
```

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1900,6 +1900,86 @@ These are not currently modeled.
...
+Note that the top row is a simulate that received a SSN at birth. The 
+third row indicates someone who immigrated at age 5 and received their 
+SSN then. 
+
+**Who to Sample** 
+
+100% of simulants with a SSN and a qualifying event in that time step 
+will be recorded. 
+
+**Data Errors/Noise**
```
zmbc, line 1958: From the SE perspective, the fact that this section is "parameterizable" means "you should implement the rest of component, but we plan to change this part." For that reason, it would make sense to me to put the bullets in this section in `..todo` blocks or similar to make them stand out.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1900,6 +1900,86 @@ These are not currently modeled.
 #. Currently mailing addresses are the same as home addresses. This is not true, especially for rural populations. We plan to add this to the model later. 
 
 
+Social Security Observer
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Social security information is kept and tracked for a range 
+of different actions. For simplicity, we will limit this 
+section to only creation and dates of death. Others could be 
+added later, if desired.
```
zmbc, line 1909: Let's be clear here that "later" = not Model 1/the initial full-US run, if that is what we mean.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Okay, I can switch them in the future. Since they are used throughout this section and match I'm going to leave them all for now.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Okay! I wasn't sure if simulants got an immigration "date" or just that they moved on a time step. If they get a date then I can delete this.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Okay, makes sense. I changed the language to be "Items NOT Included in Minimum Viable Model" which I think is more clear.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Okay, yeah, makes sense. I think this will be a tough add because to track a "correction" we first need to intentionally get something wrong and "remember" that error. I wrote a bit more to express this more fully in the note.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Thanks!

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: We actually do have this in the tax observer! So while no simulants share a SSN in the state table, they actually can "borrow" a SSN in an observer. I have clarified that here.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Added

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Sure, added these in a todo block

zmbc (approved):

PR #1034, opened 2022-10-28
Defining observer elements that are parameterizable 

SylLutze:
I also included in this PR the addition of an example on how survey data should look


beatrixkh (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1420,6 +1417,7 @@ There are two types of sampling plans:
   :widths: 20
   :header-rows: 0
 
+  * - Household number (random indicator for who is in the same household)
```
zmbc, line 1420:
```suggestion
  * - Household number (random indicator which is the same for simulants in the same household)
```

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1538,15 +1541,14 @@ survey iterations.
...
-addresses, age, and person swaps. 
 
-Some errors may be introduced within the simulation, such as 
-duplicate people or swapped people. Others may be introduced 
-at the time of sampling, such as name and address misspellings. 
+.. todo::
+    Engineers: these things are not yet defined and might be edited later - they should be parameterizable 
 
-These are not currently modeled. 
+    - Omission rates by survey, oversampling might be needed to avoid rerunning
```
zmbc, line 1548: Technically omission wouldn't require any oversampling, right?

That said, I think you should specify here a maximum number we might want to sample, so that we could change the sample rate up to that number without re-running.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1370,15 +1370,12 @@ duplicates at this time.
...
 
-Some errors may be introduced within the simulation, such as 
-duplicate people or swapped people. Others may be introduced 
-at the time of sampling, such as name and address misspellings. 
+.. todo::
+    Engineers: these things are not yet defined and might be edited later - they should be parameterizable 
 
-These are not currently modeled. 
+    - We currently have net undercounts, might want to have omissions and duplicates instead. Having a perfect census for now should be fine. 
+    - Creation of a noise function for misrecording data (names, addresses, birthdays)
```
zmbc, line 1378:
```suggestion
    - Addition of a noise function for misrecording data (names, addresses, birthdays) based only on the true values and the other columns in the survey output
```

Just being overly specific, but we need to specify that our noise function (whatever it is) will not use any currently unobserved values. Or if you don't think we can make this guarantee, I think we need to identify which columns aren't here but we might want to use.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1370,15 +1370,12 @@ duplicates at this time.
...
-addresses, age or DOB, and person swaps. 
 
-Some errors may be introduced within the simulation, such as 
-duplicate people or swapped people. Others may be introduced 
-at the time of sampling, such as name and address misspellings. 
+.. todo::
+    Engineers: these things are not yet defined and might be edited later - they should be parameterizable 
 
-These are not currently modeled. 
+    - We currently have net undercounts, might want to have omissions and duplicates instead. Having a perfect census for now should be fine.
```
zmbc, line 1377: I don't think this clearly defines what needs to be parameterizable. What we need to do is scope: if we want to have the flexibility to add omissions and duplicates, what values will the omission/duplication function need as inputs?

As a rough example (haven't thought this through), if we know we will want some flexibility about duplicating college students, we might need GQ type and tracked parent/guardian.

It's possible that we can make the ask here to be able to use all tracked columns (I don't know how expensive this could get) -- might be an engineering question.

zmbc: Re: the last sentence "Having a perfect census for now should be fine" -- it's unclear to me whether you mean for our research goals, or because we can add noise later. And "for now" -- is that Model 1/Minimum Viable Model?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1538,15 +1541,14 @@ survey iterations.
...
-Some errors may be introduced within the simulation, such as 
-duplicate people or swapped people. Others may be introduced 
-at the time of sampling, such as name and address misspellings. 
+.. todo::
+    Engineers: these things are not yet defined and might be edited later - they should be parameterizable 
 
-These are not currently modeled. 
+    - Omission rates by survey, oversampling might be needed to avoid rerunning 
+    - Creation of a noise function for misrecording data (names, addresses, birthdays) 
+    - Include person swaps or duplicates
```
zmbc, line 1550: Again, here we should define what info we would need to perform these.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1876,18 +1878,14 @@ for who files taxes:
...
-Furthermore, there are often complications with people being claimed as 
-dependents in different households, or claiming people from group quarters.
+.. todo::
+    Engineers: these things are not yet defined and might be edited later - they should be parameterizable 
 
-These are not currently modeled. 
+    - Omission rates, oversampling might be needed to avoid rerunning 
+    - Creation of a noise function for misrecording data (names, addresses, birthdays) 
+    - Further refine "sharing" of SSN's 
+    - Include person swaps or duplicates - especially relevant for taxes if dependents should be double claimed sometimes
```
zmbc, line 1888: I think we need to specify what we might need for these last two bullets especially.

aflaxman (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1429,6 +1427,11 @@ There are two types of sampling plans:
   * - Home Address 
   * - Home Zip Code 
 
+Here is an example:
```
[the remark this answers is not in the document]
SylLutze: Good catch! Fixed this

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Thanks

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Sure, changed language and gave a number here

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Okay, I think that makes sense. I have tried to add in what other parameters are needed (mainly focused on tracked guardians) across the board.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: GQ type should be in address already I think? I agree on adding tracked guardians!

zmbc: @SylLutze I don't think GQ type is included in address. Everyone with the same GQ type will have the same address, but we won't have a way to go backwards and tell which GQ type that address represents unless we have GQ type in the output.

zmbc (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Actually, I think we know the GQ type [from household ID](https://github.com/ihmeuw/vivarium_census_prl_synth_pop/blob/419f5c434eb9558389c1420c3f135bfe0ffd5a5c/src/vivarium_census_prl_synth_pop/constants/data_values.py#L10-L23), if that is included in the output. (Right now the docs say a "random identifier" representing household which doesn't guarantee us this correspondence.)

SylLutze: Okay! I will add that in more explicitly then, thanks!

PR #1036, opened 2022-10-31
Updating assigning parents/guardians to be simpler

SylLutze:


aflaxman (approved):

zmbc (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -406,29 +406,18 @@ this person as a dependent.
...
-- Child is grandchild of reference person 
-    * If there is a child (biological, adopted, foster) of reference person, assign them (if multiple, assign at random) 
-    * Otherwise assign reference person 
-- Child is brother/sister to reference person 
-    * If there is a parent of the reference person, assign them 
-    * Otherwise assign the reference person 
-- Child is other relative to reference person 
-    * Assign a relative of the reference person who is between 20 and 45 years older than the child. If there are multiple, select one of the same race/ethnicity. If there are multiple of the same race/ethnicity, assign at random. 
+- Child is any other relative to reference person (NOT roommate/housemate or other nonrelative)
+    * Assign a relative of the reference person who is between 20 and 45 years older than the child. If there are multiple, assign at random.
```
beatrixkh, line 412: could you define "relative" here?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -406,29 +406,18 @@ this person as a dependent.
 
 **For simulant under 18:**
```
beatrixkh, line 407: if this edit would be accurate, would you mind updating this to "simulant under 18 living in non-GQ" (or something similar?)

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
beatrixkh: i know this seems like it should be obvious, but the concept "relative" doesn't currently exist in the sim, so we need to pin it down exactly in order to implement it.

for example, i think you would probably mean any two people A and B who currently live in a non-GQ house together, such that person A and person B both have one of the following "relationship to reference person"s:

     "Reference person",
    "Opp-sex spouse",
    "Opp-sex partner",
    "Same-sex spouse",
    "Same-sex partner",
    "Biological child",
    "Adopted child",
    "Stepchild",
    "Sibling",
    "Parent",
    "Grandchild",
    "Parent-in-law",
    "Child-in-law",
    "Other relative",
    "Foster child"

as opposed to, say, that plus anyone who has a tracked mother or guardian who now lives in another house, or some other definition?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -406,29 +406,18 @@ this person as a dependent.
...
-    * Otherwise assign the reference person 
-- Child is other relative to reference person 
-    * Assign a relative of the reference person who is between 20 and 45 years older than the child. If there are multiple, select one of the same race/ethnicity. If there are multiple of the same race/ethnicity, assign at random. 
+- Child is any other relative to reference person (NOT roommate/housemate or other nonrelative)
+    * Assign a relative of the reference person who is between 20 and 45 years older than the child. If there are multiple, assign at random. 
     * If there is not a relative of the appropriate age available, assign the reference person 
 - Child is non-relative (roommate or other nonrelative) to reference person 
-    * Assign another non-relative of the reference person who is between 20 and 45 years older than the child. If there are multiple, select one of the same race/ethnicity. If there are multiple with the same race/ethnicity, assign at random.  
+    * Assign another non-relative of the reference person who is between 20 and 45 years older than the child. If there are multiple, assign at random.
```
beatrixkh, line 415: similar to above, i think you mean someone who lives in the same household as the child, who does not fall into the category of "relative" --- is that correct?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
beatrixkh: actually -- i think i have a recollection of Zeb writing something that defines the concept "relative". im not sure if it's been merged yet (or if i made that up?) but if that's somewhere else in the docs, would you reference or link to that from here? that way this section on its own contains all the details we need in order to implement

beatrixkh (changes requested):
one additional clarification request

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Sure, makes sense. I added some clarification on "relative", which is much easier to define in the negative (anything except XYZ), let me know if this makes sense. Since this is run on initialization only, I don't think we will encounter any tracked mothers living elsewhere?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Good callout!

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Added more definition here as well - let me know if it makes sense.

beatrixkh (approved):

PR #1039, opened 2022-11-02
Re-document domestic migration

zmbc:
My questions:
- Are there performance concerns? Particularly, the PUMA-to-PUMA file is quite large (~5M rows), and could be shrunk by implementing some of the logic in the sim: the (single-tier) smoothing procedure and the MIGPUMA-to-PUMA correspondence might be faster to do at runtime. The tradeoff is implementation complexity.
- Is there anything here that seems overly sophisticated? I have already done almost all of the work on my side, but cutting something could still save computational and implementation complexity.
- Are any of the limitations things that we need to address for Model 1/MVM?

My code for calculating everything can be found here, but it is still pretty rough: https://github.com/ihmeuw/vivarium_research_prl/pull/3. The numbers I've currently included in this PR are intended to demonstrate the schema, but not be final.


aflaxman (approved):

SylLutze (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -609,106 +612,347 @@ simulants are dying at the expected rates.
...
-#. the head of household cannot move to a new household
+* What PUMA the person lives in now (:code:`ST` and :code:`PUMA`).
+* Whether they moved domestically in the last 12 months (:code:`MIG` and :code:`MIGSP`).
+* If they moved, what "migration PUMA" (MIGPUMA) they lived in 12 months ago (:code:`MIGSP` and :code:`MIGPUMA`).
 
-#. Group Quarters address and zip code do not change
+MIGPUMAs are geographic entities created for this purpose.
+They are similar to PUMAs and many are exactly identical to a PUMA.
+However, some individual PUMAs had too few people moving from them, so they were grouped together
+with neighboring PUMAs into a single MIGPUMA for disclosure avoidance reasons.
```
beatrixkh, line 646: out of curiosity, when this is the case, what does the PUMA variable contain?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -609,106 +612,347 @@ simulants are dying at the expected rates.
...
+(e.g. all 0-15 year olds).
+We refer to these groups, which are less stratified and have larger sample size, as the "reference group"
+in the smoothing procedure description below.
+We do not include a third smoothing pass with the fully unconditional distribution (all people regardless of age),
+because there are logical relationships between age and relationship -- for example,
+children should never be spouses.
+
+Our smoothing procedure is as follows:
+
+* Take the total probability of all demographic + relationship combinations which have less than **30** sample size,
```
beatrixkh, line 737: I'd be interested to ask you over zoom if i'm understanding this correctly -- but not necessary for implementation

beatrixkh (approved):
this looks great; thank you everything that we are concerned with is really clearly laid out

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: For the input files to the simulation, I have handled this wrinkle for you. Both the source and destination columns are regular PUMAs.

(Because we don't have granularity below MIGPUMA for the source, that means I've duplicated rows for each source PUMA that lies within the MIGPUMA.)

PR #1040, opened 2022-11-03
Updating the Outreach Intervention 

SylLutze:


stevebachmeier (approved):
@ramittal Do these changes to the affected outcome tables look good? It makes more sense conceptually, though not sure if we need something else/different implementation-wise

PR #1041, opened 2022-11-03
Clarifying dependents in taxes with the new guardian information

SylLutze:
Updated taxes 

Let me know about the guardians section and how exactly we should define the columns from tracking guardians. Happy to make edits!


docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -394,7 +394,9 @@ dependent on tax forms to have a guardian. This will
 improve tracking for names, and dependent status on tax forms. 
 
 This person will be listed as ["Guardian"]. By design, most will be 
-parents, but some may be a grandparent or other relative. 
+parents, but some may be a grandparent or other relative. This will
```
zmbc, line 397: IMO, describing the state table shouldn't be necessary if we clearly outline how this works. While I'm pretty sure in this case it will be implemented the way you describe here, it could also be implemented the opposite way (each simulant keeps track of who they are the guardian **of**) and that would also work.

zmbc: I think the main thing this section needs is to consistently use the term "guardian" instead of "parent/guardian" -- because "parent" is already a concept in the simulation and we do not want them to be confused with each other.

Also, I would specify here at the top the restriction that each simulant will have a maximum of 2 guardians.

In the rules below, to make it more implementation-agnostic, we can be a bit more explicit, something like saying "assign the reference person to be this simulant's guardian" instead of "assign reference person." Or, we can add some introductory text that explains to interpret the rules this way.

A nit I didn't notice until now: "child"/"parent" are also used in the assignment rules below. This is particularly confusing on [line 421](https://github.com/ihmeuw/vivarium_research/pull/1041/files#diff-45a09ee257e7b14d6545111feb3c2f747ef32aa90038f0dbc028e55d2ff08f0dR421), where I think you are referring to the relationship attribute, not the tracked parent attribute (which of course is not set at initialization).

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1884,16 +1886,13 @@ for who files taxes:
...
-    * Not be a "housemate/roommate" or "other nonrelative" to whoever is claiming them 
-    * The dependent's income must be below $4300 
-- For simulants outside of the household, they will be claimed as a dependent by their parent until age 19 OR age 24 if they are in group quarters for college 
-    * For simulants born in the simulation, they should have at least one parent/guardian tracked 
-    * For simulants not born in the simulantion, information on initializing parent/guardians is being added shortly 
-    * For simulatns with more than one parent/guardian tracked, assign the dependent randomly 
+- All simulants eligible to be dependents will be claimed. To be eligible, the simulant must either: 
+    * Be under the age of 19 
+    * OR be less than 24, in GQ in college, and earn less than $10,000 
+    * OR be any age, but NOT "housemate/roommate" or "other nonrelative" to whoever is claiming them and earn less than $4300
```
zmbc, line 1892: There are two issues here, neither of which is new but which I am only noticing now:
- By using "whoever is claiming them" here, this part depends on the next, but the next depends on this part ("An **eligible** simulant...").
- "Whoever is claiming them" may live in a different household, in which case the "housemate/roommate" or "other nonrelative" part is not something we can check.

Happy to help think through how to clarify this rule, if you want.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1884,16 +1886,13 @@ for who files taxes:
...
-    * For simulants born in the simulation, they should have at least one parent/guardian tracked 
-    * For simulants not born in the simulantion, information on initializing parent/guardians is being added shortly 
-    * For simulatns with more than one parent/guardian tracked, assign the dependent randomly 
+- All simulants eligible to be dependents will be claimed. To be eligible, the simulant must either: 
+    * Be under the age of 19 
+    * OR be less than 24, in GQ in college, and earn less than $10,000 
+    * OR be any age, but NOT "housemate/roommate" or "other nonrelative" to whoever is claiming them and earn less than $4300 
+- An eligible simulant will be claimed by: 
+    * The defined "guardian" from initialization OR "parent" from birth (Note: if there are two parents/guardians and they are not filing jointly, assign the simulant randomly to one)
+    * If there is not a parent/guardian tracked, they will randomly be assigned to a tax-filing relative (not housemate or other non-relative) in the household
```
zmbc, line 1895: Won't there be cases where this person is e.g. in college GQ (where we can't find relatives) and not have a tracked parent or guardian? For example if they started the sim in a different type of GQ, or were living in a household where we couldn't identify a guardian (i.e. the "N/A" case). I think we need to have a last option here, where if there are no tax-filing relatives, they are not claimed at all (though this contradicts the "All simulants eligible to be dependents will be claimed" statement above).

zmbc: Also, as written, it seems like this line assumes that the simulant is a relative of the reference person. Or was your intention that even if they were e.g. "roommate" they would be claimed by a relative of the reference person?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1884,16 +1886,13 @@ for who files taxes:
...
-    * If there is not a parent, they will randomly be assigned to a tax-filing relative (not housemate or other non-relative) in the household 
-- A simulant eligble to be a dependent must: 
-    * Not be a "housemate/roommate" or "other nonrelative" to whoever is claiming them 
-    * The dependent's income must be below $4300 
-- For simulants outside of the household, they will be claimed as a dependent by their parent until age 19 OR age 24 if they are in group quarters for college 
-    * For simulants born in the simulation, they should have at least one parent/guardian tracked 
-    * For simulants not born in the simulantion, information on initializing parent/guardians is being added shortly 
-    * For simulatns with more than one parent/guardian tracked, assign the dependent randomly 
+- All simulants eligible to be dependents will be claimed. To be eligible, the simulant must either: 
+    * Be under the age of 19
```
albrja, line 1890: Is this less than or equal to 18 or less than 18?

beatrixkh:
hi, a few questions:

- will you want to use the "guardian" relationship for anything besides tax returns?
- right now, it looks like you are asking us to assign 1-2 guardians to certain sims. what is the purpose of the second guardian? will both guardians be claiming the dependent in a tax return, or would it be possible to have max 1 guardian and achieve everything you're wanting?
- at any one point in time, can a sim be both a guardian and have a guardian?
- do you care about previous guardian status? i.e., do you care who a 21 year old who was previously a child in the sim had as a guardian?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
beatrixkh: i agree with zeb --- we would like to avoid asking you guys to specifically delineate how the state table (or other within-sim machinery) is explicitly implemented, and instead have the docs outline the concepts you want to exist in the sim / maybe the needs they meet

i think sometimes those lines blur a bit, sorry about that!

i also agree disambiguation would be helpful here --- i think there are 3ish concepts that are getting conflated here:
1. a relationship within the sim we are calling "guardian". currently, the only usage i think you guys have for this is to decide who claims whom in tax returns (will it have other purposes?)
2. a relationship within the sim called "parent". currently, this is a link between the mother of a sim who was born during the simulation, and that child
3. a value in the "relationship to head of household" column called "parent"

it would be really helpful if, each time one of these was referenced, it was made explicit which one you are talking about (for example in lines 423-425)

beatrixkh:
some follow-up questions that i realize i should've asked a while ago but somehow forgot to think of:

do you want this "guardian" status to update at all? for example, if sim A is the guardian of sim B, and they live in the same household:
- if A moves out of the house, are they still guardian of B?
- if B moves out of the house, is A still guardian of B?
- if A dies, do you want a new guardian selected? (does it depend on how old B is?)

also -- in our sim, would someone ever claim their parent or parent-in-law (or other older relative) as a dependent on a tax form? if yes, would we also want that person being the "guardian" of their parent?
^ this is sort of again asking if we want a 1:1 relationship between guardians and sims-who-claim-dependents-when-filing-taxes, or if we want guardians to have other purposes (or people who claim that are not represented by the guardian column)

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Added clarification

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: You're right, I added a line where if there is not anyone to claim them they remain unclaimed 

I made sure to clarify in line 1915 that to qualify as a dependent they had to be a relative of the reference person (so a roommate only wouldn't qualify to be claimed. This WILL miss people who might be related to someone in the house but just not the reference person, but I think this will be reasonably rare.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Okay I went through and tried to clarify this section and the language used. Let me know if there are still outstanding questions!

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1884,16 +1909,14 @@ for who files taxes:
...
-    * Not be a "housemate/roommate" or "other nonrelative" to whoever is claiming them 
-    * The dependent's income must be below $4300 
-- For simulants outside of the household, they will be claimed as a dependent by their parent until age 19 OR age 24 if they are in group quarters for college 
-    * For simulants born in the simulation, they should have at least one parent/guardian tracked 
-    * For simulants not born in the simulantion, information on initializing parent/guardians is being added shortly 
-    * For simulatns with more than one parent/guardian tracked, assign the dependent randomly 
+- Simulants eligible to be dependents must qualify as ONE of the following: 
+    * Be under the age of 19 (less than or equal to 18)
+    * OR be less than 24, in GQ in college, and earn less than $10,000 
+    * OR be any age, but be either a relative (NOT "housemate/roommate" or "other nonrelative") of the reference person, or be the reference person and earn less than $4300
```
zmbc, line 1915: I think we can simplify this further (see suggestion below).
```suggestion
    * OR be any age, but earn less than $4300 
```

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1884,16 +1909,14 @@ for who files taxes:
...
-    * For simulants born in the simulation, they should have at least one parent/guardian tracked 
-    * For simulants not born in the simulantion, information on initializing parent/guardians is being added shortly 
-    * For simulatns with more than one parent/guardian tracked, assign the dependent randomly 
+- Simulants eligible to be dependents must qualify as ONE of the following: 
+    * Be under the age of 19 (less than or equal to 18)
+    * OR be less than 24, in GQ in college, and earn less than $10,000 
+    * OR be any age, but be either a relative (NOT "housemate/roommate" or "other nonrelative") of the reference person, or be the reference person and earn less than $4300 
+- An eligible simulant will be claimed in the below order of options: 
+    * The defined "guardian" (Note: if there are two parents/guardians and they are not filing jointly, assign the simulant randomly to one) 
+    * If there is not a "guardian" they will be randomly assigned to any relative (reference person OR relative of the reference person (NOT roommate or non-relative)) or a relative of themselves (if they are the reference person)
```
zmbc, line 1918:
```suggestion
    * Otherwise, if they are the reference person or a relative of the reference person (NOT roommate or non-relative) they will be randomly assigned to someone else in the household who is the reference person or a relative of the reference person (NOT roommate or non-relative)
```

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1884,16 +1909,14 @@ for who files taxes:
...
-    * For simulants not born in the simulantion, information on initializing parent/guardians is being added shortly 
-    * For simulatns with more than one parent/guardian tracked, assign the dependent randomly 
+- Simulants eligible to be dependents must qualify as ONE of the following: 
+    * Be under the age of 19 (less than or equal to 18)
+    * OR be less than 24, in GQ in college, and earn less than $10,000 
+    * OR be any age, but be either a relative (NOT "housemate/roommate" or "other nonrelative") of the reference person, or be the reference person and earn less than $4300 
+- An eligible simulant will be claimed in the below order of options: 
+    * The defined "guardian" (Note: if there are two parents/guardians and they are not filing jointly, assign the simulant randomly to one) 
+    * If there is not a "guardian" they will be randomly assigned to any relative (reference person OR relative of the reference person (NOT roommate or non-relative)) or a relative of themselves (if they are the reference person) 
+    *  If there is no one found to claim them, they will remain unclaimed. This is most common for sims in GQ's but there might be other cases
```
zmbc, line 1919:
```suggestion
    *  If there is no one found to claim them, they will remain unclaimed. This is most common for sims in GQ's or non-relatives in households but there might be other cases
```

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -449,7 +453,28 @@ Simulant will be randomly assigned to a parent/guardian based on the below rules
...
+
+A person cannot "age out" of their guardian when they become 
+an adult or have their own children. Therefore it is possible 
+to both HAVE a guardian and BE a guardian. 
+
+However, for simulants born in the simulation, they receive a 
+tracked mother (term for the person that births them, regardless 
+of gender). 
+
+The tracked mother will be assigned as a guardian. In addition, if
```
zmbc, line 474: I think it is good to have this here, but we might want to have it in the fertility section as well? Or reference it from that section? From an engineering perspective, this will be implemented in that component, I believe.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Thanks! This does look simpler

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Hmm good point, I'll add that in!

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -416,23 +422,21 @@ this person as a dependent.
...
 **For a simulant who is below 24 and in GQ at college:**
 
-Simulant will be randomly assigned to a parent/guardian based on the below rules: 
+Simulant will be randomly assigned to a guardian based on the below rules: 
 
-- 78.5% will be assigned to a parent/guardian within their state. The remainder will be assigned out of state source1_. For early versions with only one state, the out of state parent/guardians can be ignored. 
+- 78.5% will be assigned to a guardian within their state. The remainder will be assigned out of state source1_. For early versions with only one state, the out of state guardians can be ignored. 
 - Match to a person 20 to 45 years older than the child 
-- If child is not "Multiracial or Some Other Race", match parent's race. If child is "Multiracial or Some Other Race", then assign to a parent of any race
+- If child is not "Multiracial or Some Other Race", match parent's race. If child is "Multiracial or Some Other Race", then assign to a guardian of any race
```
zmbc, line 439:
```suggestion
- If child is not "Multiracial or Some Other Race", match guardian's race. If child is "Multiracial or Some Other Race", then assign to a guardian of any race
```

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -486,6 +511,10 @@ simulant is initialized, and to come up with a consistent value for
 the "reference person" relationship field, we use the following
 mapping:
 
+In addition, tracked mothers will be assigned as "guardians" based on the
```
zmbc, line 514: I think you'll want to move this above the previous paragraph, so it doesn't go between `the following mapping:` and the mapping.

zmbc (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Good catch!

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Thanks!

albrja (approved):

PR #1043, opened 2022-11-04
Update wasting relapse literature and effect values

alibow:


docs/source/gbd2019_models/risk_effects/x_factor/index.rst
```diff
@@ -55,9 +103,11 @@ Zone, Amhara Region, Ethiopia of children discharged from community management o
 
 [Adegok-et-al-2020]_ performed a prospective matched cohort study of SAM recurrance in Northern Nigeria. The study recruited children discharged from OTP treatment and matched community controls and followed them for 6 months. At the end of follow-up 24% of the OTP discharged children relapsed to SAM and 0.6% of community controls developed SAM.
 
+Additionally, a study protocol for a prospectiv matched cohort study of acute malnutrition relapse (similar to [Girma-et-al-2022]_) conducted in Mali, Somalia, and South Sudan has been published by [King-et-al-2022]_, with data collection expected to conclude in January of 2023.
```
aflaxman, line 106:
```suggestion
Additionally, a study protocol for a prospective matched cohort study of acute malnutrition relapse (similar to [Girma-et-al-2022]_) conducted in Mali, Somalia, and South Sudan has been published by [King-et-al-2022]_, with data collection expected to conclude in January of 2023.
```

aflaxman (approved):
How did we pick sqrt(10) for the stand-in value?  It is pretty close to the one from this new study!  Not a bad pick.

alibow:
> How did we pick sqrt(10) for the stand-in value? It is pretty close to the one from this new study! Not a bad pick.

It came from essentially the same study - just before it was published. There is a systematic review of wasting relapse that was published by the lead author of the wasting relapse study that Indi Trehan is working on. This review referenced some unpublished work by a group at Jimma University in Ethiopia that we were using for our sqrt(10) value. Now that work has been published so we're using those values instead. There's a lot more detail in the published article than was reported in the review so we can be more precise now - particularly on the difference in the RR between the mild->MAM transition and the MAM->SAM transition (we had previously assumed that these were equal).

SylLutze (approved):

PR #1045, opened 2022-11-07
Document correspondence between PUMA and ZIP

zmbc:
@aflaxman it looks like you did some pre-processing on the deepparse address data to generate deepparse_address_data_usa.csv.bz2 -- I added a TODO here because I don't see a description of, or code for, this pre-processing.


aflaxman (approved):
Thanks for this.  I will dig up the code I used to process the address data soon, sorry for that oversight!

SylLutze (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -124,23 +124,22 @@ of domestic (:ref:`8 <census_prl_domestic_migration>`) and international (:ref:`
...
-On top of this, we will layer attributes relevant to PRL: mailing
-addresses for each household (11); first, middle, and last names for
-each simulant (12); date of birth (13); intended-to-be-unique
+On top of this, we will layer attributes relevant to PRL: residential (11)
+and mailing addresses for each household (12); first, middle, and last names for
+each simulant (13); date of birth (14); intended-to-be-unique
 identification number modeling SSN that is missing for some and not
-actually unique for others (14); and periodic survey, census, and registry
-observations with realistic noise (15).
+actually unique for others (15); and periodic survey, census, and registry
```
beatrixkh, line 131: this might be in an update that i missed -- but are you guys still intending this? that SSNs are not unique for some sims? or is this referring only to the outputs in a census or set of tax forms?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -927,62 +926,111 @@ Limitations
...
+.. todo::
 
-    def my_fake_address():
-        orig_address = fake.unique.address()
-        address = orig_address.split('\n')[0]
-        return address
+  Document the pre-processing of the deepparse address data.
 
-    address_dict = {hh_id: my_fake_address() for hh_id in df_population.household_id.unique()}
+To make PUMA correspond to ZIP code, we use a crosswalk generated by the
```
beatrixkh, line 962: oh yay i'm glad you found this!

beatrixkh (approved):
it sounds like we need PUMAs within the sim, but zipcodes could be added in a post-procesing step after the sim has run; is that correct?

zmbc:
> it sounds like we need PUMAs within the sim, but zipcodes could be added in a post-procesing step after the sim has run; is that correct?

Yes -- state and PUMA will be required for the zipcode generation process, but that process can happen after the sim as long as state and PUMA have been recorded.

aflaxman:
@zmbc Here is the code to generate the address data: https://gist.github.com/aflaxman/dced6d09069bf9e14429bfb0500c7a7b

Can you get it documented in a better way when you have a chance?

PR #1046, opened 2022-11-08
Updating SBP and LDL-C ramps to allow non-adherent folks on interventions

SylLutze:


stevebachmeier (approved):

stevebachmeier (approved):

rmudambi (commented):
Can't we just say that anyone who is already on treatment and has LDL-C >= 1.81mmol/L (or SBP >= 140) has the outreach/polypill enrollment check? I think that's equivalent to what you have put here, since enrollment has no impact on people who are adherent.

aflaxman (approved):

PR #1047, opened 2022-11-08
Creating a WIC Observer

SylLutze:
Two questions I am struggling with: 
- Should your siblings being on WIC make you more or less likely to be on WIC? 
- Should there be any continuity between who is on WIC in subsequent years?


docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1612,6 +1612,119 @@ Note/limitations:
...
+  * - 4 
+    - $51,338
+  * - +1 person 
+    - +$8,732 
+
+Source: [WIC_eligibility]_
+
+**Covered by WIC**
+
+Once someone is found to be eligible, the are then randomly assigned to be
```
zmbc, line 1692:
```suggestion
Once someone is found to be eligible, they are then randomly assigned to be 
```

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1612,6 +1612,119 @@ Note/limitations:
...
+- Have a total household income less than the limits below (based on number of people in the household)
+
+.. list-table:: Income Maxs  
+  :widths: 10 10
+  :header-rows: 1
+
+  * - Number of people in household 
+    - Maximum Income 
+  * - 1 
+    - $25,142
```
zmbc, line 1678: These are all evenly spaced. So, it would be simpler to say $16,410 + $8,732 * number of people in the household.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1612,6 +1612,119 @@ Note/limitations:
...
+**Who to Sample** 
+
+Selection for the observer will be in 2 steps: 
+
+#. Eligibilty for WIC 
+#. Covered by WIC 
+
+Please note that multiple people in the same household can be selected and selection 
+should be independent (e.g., having one child in the household on WIC does not mean a 
+second child is or is not on WIC). Inclusion is at the simulant level. Do not include
```
zmbc, line 1656: Am I understanding this right that parents/guardians are also selected independently of children? So, you could have a tracked parent sampled here but not the <1 year old child that made them eligible?

Not saying this is an issue, it could just be a limitation, but that would seem to be the more important example to use here instead of/in addition to siblings.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1612,6 +1612,119 @@ Note/limitations:
...
+
+    - Addition of a noise function for misrecording data (names, addresses, birthdays) 
+    - To create these noise functions, in addition to the above survey outputs, please include: tracked guardian(s) and the tracked guardian(s) addresses
+
+**Limitations and Possible Future Adds** 
+
+#. Currently the model does not track pregnant people, therefore they are not included despite being eligible 
+#. It is unclear if having one child on WIC would increase or decrease the chance of having multiple children on WIC. This could probably be improved upon with more data. 
+#. Selection for WIC is random after eligibility. In practice, lower income individuals are more likely to sign up for WIC and the selection is therefore biased. 
+#. Some states use different income cutoffs, but the ones listed are used in the majority of cases and so are implemented here
```
zmbc, line 1727: Can you add a limitation about the independence between years? I think this is fine, but should be noted.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Fair enough!

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Yes, that is my understanding of the program. For example, if a mom is enrolled you get food like bread and cheese, but if an infant is enrolled you get baby formula and stuff. So I think it is independent in WIC, which I continued here. 

I agree, I should have written that more explicitly and added here!

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Good callout!

zmbc (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1612,6 +1612,109 @@ Note/limitations:
...
+
+**Who to Sample** 
+
+Selection for the observer will be in 2 steps: 
+
+#. Eligibilty for WIC 
+#. Covered by WIC 
+
+Please note that multiple people in the same household can be selected and selection 
+should be independent. This applied to siblings (e.g., having one child in the household
```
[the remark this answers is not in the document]
SylLutze: Thanks!

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1612,6 +1612,109 @@ Note/limitations:
...
+
+For example, a one person household would have a maximum income of $25,142 and 
+a two person household would have a maximum income of $33,874. 
+
+Source: [WIC_eligibility]_
+
+**Covered by WIC**
+
+Once someone is found to be eligible, they are then randomly assigned to be 
+covered or not covered by WIC. This is a random sample based on the age of
```
aflaxman, line 1682: I think we should either include both parent and child or neither --- it would be weird for the parent to get the benefit but not the child.

aflaxman (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Okay, I can see saying if the parent has benefits then the infant definitely does, especially since infant rates are higher than postpartum women. But the other direction (child has benefits --> mother must also) I don't think we should specify. The rates for mothers are lower and they requirements don't always match up (breastfeeding for moms; also having "nutritional risk" is a requirement that might be there for an infant but not a mother)

zmbc: I feel like you might need to spell out in more detail how you are introducing this dependency. You sort of hint at it with "... additional infants can be selected to reach the higher coverage rate for this group" but then in this section you are still saying it is a random sample at the coverage rates.

I think it is something like:
- Randomly select eligible tracked parents or guardians to enroll based on their coverage rate (84.7%)
- Also enroll _all_ (including multiple per parent/guardian) infants of these parents/guardians
- Calculate the infant coverage after the previous step
- Enroll more infants randomly, among those eligible but not already enrolled, _until_ the overall infant coverage is 98.4%.

For that last step, there is some math needed to figure out the coverage rate; if the engineers implement it as a rate, which I expect they will, we might want to do the math on our end just to make sure everyone is on the same page.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1612,6 +1612,119 @@ Note/limitations:
...
+
+.. todo::
+
+    - Addition of a noise function for misrecording data (names, addresses, birthdays) 
+    - To create these noise functions, in addition to the above survey outputs, please include: tracked guardian(s) and the tracked guardian(s) addresses
+
+**Limitations and Possible Future Adds** 
+
+#. Currently the model does not track pregnant people, therefore they are not included despite being eligible 
+#. It is unclear if having one family member on WIC would increase or decrease the chance of having multiple family members on WIC. This could probably be improved upon with more data.
```
zmbc, line 1724: This limitation should also be updated for the dependency.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Also @aflaxman it seems worth noting that this relationship between infant enrollment and parent enrollment shouldn't matter for our case study of counting under-5s, since the parent enrollment is not used in the case study (IIUC).

aflaxman: > Also @aflaxman it seems worth noting that this relationship between infant enrollment and parent enrollment shouldn't matter for our case study of counting under-5s, since the parent enrollment is not used in the case study (IIUC).

Let's check in with @NathanielBlairStahn about that... I thought that the strategy would be to link the grownups and then see if the children were also included in the decennial --- any grownup/child pair in WIC that matches just the grownup in decennial is an opportunity to add a kid that was missed in the right place.

NathanielBlairStahn: @aflaxman Seems like we could also just look for all the kids receiving WIC, regardless of whether their parent does, and then try to match the child's parent from the WIC data (will it have info on children's parents even if the parents are not covered?) to the census data.

SylLutze: Currently, no. The way its set up, it tracks only people receiving benefits. We could change this to always include guardians or always include the whole household. 

Both of these have their own issues: guardians in our current sim can die or move and not be replaced. And the household might include people not in the same family.

NathanielBlairStahn: @SylLutze If I'm understanding correctly, data on the child's parent will **not** be included in the WIC observer unless the parent is also on WIC? In that case does the observer record the fact that the two recipients have a parent-child relationship?

If we *don't* have parent information in the WIC data, then we can still try to link all the children to the census using other fields such as last name and address (and also full name and age/DOB for children who _are_ in the census data).

SylLutze: That is correct. We do have a "household" indicator which says if 2 sims are in the same household - this could be parent/child or siblings. 

Again, we can change that to something else still if we need additional information.

NathanielBlairStahn: Ah, right, I hadn't paid attention to the household ID since I was just looking through this quickly. I think household number will be sufficient for the linking task when the child has a parent or other household member who is also covered by WIC. So I think the remaining question is whether the actual WIC data used by Census would have information about a covered child's parent(s) even if the parent is not covered; if so, that would make the linking task easier.

NathanielBlairStahn: Also, does WIC record other demographic data such as race/ethnicity that would also be captured by the decennial census? If so, that would also help with linking.

NathanielBlairStahn:
Regarding my [comment in the thread above](https://github.com/ihmeuw/vivarium_research/pull/1047#discussion_r1019704749):
> Also, does WIC record other demographic data such as race/ethnicity and sex that would also be captured by the decennial census? If so, that would also help with linking.

It looks like WIC does track race/ethnicity (https://www.fns.usda.gov/wic/2019-eligibility-coverage-rates):

image

So it seems like we should record that in the observer. 

**Question:** Do we want to accurately capture eligibility and coverage rates stratified by race/ethnicity?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1612,6 +1612,154 @@ Note/limitations:
...
+
+- Sample compiled on the time step containing Jan 1st of each year (the time step might end on Jan 2nd, Jan 15th, Jan 27th, etc.)
+
+**What to Sample** 
+
+.. list-table:: Simulant Attribute to Sample 
+  :widths: 20
+  :header-rows: 0
+
+  * - Household number (random indicator which is the same for simulants in the same household)
```
NathanielBlairStahn, line 1633: Do you mean "random" or do you mean "arbitrary"? I.e. do we need this number to have any particular statistical properties, or can it just be anything as long as it's unique for each household?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1612,6 +1612,154 @@ Note/limitations:
...
+  * - DOB 
+  * - Home Address 
+  * - Home Zip Code 
+  * - Race/ethnicity 
+
+Here is an example: 
+
+.. image:: WIC_example.png
+
+Note: for household number, if a tracked mother/guardian is in the sim along with their
```
NathanielBlairStahn, line 1648: I had been imagining that the "household number" or "household ID" could be the same as the "address ID" (since I believe we're going to stop generating addresses in the simulation and have a unique identifier for each address, adding the addresses in the first post-processing/observer step).

But it sounds like what you're saying here is that the household ID cannot necessarily be the same as the address ID -- it sounds like they could be the same thing *except* in the case where the mother and child live at different addresses, in which case a new ID would need to be generated. Is that correct?

Also, does "tracked mother/guardian" here mean tracked by the sim, or tracked by the WIC observer?

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1612,6 +1612,154 @@ Note/limitations:
...
+  * - Home Zip Code 
+  * - Race/ethnicity 
+
+Here is an example: 
+
+.. image:: WIC_example.png
+
+Note: for household number, if a tracked mother/guardian is in the sim along with their 
+infant, please assign them the same household number even if they do not live together 
+currently.
```
NathanielBlairStahn, line 1650: I had been imagining that the "household number" or "household ID" could be the same as the "address ID" (since I believe we're going to stop generating addresses in the simulation and have a unique identifier for each address, adding the addresses in the first post-processing/observer step).

But it sounds like what you're saying here is that the household ID cannot necessarily be the same as the address ID -- it sounds like they could be the same thing *except* in the case where the mother and child live at different addresses, in which case a new ID would need to be generated. Is that correct?

Also, does "tracked mother/guardian" here mean tracked by the sim, or tracked by the WIC observer?

SylLutze: Hmm I didn't realize we had address ID tracked in the sim! I had added this so we could tell who was related, which generally means within the same household. You had pointed out yesterday that sometimes a parent might be a different household and so I updated this to more mean "family ID" not a true "house ID". 

"Tracked mother/guardian" means tracked by the sim. We have had a lot of conversation about the overlapping concepts of "tracked mother (person who gives birth)", "guardian" and "parent (from the relationship attribute)". So this language is intended to be quite specific.

PR #1050, opened 2022-11-09
Vivarium modeling overview filesystem restructure

alibow:
Changes as discussed in this slack thread: https://ihme.slack.com/archives/CQ8MGENNA/p1667868607331959

Notably this PR only contains changes to the "vivarium modeling overview" section and has not made changes to the "vivarium models" section. This is mostly because this PR affects so many files as is and I am thinking that there will be more potential for merge contents with the vivarium models section so I'll follow up with that one after getting this one merged.

There are a ton of files changed in this PR, so I suggest building the html locally and navigating through the file structure that way to do your review. Please let me know if you have any questions and/or suggestions!


zmbc (approved):
This seems like a step in the right direction. I think we will continue to move pages and reorganize a bit as we fill out the content of these pages and add new ones.

**One nit**: it seems like the page "Monte Carlo Uncertainty in Vivarium" is pretty Vivarium-specific and would fit better in either the "Overview..." or "Considerations..." section.

I think long-term the "Considerations in the design of Vivarium models" section will probably be broken up -- perhaps some of it will be filed under a type of model component (e.g. "Accounting for the Baseline Coverage of an Intervention" can become part of the docs on modeling interventions), and other parts will become part of "Overview" or the tutorial we were talking about including. As it is, it's a bit of a miscellaneous section, but I don't think that can be avoided without significantly reworking the content.

aflaxman (approved):

PR #1051, opened 2022-11-09
Document employment as currently implemented

zmbc:
Nothing new -- all of this is intended to reflect what has already been done. Hoping to get SE review to verify that this is the case.


docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2157,53 +2157,127 @@ limitation.
...
+We consider all simulants age 18 and over to be working-age; all such
+simulants either have an employer or are considered unemployed.
+We only allow a single employer at a time for each simulant.
+
+Each employer has a single street address.
+
+Initialization
+^^^^^^^^^^^^^^
+
+We initialize 58% of the working-age population to be unemployed,
```
beatrixkh, line 2176: 57.6%. we technically assign "unemployed" with probability .576, so it won't be precisely this number.

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2157,53 +2157,127 @@ limitation.
...
+regardless of demographics.
+We initialize 3% of the working-age population to be employed by the military,
```
beatrixkh, line 2178: similar to above --- with probability 0.03

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2157,53 +2157,127 @@ limitation.
...
+no employment change occurs;
+they remain employed by the military.
+This means that in practice, the actual rate of employment changes will be a bit
+less than 50 per 100 person-years.
+
+When an employed simulant changes employment, they have a 58% chance of
+becoming unemployed and a 3% chance of becoming employed by the
+military.
+Otherwise, they sample a new non-military employer with probability
+proportional to employer size, as at initialization.
```
beatrixkh, line 2237: note that this is specifically proportional to the employer sizes as they were at initialization. Over time this results in the employer sizes moving towards ... something less skewed than they were to start. I'm not sure precisely where it hits equilibrium

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2157,53 +2157,127 @@ limitation.
...
+this whole procedure is repeated (or an equivalent mechanism
+to prevent job "changes" that don't change employment and redistribute
+the probability among all other options).
+
+The same procedure is done for an unemployed simulant selected for an employment change,
+except that instead of not being allowed to "change" to the same employer,
+they are not allowed to "change" to remain unemployed.
+
+This approach to selecting a new employer ensures that at the population level,
+the *relative* sizes of different employers will remain roughly constant (and
```
beatrixkh, line 2249: oh oops you commented on this down here. 

i think ^ that that's not true and that the employer sizes shifted around; for example i think unemployment steadily grows(?). i ran a mini sim to test this / to show abie the weird dynamics we were going to get, but i'm sorry i don't know if i saved the results because this was originally a placeholder implementation

beatrixkh (approved):

aflaxman (approved):

docs/source/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Ah, okay. From reading the Slack archives, I had thought that this issue was fixed.

I can see why what we've done here will lead to a _less_ skewed distribution. If there is any easy way to check how drastic of an effect this is, let's check. But I'm not too worried about it.

However, that seems directly at odds with unemployment growing (unemployment, the single largest "employer," should shrink in a less skewed distribution), especially "steadily" which would imply no equilibrium. I can't see why what we've implemented would cause that, but if it is, I think we need to fix it.

SylLutze (approved):

PR #1052, opened 2022-11-10
Custom parameterized draw feature overview

alibow:


docs/source/model_design/vivarium_features/custom_parameterized_draw/index.rst
```diff
@@ -0,0 +1,130 @@
...
+---------------------------------------------------
+
+Historical behavior
++++++++++++++++++++
+
+For GBD parameters with draw-level estimates (this notably excludes GBD covariate estimates), we have randomly selected a numbered draw between 0 and 999 and then used the corresponding value for each GBD draw-level parameter used in a simulation. The draw number used is reported to researchers who therefore can easily look up the specific parameter values used for that simulation run. Notably, we have not documented draw-level correlation patterns within GBD estimates to understand the potential implications of this strategy, but it is consistent with GBD methodology.
+
+For parameters without draw-level estimates, including GBD covariate estimates and values obtained from the literature, we have randomly sampled a value from withing the parameter's uncertainty distribution for each run of the simulation. The parameter value used for each simulation run has not historically been reported with simulation outputs, so researchers therefore cannot easily look up the specific values used for a given simulation run for parameters without draw-level estimates.
```
aflaxman, line 45:
```suggestion
For parameters without draw-level estimates, including GBD covariate estimates and values obtained from the literature, we have randomly sampled a value from within the parameter's uncertainty distribution for each run of the simulation. The parameter value used for each simulation run has not historically been reported with simulation outputs, so researchers therefore cannot easily look up the specific values used for a given simulation run for parameters without draw-level estimates.
```

aflaxman (approved):

docs/source/model_design/vivarium_features/custom_parameterized_draw/index.rst
```diff
@@ -0,0 +1,138 @@
...
+Selecting the "average" value for a parameter may be challenging in the case of skewed parameter distributions such as relative risks where the mean and median values differ.
+
+For sensitivity analysis and tornado diagrams
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Our simulations contain a high number of parameters, so running multiple values for each parameter may be onerous and require significant computational resources.
+
+Identifying reasonable upper and lower bounds for each parameter may be difficult. For instance,
+
+- Selecting upper and lower 95% uncertainty intervals may bias our assessment of influential parameters towards parameters with more parameter uncertainty.
```
zmbc, line 117: Whether or not this is a bias depends on what you are using the tornado diagram for, I think. For a standard sensitivity analysis, this seems exactly right and not biased at all. It answers the question: "how does our uncertainty about the inputs contribute to our uncertainty about the output?"

If you want to use the tornado diagram for more of an "effect size" and making preliminary guesses about how things vary by location/sub-population, instead of UIs here you'd want to use some measure of how much you think that parameter varies between locations/sub-populations.

I could imagine a parameter where we're fairly unsure of the true value but for biological reasons we think it is likely to be almost constant across sub-populations; in that case, the two approaches would significantly differ.

zmbc (approved):

docs/source/model_design/vivarium_features/custom_parameterized_draw/index.rst
alibow: Good point! How do you feel about the term "will favor" instead of bias? Something like...

`Selecting upper and lower 95% uncertainty intervals will favor parameters with greater parameter uncertainty as more influential parameters. These values may be appropriate if the aim is to assess how input parameter uncertainty influences output uncertainty. However, if the goal is to assess parameter influence across a range of plausible values (across locations or subpopulations, for example), the upper and lower uncertainty bounds may not be the most appropriate parameter values to select.`

zmbc: Sounds good!

PR #1060, opened 2022-11-21
Cause model for other causes with YLDs

alibow:
This is a proposal to model YLDs in addition to YLLs attributable to "other_causes." This will be useful for two reasons:
1. More accurate estimation of intervention impact on DALYs
2. Can allow us to model risk factor impact on all causes rather than specific causes and remove the need for cause model transitions at every (frequent) time step, which could really speed up our sims when this model structure is appropriate! This could work very well for the IV iron child sim and other similar models.

The big question mark in terms of implementation for me is whether we can have risks modify the cause state disability weight as I have proposed here. Perhaps we can run through this in an upcoming coworking session or separate meeting to discuss feasibility and best approaches between researchers and engineers!


aflaxman (approved):
Was including figure.drawio intentional?

Would it be simpler to make this component *only* handle YLDs, since we already have other-cause deaths (and therefore YLLs)?

alibow:
@aflaxman I did it this way because I think it will more easily allow us to include it as a risk outcome for risk factors and it shouldn't be too much of a lift because there it fits within the structure of a typical cause model. That would be possible though, yes!

SylLutze (approved):

PR #1072, opened 2022-12-05
Updates to Heart Failure

SylLutze:
@alibow I would appreciate if you can help me check the equations for the transitions in particular. Happy to chat about how I got these, but it is a bit confusing and I would appreciate another set of eyes.


alibow:
Ok I think this all looks good to me, except I am confused about why angina does not show up anywhere here. Should the angina sequelae show up in the PostMI sequelae list? I only see 1040 there currently.

alibow (approved):

alibow:
Also, there's no morbidity associated with heart failure? Is this just definitional to heart failure or is it an assumption we are imposing?

SylLutze:
So, we removed angina because of Greg's request. Why do you say there is no morbidity? Heart failure has disability weights associated with it, do I need to include more information?

alibow:
Ok, thanks for the clarification! 

I think that the "State definitions" table should be updated then to make it clear that angina is part of the S1 rather than P state and then maybe add a point about this in the assumptions/limitations section with a brief reminder about why this decision was made.

And sorry, my comment about the disability weights was not well explained... it stemmed from noticing that there's no difference in disability weight for AMI and for AMI_HF states. It looks to me like all of the disability for heart failure due to IHD is attributed to the IHD_HF state and none in the AMI_HF state. Do we want some to be experienced in the AMI_HF state as well? Either way, I think stating the exact assumption made would be a good addition to the documentation.

PR #1074, opened 2022-12-05
Splitting general epi into two pages

SylLutze:
A couple things I want to note: 
1. This is just a reorganization into two pages rather than the one giant "epi" page. I have not significantly gone through the material or edited content 
2. Once we are aligned on this, I will delete the original General Epi page completely since it is now in two separate sections


alibow (approved):
I think this is a really helpful way to break these pages apart - thanks :)

aflaxman (approved):

PR #1088, opened 2022-12-12
Defining Column Noise

SylLutze:
Note that this is column noise only, it does not include row noise right now (duplicates and omissions)


docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,268 @@ have not been included. These are:
...
+
+  * - Data in Observer
+    - Observers Present 
+    - Maximum Noise Level 
+    - Types of Noise 
+    - Notes
+  * - First Name
+    - Census, Household Surveys, WIC, Taxes (both), SSA  
+    - 5%
+    - OCR, phonetic, typographic, fake names
```
aflaxman, line 2838: let's have Nicknames, too

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,268 @@ have not been included. These are:
...
+available and information for implimentation. **Software engineering team - please alert the research team if any of the below looks to be particularly challenging for further discussion.**
+
+**OCR**
+
+Optical character recognition is when a character is misread for another character that 
+looks similar. A common examples is 'S' and '5'. In order to emulate 
+that, there is a `GeCo like corrupter and related list of possible changes in the ocr_variations_upper_lower csv found here <https://github.com/ihmeuw/vivarium_research_prl/tree/main/noise>`_. 
+
+Currently, the user defined rate will be a rate 'eligible' for noise. The actual rate of noise 
+will be determined at the character level as XX errors per character. We think this approximates to 1
```
aflaxman, line 2961: What is XX here?

aflaxman: oh, I see the todo about it below.

aflaxman (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Yep! I think Nathaniel and I still need to work out what value this is going to be so its a placeholder for now.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Added!

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,284 @@ have not been included. These are:
...
+For all numeric miswriting, first select the sample to have noise added. 
+Once the sample is selected, change 1 digit of the number at random to any 
+other random number. 
+
+For example: if the correct SSN is 123456789, a miswriting might look like 123416789. 
+Please confirm that the new digit is an actual error. 
+
+Limitations: 
+
+- This might lead to illogical data, especially for age and dates (e.g., a person who's birthday is 12/87/2000). It is more likely that someone lists an incorrect but still possible birthday/age. However, since the main goal is noise for PRL, we think this is still acceptable.
```
zmbc, line 3045: I think we need to clarify how this applies to DOB. It seems as though we are thinking of DOB as a string here. If that is the case, we need to specify the format (for each observer? is it always MM/DD/YYYY?). A similar question applies to month/day swapping.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,284 @@ have not been included. These are:
...
+noise added to the data. Noise is generally divided into two types: 
+
+#. **Column based noise:** errors in individual data entry, such as miswriting or incorrectly selecting responses 
+#. **Row based noise:** errors in the inclusion or exclusion of entire rows of data, such as duplication or omission 
+
+To begin, we will start with defining column based noise. Some general rules 
+for all column based noise include: 
+
+- The amount of each type of noise is individually configurable for each observer up to the max listed in the table below. This means that the end user can select that: census first names have 5% typographic errors. The minimum noise is 0 for all column noise. 
+- Simulants are selected for noise at random. Selection is not based on any attributes and simulants do not have a higher or lower propensity for noise that would carry across observers (e.g., there are not "messy" simulants who make errors on all forms).
```
zmbc, line 2823: Can you explain a bit more explicitly that a "noise level" is the probability of selecting each row to have that form of noise and that the selection for each form of noise is independent?

Also, we will need to define the order in which these noise types are applied. I don't think it matters much, but it will need to be chosen at some point so we might as well specify it now.

zmbc, line 2823:
```suggestion
- Simulants are selected for noise at random. Selection is not based on any attributes and simulants do not have a higher or lower propensity for noise (e.g., there are not "messy" simulants who are more likely to make errors on all fields/forms).
```
Noting that it is independent between columns in a single observer as well.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,284 @@ have not been included. These are:
...
+Noise Functions
+^^^^^^^^^^^^^^^
+
+In order to have a realistic challenge with PRL, it is essential to have 
+noise added to the data. Noise is generally divided into two types: 
+
+#. **Column based noise:** errors in individual data entry, such as miswriting or incorrectly selecting responses 
+#. **Row based noise:** errors in the inclusion or exclusion of entire rows of data, such as duplication or omission 
+
+To begin, we will start with defining column based noise. Some general rules
```
zmbc, line 2819: Can you include a TODO for row noise, maybe have a subsection for it that is empty?

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,284 @@ have not been included. These are:
...
+In order to have a realistic challenge with PRL, it is essential to have 
+noise added to the data. Noise is generally divided into two types: 
+
+#. **Column based noise:** errors in individual data entry, such as miswriting or incorrectly selecting responses 
+#. **Row based noise:** errors in the inclusion or exclusion of entire rows of data, such as duplication or omission 
+
+To begin, we will start with defining column based noise. Some general rules 
+for all column based noise include: 
+
+- The amount of each type of noise is individually configurable for each observer up to the max listed in the table below. This means that the end user can select that: census first names have 5% typographic errors. The minimum noise is 0 for all column noise.
```
zmbc, line 2822:
```suggestion
- The amount of each type of noise is individually configurable for each column in each observer. This means that the end user can select that: census first names have 5% typographic errors. 
```

I don't think having a maximum value here makes things any easier on the engineering side, so I'd switch this to just being a default.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,284 @@ have not been included. These are:
...
+- Simulants are selected for noise at random. Selection is not based on any attributes and simulants do not have a higher or lower propensity for noise that would carry across observers (e.g., there are not "messy" simulants who make errors on all forms). 
+- As noise functions for certain data types are common across observers, the table below is organized by data type (first name). Below the table, there is further information and defintion on each noise type. 
+
+.. list-table:: Type of Noise Included and Maximum Level by Data Included
+  :widths: 20 20 20 20 20 
+  :header-rows: 0
+
+  * - Data in Observer
+    - Observers Present 
+    - Maximum Noise Level
```
zmbc, line 2832:
```suggestion
    - Default Noise Level 
```
See above.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,284 @@ have not been included. These are:
...
+To do this, first determine who is eligible for copying. This is essentially simulants 
+with at least one other person in their household. Note that GQ is not eligible for this 
+type of noise. 
+
+From the eligble simulants, select the sample to have noise added. For those individuals, 
+copy the relevant piece of data from another person in the household. 
+
+Limitations: 
+
+- This oversimplifies some swapping of ages or birthdays between family members. However, it allows up better control over the percent of simulants to receive incorrect information and will likely pose a similar PRL challenege.
```
zmbc, line 3061:
```suggestion
- This oversimplifies some swapping of ages or birthdays between family members. However, it allows better control over the percent of simulants to receive incorrect information and will likely pose a similar PRL challenge. 
```

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,284 @@ have not been included. These are:
...
+
+Month and day swaps apply to dates. For this, select the sample to have noise added. 
+For those selected, swap the month and day to be in the incorrect position (e.g., December 8, 2022 
+would be listed in MM/DD/YYYY format as 08/12/2022). 
+
+**Incorrect Select** 
+
+Incorrect select applies to a range of data types. For this, select the sample to 
+have noise added. For those selected, randomly select a different option from the 
+others available. Note that for relationship to head of household, this includes
```
zmbc, line 3073: I predict that there may be engineering questions about what "others available" means. Is it the unique values present in the observer output? Do we plan to have set lists of the options available for each column?

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,284 @@ have not been included. These are:
...
+#. **Row based noise:** errors in the inclusion or exclusion of entire rows of data, such as duplication or omission 
+
+To begin, we will start with defining column based noise. Some general rules 
+for all column based noise include: 
+
+- The amount of each type of noise is individually configurable for each observer up to the max listed in the table below. This means that the end user can select that: census first names have 5% typographic errors. The minimum noise is 0 for all column noise. 
+- Simulants are selected for noise at random. Selection is not based on any attributes and simulants do not have a higher or lower propensity for noise that would carry across observers (e.g., there are not "messy" simulants who make errors on all forms). 
+- As noise functions for certain data types are common across observers, the table below is organized by data type (first name). Below the table, there is further information and defintion on each noise type. 
+
+.. list-table:: Type of Noise Included and Maximum Level by Data Included
```
zmbc, line 2826:
```suggestion
.. list-table:: Type of Noise Included and Default Level by Data Included
```

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,284 @@ have not been included. These are:
...
+  :header-rows: 0
+
+  * - Data in Observer
+    - Observers Present 
+    - Maximum Noise Level 
+    - Types of Noise 
+    - Notes
+  * - First Name
+    - Census, Household Surveys, WIC, Taxes (both), SSA  
+    - 5%
```
zmbc, line 2837: Just to confirm I am understanding: this means 5% for _each_ error type? So roughly 25% of people getting their name wrong?

This seems very high. We should pay more attention to these defaults but I don't think they are blocking, so maybe add a TODO that they are not final.

zmbc (commented):
I realize this is quite a long review; feel free to reach out if you want to talk through anything in real time.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,284 @@ have not been included. These are:
 #. We sample 100% of events. This is likely unrealistic, but the percent is probably very high still. 
 #. There are errors in SSN data, which are not modeled here. 
 
+Noise Functions
+^^^^^^^^^^^^^^^
+
+In order to have a realistic challenge with PRL, it is essential to have 
+noise added to the data. Noise is generally divided into two types:
```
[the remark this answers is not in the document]
SylLutze: Yeah, Nathaniel pointed out that there is data type or data formatting noise that we haven't included. Like dates or zip codes can be formatted in different ways. I will adjust the language here in case we want to add it later!

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,284 @@ have not been included. These are:
...
+
+#. **Column based noise:** errors in individual data entry, such as miswriting or incorrectly selecting responses 
+#. **Row based noise:** errors in the inclusion or exclusion of entire rows of data, such as duplication or omission 
+
+To begin, we will start with defining column based noise. Some general rules 
+for all column based noise include: 
+
+- The amount of each type of noise is individually configurable for each observer up to the max listed in the table below. This means that the end user can select that: census first names have 5% typographic errors. The minimum noise is 0 for all column noise. 
+- Simulants are selected for noise at random. Selection is not based on any attributes and simulants do not have a higher or lower propensity for noise that would carry across observers (e.g., there are not "messy" simulants who make errors on all forms). 
+- As noise functions for certain data types are common across observers, the table below is organized by data type (first name). Below the table, there is further information and defintion on each noise type.
```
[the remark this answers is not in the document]
SylLutze: Sure - happy to adjust

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,284 @@ have not been included. These are:
...
+
+  Include character based noise level (applies to phonetic and typographic too)
+
+Limitations: 
+
+- The result of having parameters at the character level is that we cannot tell exactly how many data points have noise in them. With the current settings, it is likely similar to the user selected value, but we will always have fewer errors than expected. This could be improved at a later time. 
+
+**Phonetic**
+
+Phoentic errors are when a character is misheard. This could similar sounding letters when
```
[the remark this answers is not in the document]
SylLutze: Thanks

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,284 @@ have not been included. These are:
...
+determine who is eligible for a nickname. Then select simulants for noise. Lastly, replace their 
+name with any of the nicknames included in the csv. If there are multiple options, 
+select at random. 
+
+.. note::
+
+  If the process of determining elgibility prior to selecting simulants for noise is challenging, we can work on finding a simpler approach. 
+
+
+**Miswriting**
```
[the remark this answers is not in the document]
SylLutze: Sure

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Sure

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Hahah well I was thinking of this as a max! Yes, that would be insane. When I thought of a max, it was more if someone wanted to test very high noise of a specific type to see how their PRL would handle it. If it is a default then yes, I will definitely turn this down a good amount.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Sure, if that's easier!

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Fair, I tried to clarify this a bit. In my head it is any other option that was available in the original observer. If this is still not clear enough and I should define it explicitly I can

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Okay, I expanded and added order to this

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,284 @@ have not been included. These are:
...
+Limitations: 
+
+- Many of the fake first names include some information about the simulant (daughter, child f, minor) all specify something about the simulant. We will not try to match this information, which might lead to illogical information (an older man being labeled as 'daughter') but will not impact PRL. 
+- Someone who is likely to use a fake name might well do so across multiple observers. This would likely increase PRL challenges but will not be included here. 
+
+**Nicknames**
+
+Many people choose to use nicknames instead of their "real" names. A common example is an 
+Alexander who chooses to go by Alex. These individuals might write their nicknames on forms 
+which should be recorded. Here is a list of 1080 names and their `relevant nicknames <https://github.com/carltonnorthern/nicknames/blob/master/names.csv>`_.
```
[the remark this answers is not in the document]
SylLutze: Makes sense!

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,284 @@ have not been included. These are:
...
+available and information for implimentation. **Software engineering team - please alert the research team if any of the below looks to be particularly challenging for further discussion.**
+
+**OCR**
+
+Optical character recognition is when a character is misread for another character that 
+looks similar. A common examples is 'S' and '5'. In order to emulate 
+that, there is a `GeCo like corrupter and related list of possible changes in the ocr_variations_upper_lower csv found here <https://github.com/ihmeuw/vivarium_research_prl/tree/main/noise>`_. 
+
+Currently, the user defined rate will be a rate 'eligible' for noise. The actual rate of noise 
+will be determined at the character level as XX errors per character. We think this approximates to 1
```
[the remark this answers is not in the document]
NathanielBlairStahn: >It'd be a bit cleaner in my opinion to ask the pseudopeople user to supply a per-character error probability, but it would make everything a bit more complicated.

@zmbc Are you suggesting having the end user supply *only* a per-character probability, e.g., applying to all characters in the column (or dataset?), *instead* of having them supply a row-eligibility probability? Or are you suggesting having them supply a per-character probability *in addition* to the row probability, so that it would represent the conditional probability of a character being corrupted, given that it is in a row that is selected?

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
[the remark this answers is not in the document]
NathanielBlairStahn: > Yeah, Nathaniel pointed out that there is data type or data formatting noise that we haven't included. Like dates or zip codes can be formatted in different ways.

I think this type of noise would be column noise in our terminology. If we haven't figured out how to do it yet, maybe we should add a to-do noting that we will want to include this additional type of column noise later.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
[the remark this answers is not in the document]
zmbc: @NathanielBlairStahn good question. I was saying that it would be cleaner to have them _only_ supply a per-character probability. But, I get it if we don't want to do that.

On the other hand, having the option to customize the per-character probability conditional on the row being selected sounds like a good idea, if it isn't too hard to implement.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,284 @@ have not been included. These are:
...
+
+.. note::
+
+  If the process of determining elgibility prior to selecting simulants for noise is challenging, we can work on finding a simpler approach. 
+
+
+**Miswriting** 
+
+For all numeric miswriting, first select the sample to have noise added. 
+Once the sample is selected, change 1 digit of the number at random to any
```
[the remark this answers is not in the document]
NathanielBlairStahn: I haven't thought very carefully about how to do this well, but I was imagining we'd use a character-level probability of corruption like in the other noise functions.

And we might want different probabilities for different characters, depending on what the number represents. For example, for the WIC case study, I implemented a noise function for zipcodes that has a higher probability of corruption for the last 2 characters than for the first 3 characters, as it seems like the final two characters would be more likely to be wrong. Doing it that way should also be interesting for linkage since one blocking strategy is to block on the first 3 digits of zipcode.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
NathanielBlairStahn: I think what you're talking about is what I ran into when trying to add noise to dates for the case study.

The dates are currently stored as pandas datetime objects in the state table and census files. In this format, it is impossible to replace a date with an invalid date because the datatype only supports real dates.

So yes, in the observers, I think we need to specify that the dates are stored as strings so that noise can be added, and we need to specify how the date should be formatted for each observer. For now, I think it will be sufficient to pick a standard format to use for all observers, but eventually, I think we we will want to match the format to whatever format(s) we see in real data.

NathanielBlairStahn:
Another type of noise that is not on our list is swapping two adjacent characters. For example, writing "teh" instead of "the." That seems like maybe a kind of typo or "keyboard error" that is not captured in the `keyboard_corrupt` function we currently have.

Actually, it seems like this could fall into a new category of "misspelling errors," which may overlap with keyboard errors and phonetic errors, but I don't think it's exactly the same. Although perhaps with enhanced versions of those functions we could capture most cases. Another option could be to find a list of commonly misspelled words, similar to the other data files we have found for adding noise.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3070,7 +3076,9 @@ would be listed in MM/DD/YYYY format as 08/12/2022).
 
 Incorrect select applies to a range of data types. For this, select the sample to 
 have noise added. For those selected, randomly select a different option from the 
-others available. Note that for relationship to head of household, this includes 
+correct one. This is randomly chosen from the original list of options 
+(e.g., for state, if WA is the true answer randomly pick any other state).
```
zmbc, line 3080: The issue is that even though this "list of options" makes perfect intuitive sense, it doesn't really exist anywhere.

For example, when we initialize simulants to be in a state, we don't choose from a list of states; instead, we sample a row from the 15M rows in the ACS PUMS to base that simulant on, and that row already has a state attribute.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3019,19 +3021,23 @@ Limitations:
 
 Many people choose to use nicknames instead of their "real" names. A common example is an 
 Alexander who chooses to go by Alex. These individuals might write their nicknames on forms 
-which should be recorded. Here is a list of 1080 names and their `relevant nicknames <https://github.com/carltonnorthern/nicknames/blob/master/names.csv>`_. 
+which should be recorded. Here is a list of 1080 names and their `relevant nicknames <https://github.com/ihmeuw/vivarium_research_prl/blob/main/noise/nicknames.csv>`_.
```
zmbc, line 3024: Oops sorry for the back and forth, but we will need to still cite where we got this.

zmbc:
> Another type of noise that is not on our list is swapping two adjacent characters. For example, writing "teh" instead of "the." That seems like maybe a kind of typo or "keyboard error" that is not captured in the `keyboard_corrupt` function we currently have.

I'd be in favor of calling this a keyboard error (since it is an error generated by someone knowing what they want to type but typing it wrong) and adding it to `keyboard_corrupt`.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
NathanielBlairStahn: Do we want approximately 1 error per selected string? That seems like it might be a good reasonable default, but I hadn't really thought about it. On the other hand, @zmbc [suggested](https://github.com/ihmeuw/vivarium_research/pull/1088/files#r1047510209) that maybe it's better to have the user specify the per-character error probability.

If we *do* want approximately 1 error per selected string, we could have a column-dependent value of XX, determined something like this (some of Zeb's comments gave me this idea):

1. Compute the average length of strings in the column (or in a random sample of strings in the column that will be eligible for error)
2. Choose XX to be 1/(average string length)
3. Choose the fraction of eligible rows to be
```
(desired fraction of rows with error) / (probability of at least one error),
```
which, as Zeb pointed out [below](https://github.com/ihmeuw/vivarium_research/pull/1088#discussion_r1047504927), is approximately
```
(desired fraction of rows with error) / 0.65
```
if the average string length is at least 5. The denominator would be larger for shorter strings (1 for length 1, 0.75 for length 2,...) and approaches 1-1/e = 0.632... as the average string length approaches infinity.

In this case the average number of corrupted characters in a string ,given that there is at least one error, would be approximately 1/0.65 = 1.53.... Note that there's no way to make this value exactly equal to 1 since the minimum number of errors is 1. (However, we could drive the average arbitrarily close to 1 by artificially inflating the number of strings with exactly 1 error. That is, we could modify the individual noise functions to make errors in exactly one character more likely than would occur when all characters independently have the same probability of corruption. But that seems too complicated for a first pass.)

With this strategy, the user could specify the actual fraction of rows they want errors in, instead of just the eligible fraction.

Except that this strategy won't quite work for OCR and phonetic noise, which have a "per-token" error probability instead of a per-character probability (tokens are up to 3 characters for OCR and up to 7 characters for phonetic).

NathanielBlairStahn: So yeah, it's actually not accurate to say that we have a "rate of XX errors per character," because it's really a rate per *token*, and tokens have variable length. Sorry, @SylLutze, for not being more precise when I described how the noise functions work!

I think the variable-length token problem was one of the main reasons I was saying we should leave the probabilities for the noise functions as an implementation detail to figure out later.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,297 @@ have not been included. These are:
...
+
+
+**Typographic** 
+
+Typographic errors occur due to mistyping information. The commonality of errors are therefore 
+based on the QWERTY keyboard layout. These errors can include added characters, missed characters, 
+and replaced characters. In order to emulate that, there is a `GeCo like corrupter and related layout of the QWERTY keyboard csv found on github <https://github.com/ihmeuw/vivarium_research_prl/tree/main/noise>`_. 
+
+Currently, the user defined rate will be a rate 'eligible' for noise. The actual rate of noise 
+will be determined at the character level as XX errors per character. We think this approximates to 1
```
NathanielBlairStahn, line 2999: The [typographic noise function](https://github.com/ihmeuw/vivarium_research_prl/blob/main/src/vivarium_research_prl/noise/corruption.py#L119) actually has two user-specified probabilities instead of one: (1) the probability that a token is corrupted, and (2) the probability that a corrupted token contains the original token (a single character in this case) in addition to the corruption.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,297 @@ have not been included. These are:
...
+    - Incorrect select 
+    - 
+  * - Date of SSA Event 
+    - SSA 
+    - 1%
+    - Numeric miswriting, month and day swap 
+    - 
+
+The below section further describes types of noise including any code 
+available and information for implimentation. **Software engineering team - please alert the research team if any of the below looks to be particularly challenging for further discussion.**
```
NathanielBlairStahn, line 2954:
```suggestion
available and information for implementation. **Software engineering team - please alert the research team if any of the below looks to be particularly challenging for further discussion.**
```

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Hahah sure, makes sense

SylLutze:
Okay @NathanielBlairStahn and @zmbc I think I have addressed almost all of the comments. Even if that is just to add a todo note! Would love if someone can check my math on this one. Also for the dates comments - I am going to define the dates in the observers as strings for now in a particular format, but was going to do that in a separate PR.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,349 @@ have not been included. These are:
...
+
+Next, once a row is selected to receive noise, we define how much noise it should receive. 
+The default value here is 1 error per row, but can be changed. To calculate the error rate 
+per character/token, we can use this equation: 
+
+Error rate per Character/Token = (user defined errors per row)/(average string length)
+
+This approach has a number of assumptions and limitations: 
+
+- We assume that there are a similar number of tokens and characters in each string. This is approximately true for longer strings. While it is incorrect for short strings, we feel it is still reasonable overall.
```
NathanielBlairStahn, line 3013: I don't think it's true that the number of tokens is approximately equal to the number of characters for long strings. It's plausible that the ratio of tokens to characters could approach some ratio as the strings get longer, but that's not even clear because it depends on what type of data the strings represent (e.g., the ratio for names would probably be different from the ratios for addresses, birthdays, or zipcodes).

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,349 @@ have not been included. These are:
...
+Next, once a row is selected to receive noise, we define how much noise it should receive. 
+The default value here is 1 error per row, but can be changed. To calculate the error rate 
+per character/token, we can use this equation: 
+
+Error rate per Character/Token = (user defined errors per row)/(average string length)
+
+This approach has a number of assumptions and limitations: 
+
+- We assume that there are a similar number of tokens and characters in each string. This is approximately true for longer strings. While it is incorrect for short strings, we feel it is still reasonable overall. 
+- The user defined values for errors will not be exact, but rather approximated with this method.
```
NathanielBlairStahn, line 3014: I think in the real pseudopeople package, we need to define the parameters very clearly so the user is not confused. So we should come up with clear, descriptive names for whatever parameters we decide the user should specify, and tell the user whether it represents an exact or approximate quantity, and how different we expect the actual values to be if it's approximate.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,349 @@ have not been included. These are:
...
+The default value here is 1 error per row, but can be changed. To calculate the error rate 
+per character/token, we can use this equation: 
+
+Error rate per Character/Token = (user defined errors per row)/(average string length)
+
+This approach has a number of assumptions and limitations: 
+
+- We assume that there are a similar number of tokens and characters in each string. This is approximately true for longer strings. While it is incorrect for short strings, we feel it is still reasonable overall. 
+- The user defined values for errors will not be exact, but rather approximated with this method. 
+- While the user provides an overall rate of errors per rows, in reality the number of expected errors per string is a function of the length of the string.
```
NathanielBlairStahn, line 3015: Since it's an average over all strings, I don't think the expected number of errors per string is a function of the length of the string. Rather, the conditional expectation of the number of errors, given that there is at least one error, is a function of the average number of tokens per string, which is unknown. The overall average number of errors per string also depends on the user-selected fraction of rows with errors.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,349 @@ have not been included. These are:
...
+
+#. Compute the average length of strings in the column 
+#. Probability of at least 1 error = 1 - ((1 - (user defined errors per row / average string length))^(average string length))
+#. Rows eligible for errors = User defined rows with errors / probability of at least 1 error 
+
+Next, once a row is selected to receive noise, we define how much noise it should receive. 
+The default value here is 1 error per row, but can be changed. To calculate the error rate 
+per character/token, we can use this equation: 
+
+Error rate per Character/Token = (user defined errors per row)/(average string length)
```
NathanielBlairStahn, line 3009: @SylLutze @zmbc @aflaxman This was just an idea I wrote down on the fly, and I'm not entirely sure it is the best strategy, for a few different reasons.

I was going to send a message in Slack this morning, but then I was out because of sick kids: I suggest that the researchers meet to discuss how we want to parameterize the noise functions for the end user in pseudopeople, and whether the version we specify for the minimal model should be the same as what we want in the final product or whether we want to do something simpler for now. (I think how we choose to simplify things for the MVM might depend on what we eventually want to do, but I'm not clear on what the best strategy is for the final product from a usability perspective.)

NathanielBlairStahn: My original thought was that the simplest thing to do for for the MVM would be essentially what Syl wrote down in the first place: Allow the user to specify the fraction of eligible rows, and for now have the researchers choose fixed values for the per-token error probabilities (which was left as a to-do because I hadn't thought about what good default values would be). The main problem with the original writeup was that we need to be clear for the engineers that it's token-level probabilities instead of character-level, and we should be clear for the engineers and the end user that the user is specifying the probability of *eligibility* for errors rather than the probability of actually getting an error, and we could give some estimates of what the difference would be under certain assumptions.

For the final product, I think we should do something more similar to what is proposed here, but I think we can do a better job about having parameters that are useful and intuitive to the user, and having code to accurately match these instead of just saying that everything is approximate and that's a limitation. But there are several details to work out.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Ah, you are right; I was misunderstanding what "tokens" really meant (I thought they were dense, i.e. every 3-character string was a token).

This can be solved pretty simply: sample 100 values and calculate the number of tokens in each, then take the average.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: @NathanielBlairStahn First of all, I agree with you that this doesn't need to be worked out right now. I think we are confident that this will work "under the hood" using a probability of eligibility, and then a per-character or per-token (depending on noise type) error rate among the eligible. I **do** think that both (not just the former) should be customizable by the user.

All the additional stuff is for usability: allowing the user to set parameters with more intuitive meaning than those two. What I proposed yesterday to Syl was to have the user set these two parameters:

- Percent of rows with any error
- Mean number of errors per incorrect row

I think these have intuitive significance in a way that "eligibility" does not.

The math of how to get from those to probability of eligibility and per-token error rate is not simple and I believe is not correct in this PR. I _think_ I've mostly worked it out on paper (so not in a form yet that I can share). But this will take more work and more in-depth review.

@SylLutze For this PR I think we should just explain the version with eligibility probability and per-token error rate and add a TODO that we want to allow the user to specify more intuitive parameters from which we would back-calculate the easier-to-implement ones.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Fair point. I believe this was referring to the conditional expectation. We are not going to be able to do any of these calculations without an estimate for average number of tokens per string (see above where I have proposed how to get this). However, we'll also need to work out the _conditional_ expectation of tokens per string among those with at least one error, which (I think?) will require also the variance of the tokens-per-string distribution.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,349 @@ have not been included. These are:
...
+
+The below section further describes types of noise including any code 
+available and information for implementation. **Software engineering team - please alert the research team if any of the below looks to be particularly challenging for further discussion.**
+
+**Converting from Noise Inputs to Noise Function Parameters for OCR, Phonetic, Typographic and Numeric Miswriting** 
+
+For simplicity for the end user, they will input two values for noise level: 
+
+#. Rows with Errors 
+#. Errors per Row
```
zmbc, line 2986: To be clear, I thought this would be errors per row **with at least one error**. But if this is overall, maybe that is also intuitive and makes the math simpler.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2807,6 +2807,349 @@ have not been included. These are:
...
+
+These parameters, while related are different in key ways and therefore we need to 
+provide a way to convert between them. 
+
+Starting with #1 from both lists, we will calculate the number of rows eligible 
+for errors to approximate the user inputted rows with errors given. To calculate 
+this we can: 
+
+#. Compute the average length of strings in the column 
+#. Probability of at least 1 error = 1 - ((1 - (user defined errors per row / average string length))^(average string length))
```
zmbc, line 3002: This is the point where I think this math is incorrect: (user defined errors per row / average string length) is not the character/token error rate that leads to the user-defined number of errors per row (whether that is per row overall or per row with at least one error), because this rate is applied to characters in _eligible_ rows.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2910,6 +2910,479 @@ have not been included. These are:
...
+
+Column Noise
+''''''''''''
+
+To begin, we will start with defining column based noise. Some general rules 
+for all column based noise include: 
+
+- Rows eligible for errors is the probability of selecting a row to have a particular type of noise added. For example, 1% noise level for incorrect selection to type of tax form means 1% of rows will be selected to have the wrong value selected. This is somewhat more complicated for: OCR, phonetic, typographic, and numeric miswriting which is elaborated on below. 
+- Token noise level is a noise parameter that only applies to certian noise types and defines the amount of errors expected once a string is selected for noise. This parameter is also elaborated on below. 
+- The amount of each type of noise is individually configurable for each column in each observer. This means that the end user can select that: in the census sex has 2% incorrect selections. The minimum noise is 0% and the maximum if 5% for all column noise. For errors per row, the minimum is 0 and maximum is 5.
```
NathanielBlairStahn, line 2930:
```suggestion
- The amount of each type of noise is individually configurable for each column in each observer. This means that the end user can specify, for example, that in the census, sex has 2% incorrect selections. The minimum noise is 0% and the maximum if 5% for all column noise. For errors per row, the minimum is 0 and maximum is 5. 
```
This was a bit confusing when I first read it because you say that noise is configurable, but then it sounds like you're specifying a definite value, so here's an attempt to make it more clear.

Also, what's the reason we're specifying a maximum noise level, and how did we choose it? It seems like there's no reason it would be easier to have a maximum than not to have a maximum, except for the sake of not having the code take forever to run if the noise level is very high.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2910,6 +2910,479 @@ have not been included. These are:
...
+particular character/token receiving noise, not all rows that are selected to be 
+eligble for noise will actually receive noise. Another limitation is that token 
+error rate is not very intuitive for the end user. 
+
+At some point in the future, we might make this more user friendly. However, for the 
+sake of a minimum functional model, this is satisfactory. 
+
+One function has an additional parameter. The typographic function needs an input for 
+the probability that a corrupted token contains the original token. For simplicity, 
+we will use 0.5 and the user will not be able to edit this value.
```
NathanielBlairStahn, line 3103: I would lead towards a smaller value than 0.5 for the default, maybe 0.1 or 0.2? Not sure it will make a huge difference for PRL, but the string comparison algorithms may behave slightly differently for different values.

I think it's fine to keep this fixed for the MVM, and I know we said we'd just have two user parameters in the final version, but I think it might be good to keep the option open to allow more fine-grained user inputs like this one in the final version.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2910,6 +2910,479 @@ have not been included. These are:
...
+- Token noise level is a noise parameter that only applies to certian noise types and defines the amount of errors expected once a string is selected for noise. This parameter is also elaborated on below. 
+- The amount of each type of noise is individually configurable for each column in each observer. This means that the end user can select that: in the census sex has 2% incorrect selections. The minimum noise is 0% and the maximum if 5% for all column noise. For errors per row, the minimum is 0 and maximum is 5. 
+- Simulants are selected for noise at random. This is true for each type of noise, each column, and each observer. Selection is not based on any attributes and simulants do not have a higher or lower propensity for noise that would carry with them (e.g., there are not "messy" simulants who are more likely to make errors on all fields/forms). 
+- As noise functions for certain columns are common across observers, the table below is organized by column (e.g., first name). Below the table, there is further information and definition on each noise type. 
+- The order of different noise types should not matter, but will go in the order they are listed in the table below.
```
NathanielBlairStahn, line 2933: I'm not sure it will make much difference for PRL, especially if the error rates are small, but I think there is often a sort of logical order to apply the noise in. For example, for names in the census, I think an order that would make sense would be

1. nicknames
2. fake names
3. phonetic noise
4. OCR noise
5. keyboard noise

E.g., first someone would decide to use a nickname or fake name, then there might be a misspelling because of phonetic confusion, then there might be a keyboard error in typing in the phonetically misspelled name.

Anyway, I think it's fine to have an arbitrary order for the MVM, but we might want to revisit this for the final version.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2910,6 +2910,479 @@ have not been included. These are:
...
+the numeric miswriting noise function to all strings with the user defined 
+character error rate. 
+
+Limitations: 
+
+- This might lead to illogical data, especially for age and dates (e.g., a person who's birthday is 12/87/2000). It is more likely that someone lists an incorrect but still possible birthday/age. However, since the main goal is noise for PRL, we think this is still acceptable. 
+
+.. todo::
+
+  Decide if any numerical strings need to be further refined in approach. Example is zip code where there would be an option to separate the first 3 and last 2 digits.
```
NathanielBlairStahn, line 3197: It would probably be good to add links to the [noise package](https://github.com/ihmeuw/vivarium_research_prl/tree/main/src/vivarium_research_prl/noise) and [corruption module](https://github.com/ihmeuw/vivarium_research_prl/blob/main/src/vivarium_research_prl/noise/corruption.py) I'm writing for the case studies. For example, I have implemented a [noise function for zipcodes](https://github.com/ihmeuw/vivarium_research_prl/blob/main/src/vivarium_research_prl/noise/corruption.py#L172) that take 3 probabilities as input (for the first 2 digits, 3rd digit, last 2 digits) and a [function to swap month and day](https://github.com/ihmeuw/vivarium_research_prl/blob/main/src/vivarium_research_prl/noise/corruption.py#L143), as well as functions for incorrect selection and to make some values missing.

NathanielBlairStahn (approved):
Approving with a few comments. Thanks for all your work, Syl!

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Thanks, I see your point on clarity! 

For the max, we can reassess this later. I just thought it might be helpful for the engineers to have a max value they should account for in designing things. But you're right, we could set the max to 100%.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: That is helpful! I changed it to be 0.1

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: I agree, I was going to make a separate (less giant) PR with ordering of the noise. This also applies to row noise. I will keep this order in mind for that.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Thanks!

PR #1090, opened 2022-12-13
Do a first pass of improvement on risk correlation

zmbc:
* Split risk correlation from mediation
* Remove unnecessary justification from docs; replace with closed-form solution, and link to Monte Carlo experiments in notebook.


zmbc:
@alibow I'd like to include more examples here and I'm also fairly unsure how these things work in practice (see [this TODO](https://github.com/ihmeuw/vivarium_research/pull/1090/files#diff-2ad933f4c1ae75297fe0a6b58aa855bc15980d8fe595c30e34145c8299e8e903R130-R133)) -- hoping to get your help with this.

aflaxman (approved):

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -1,1213 +1,152 @@
...
-  &= \int_{e_1} \int_{e_2} i_0 \cdot \left(\text{RR}_1\right)^{e_1}\cdot \left(\text{RR}_2\right)^{e_2}
-  p_{e_1,e_2} de_1 de_2\\
-  &= \int_{e_1} \int_{e_2} \left(1-\text{PAF}\right) \cdot i_{\text{GBD}} \cdot \left(\text{RR}_1\right)^{e_1}\cdot \left(\text{RR}_2\right)^{e_2}
-  p_{e_1,e_2} de_1 de_2
-  \end{align*}
+The population-attributable fractions (PAFs) must be calculated correctly,
+or the population-level incidence of the outcome will
+not validate to GBD.
+**The usual Vivarium strategy to calculate joint PAFs (PAFs of more than one risk factor on an outcome)
+is biased when the risk factors are correlated.**
```
alibow, line 83: I think it would be worthwhile to state that the "usual Vivarium strategy" is the same one the GBD uses, so GBD overestimates PAFs when risk factors are positively correlated.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -1,1213 +1,152 @@
...
-
-
-Define what we will call a joint PAF:
+  O = O_{tmrel} * RR_{r1} * RR_{r2}
 
 .. math::
+  PAF = \frac{E(O) - E(O_{tmrel})}{E(O)} = 1 - \frac{O_{tmrel}}{E(O_{tmrel} * RR_{r1} * RR_{r2})} = 1 - \frac{1}{E(RR_{r1} * RR_{r2})}
 
+where :math:`O_{tmrel}` is a constant representing the outcome (rate) among those at the
+theoretical minimum risk exposure level.
```
alibow, line 95: TMREL across the joint distribution of r1 and r2 here.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -1,1213 +1,152 @@
...
-.. image:: rate_dotplot.jpg
-
-Proposal to test
------------------------
-We would like to test these two changes for exposures and attributable fractions in the Balanced Energy Protein model as sensitivity analyses supplementing the primary analysis. The risks that we want to test are childhood growth failure: stunting, underweight, and wasting. We are proposing using correlation structures for these exposures from analyses of individual-level data.
-
-Also in the BEP model, we would like to attempt creating correlated changes to propensity scores for stunting, underweight, and wasting within individuals over time. This would involve implementing multiple samples for these risk factors to determine propensity and risk exposure for each simulant. There would not be any changes to the attributable fractions for these risks, the only change would be in allowing these propensity values to change at defined time points, corresponding with GBD age groups.
-
-Illustrative code example
+Risk exposure correlation
```
alibow, line 29: I think we could add a todo note here to include some info about how this can be done for non-continuous risks as well. Some approaches we've taken include:
- Combining two dichotomous risks into one risk with four categories of the joint distribution (which could scale beyond dichotomous)
- Assigning one risk exposure first and using that to inform conditional probability of second risk exposure assignment

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -1,1213 +1,152 @@
...
-  \begin{align*}
-  i_{\text{GBD}} &= \int_{e_1} \int_{e_2} i_{e_1, e_2} p_{e_1,e_2} de_1 de_2\\
-  &= \int_{e_1} \int_{e_2} i_0 \cdot \left(\text{RR}_1\right)^{e_1}\cdot \left(\text{RR}_2\right)^{e_2}
-  p_{e_1,e_2} de_1 de_2\\
-  &= \int_{e_1} \int_{e_2} \left(1-\text{PAF}\right) \cdot i_{\text{GBD}} \cdot \left(\text{RR}_1\right)^{e_1}\cdot \left(\text{RR}_2\right)^{e_2}
-  p_{e_1,e_2} de_1 de_2
-  \end{align*}
+The population-attributable fractions (PAFs) must be calculated correctly,
+or the population-level incidence of the outcome will
+not validate to GBD.
```
alibow, line 81: I think we should at context to what we mean by "correctly" here. Here's my outline:

- If we do not model correlation between risks that are truly correlated and use the multiplicative PAF, we will validate to GBD's outcome rate, but we may overestimate burden attributable to these two risks (although it will match GBD's estimate).
- If we model correlation between risks that are truly correlated and use the multiplicative PAF, we may not validate to GBD's outcome rate
- If we model correlation between risks that are truly correlated and use the joint PAF, we will validate to GBD's outcome rate and will have less bias in our estimate of burden attributable to these two risks (although it will be lower than GBD's estimate of attributable burden)

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -1,1213 +1,152 @@
...
+directly estimate :math:`E(RR_{r1} * RR_{r2})` using our joint distribution between
+:math:`r1` and :math:`r2`.
+This generally needs to be performed for each combination of age, sex, and location.
 
-   i_0 = \left(1-\text{PAF}\right) \cdot i_{\text{GBD}}
+.. todo::
+  Is this usually/always calculated within the simulation, or is it sometimes done on the research side?
+  Is there shared code or components that implement this on the engineering side?
+  If the latter were true, we could simplify this section to "tell engineering to use the correlation-corrected
+  joint PAF calculation" and put the mathematical details elsewhere.
```
alibow, line 133: I am not sure that we have a standard or desired process on this point. I think talking this over with @rmudambi, perhaps using the CVD sim as a case study could be a good opportunity to develop one, but I think we can leave this as a todo note until then.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
alibow: Also, I think it would be helpful to specifically reference the "multiplicative PAF" for the usual vivarium strategy for calculating PAFs when multiple risks affect the same outcome in case the reader doesn't already have that understanding.

alibow (approved):
Approving with comments - let me know if you'd like to talk any of this through in real time!

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
alibow: Here's an example of the latter: https://vivarium-research.readthedocs.io/en/latest/models/risk_exposures/maternal_bmi_hgb/index.html

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
SylLutze: I agree with this ^^ We are going to encounter this shortly with CVD so can decide then and add here.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
SylLutze: I know there was once a proof of this idea, but is there a 1 sentence intuitive "why" for this concept?

SylLutze (approved):

alibow:
@SylLutze I did not do it in one sentence, but I like to think through it this way, so here is my two paragraph explanation :) 

It has to do with how the assumption that risk factors are causally related to outcomes on a multiplicative scale no matter what the magnitude of that outcome is. Say we have RRs that say that R1 is responsible for 20% of outcome O and R2 is also responsible for 20% of outcome O. Now if we remove all of R1 through some intervention or something, we have 20% less O in our population and R2 is still responsible for 20% of that remaining O, although that 20% of O is less _absolute_ risk than it was before because there is less total O. 

So if we think about two individuals, let's say one has R1 and the other has R2. On average between these two simulants, 20% of their O is attributable to R1 and R2 combined. But if we have one simulant with no risk exposure and another with both R1 and R2. The first simulant has zero attributable to R1 and R2 combined. For the second simulant, 20% of their O is attributable to R1, leaving 0.8O, of which 20% is attributable to R2, giving us 0.64O (36% attributable). Now if we average the percent attributable between these simulants we get 18% rather than 20%. So making sure that we account for how many people have both risks in our calculation of the PAF is important because it impacts the magnitude through this mechanism.

alibow:
There are lots of causality assumptions baked in here that may not stand in all situations, but this is the central assumption we rely on. An example of when this assumption does not hold would be when some absolute proportion of risk is _mechanistically_ related to the outcome. For instance, some gene causes X% of anemia through hemoglobinopathies.

SylLutze:
@alibow thanks for the explanation! That is a helpful way to think about it. I think it would be nice @zmbc to include this or similar in the writing as a conceptual idea for users.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
zmbc: @alibow Is it not the case that the TMRED on the joint distribution is always where r1 = TMREL(r1) and r2 = TMREL(r2) for all individuals?

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
zmbc: Here's how I'm thinking about it:

- There is no reason I can think of to ever do the second option you listed -- we always want to treat correlation and joint PAF as a package deal.
- The choice between the first and third options is about the biases that can be introduced by omitting a real-life risk correlation in our model structure. This gets to what we were talking about yesterday (when/why should you include a risk correlation?) and falls under the TODO I put above about this.

That said, I don't think the word "correctly" is very clear here so I will revise.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -1,1213 +1,152 @@
...
+We generally cannot get multivariate distributions from the GBD;
+instead, we seek out auxiliary data sources.
+These may report summary statistics such as a correlation coefficient, or
+have full microdata we can use to summarize the joint distribution in a more detailed way.
+However, no matter how we characterize the relationship, we prefer to continue to validate to the GBD marginal distributions,
+since we trust them more than any single data source.
 
+This plot shows data from a joint distribution reflecting the variance and covariance between height and weight for infants at age one month using
+data from a cohort study.
+The top panel shows the values for the height and weight z-scores and the bottom panel shows the percentiles
```
NathanielBlairStahn, line 46: Technically, the top panel shows histograms of the percentiles/quantiles (i.e. the actual values of WAZ and HAZ), just not labeled as such, and the bottom panel shows histograms of the **percentile ranks**/**quantile ranks**. See these Wikipedia pages:
- https://en.wikipedia.org/wiki/Percentile
- https://en.wikipedia.org/wiki/Quantile
- https://en.wikipedia.org/wiki/Percentile_rank

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -1,1213 +1,152 @@
...
-
-I propose we continue to assume that the relative risks have a “log
-linear” relationship with incidence:
-
-.. math::
-
-
-   i_{e_1, e_2} = i_0 \cdot \left(\text{RR}_1\right)^{e_1}\cdot \left(\text{RR}_2\right)^{e_2},
+Maintaining the correlation is tricky when the risk exposures for a single simulant change over time.
+In general, we have done this by sampling correlated *percentiles* at initialization and then keeping those *percentiles*
```
NathanielBlairStahn, line 55: Again, these are not percentiles but *quantile ranks*.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -1,1213 +1,152 @@
...
+static throughout a simulant's life.
+These percentiles, also known as propensities, are applied to different distributions as the simulant changes age group,
+which means that a simulant's risk exposure in one age group and the next are perfectly auto-correlated.
 
-where we have written :math:`e_1` and :math:`e_2` as “standardized”
-exposures, after scaling and shifting so that the TMREL is 0.0 and a one
-unit increment gives the relative increase in disease burden reported by
-GBD.
+For example, a simulant might have their propensities for height and weight sampled at birth from a
+multivariate uniform distribution with correlation.
```
NathanielBlairStahn, line 61: What you mean here is a [copula](https://en.wikipedia.org/wiki/Copula_(probability_theory)), right?

Generally, "multivariate uniform distribution" means a uniform distribution in the product space (or possibly on a subset of the product space), which always equals the independent coupling of univariate uniforms (unless it's on a subset of the product space, in which case the marginals aren't uniform), so this language might be confusing if people are used to the standard usage.

Perhaps you could say "...height and weight sampled at birth from a bivariate distribution in wich the two components are correlated and are each uniformly distributed (i.e., a bivariate [copula](https://en.wikipedia.org/wiki/Copula_(probability_theory))" or something like that, linking out to the definition of copula. I think it would be good to explicitly mention copulas since that seems to be how statisticians usually talk about coupling two marginal distributions in a non-independent way.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -1,1213 +1,152 @@
...
+In this way, their height and weight can change over time, but the correlation introduced in the initial sampling
+of propensities remains.
 
-With the log-linear assumption, we can calculate the PAF from first
-principles as a double integral (below :math:`p_{e_1,e_2}` denotes the
-joint probability density of exposure level :math:`(e_1,e_2)` in the
-population:
+.. todo::
+  Be more exact about the properties of the propensity correlation approach.
+  I believe that it only exactly maintains the correlation coefficient with some distributional assumptions.
```
NathanielBlairStahn, line 71: Not strictly necessary to elaborate now, but I'm curious about exactly what approach you're referring to and why you think it might not be exact.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -1,1213 +1,152 @@
...
 
-  \begin{align*}
-  i_{\text{GBD}} &= \int_{e_1} \int_{e_2} i_{e_1, e_2} p_{e_1,e_2} de_1 de_2\\
-  &= \int_{e_1} \int_{e_2} i_0 \cdot \left(\text{RR}_1\right)^{e_1}\cdot \left(\text{RR}_2\right)^{e_2}
-  p_{e_1,e_2} de_1 de_2\\
-  &= \int_{e_1} \int_{e_2} \left(1-\text{PAF}\right) \cdot i_{\text{GBD}} \cdot \left(\text{RR}_1\right)^{e_1}\cdot \left(\text{RR}_2\right)^{e_2}
-  p_{e_1,e_2} de_1 de_2
-  \end{align*}
+The population-attributable fractions (PAFs) must be calculated correctly,
+or the population-level incidence of the outcome will
```
NathanielBlairStahn, line 80: Well, that's assuming the risks affect the incidence rate. They could also be affecting the excess mortality rate or perhaps some other property of the outcome.

NathanielBlairStahn: If the risks are affecting an incidence rate, then I would say that the incidence rate itself is the outcome they're affecting (not that they affect the incidence of some other outcome).

For example, if vitamin A deficiency affects the incidence rate of measles, I'd say the affected outcome is "the incidence rate of measles" (not the whole GBD cause Measles). Similarly, if VAD affects the remission rate of measles, I'd say the affected outcome is "the remission rate of measles."

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -1,1213 +1,152 @@
...
+not validate to GBD.
+**The usual Vivarium strategy to calculate joint PAFs (PAFs of more than one risk factor on an outcome)
+is biased when the risk factors are correlated.**
 
-Dividing through and cancelling terms yields a solution:
+The true population attributable fraction for the combined effect of two risk factors,
+:math:`r1` and :math:`r2`, on an outcome :math:`O` such as a mortality rate is (assuming no effect modification, as we typically do):
```
NathanielBlairStahn, line 86: Oh, are you saying "no effect modification" is equivalent to the assumption that

RR(r1, r2) = RR(r1)*RR(r2)

I hadn't thought about it that way before. Do you mean that r1 is not an effect modifier for r2, and vice versa?

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
NathanielBlairStahn: Well, I wouldn't be surprised if that's what GBD actually does, but in general, I don't think there's any reason that this should always be true (I'd have to think about it more to be sure).

NathanielBlairStahn: But for the only pair of risks that GBD actually models jointly, _(birth weight, gestational age) = Low Birth Weight and Short Gestation risk factor_, they first define the TMREL for the joint (bw, ga) distribution, then they have some confusing and conceptually suspicious process of trying to identify separate univariate TMRELs for BW and GA. I think their strategy for identifying the univariate TMRELs changed from GBD 2017 to GBD 2019, and I'm pretty sure that in either case, the univariate TMRELs don't  necessarily coincide with the projections of the joint TMREL onto the two components.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -1,1213 +1,152 @@
...
-
-
-
-Define what we will call a joint PAF:
+  O = O_{tmrel} * RR_{r1} * RR_{r2}
 
 .. math::
+  PAF = \frac{E(O) - E(O_{tmrel})}{E(O)} = 1 - \frac{O_{tmrel}}{E(O_{tmrel} * RR_{r1} * RR_{r2})} = 1 - \frac{1}{E(RR_{r1} * RR_{r2})}
 
+where :math:`O_{tmrel}` is a constant representing the outcome (rate) among those at the
```
NathanielBlairStahn, line 94: Again, if you're thinking of the RR's affecting a rate, then I think the rate *is* the outcome.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
zmbc: I am not sure I follow the LBWSG example, but I'd expect that to be the one _exception_ since that is the one case where they use sophisticated methods to get a joint PAF. In the usual case, they use the multiplicative approach `PAF_joint = 1 - (1 - PAF_1) * (1 - PAF_2)`. This would seem to make no sense at all if the joint TMREL was unrelated to the univariate TMRELs.

NathanielBlairStahn: Oh, but I guess if we're assuming RR(r1,r2) = RR(r1)*RR(r2) and RR's are >=1, then the points where RR(r1,r2) = 1 are precisely the points (r1,r2) where RR(r1)=1 and RR(r2)=1.

So maybe it would be good to explicitly say that the joint TMREL coincides with the Cartesian product of the univariate TMRELs because of our assumption about the form of the joint relative risk.

NathanielBlairStahn: > I am not sure I follow the LBWSG example, but I'd expect that to be the one _exception_ since that is the one case where they use sophisticated methods to get a joint PAF.

The (BW, GA) coordinates are binned into 58 categories (give or take), and then they estimate the RR for each LBWSG exposure category, just like any other categorical risk. In particular, the relative risks do **not** satisfy the assumption RR(BW, GA) = RR(BW)*RR(GA) for the joint risk.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
zmbc: I do not understand what you mean when you say the top panel uses quantiles. It shows Z scores, which are still normally distributed.

I think that "percentile rank" is very commonly referred to as "percentile" (e.g. by the person who made this graph) but the difference makes sense to me and I will add it to the text here, but not mess with the graph.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -1,1213 +1,152 @@
...
 
-And also what we will call a multiplicative PAF:
+By default in Vivarium, a joint PAF is calculated with the "multiplicative" approach:
 
 .. math::
+	PAF_{multiplicative} = 1 - (1 - PAF_{r1}) \cdot (1 - PAF_{r2}) = 1 - \frac{1}{E(RR_{r1})} \cdot \frac{1}{E(RR_{r2})}
 
+When :math:`RR_{r1}` and :math:`RR_{r2}` are independent, :math:`E(RR_{r1}) * E(RR_{r2}) = E(RR_{r1} * RR_{r2})`
+so :math:`PAF = PAF_{multiplicative}`; this is not the case when the risks are correlated.
+The bias gets larger the more correlated the risks are, and the larger the true PAF is.
```
NathanielBlairStahn, line 107: We could link to James's .pdf writeup computing the error as a function of the PAF.
[Why_is_getting_the_paf_right_.pdf](https://github.com/ihmeuw/vivarium_research/files/10273724/Why_is_getting_the_paf_right_.pdf)

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
NathanielBlairStahn: I don't think this section needs to be "simplified" at all, and I like having the mathematical details here. Where else would we put them?

But yeah, once we figure out if there's a standard process or if we create one with the engineers, then we should add that along with the math.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
NathanielBlairStahn: > I know there was once a proof of this idea, but is there a 1 sentence intuitive "why" for this concept?

@SylLutze I see @alibow  added an explanation below, and I think @zmbc had a good intuitive explanation yesterday at the documentation session.

The mathematical proof that correlation induces bias in the PAF is also pretty simple and allows us to easily deduce the direction of bias:

By definition, the relative risks are positively correlated if their covariance is positive, which [means precisely that](https://en.wikipedia.org/wiki/Covariance#Definition) `E[RR(r1) * RR(r2)] > E[RR(r1)] * E[RR(r2)]`. Therefore,
```
PAF_joint = 1 - 1 / E[RR(r1) * RR(r2)]
          > 1 - 1 / ( E[RR(r1)] * E[RR(r2)] )
          = PAF_mult,
```
since `1 - 1/x` is a monotonically increasing function of `x` for `x>0`. That is, the multiplicative PAF (PAF_mult) underestimates the true joint PAF (PAF_joint) when the risks are positively correlated. Similarly, switching the inequality shows that PAF_j < PAF_m precisely when RR(r1) and RR(r2) are are negatively correlated, and PAF_joint = PAF_mult when the relative risks are uncorrelated.

We can translate the mathematical proof into more natural language as follows: Since we assumed that the relative risks of r1 and r2 are multiplicative (i.e., RR(r1,r2) = RR(r1)*RR(r2)), if r1 and r2 are positively correlated, then the average joint relative risk of (r1,r2) is greater than the product of the average relative risks of r1 and r2, which implies that the joint PAF is greater than the multiplicative PAF. The only remaining intuition to fill in for this explanation is why positive correlation of two random variables X and Y is equivalent to the inequality E(XY) > (EX)(EY), if we don't take this as the definition of positive correlation; this seems related to what Ali was getting at in her explanation.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
NathanielBlairStahn: > I do not understand what you mean when you say the top panel uses quantiles. It shows Z scores, which are still normally distributed

Yeah, that's kind of confusing, I was just trying to make a point. What I mean is that if your height (H) is 5'10", and that's the 70th percentile (say) of the distribution of H, then the 70th percentile of H is 5'10", and the percentile rank of 5'10" is 70%. So if you draw a histogram of H, then every value on the x-axis is some percentile of H, because the percentiles of H are simply *values* of H, just labeled with their percentile rank (e.g., the value h=5'10" can be labeled as the 70th percentile). So you're drawing a histogram with "percentiles" on the x-axis, except that we don't call them percentiles in the context of drawing a histogram. (I don't know if that really makes sense, but I don't see any flaws in the logic...)

> I think that "percentile rank" is very commonly referred to as "percentile" (e.g. by the person who made this graph) but the difference makes sense to me and I will add it to the text here, but not mess with the graph.

Yeah, when we started writing the docs, some confusion arose because we were referring to both the propensities and the corresponding exposure values as percentiles, so we consciously tried to make the distinction between quantiles and quantile ranks in order to be able to communicate more clearly with each other.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -1,1213 +1,152 @@
...
-    df['bin_a'] = np.random.binomial(1, 0.3, size=N)
-    df['bin_b'] = np.random.binomial(1, 0.1, size=N)
-
-    rr_1 = 2.5
-    rr_2 = 5
-
-
-
-Define what we will call a joint PAF:
+  O = O_{tmrel} * RR_{r1} * RR_{r2}
```
NathanielBlairStahn, line 89: Just a comment that I'm pretty sure this equation is where the causality and no confounding assumptions are needed when dealing with PAFs and RRs.

Namely, O_{tmrel} is the outcome in the *counterfactual* scenario, while O is the outcome in the *factual* scenario. In order to relate them via this equation, we need the assumption that the the joint relative risk RR(r1,r2) is causal (hence equal to the population RR in the counterfactual scenario, I think) and equals the population relative risk in the factual scenario (which I think is equivalent to no confounding of the factual population's relative risks).

There's nothing to add or change here, I just thought I'd point this out since I've been thinking about it, and see if anyone wants to disagree.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -1,1213 +1,152 @@
...
-And also what we will call a multiplicative PAF:
+By default in Vivarium, a joint PAF is calculated with the "multiplicative" approach:
 
 .. math::
+	PAF_{multiplicative} = 1 - (1 - PAF_{r1}) \cdot (1 - PAF_{r2}) = 1 - \frac{1}{E(RR_{r1})} \cdot \frac{1}{E(RR_{r2})}
 
+When :math:`RR_{r1}` and :math:`RR_{r2}` are independent, :math:`E(RR_{r1}) * E(RR_{r2}) = E(RR_{r1} * RR_{r2})`
+so :math:`PAF = PAF_{multiplicative}`; this is not the case when the risks are correlated.
+The bias gets larger the more correlated the risks are, and the larger the true PAF is.
+When exposures (with RR greater than 1) are positively correlated, the multiplicative approach underestimates the PAF.
```
NathanielBlairStahn, line 108: Maybe add a short explanation of why:

```suggestion
When exposures (with RR greater than 1) are positively correlated, :math:`E(RR_{r1}) * E(RR_{r2}) < E(RR_{r1} * RR_{r2})`, so the multiplicative approach underestimates the PAF.
```

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
SylLutze: Thanks! Yeah, this is helpful for me to understand better. I think similarly it would be helpful to add this to the page for future folks.

NathanielBlairStahn (approved):
I think this looks good! I just had some minor quibbles for clarity's sake.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
alibow: Agreed on your points here, sounds good.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
NathanielBlairStahn: Though I guess maybe it does make sense to say that we keep the same percentiles throughout the simulant's life if we want to be a bit loose with language, e.g.,

"In general, we have done this by sampling correlated *percentile ranks* at initialization and then keeping the simulant's *percentiles* in the exposure distribution static throughout the simulant’s life."

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
alibow: Yeah, I just meant that as written, the TMREL here should be RR_r1=1 and RR_r2=1, and we typically think of TMRELs in terms of a single risk exposure, so I just thought we could clarify for this situation that we mean for both risks.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
zmbc: I think if we had a standard way to do it that was entirely within Vivarium (the researcher did not even need to be aware of the PAF calculation method, because it was automatically applied whenever a correlation was introduced), then the math would belong in the Vivarium documentation.

But, it seems like we are a long way from that. Keeping this as-is.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
zmbc: Yes, you're absolutely right: I was going for a more layman's-terms version, but what I wrote was not remotely correct. I like your idea of briefly explaining what it is and also introducing the word copula.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
zmbc: I am having a hard time finding the right google search for this, but my instinct is that if the marginals are not Gaussian, the correlation coefficient in the copula and the correlation coefficient in the joint distribution might not be equal. In that case, if we use a single copula combined with different marginals in different age groups, the correlation coefficient might not be the same in every age group.

zmbc: I think this is **very** far down the list in terms of important questions about risk correlation :)

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
zmbc: Yes, this is also how I am thinking about it. I will clean up this phrasing.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
zmbc: Yes -- in my brain that is the definition of "no effect modification" though I admit I am not too familiar with how people usually use the term.

zmbc: Which is part of why I have always been a bit confused by "X modifies the effect of Y" constructions -- it should always be reciprocal, I think.

alibow: Yeah, another way to say it would be there is a statistical interaction between the effect of R1 and R2 on O. While they are the synonymous, I think effect modification is used more often when there's a focus on a specific exposure/outcome pair and interaction is used more often in other contexts. We've tended to use the term effect modification, but interaction might be more intuitive because were rarely focusing on a single exposure/outcome like you would in some epi investigation of causality.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
NathanielBlairStahn: And maybe we should ask @collijk for the source LaTeX as well...

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
zmbc: I've left out your first bullet point for now, because the rest of this page wouldn't really apply in that case (from a Vivarium perspective, it would be a single risk).

zmbc:
@aflaxman @alibow @NathanielBlairStahn @SylLutze I have made pretty significant updates to this PR, especially including a section to try to explain the intuition behind the bias we've been talking about, based on Ali's explanation above. I also discovered that the content of this page is very closely related to [existing documentation that I hadn't looked at in a while](https://vivarium-research.readthedocs.io/en/latest/model_design/vivarium_model_components/risk_factors/population_attributable_fraction/index.html#bias-in-paf-calculation), so I tried to integrate that.

My sense now is that I've made the page too long so any suggestions of how to make it more succinct would be appreciated.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -33,83 +36,185 @@ Instead of sampling from a univariate distribution for each risk exposure at ini
...
+
+By default, Vivarium uses GBD-calculated PAFs for each risk-outcome pair and combines them using
+a multiplicative approach (described below).
+This is correct when risks are all independent, but it has two problems in the presence of correlated risks:
+
+* The GBD approach to calculating PAFs is biased in the presence of any confounding.
+  If we are choosing to model correlated risk exposures with effects on the same outcome,
+  that means we believe there is **at least** one confounding variable in each PAF calculation.
+  If risks (with RR > 1) are positively correlated, the GBD approach will underestimate the PAF of each;
+  this issue is explained in more detail :ref:`on the PAFs page <pafs>`.
```
alibow, line 98: This is maybe nit-picky, but it could be helpful to note that this bias is present even if the risk effects are adjusted for confounding. My first reaction when I was reading this was to say, but GBD already adjusts for that confounding so it shouldn't matter! Until I followed the link and remembered the more nuanced point.

alibow:
> @aflaxman @alibow @NathanielBlairStahn @SylLutze I have made pretty significant updates to this PR, especially including a section to try to explain the intuition behind the bias we've been talking about, based on Ali's explanation above. I also discovered that the content of this page is very closely related to [existing documentation that I hadn't looked at in a while](https://vivarium-research.readthedocs.io/en/latest/model_design/vivarium_model_components/risk_factors/population_attributable_fraction/index.html#bias-in-paf-calculation), so I tried to integrate that.
> 
> My sense now is that I've made the page too long so any suggestions of how to make it more succinct would be appreciated.

I think the changes look great and I think the page is well organized, so I don't think we need to cut significant length!

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
```diff
@@ -33,83 +36,185 @@ Instead of sampling from a univariate distribution for each risk exposure at ini
...
-^^^^^^^^^^^^^^^^
+where :math:`O_{tmrel}` is a constant representing the outcome (e.g. incidence rate) among those at the
+theoretical minimum risk exposure level for both risk factors.
 
-By default in Vivarium, a joint PAF is calculated with the "multiplicative" approach:
+In the default configuration of Vivarium, using GBD PAFs with the multiplicative approach:
 
 .. math::
-	PAF_{multiplicative} = 1 - (1 - PAF_{r1}) \cdot (1 - PAF_{r2}) = 1 - \frac{1}{E(RR_{r1})} \cdot \frac{1}{E(RR_{r2})}
+	PAF_{vivarium} = 1 - (1 - PAF_{r1}) * (1 - PAF_{r2}) = 1 - \frac{1}{E(RR_{r1})} * \frac{1}{E(RR_{r2})}
```
alibow, line 203: Just noting that we can only assume that E(O) = O_tmrel * E(RR_r1 * RR_r2) because we've stated that there are no other confounders between r1 and r2 and O, but in most realistic situations this would not be the case because of the confounding bias. I am not sure how to approach this mathematically in a simple way without running involved nanosim tests, but I expect that this could complicate the conclusion that the underestimation is always greater than the overestimation - perhaps this would be worth discussing in a doc session or at least noting in the docs.

SylLutze:
> @aflaxman @alibow @NathanielBlairStahn @SylLutze I have made pretty significant updates to this PR, especially including a section to try to explain the intuition behind the bias we've been talking about, based on Ali's explanation above. I also discovered that the content of this page is very closely related to [existing documentation that I hadn't looked at in a while](https://vivarium-research.readthedocs.io/en/latest/model_design/vivarium_model_components/risk_factors/population_attributable_fraction/index.html#bias-in-paf-calculation), so I tried to integrate that.
> 
> My sense now is that I've made the page too long so any suggestions of how to make it more succinct would be appreciated.

I actually don't think this is too long and I was able to read and follow the logic, so would keep as is! It is in good shape!

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/index.rst
zmbc: I think we discussed this in the docs session, but just to leave a paper trail:

This is far from our only problem in the (extremely common) case where we leave known confounders out of a model because there are too many to simulate. In that case, the conclusions we make about the impact of intervening on a risk factor are all biased by not capturing the non-causal portion of the correlation between risk factor and outcome.

Long-term, the solution to this may be something along the lines of simulating a correlated risk "X-factor". If that is the solution, we would be moving more toward *making* the assumptions of this section true, rather than handling when they are not.

zmbc:
I'm merging this, because it is already quite large and has been open for a while. I know in particular @NathanielBlairStahn had some additional ideas for how to make this more explicit and easier to follow; let's plan to discuss these and follow up with more PRs as necessary.

PR #1092, opened 2022-12-14
Requiring high ASCVD risk to move up LDL-C med ladder

SylLutze:
Adding another qualification to moving up LDL-C treatments - high ASCVD. Hopefully this will help fix the issue of adults getting prescribed very high treatments very quickly.


alibow (approved):

aflaxman (approved):
Is the only change an addition of an ASCVD Calculated box in the right-hand path?

SylLutze:
> Is the only change an addition of an ASCVD Calculated box in the right-hand path?

Yes!

PR #1099, opened 2022-12-19
Adding duplicates

SylLutze:


aflaxman (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3214,6 +3214,69 @@ In general, the same logic as is outlined for the census can be applied to all h
...
+guardian living at a different address will be calculated from the sim. 
+We will use 10% as a placeholder value. The user can then pick a rate of 
+duplication between 0 and 10%.  
+
+Based on this rate, the simulant will be duplicated, but with their address 
+changed to the guardian’s address instead of their “correct” address. 
+
+.. todo:: 
+
+    Find the maximum noise level from the simulation. If it is above ~25% or has a significant amount of variation over time or across ages, reconsider how to define the noise parameter.
```
zmbc, line 3256: Could we just have it be a conditional probability? i.e. the probability of being duplicated _given_ that you have a guardian in a different household. That would be naturally bounded between 0 and 100%.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3214,6 +3214,69 @@ In general, the same logic as is outlined for the census can be applied to all h
...
+    Find the maximum noise level from the simulation. If it is above ~25% or has a significant amount of variation over time or across ages, reconsider how to define the noise parameter.  
+
+For college GQ simulants aged less than 24, all are assigned to a guardian who 
+by definition lives at a different address. This means that theoretically 
+the maximum noise level is 100% for this group, however, that would add 
+significantly to the dataset and so it not allowed here. The default rate 
+will be set to 5%, with a minimum of 0% and a maximum of 25%.  
+
+To create duplicates, the college student will be included in the final 
+dataset twice, once at their college GQ and once at their guardian's home.
```
zmbc, line 3265: Does this duplication happen before or after column noise? I assume before.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3214,6 +3214,69 @@ In general, the same logic as is outlined for the census can be applied to all h
...
 For the two surveys currently outlined in the model, the default rates are 1.45% for ACS, and 29.05% for CPS.  
 
+**Duplicates:** 
+
+Duplication is most relevant for the census, which has a significant 
+amount of known duplicates. Often, the first PRL task is to match within 
+the same dataset in order to remove duplicates.  
+
+As a simplifying assumption in our initial model, we will be including 
+duplicates in the census only and limiting it to guardian-based duplication.
```
zmbc, line 3224: I'm curious about the decision not to include simple duplication with a certain rate across the whole population. It does not seem like it would be much more difficult to include here. It might be a bit too easy to de-duplicate, especially because address does not have "incorrect choose" column noise, but with the current docs anyone >24 years old would be guaranteed not to be a duplicate.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3214,6 +3214,63 @@ In general, the same logic as is outlined for the census can be applied to all h
...
+
+To facilitate this type of error, we have simulants assigned to guardians 
+within the simulation. Sometimes, those guardians may move and live at 
+different addresses than their dependents. In this case, there is an 
+opportunity for duplication. Since this mechanism occurs within the 
+simulation, there is a natural maximum that we will impose in the 
+noise function.  
+
+Guardian based duplication is further divided here into two types: simulants 
+younger than 18 (<18), and those at college GQ less than 24 (<24).
```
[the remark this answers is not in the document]
SylLutze: True, I added that the less than 18 are less than 18 and NOT in college GQ to make them separate

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3214,6 +3214,69 @@ In general, the same logic as is outlined for the census can be applied to all h
...
+opportunity for duplication. Since this mechanism occurs within the 
+simulation, there is a natural maximum that we will impose in the 
+noise function.  
+
+Guardian based duplication is further divided here into two types: simulants 
+younger than 18 (<18), and those at college GQ less than 24 (<24).  
+
+For simulants younger than 18, an average percent who have a 
+guardian living at a different address will be calculated from the sim. 
+We will use 10% as a placeholder value. The user can then pick a rate of
```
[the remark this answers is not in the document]
SylLutze: Yep! I changed the language here to remove that for clarity

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: I did consider that, and it would certainly be easier. I felt that it would be too difficult for an end user to understand and lead to the obvious question of "well how many people are duplicated at 100%?". 

If finding an average proves too difficult, I think we will default to the conditional probability anyways. 

Do you think this is overly complex as is?

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: I think before, but was planning to make a separate PR defining how all the noises work together since its not obvious. For example, should we omit or duplicate first? We need to decide and tell the engineers

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Yes, that is definitely a limitation currently. I felt that having the most basic duplication was very easy for PRL. So then we discussed having noise parameters for duplicates only indicating how much noise they should have. But Abie felt this was complex and not needed in the simple model. 

So we ended with just having guardian based duplicates. If we want to revisit this with Abie, I am happy to add other duplication later.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Hm I could see somebody thinking about it either way, actually.

One idea: we could have the pseudopeople package dynamically calculate this maximum, and maybe make it only a warning if the user tries to set a value above the maximum.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Makes sense. I agree it doesn't matter much either way since it will be quite easy to deduplicate those without cranking up the column noise on duplicates.

PR #1102, opened 2022-12-19
Risk exposure page doc cleanup

alibow:


docs/source/model_design/vivarium_model_components/risk_factors/risk_exposure/index.rst
```diff
@@ -61,6 +58,35 @@ Risk exposure distributions can be:
...
+
+  * - Distribution
+    - Example
+    - Possible exposure values
+  * - Dichotomous
+    - Zinc deficiency
+    - * Not deficient
+      * Deficient
+  * - Unordered polytomous
+    - Unsafe water source
```
aflaxman, line 74: This exposure is partially ordered---it has two dimensions so some levels are comparable, but others are not.  Not sure if that is too much information to pack in here, but it could potentially reduce confusion to make a note of it.

aflaxman (approved):

docs/source/model_design/vivarium_model_components/risk_factors/risk_exposure/index.rst
alibow: Thanks for noting this. I changed it to tobacco use (none/chewing/cigarettes/cigars) which I think is a better fit. I also added a note about risk exposures being mutually exclusive and collectively exhaustive because it made me think of that.

aflaxman:
Nice

SylLutze (approved):

PR #1110, opened 2023-01-10
Creating CVD risk correlation page

SylLutze:


aflaxman (approved):

docs/source/models/risk_correlation/sbp_ldlc_fpg_bmi/index.rst
```diff
@@ -0,0 +1,89 @@
...
+multivariate, a cdf function is used to convert the values into a uniform distirbution 
+between 1 and 0 with the same correlation between variables. 
+
+The calculations are done `in this notebook <https://github.com/ihmeuw/vivarium_research_nih_us_cvd/blob/main/Correlation_Testing.ipynb>`_. This code can be used to generate propensity 
+values for any number of simulants. 
+
+Causation
++++++++++
+
+We are not currently modeling a direct causal relationship between changes in
```
alibow, line 74: Except for via mediation, right? Maybe we should just note that this is outside of the mediation schema.

docs/source/models/risk_correlation/sbp_ldlc_fpg_bmi/index.rst
```diff
@@ -0,0 +1,89 @@
...
+We are not currently modeling a direct causal relationship between changes in 
+any risk exposure values. 
+
+Assumptions and Limitations
+++++++++++++++++++++++++++++++
+
+#. We assign a correlation at initialization that does not change over time. For SBP and BMI which we model to change over time, this means that as simulants age, they will retain their original correlation and increasly differ from the true correlation. 
+
+#. To calculate the correlation, we limited the population to only untreated individuals. However, as our simulants receive treatments their risk exposure values will change and likely make validation more challenging. 
+
```
alibow, line 83: Have we examined how the treated correlation coefficients differ from the untreated? I'm mostly just curious, but we could use this to inform what to expect here and/or even use that data directly if needed

docs/source/models/risk_correlation/sbp_ldlc_fpg_bmi/index.rst
```diff
@@ -0,0 +1,89 @@
...
+- :ref:`SBP risk exposure <2019_risk_sbp>`
+
+- :ref:`LDL-C risk exposure <2019_risk_exposure_ldl>`
+
+- :ref:`FPG risk exposure <2019_risk_exposure_fpg>`
+
+- :ref:`BMI risk exposure <2019_risk_bmi>`
+
+Vivarium Modeling Strategy
+----------------------------
```
alibow, line 33: Maybe this is coming after, but I think we might need some more detail on _how_ we intend to enforce these correlations. Mainly...

- If we enforce correlation between risk A and B as well as B and C, then there's no guarantee that the resulting correlation between A and C will match what we got out of NHANES. We might need to make a ranked priority between risk pairs to get ahead of this (hopefully it all matches up given that we're using the same data source for everything, but we'll see!).
- I assume that the correlation should be applied to the *untreated* exposure values for these risks and then treatment effects that are present at initialization will be layered on top. Is that right? We should make that clear if so!

alibow (approved):

docs/source/models/risk_correlation/sbp_ldlc_fpg_bmi/index.rst
SylLutze: Okay! Yeah I got it from another page. I will delete it now.

docs/source/models/risk_correlation/sbp_ldlc_fpg_bmi/index.rst
SylLutze: Addressing your first bullet point - maybe I am not understanding something, but I thought we were enforcing A to B, B to C AND A to C. The multivariate normal function has correlations for SBP to BMI and BMI to LDL-C but also SBP to LDL-C. Code is below - is this not doing what I think it is doing? 

`probit_propensity = np.random.multivariate_normal([0,0,0,0], [[1, sbp_ldl, sbp_bmi, sbp_fpg],
                                                         [sbp_ldl, 1, bmi_ldl, ldl_fpg],
                                                         [sbp_bmi, bmi_ldl, 1, bmi_fpg],
                                                             [sbp_fpg, ldl_fpg, bmi_fpg, 1]], size=N)`

docs/source/models/risk_correlation/sbp_ldlc_fpg_bmi/index.rst
SylLutze: I looked again at the untreated values. They are not very different - in general there is actually a bit higher correlation and fewer very low outliers in the older age groups. The outliers is probably since so many people are on treatment by the time they are 80+ that we don't get great results on untreated only. 

However, the bigger issue you pointed out is that we actually assign _treated_ exposure values to the population from GBD. Then afterwards, we pick people to be on treatment and create artificial _untreated_ exposures, and then reduce them with treatment to their original value obtained through GBD. [More information](https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_us_cvd/concept_model.html#creation-of-untreated-sbp-values-on-initializaiton) 

Therefore, I actually think we should use the full NHANES population in finding the correlation coefficients. Conveniently, this is also how we decide how many folks are on treatment, so by definition it will match the model. Does that make sense to you? If so, I can update the values but I don't expect things to change dramatically.

docs/source/models/risk_correlation/sbp_ldlc_fpg_bmi/index.rst
alibow: Oh yeah sorry, just took a look and this is a smarter function than I thought. Yeah disregard my initial point then!

docs/source/models/risk_correlation/sbp_ldlc_fpg_bmi/index.rst
alibow: Yeah, that makes sense to me! Good to know it won't make too much of a difference, but yeah seems like it would be best to update then

SylLutze: Thanks!!

PR #1111, opened 2023-01-11
Excluding BMI standard deviation outliers

SylLutze:


aflaxman (approved):

alibow (approved):
Saving the custom values to a csv file for the engineers to use as input data has worked well for me on other projects if there's interest in doing it that way, but otherwise looks good!

SylLutze:
> Saving the custom values to a csv file for the engineers to use as input data has worked well for me on other projects if there's interest in doing it that way, but otherwise looks good!

Good idea! I added a csv, that will probably be much easier

PR #1113, opened 2023-01-11
Risk correlation bias quantification

alibow:


aflaxman (approved):
It might be helpful to include more text in markdown cells in the notebook to help the busy reader focus on the key points.

SylLutze:
For the first 2 graphs in scenario 1, it would help to force the same y-axis so they are more visually comparable.

zmbc (approved):

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/bias_quantification.ipynb
```diff
@@ -0,0 +1,1387 @@
+{
```
zmbc, line 1: <p>Typo: *associated</p><p> </p><p>I think it would be clearer to remove the outcome bubble from this diagram, as it is not included in the below functions.</p><p> </p><p>Let's add a note here: in this situation, we normally do not have to worry about PAFs, because our baseline intervention coverage is 0. When baseline intervention coverage is not zero, a "joint PAF" of R1 and R2 on the intervention would need to account for correlation. This <strong>cannot</strong> be done using the method described on the risk-risk correlation page, because there is an interaction effect/effect modification between R1 and R2. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

zmbc, line 1: <p>Can we include an example of a single call to this function, with interpretation of the output? While the plots below are very cool, I think future researchers are more likely to want to call the function with a specific set of parameters.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

zmbc, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    val = 0.05
</pre><p>This section at the beginning can be replaced with vals = np.linspace(0.05, 1.0, num=20)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

zmbc, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    val = 0.05
</pre><p>As above, let's use np.linspace.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

zmbc, line 1: <pre class="ql-syntax" spellcheck="false">Line #11.        #plt.scatter(exp, relative_bias, color='tab:orange')
</pre><p>Let's delete these commented lines; same goes for the next cell.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

zmbc, line 1: <p>Are there no examples here because it turned out to be equivalent to the above? If so, can we note in the text introducing the section what the equivalence is between dichotomous and continuous risks?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

zmbc, line 1: <p>A few notes on this diagram:</p><p> </p><ul><li>I don't find the pre- and post-intervention bubbles helpful in thinking through this. In general, causal diagrams already imply the passage of time (a causal arrow must take place over time).</li><li>As above, I don't think we need O1 and O2 in this diagram, especially since they are not included in the functions below.</li><li>I don't think post-intervention R1 adds anything here, since we are focused here on the correlations between <em>two</em> risks.</li></ul><p>The isolated form of this diagram, I think, is R1 -&gt; intervention -&gt; R2 with a correlation between R1 and R2. In that situation the correlation between R1 and R2 matters to the result (as long is it is downstream of R2).</p><p> </p><p><strong>And</strong>, let's add a note that in this situation, if there is non-zero baseline intervention coverage, determining the rho to initialize with in order to validate to the real-life correlation is hard. I believe it needs to be done using something like our methods for mediation.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

zmbc, line 1: <p>As above, I think this can just be wasting -&gt; SQ-LNS -&gt; stunting with a correlation between wasting and stunting.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

zmbc, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.    Calculates absolute and relative error in the change in pre and post intervention
</pre><p>This isn't necessarily a helpful comment, but this sentence is pretty long; maybe could be broken up?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

zmbc, line 1: <pre class="ql-syntax" spellcheck="false">Line #4.    R1 and R2 are dichotomour risk factors with the
</pre><p>*dichotomous</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

zmbc, line 1: <pre class="ql-syntax" spellcheck="false">Line #23.    def scenario_3_dichotomous_risks(r1_exp, r2_exp, rho, int_r2_rr, N):
</pre><p>*scenario_2</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

zmbc, line 1: <p>Great! These are exactly the kinds of examples I had in mind for Scenario 1.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

zmbc, line 1: <pre class="ql-syntax" spellcheck="false">Line #3.        result = scenario_3_dichotomous_risks(r1_exp=val, 
</pre><p>Let's do <strong>absolute_bias, relative_bias = </strong>as you did in scenario 1 to make this clearer.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

zmbc, line 1: <p>Cool! I thought this would be the case but didn't get very far on the math.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

zmbc, line 1: <p>As I mentioned in the last documentation session, I think the intervention bubble in this diagram is misleading because you could conclude that you are fine if the intervention does not affect <strong>both</strong> risks! I'd vote to remove the bubble altogether, with the implied understanding that R1 or R2 should be impacted by the intervention (directly or indirectly).</p><p> </p><p>I'm not sure I agree with the text here. It seems to imply that if not for the difference between PAF bias and intervention impact bias, we'd just accept the same limitation as GBD. But I don't think it's always true that we want to accept all of GBD's limitations: GBD is restricted in what it can do by making estimates for all years/locations/causes, while we try to make a higher-fidelity model with a much smaller scope.</p><p> </p><p>I think this section should call out that when this type of correlation <em>is</em> included in a model, the correlation-aware joint PAF calculation should be used, as described on the risk-risk correlation page.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

zmbc, line 1: <pre class="ql-syntax" spellcheck="false">Line #66.        return paf_bias, O_bias
</pre><p>Is PAF bias ever of primary interest to us? I can't think of when it would be. I'd favor switching this to the absolute and relative biases as in the other scenarios.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

zmbc, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    for val in [-2, -1.5, -1, -0.5, 0, 0.5, 1, 1.5, 2]:
</pre><p><strong>np.linspace(-2, 2, num=9)</strong></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

zmbc, line 1: <pre class="ql-syntax" spellcheck="false">Line #8.    def calculate_paf_and_outcome_bias_dichotomous_risks(r1_exp, r1_rr, r2_exp, r2_rr, rho,
</pre><p>Can we start with dichotomous, as in the other scenarios?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

zmbc, line 1: <p>Let's stick to one example here, or add interpretations of these results.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1113/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='32'/>

zmbc:
@alibow Thanks for all your work on this!

Since this is the first time we are using ReviewNB, I'll just call out that you can see my comments above inline with the rendered notebook by opening this PR in ReviewNB and going to the "Changes" tab.

docs/source/model_design/vivarium_model_components/risk_factors/risk_correlation/bias_quantification.ipynb
alibow: <p>So I agree with your comment above that we should simplify the generic diagram to R1 -&gt; intervention -&gt; R2 with a correlation between R1 and R2. I'm thinking it might be worthwhile to include these sorts of applied examples that might appear different than that exact structure of the generic diagram but do contain that same underlying structure (in addition to potentially distracting details) to demonstrate that it might not always to be obvious in the context of an actual project. I'm thinking I'll include <em>both </em>versions here with some commentary about this. Does that sound alright to you?</p>

zmbc: Ah, good idea! I think that sounds great if you clearly mark which section of the messy, real-world example has the pattern. For example, having everything but putting the part I suggested in bold or something.

alibow: <p>Hmm, I was thinking it was worthwhile to include the PAF bias simply to show that it is often different than the cases averted bias to encourage not to stopping bias assessment just with the PAF. Do you agree?</p><p> </p><p>In terms of absolute/relative bias, I had a hard time thinking through the best way to present this here. I'm currently presenting absolute bias (difference in cases averted due to intervention) normalized to the estimate of cases averted while considering correlation. </p><p> </p><p>I've been thinking of the relative bias as the difference in the population outcome rate in the intervention scenario as a proportion of the outcome rate without the intervention, which doesn't really seem like something that we would typically measure.</p><p> </p><p>Either way, I think I should add some description to what is actually being measured here. Do you have thoughts on whether we should be evaluating the relative scale I described - or do you have alternative ideas on how to estimate bias on the relative scale that I am missing?</p>

zmbc: > to encourage not to stopping bias assessment just with the PAF

I see your point, but I think this could just be a note (which I believe you already have) not to only look at PAF bias. I also think a similar note could be added to the main risk/risk correlation page, or maybe it should just be revised to show how the outcome is biased and not only the PAF. When I wrote it I was thinking more binary (is this biased or not?) which _is_ the same between PAF and outcome, and not about magnitude, which is _not_ the same.

> I've been thinking of the relative bias as the difference in the population outcome rate in the intervention scenario as a proportion of the outcome rate without the intervention, **which doesn't really seem like something that we would typically measure.**

I feel like I have seen plots of e.g. % of SAM/MAM cases averted -- is that not the same thing? But if it's really true that we don't care about relative bias, then maybe we don't need to show that either! The less things we plot, the better IMO -- it makes our story clearer.

PR #1114, opened 2023-01-12
Updates from wasting paper model 3.1

alibow:


aflaxman (approved):
I bet Laura will be interested in the switch to age-specific effects you documented here.  You could include a slide about this with the before-and-after plots from your nanosim in our next meeting.

SylLutze (approved):

alibow:
> I bet Laura will be interested in the switch to age-specific effects you documented here. You could include a slide about this with the before-and-after plots from your nanosim in our next meeting.

Great, will do!

hussain-jafari (approved):

PR #1115, opened 2023-01-13
Additions to noise functions

SylLutze:


docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3118,9 +3119,33 @@ error rate is not very intuitive for the end user.
 At some point in the future, we might make this more user friendly. However, for the 
 sake of a minimum functional model, this is satisfactory. 
 
-One function has an additional parameter. The typographic function needs an input for 
-the probability that a corrupted token contains the original token. For simplicity, 
-we will use 0.1 and the user will not be able to edit this value. 
+A few functions have additional parameters that allow the user more control. These 
+inputs will not be specified in the main settings for pseudopeople listed above. We 
+imagine that a yaml file (or similar file type) will be available for users to download
```
aflaxman, line 3124: I was imagining the interface would be through python and R functions in a module called pseudopeople:
```
import psuedopeople as pp
pp.simulate_decennial_data(first_nickname_pr=.2)
```

I guess this might get unwieldly with the number of options we are heading towards, so I'm not set on it.  But I wanted to call it out, in contrast to the specification-file approach you are proposing.

aflaxman (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: We could do something like this! This is similar to what I imagined above for the "basic" parameters. I have a hunch the list of optional/additional parameters might expand over time and having them in a separate file might make it cleaner. But we can definitely adjust this once we have a better picture of what pseudopeople will look like.

zmbc: What do we gain by separating the parameters users can pass into "basic" and "additional" parameters? It seems like it would be simpler to me to have a single set of parameters which can be specified via config file, Python dictionary/R list, or directly as function arguments as in Abie's example.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3118,9 +3119,33 @@ error rate is not very intuitive for the end user.
...
+imagine that a yaml file (or similar file type) will be available for users to download 
+and adjust the additional inputs if desired. The table below outlines these inputs, 
+their purpose, and the default value for the yaml file. 
+
+.. list-table:: Additional Inputs and Default Values
+  :widths: 20 20 20 20
+  :header-rows: 0
+
+  * - Noise Function Affected
+    - Observer Data Affected
```
zmbc, line 3133: Are these inputs being set for _each_ column they are affecting, or is there one value for all columns affected by that noise function?

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3118,9 +3119,33 @@ error rate is not very intuitive for the end user.
...
+
+  * - Noise Function Affected
+    - Observer Data Affected 
+    - Additional Inputs 
+    - Default Input Value 
+  * - Typographic Noise
+    - First name, middle initial, last name, street name, city name, employer name 
+    - Probability that a corrupted token contains the original token
+    - 0.1 
+  * - Numeric miswriting
```
zmbc, line 3140: This implies that numeric miswriting works significantly differently between age, ZIP, and other numbers. Why not just separate these into different noise functions? IIUC, that would make the second column of this table unnecessary.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3242,21 +3263,32 @@ would be listed in MM/DD/YYYY format as 08/12/2022).
...
+have noise added. For those selected, randomly select a new option. This is chosen 
+from the list of options in `this csv <https://github.com/ihmeuw/vivarium_research_prl/blob/main/src/vivarium_research_prl/noise/incorrect_select_options.csv>`_. Note that for relationship to head of household, this includes the full list of options, not just those seen in the household. 
 
 Limitations: 
 
 - For single person homes, incorrectly selecting relationship to head of household does not make as much sense. However, we continue with it here anyways. 
 - Incorrect selection likely takes place in a logical way, and might persist across observers (e.g., trans or nonbinary people "incorrectly" selecting a sex; confusion with different race/ethnicity groups; selecting a state from a prior address) however, we are not including this complexity. 
+- The current version of the function does not enforce that the new selection is different than the original selection. This means that some rows designated for incorrect selection will not actually be incorrect. This is especially important for sex as there are only 2 options to select.
```
zmbc, line 3273: What does this refer to? The version in the vivarium_research_prl repo? Having this as a limitation is a bit confusing, if we plan to implement it differently for the real package.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: There are a lot of parameters right now, and being realistic I expect the list to expand as these functions are further built out (software used, blocking requirements, confidence level needed, etc.). Just with the current "basic" options, I counted 58 parameters for the census. 

I thought there would be value in starting simpler and having some "basic" options for a user who is new or has a more simplistic need. Then "additional" parameters for users with more specific needs and who is not intimidated by the number of inputs! It also sounds to me like we should have a conversation about the user interface soon.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Currently, parameters are at the column specific level. So you can specify for typographic errors that the rate for first names is higher than for street name.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: Yeah, I can split this into multiple functions. I think the second column is still helpful for clarity even if I agree its technically redundant now.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
NathanielBlairStahn: As @zmbc says, I was imagining a single configuration file that includes all parameters -- there would be no distinction between "basic" and "additional" in the full parameter specification. I was imagining the config file would get read into some in-memory data structure like a Python dictionary, which would then get passed to a top level function that adds noise to the dataset (so if desired, the user could instead specify parameters directly in a dictionary instead of in a file).

The reason we were talking about "basic" vs. "additional" is because, for the sake of simplicity, we had all previously agreed to have the user specify exactly two parameters for each column (the "overall fraction of rows with noise," and the "amount of noise within each corrupted row," or something along those lines -- except that the second of these doesn't make sense for some types of noise such as incorrect selection), and have any additional parameters be implementation details.

But there are a **lot** of parameters, and the list will probably grow as we improve the noise functions. I think it could be useful for users to be able to tweak the different parameters to see how variations in noise affect the linking, and as @SylLutze says, we probably (I think?) want to have a simplified set of parameters (e.g., the two "basic" ones we agreed to previously) with reasonable defaults for everything, so as not to overwhelm a new user. 

As for @aflaxman's suggestion, it seems to me like having a function with 58 or 100 parameters would be rather unwieldy, which is why I suggested a configuration file (or dictionary). But I guess if we *really* want to have all the parameters as arguments in some top-level function, I guess we could... although that would be harder to maintain if we ever change the underlying noise functions.

zmbc: > it seems to me like having a function with 58 or 100 parameters would be rather unwieldy

We could just have the function accept `**kwargs` and treat that similar to the config file/data structure. Only downside is the lack of nesting but we could figure out an approach for that too (e.g. `{observer}_{column}_{param}` or something).

> we probably (I think?) want to have a simplified set of parameters (e.g., the two "basic" ones we agreed to previously) with reasonable defaults for everything

My vote would be to have reasonable defaults for **everything** -- the user can run pseudopeople without specifying anything regarding noise. Then, all those parameters can be overridden. Separating into "basic" and "additional" seems like more of a documentation question to me, rather than a distinction we need to make in code.

aflaxman: > My vote would be to have reasonable defaults for **everything**

👍 

Regarding how we go beyond the defaults, I find this discussion very helpful, but I'm not sure what will work best.

NathanielBlairStahn: For reference, [here](https://github.com/ihmeuw/vivarium_research_prl/blob/main/src/vivarium_research_prl/find_kids/noisify_data.py) is the code where I'm actually adding noise to the census and WIC data, which is essentially a hard-coded version of what I imagined a config file would look like.

And [here](https://github.com/ihmeuw/vivarium_research_prl/blob/main/src/vivarium_research_prl/noise/noisify.py) are some functions I started to write to parse the input parameters and add noise in the correct way (the first of these gets called repeatedly in the module above; the second isn't used yet).

If we went with Zeb's suggestion of passing `**kwargs`, my current approach would require additional nesting, e.g., `{observer}_{column}_{noise_function}_{param}`, and you would have to simultaneously pass `row_fraction=0.01`, for example. But maybe we could come up with a way to simplify that? Maybe `obeserver`, `column`, and `row_fraction` should all just be separate required arguments because those are needed for everything, and then the noise is specified by `{noise_function}_{param}`. Or `{noise_function}` and `{param}` could even be split into two arguments. Anyway...

I agree that we should come up with reasonable defaults for everything, and that it's not necessary to make a distinction between "basic" and "additional" for the code, so we don't need to specify that for the engineers.

As for what we should tell the engineers to implement, it seems to me like the configuration file / nested dictionary approach is the easiest and most flexible way to start, and if we decide we want to put a nicer wrapper on the user interface such as being able to pass keyword arguments, that's something we could add on later. What do you think, @aflaxman and @zmbc?

zmbc: > it seems to me like the configuration file / nested dictionary approach is the easiest and most flexible way to start

Agree, I think we can add Abie's version later pretty easily.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2947,180 +2947,224 @@ for all column based noise include:
...
 - Token noise level is a noise parameter that only applies to certian noise types and defines the amount of errors expected once a string is selected for noise. This parameter is also elaborated on below. 
-- The amount of each type of noise is individually configurable for each column in each observer. This means that the end user can can specify, for example, that in the census, sex has 2% incorrect selections. The minimum noise is 0% and the maximum if 5% for all column noise. For errors per row, the minimum is 0 and maximum is 5. 
+- A few noise types have additional parameters which can be specified by the user separately. This is elaborated on in the section on notes about inputs to the functions. 
+- The amount of each type of noise is individually configurable for each column in each observer. This means that the end user can can specify, for example, that in the census, first names have a 2% error rate for typographic noise. The minimum noise is 0% and the maximum if 5% for all column noise. For errors per row, the minimum is 0 and maximum is 5.
```
zmbc, line 2951: I realize this wasn't introduced in this PR, but at the end of this line where it says "For errors per row, the minimum is 0 and maximum is 5" is that a holdover from a previous version that had errors per row instead of token noise level?

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2947,180 +2947,224 @@ for all column based noise include:
...
+The exact method for how a user will input the parameters has not been finalized. We 
+anticipate a nested parameter dictionary with the levels as: :code:`{observer}_{column}_{noise_function}_{param}`. 
 
-One limitation of this strategy is that since there is a random probability of a 
-particular character/token receiving noise, not all rows that are selected to be 
-eligble for noise will actually receive noise. Another limitation is that token 
-error rate is not very intuitive for the end user. 
+One limitation of having a row probability and then token probability is that since 
+there is a random probability of a particular character/token receiving noise, not 
+all rows that are selected to be eligble for noise will actually receive noise.
```
zmbc, line 3142:
```suggestion
all rows that are selected to be eligible for noise will actually receive noise. 
```

zmbc (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: yeah, good catch!

PR #1116, opened 2023-01-13
Data formatting specifications

SylLutze:


aflaxman (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2371,7 +2379,7 @@ in the home.
   * - Middle initial 
   * - Last name
   * - Age 
-  * - DOB (stored as a string in MM/DD/YYYY format)
+  * - DOB (stored as a string in MMDDYYYY format)
```
zmbc, line 2382: I have no preference, but what is the reason behind this change?

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2025,6 +2025,11 @@ Census
   * - Tracked Guardian Address(es) (for noise functions ONLY)
   * - Type of GQ (for noise functions ONLY)
 
+.. note::
+
+  Please save age as an integer and all other data as strings. This will make the data compatible with existing noise functions.
```
zmbc, line 2030: It is not entirely clear to me what is meant by "save" -- I am guessing based on the second sentence you are referring to the output of post-processing, before noise is applied? But it's not clear to me that we care about data types at this step.

If you're talking about data types in the final output of pseudopeople, that does seem like something we care about -- but in that case, I don't know what you mean by this second sentence.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: We are trying to match the data to the actual data source, when possible. So some googling of data structure lead to this change so we can match real life.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: This is meant to be at the end of post-processing it should have these data types. Based on a conversation with Nathaniel, he thought these data types would be best for inputs to the noise functions. I know that noise functions are not finalized, but I thought there was value in defining data based on what we know today of the noise functions.

zmbc: I guess this will get SE review anyway, but I expect this may be too prescriptive. It sounds like what we are saying here is "if the post-processing outputs these data types, engineering will be able to more easily re-use the noise functions that Nathaniel has written." Which is probably a good thing, but if they want to change these data types and implement their noise functions slightly differently, that's up to them and doesn't really affect us. So, this is less of a requirement and more of an implementation pointer.

What this _doesn't_ specify, then, is the final pseudopeople output data types (except for dates, where it is specified that they are strings). That _does_ seem like a requirement that should be in this doc (whether in this PR or not). Or is it already there and I'm missing it?

albrja: @SylLutze Since this hasn't been merged yet, WIC observer DOB will be MMDDYYYY format, Social Security observer will be YYYYMMDD format, and all other observers will still be MMDDYYYY or MM/DD/YYYY?

SylLutze: Yep! WIC and SSA are the two we found very specific formats for, so those were updated. The others can remain as they were.

SylLutze: Yep! WIC and SSA are the two we found very specific formats for, so those were updated. The others can remain as they were.

albrja (approved):

PR #1118, opened 2023-01-14
Use a single state and PUMA per GQ type

zmbc:
I probably should have run this by folks on the research side before starting on this PR, but I didn't think to do that. Hopefully nobody strongly objects; if someone does, we will need to think through the best path forward.

This change comes out of a conversation with @rmudambi and @stevebachmeier.

We already model no household "structure" for GQ and we have no concept of e.g. different nursing homes in different places. Even before this change, all simulants in a single GQ type would have a single **address** (physical and mailing). However, previously we allowed state and PUMA to be different between different individuals in the same GQ type. This was on its face contradictory, was pretty hard to reason about, and led to some weird edge cases (#1068). It also violates constraints that simplify things on the engineering side, including the implementation of address_id.

This PR changes that, so that each GQ type has a single state and PUMA, which is chosen randomly.

I have also cleaned up the perturbation section a bit, since that was edited here anyway, to address the confusion between "PUMA" and "state and PUMA," which has come up a few times.


zmbc:
@SylLutze one outstanding question: in the docs it says about assigning guardians for college GQ that

> 78.5% will be assigned to a guardian within their state.

With college GQ being in a single state, this would be pretty weird. I am wondering how important you think this is? Would you rather keep it or drop it if all college students were in the same state? Hopefully this isn't a dealbreaker for this PR overall 😬

SylLutze:
> @SylLutze one outstanding question: in the docs it says about assigning guardians for college GQ that
> 
> > 78.5% will be assigned to a guardian within their state.
> 
> With college GQ being in a single state, this would be pretty weird. I am wondering how important you think this is? Would you rather keep it or drop it if all college students were in the same state? Hopefully this isn't a dealbreaker for this PR overall 😬

I think it is fine to remove this requirement. I wanted to place guardians close to their dependents, but if we are okay with having a single giant "college" for the whole country, then I think it is fine to spread out guardians more as well! Could you please add this to the PR? Just deleting the requirement and having guardians pull from anywhere in the country? 

My generally, I think this is an acceptable simplification. I just want to check that deviating from the "true" population in a state/PUMA won't be an issue? For example, if a couple of GQs got places in Wyoming, it could change the population size in that state quite considerably. I think that is fine, but just something to keep in mind.

zmbc:
> More generally, I think this is an acceptable simplification. I just want to check that deviating from the "true" population in a state/PUMA won't be an issue? For example, if a couple of GQs got places in Wyoming, it could change the population size in that state quite considerably. I think that is fine, but just something to keep in mind.

Yes, this is a good point. I think we don't care _too_ much about getting the state populations exactly right. The GQ population as a whole is 3% of the general population (at the beginning of the simulation, anyway) -- big enough to cause some large misses for a few states as you mentioned, but not big enough to substantially change the overall population distribution.

aflaxman (approved):

PR #1119, opened 2023-01-18
Creating HF PAFs for unmediated and uncorrelated runs

SylLutze:


docs/source/models/risk_effects/bmi/index.rst
```diff
@@ -324,6 +324,24 @@ incidence(i) = incidence*(1-PAF\ :sub:`r370`\)*RR^{max((BMI_i - TMREL),0)}
 The relative risk for heart failure is per 1-unit increase in BMI. 
 **Please note that this is different than the other relative risks.** 
 
+The PAF for heart failure for the unmediated and uncorrelated runs will be 
+calculated at initialization. To do this, first the simulant level relative risk 
+will be found based on their BMI exposure using the below formula: 
+
+Simulant_RR(i) = RR^{max((BMI_i - TMREL),0)}
```
aflaxman, line 331: maybe typeset as math so it is prettier?

docs/source/models/risk_effects/bmi/index.rst
```diff
@@ -324,6 +324,24 @@ incidence(i) = incidence*(1-PAF\ :sub:`r370`\)*RR^{max((BMI_i - TMREL),0)}
...
+
+Then, the individual PAF can be found using this formula: 
+
+PAF(i) = (Simulant_RR(i) - 1) / Simulant_RR(i)
```
aflaxman, line 335: maybe typeset as math so it is prettier? or as code?

alibow (changes requested):
Hmm, we will get different answers if we take the mean of the (RR-1)/RR using simulant-specific RRs rather than use population mean RR for a single (RR-1)/RR equation. I think we should take the latter approach, described below.

1) assign simulant-specific relative risks as already instructed
2) calculate the population mean relative risk across all simulants in a demographic group
3) calculate the population attributable fraction for that demographic group as (mean_RR - 1) / mean_RR, where mean_RR is the average population relative risk from step 2

This is the approach we typically take, but I am not sure how to think through why this is preferable. @aflaxman @NathanielBlairStahn @zmbc, do you have any thoughts?

aflaxman:
> This is the approach we typically take, but I am not sure how to think through why this is preferable. @aflaxman @NathanielBlairStahn @zmbc, do you have any thoughts?

I think Ali is right about this, although now I am confused.  Can someone walk me through the implications of doing it each way?

alibow:
I just ran a small test calculation in excel, and it looks like we do not accurately replicate the population outcome rate if we use a PAF calculated from the average of individual attributable fractions rather than using the population mean relative risk to calculate the PAF.

I'm not sure exactly how to reason this conceptually or mathematically, but I am now more certain we should use the approach outlined in the steps I commented above.

[PAF calculation test.xlsx](https://github.com/ihmeuw/vivarium_research/files/10449839/PAF.calculation.test.xlsx)

SylLutze:
Thanks for the math check! I thought that the averages would work if the risks were uncorrelated, I didn't realize that it would still be an issue. I updated this to change to average the RRs first and THEN take the PAFs, both in the python notebook and in the docs.

PR #1121, opened 2023-01-24
Switch from ssh to https in contributing instructions

NathanielBlairStahn:
Edit instructions in `README.rst` to suggest cloning the repo using `https` rather than `ssh`, as that's GitHub's recommended method and doesn't require the user to set up an SSH key.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,7 +23,7 @@ install this repository.  As always, you should do this in a clean environment.
 
    $> conda create -y --name=vivarium_research python=3.8
    $> conda activate vivarium_research
-   (vivarium_research) $> git clone git@github.com:ihmeuw/vivarium_research.git
+   (vivarium_research) $> git clone https://github.com/ihmeuw/vivarium_research.git
    (vivarium_research) $> cd vivarium_research
    (vivarium_research) $> pip install -r requirements.txt
 


aflaxman:
Is this really what GitHub recommends now?

zmbc:
I think GitHub has gone back and forth at least 5 times on whether or not HTTPS is the preferred method; right now their documentation doesn't state a preference. But I do think the majority of us use HTTPS and it is easier to set up in my opinion (creating a PAT in the GitHub interface vs. creating an SSH key on the command line).

aflaxman:
> the majority of us use HTTPS and it is easier to set up in my opinion (creating a PAT in the GitHub interface vs. creating an SSH key on the command line).

Ok, as long as this is what you are finding easier these days, I have no problem with changing the docs.

aflaxman (approved):

zmbc (approved):

NathanielBlairStahn:
> > the majority of us use HTTPS and it is easier to set up in my opinion (creating a PAT in the GitHub interface vs. creating an SSH key on the command line).
> 
> Ok, as long as this is what you are finding easier these days, I have no problem with changing the docs.

I agree -- I find the GitHub Personal Access Token easier than SSH as well. I can easily add the token to macOS Keychain on my laptop, and on the cluster I think I set up GitHub CLI so that I never have to do anything except update the token when it expires.

@pletale FYI for when you're cloning the repo, and @SylLutze FYI when you're updating the documentation on how to get started with all the computer stuff. We should probably add documentation and links about how to set up a GitHub personal access token and set up GitHub CLI to make it easy to access GitHub. Let me or @zmbc know if you have any questions.

PR #1123, opened 2023-01-30
Clean up last name, include race/ethnicity logic

zmbc:
This section was primarily documented with sample code, much of which was outdated/misleading; I have replaced that with prose and explained how the race/ethnicity for sampling is determined.

### DIFF
--- a/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
@@ -1610,42 +1610,17 @@ Note: RAND used something like this for their BISG project
 https://www.rand.org/pubs/external_publications/EP20090611.html
 https://www.rand.org/health-care/tools-methods/bisg.html
 
-.. sourcecode:: python
-
-    # last name can be race/ethnicity specific
-    df_census_names = pd.read_csv('/home/j/Project/simulation_science/prl/data/Names_2010Census.csv', na_values=['(S)'])
-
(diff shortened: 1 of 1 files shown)


aflaxman (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1610,57 +1610,44 @@ Note: RAND used something like this for their BISG project
...
 
-For now, assign all simulants within the same household who are 
-relatives of the reference person the same last name. 
+For now, assign all simulants initialized into the same household who are 
+relatives of the reference person (including the reference person themselves) the same last name.
 This excludes "roommate/housemate" and "other nonrelative" to the 
-reference person. This is an oversimplification as some relatives might have 
-different last names, but works for the initial model. 
+reference person.
+Simulants initialized into group quarters will not be assigned matching last names.
```
SylLutze, line 1638: We don't have to do this - but for college students assigned guardians, we could match their last names to their guardian's. I imagine that would be relatively straightforward to do if we wanted to include it.

SylLutze (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: I think we won't do this for this model version, but I'll add a note that we might consider it in the future.

PR #1124, opened 2023-01-31
Clean up PRL introduction, fertility docs

zmbc:
- Update intro to reflect current state of the model
- Remove component numbering which was only used by the old intro
- Make the TOC depth-limited so it is not huge
- Revise the fertility component documentation

There is no (intentional) change to the simulation here!


aflaxman (approved):
Thanks for the cleanup!

SylLutze (approved):

PR #1128, opened 2023-02-06
Clarify noise order, add OCR and typographic noise

SylLutze:
- Also changing data types to be all strings 
- and clarifying typographic noise parameters

### DIFF
--- a/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
@@ -1990,7 +1990,7 @@ Census
 
 .. note::
 
-  In the final version of the observers, following the noise functions, please have all data as strings. Age must be rounded to the nearest whole number.
+  In the final version of the observers, following the noise functions, please have all data as strings. Age must be rounded down to a whole number before applying noise.
 
 
 
@@ -2166,7 +2166,7 @@ There are two types of sampling plans:
 
 .. note::
 
-  In the final version of the observers, following the noise functions, please have all data as strings. Age must be rounded to the nearest whole number.
+  In the final version of the observers, following the noise functions, please have all data as strings. Age must be rounded down to a whole number before applying noise.
 
 
 Here is an example: 
@@ -2362,7 +2362,7 @@ Here is an example:
 
 .. note::
 
-  In the final version of the observers, following the noise functions, please have all data as strings. Age must be rounded to the nearest whole number.
+  In the final version of the observers, following the noise functions, please have all data as strings. Age must be rounded down to a whole number before applying noise.
 
 
 **Who to Sample** 
@@ -2518,7 +2518,7 @@ W2 and 1099 Forms
 
 .. note::
 
-  In the final version of the observers, following the noise functions, please have all data as strings. Age and wages must be rounded to the nearest whole number.
+  In the final version of the observers, following the noise functions, please have all data as strings. Age must be rounded down to a whole number before applying noise. Wages must be rounded to the nearest whole number before applying noise.
 
 
 If a simulant does not have a social security number but is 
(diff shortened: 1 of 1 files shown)


docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2906,22 +2911,26 @@ noise added to the data. We currently divide noise into two types:
...
+#. Copy from Within Household 
+#. Month and day swaps (applies to dates only)
+#. Zip Code Miswriting (applies to Zip Code only)
+#. Age Miswriting (applies to age only)
+#. Nicknames 
+#. Fake names
+#. Phonetic 
+#. OCR 
+#. Typographic 
+#. Numeric miswriting
```
NathanielBlairStahn, line 2933: It seems like we would want numeric miswriting to happen before OCR or typographic noise, e.g. around the same stage that the zipcode miswriting or age miswriting happens. @SylLutze, is there a reason you put it last?

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1990,7 +1990,8 @@ Census
 
 .. note::
 
-  In the final version of the observers, following the noise functions, please have age as an integer and all other data as strings. 
+  In the final version of the observers, following the noise functions, please have all data as strings.
```
aflaxman, line 1993: Is this for the observers, or for the output of pseudopeople?  Maybe it is the same thing.

aflaxman (approved):
We should probably specify somewhere that age is "nominal", meaning an integer rounded down from the precise floating point number, and income is rounded to the nearest integer (or nearest penny, as you prefer).  It's not really noise, but the precision you have added about the numeric noise made me think of it.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: I just assumed it didn't matter the order if it was also a type of noise applied to strings. I can move it to where age and zip are. Those are higher since I thought they required a different non-string datatype and therefore should go before OCR/typographic. I can just numeric there are well.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: This is the output of pseudopeople. I think we are considering all of the in between datatypes as engineering decisions not specified by research team.

SylLutze:
> We should probably specify somewhere that age is "nominal", meaning an integer rounded down from the precise floating point number, and income is rounded to the nearest integer (or nearest penny, as you prefer). It's not really noise, but the precision you have added about the numeric noise made me think of it.

Sure, we can include this somewhere

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -1990,7 +1990,8 @@ Census
 
 .. note::
 
-  In the final version of the observers, following the noise functions, please have age as an integer and all other data as strings. 
+  In the final version of the observers, following the noise functions, please have all data as strings. Age must be rounded to the nearest whole number.
```
zmbc, line 1993:
```suggestion
  In the final version of the observers, following the noise functions, please have all data as strings. Age must be rounded down to a whole number before applying noise.
```

Alternatively, it might be clearer to add this note directly to the table row for age.

zmbc (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
NathanielBlairStahn: Maybe it's not a big deal, but just thinking about the data generation process in terms of how the data might actually get corrupted, it seems to me like OCR and typographic should always come last.

SylLutze: Sure, makes sense. I swapped the order.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
SylLutze: This clarification helps, thanks!

aflaxman:
But don’t ever round age up!  For engineers to be clear, you could say that “age is floored to the nomial age as an integer”

zmbc:
@aflaxman This is what I intended with my suggestion "Age must be rounded **down** to a whole number..." Do you think that is still unclear?

aflaxman:
Oops, no totally clear, I was just reading too fast.

PR #1129, opened 2023-02-06
Adding pop size and some adjustments to PAF calculations

SylLutze:
@alibow following our conversation on Friday, I would appreciate if you could give this a sanity check and look at the workbook as well to make sure this looks alright! 
https://github.com/ihmeuw/vivarium_research_nih_us_cvd/blob/main/PAF_BMI_to_HF.ipynb 

Also tagging this as SE review required since the engineers will need to complete this calculation for us so I want to ensure they have everything they need.


docs/source/models/risk_effects/bmi/index.rst
```diff
@@ -324,19 +324,27 @@ incidence(i) = incidence*(1-PAF\ :sub:`r370`\)*RR^{max((BMI_i - TMREL),0)}
...
 
-The RR can then be averaged for each age/sex group. Finally, the PAF for 
-each age/sex group can be found as: 
+To do this, follow the below steps: 
 
-:math:`PAF(i) = (RR\text{mean}(i) - 1) / RR\text{mean}(i)`
+#. Initialize a population of 100,000 
+#. Truncate the exposure of BMI at 40.8* 
+#. Find the simulant level RR with this equation: :math:`RR\text{simulant} = RR^{max((BMI_i - TMREL),0)}` 
+#. Find the median* RR for each age/sex group
```
aflaxman, line 338:
```suggestion
#. Find the median** RR for each age/sex group 
```

docs/source/models/risk_effects/bmi/index.rst
```diff
@@ -324,19 +324,27 @@ incidence(i) = incidence*(1-PAF\ :sub:`r370`\)*RR^{max((BMI_i - TMREL),0)}
...
+#. Truncate the exposure of BMI at 40.8* 
+#. Find the simulant level RR with this equation: :math:`RR\text{simulant} = RR^{max((BMI_i - TMREL),0)}` 
+#. Find the median* RR for each age/sex group 
+#. Find the PAF for each age/sex group with this equation: :math:`PAF(i) = (RR\text{median}(i) - 1) / RR\text{median}(i)`
 
 An example of this calculation can be found in the `workbook here <https://github.com/ihmeuw/vivarium_research_nih_us_cvd/blob/main/PAF_BMI_to_HF.ipynb>`_
 
+Notes: 
+
+- We truncate the exposures of BMI as this calculation is based on literature values that have limited applicability in our model. 40.8 is 3 standard deviations above the mean BMI exposure for obese individuals in the paper being used. [Kenchaiah_2008]_ Without this truncation, there would be RR's that are 2000+ which makes mean PAF values very close to 1. We do not want to assume a continued relationship in BMI to RR for values 40 BMI units above the max used in the paper.
```
aflaxman, line 345:
```suggestion
- (*) We truncate the exposures of BMI as this calculation is based on literature values that have limited applicability in our model. 40.8 is 3 standard deviations above the mean BMI exposure for obese individuals in the paper being used. [Kenchaiah_2008]_ Without this truncation, there would be RR's that are 2000+ which makes mean PAF values very close to 1. We do not want to assume a continued relationship in BMI to RR for values 40 BMI units above the max used in the paper. 
```

docs/source/models/risk_effects/bmi/index.rst
```diff
@@ -324,19 +324,27 @@ incidence(i) = incidence*(1-PAF\ :sub:`r370`\)*RR^{max((BMI_i - TMREL),0)}
...
+#. Find the simulant level RR with this equation: :math:`RR\text{simulant} = RR^{max((BMI_i - TMREL),0)}` 
+#. Find the median* RR for each age/sex group 
+#. Find the PAF for each age/sex group with this equation: :math:`PAF(i) = (RR\text{median}(i) - 1) / RR\text{median}(i)`
 
 An example of this calculation can be found in the `workbook here <https://github.com/ihmeuw/vivarium_research_nih_us_cvd/blob/main/PAF_BMI_to_HF.ipynb>`_
 
+Notes: 
+
+- We truncate the exposures of BMI as this calculation is based on literature values that have limited applicability in our model. 40.8 is 3 standard deviations above the mean BMI exposure for obese individuals in the paper being used. [Kenchaiah_2008]_ Without this truncation, there would be RR's that are 2000+ which makes mean PAF values very close to 1. We do not want to assume a continued relationship in BMI to RR for values 40 BMI units above the max used in the paper. 
+- As we are having trouble with high outlier values in BMI, we are choosing to use a median here instead of the mean. If we improve the quality of the BMI data in the future, we can reassess this decision.
```
aflaxman, line 346:
```suggestion
- (**) As we are having trouble with high outlier values in BMI, we are choosing to use a median here instead of the mean. If we improve the quality of the BMI data in the future, we can reassess this decision. 
```

aflaxman (approved):
I suggested a way to make the notes easier to follow, but you could also use some other symbols if you think that looks nicer that double-asterisks.

docs/source/models/risk_effects/bmi/index.rst
SylLutze: I do like this more! Thanks

alibow:
Do you think we need *both* truncate *and* use the median? I was initially thinking that we would just truncate and then use the mean as usual. Are we still seeing surprising results after truncating that make it seem like the median would be preferable?

alibow:
Ok, thanks for clarifying. It sounds like this is reasonable, but it does seem rather subjective especially if we're not using the median for the other BMI PAFs so I just want to make sure I'm understanding. This is because we are taking the categorical RR from the heart failure paper and extrapolating a continuous relative risk, right? Or am I misremembering? 

If so, might we want to just use the categorical values instead?

SylLutze:
I agree it does feel subjective which is a concern I have. The literature value we are using is continuous. SBP -> heart failure is categorical and we use the categorical data to find the PAFs instead. 

For BMI -> heart failure, I think the issue is related to the issues we have been having with BMI more generally. There is an overexposure of very high BMIs in the population, leading to a relatively large set of outliers in simulant level RRs and therefore PAFs. I am still hoping that switching to a better BMI data source would solve the issue and we are investigating that swap currently. But in the mean time (ha, puns) this seemed like a reasonable solution to mitigate the issue. Do you think we should just use means still anyways?

alibow (approved):

SylLutze:
That makes sense! I will change back to the mean and merge it in. I do expect with will also change in the future

PR #1139, opened 2023-02-15
Adding transitions definitions to data defs table

pletale:
Adds definitions and specific uses of severity splits and progression in Data Definitions table.


SylLutze (approved):

zmbc (approved):
Am I understanding right that these are intended to be minimal summaries, and we plan to fill in more detail in the linked sections below?

pletale:
> Am I understanding right that these are intended to be minimal summaries, and we plan to fill in more detail in the linked sections below?

Yes that was my understanding, Zeb!

PR #1140, opened 2023-02-16
Adding mediation to FPG through LDL

SylLutze:


aflaxman (approved):

alibow (approved):
Do we also need documentation of the delta between FPG on LDL as a risk effect?

SylLutze:
> Do we also need documentation of the delta between FPG on LDL as a risk effect?

So currently in the model we don't include risks having effects on other risks directly anywhere. We could add this for the lifestyle intervention (your drop in FPG from the intervention triggers a drop in LDL-C). I looked at the deltas and they aren't huge but probably are impactful. 

I am not convinced that added a new mechanism to the model for this is worth it, and if we are concerned about the missed impact, I would prefer to add LDL-C data directly as an effect of the lifestyle intervention, but interested to hear your thoughts.

aflaxman:
> > Do we also need documentation of the delta between FPG on LDL as a risk effect?
> 
> So currently in the model we don't include risks having effects on other risks directly anywhere. We could add this for the lifestyle intervention (your drop in FPG from the intervention triggers a drop in LDL-C). I looked at the deltas and they aren't huge but probably are impactful.

If we add, let's add directly --- risks changing risks is a whole additional level of complication, and I would like to do it soon, but not here.

SylLutze:
Sounds good! I will merge this in then and further investigate a lifestyle -> FPG direct impact outside of this PR

PR #1149, opened 2023-03-13
Cause morbidity impacts

pletale:
Writing up documentation about morbidity impacts:
- Describes what disability weights and YLDs are and how the GBD calculates them.
- Explains the difference between incidence- and prevalence-based YLDs


docs/source/model_design/vivarium_model_components/causes/index.rst
```diff
@@ -827,5 +827,43 @@ Morbidity Impacts
 Disability Weights
 ^^^^^^^^^^^^^^^^^^
 
+Disability weights (DWs) represent the magnitude of health loss associated with specific health states. The weights are measured on a scale from 0 (full health) to 1 (complete loss of health; equivalent to death). GBD assigns DWs based upon household and internet surveys for which respondents are presented with paired comparison questions for different health problems. These paired comparison questions include lay descriptions of health states and ask the respondent to choose which has the greatest functional consequences and symptoms associated with the given health state. For more information on the 2010 DW Measurement study in which these surveys were carried out, see this `recording of the GBD Science Seminar from 2/7/2023 <https://hub.ihme.washington.edu/display/GBD2020/GBD+Science+Seminar+series?preview=/87186031/192089713/GBD%20Science%20Seminar_Disability%20weights%20(DWs)_Feb%202023.pdf>`_ or see Section 4.8.1 of the GBD 2019 Methods Appendix.
```
zmbc, line 830: This is a great concise description!

docs/source/model_design/vivarium_model_components/causes/index.rst
```diff
@@ -827,5 +827,43 @@ Morbidity Impacts
...
+
+
+Residual YLDs
+^^^^^^^^^^^^^
+
+For less common diseases and their sequelae, GBD may not currently estimate disease prevalence and YLDs, and have thus been included in residual categories. For these residual categories, we estimate YLDs by multiplying the residual YLL estimates by the ratio of YLDs to YLLs from the estimates Level 3 causes in the same disease category that were explicitly modelled. This scaling is done for each country-sex-year.
```
zmbc, line 843: "we" -> GBD

Also, this is super interesting! Did not know this.

docs/source/model_design/vivarium_model_components/causes/index.rst
```diff
@@ -827,5 +827,43 @@ Morbidity Impacts
...
+
+Residual YLDs
+^^^^^^^^^^^^^
+
+For less common diseases and their sequelae, GBD may not currently estimate disease prevalence and YLDs, and have thus been included in residual categories. For these residual categories, we estimate YLDs by multiplying the residual YLL estimates by the ratio of YLDs to YLLs from the estimates Level 3 causes in the same disease category that were explicitly modelled. This scaling is done for each country-sex-year. 
+
+Incidence- vs. Prevalence-Based YLDs
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+*Incidence-based YLDs* provide a measure of the disease burden experienced by an individual over the course of their lifetime. Incidence-based YLDs are calculated by adding together YLD 1 (i.e., the product of the DW associated with disease 1 and the time between onset and survival) and YLD 2 (i.e., the product of the DW associated with disease 2 and the time between onset and survival).
```
zmbc, line 848: I'm not sure I follow the example here (YLD 1 and YLD 2).

The key difference between the two, in my understanding, is that incidence-based YLDs immediately count all of the future disability caused by a disease at the time of incidence, while prevalence-based YLDs only count the disability at the time it is experienced. For example, an incident case of an incurable disease would only accrue one year of YLDs in the first year it is experienced if using prevalence-based YLDs, while it would accrue many years of YLDs if using incidence-based YLDs (as many years as are left in the person's life expectancy, because we know they cannot be cured).

zmbc (approved):

docs/source/model_design/vivarium_model_components/causes/index.rst
pletale: Yes, this aligns with my understanding as well! I can rephrase this part and take out the example of YLD 1 and 2. I wanted to give an example of how incidence-based YLDs can be the sum of all future disability caused by multiple diseases (rather than just representing the disability caused by a single disease) but I don't think I made that clear

docs/source/model_design/vivarium_model_components/causes/index.rst
pletale: Oh yeah I'll make the change from 'we' to GBD, thanks for pointing that out!

docs/source/model_design/vivarium_model_components/causes/index.rst
```diff
@@ -827,5 +827,43 @@ Morbidity Impacts
...
+
+In order to compute **years lived with disability (YLDs)** for a particular health outcome in a given population, the number of people living with that outcome is mulitplied by the DW for the health state associated with that sequela. Ultimately, YLDs are used to indicate burden of disease: DALYs are calculated as the sum of YLLs and YLDs. The DALY-based estimation of the burden of disease is important because it simultaneously considers the reduced health state due to disability before death and the decline in life expectancy due to death. It thus moves away from conventional measurements of the burden of disease that use readily available data on mortality, prevalence, and incidence (`Kim et al., 2022 <https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8841194/>`_).
```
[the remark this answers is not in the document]
pletale: I think this all is intended to describe GBD methods more than Vivarium, but maybe that's not what we want anyways - happy to change!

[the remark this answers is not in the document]
zmbc: Oh okay, ignore my suggestion in that case. I think we don't know exactly how this works on the GBD side with respect to time: how is the total disease duration/person time calculated? Maybe we should do some research to figure out how they do it. But I think it is fine how it is for this PR.

aflaxman (approved):

docs/source/model_design/vivarium_model_components/causes/index.rst
SylLutze: I like the example, but maybe separate it from the other parts of the text more? Clearly define the incidence vs prevalence YLDs first and then say if you have multiple diseases, you would calculate it by XYZ

docs/source/model_design/vivarium_model_components/causes/index.rst
```diff
@@ -827,5 +827,43 @@ Morbidity Impacts
...
+
+.. list-table:: Incidence vs. Prevalence-based YLDs (`Kim et al., 2022 <https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8841194/>`_)
+   :widths: 30 30 30 
+   :header-rows: 1
+
+   * - Measure
+     - Advantages
+     - Disadvantages
+   * - Incidence-based YLDs
+     - YLDs and YLLs are measured consistently; Provide a more valid data source for policies or interventions focused on disease prevention.
```
SylLutze, line 861: With incidence based, do we risk overestimating this? Since we assume you will lose all of the possible YLDs without considering other possible comorbidity impacts. For example, if a very comorbid person had a heart attack, would we still say you get 10 YLDs from this event, regardless of if they also have diabetes, stroke, etc etc which would decrease the YLDs in practice. 

This might already by considered in the value for YLDs we assign on that time step though.

SylLutze (approved):

docs/source/model_design/vivarium_model_components/causes/index.rst
zmbc: I think this is why one of the advantages of prevalence-based YLDs below is "Incorporation of comorbidities is easier."

If using incidence-based, we could at least consider the chronic/incurable comorbidities already present, but there would be no easy way to adjust the YLDs to account for e.g. a risk factor that is not causing any disability right now but means they are more likely to have more comorbidities later in life.

PR #1152, opened 2023-03-22
Remove depth from PRL concept model

NathanielBlairStahn:
It was very annoying trying to find specific sections without having them listed and linked in the contents, so I removed the depth limit in order to display every subsection at the top of the file.


aflaxman (approved):

zmbc (approved):
Oops sorry it ended up being confusing, [I did this](https://github.com/ihmeuw/vivarium_research/pull/1124/files#diff-15207000bd94de788f2a25f7a11177d70c14251b64748b463357d9646c7a7e2eR61) thinking it would help! There is probably no way to make a document this long _easy_ to navigate but yep, let's add back all the TOC detail if it makes things easier 😃

PR #1155, opened 2023-03-24
Describe numeric miswriting

zmbc:


docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3272,15 +3272,16 @@ Limitations:
...
 
-**Numeric Miswriting** 
+**Numeric Miswriting**
 
-To implement this, select the strings eligible for noise and apply 
-the numeric miswriting noise function to all strings with the user defined 
-character error rate. 
+To implement this, select the strings eligible for noise,
+and then select, for each each digit character (0-9) in each eligible string, whether to change that character, according to the per-character error probability.
+The digits that are changed should be changed to a *different* digit character, selected uniformly at random.
```
NathanielBlairStahn, line 3279: Requiring a different character seems like the same problem I told the engineers not to worry about yesterday: requiring a new selection with incorrect select. But maybe since we have a standard pre-defined set of choices [0,1,...,9], it will be easier to implement?

NathanielBlairStahn (approved):

aflaxman (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Yes, good point. I'll add a note that this is a nice-to-have.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3272,15 +3272,20 @@ Limitations:
...
 
-**Numeric Miswriting** 
+**Numeric Miswriting**
 
-To implement this, select the strings eligible for noise and apply 
-the numeric miswriting noise function to all strings with the user defined 
-character error rate. 
+To implement this, select the strings eligible for noise,
+and then select, for each each digit character (0-9) in each eligible string, whether to change that character, according to the per-character error probability.
+The digits that are changed should be replaced with a digit character selected uniformly at random.
```
stevebachmeier, line 3279: Can the digit be replaced by the same digit, eg 9 -> 9?

@rmudambi @albrja Is this what made modifying vectorized_choice difficult?

stevebachmeier: Oh wait - one of the limitations below implies that, yes, a digit can be replaced by itself. Please confirm.

NathanielBlairStahn: I had the same thought after yesterday's conversation. However, the OCR, typographic, and phonetic noise functions work in a similar manner to what's being described here -- every token selected for noise will be replaced by a _different_ token. In those functions, that's accomplished using a lookup table for the possible values of the noisy token. If you used the same approach here, you could have a table / dictionary that looks something like this:
```
0: 1 2 3 4 5 6 7 8 9
1: 0 2 3 4 5 6 7 8 9
2: 0 1 3 4 5 6 7 8 9
...
9: 0 1 2 3 4 5 6 7 8
```
That is, for each digit, you just look up what are the legitimate values to replace it with.

NathanielBlairStahn: I'm still not clear why you need _vectorized_ choice for things like this rather than just something like [numpy.random.Generator.choice](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.choice.html), which is already vectorized over the `size` parameter.

NathanielBlairStahn: Although, I haven't carefully thought through how to vectorize the OCR, phonetic, and typographic noise functions, which seems like it could be challenging...

NathanielBlairStahn: > Oh wait - one of the limitations below implies that, yes, a digit can be replaced by itself. Please confirm.

@stevebachmeier Yes, in the current version, a digit is allowed to be replaced with itself. The original version specified that the digit must be different, but I think Zeb changed it after I pointed out that requirement seems to have the same difficulty we were discussing for incorrect select.

SylLutze (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: @stevebachmeier Basically, I think we don't understand how much of a lift it will be for engineering to avoid same-digit replacement, so we want to leave it up to you. Hopefully my [latest commit](https://github.com/ihmeuw/vivarium_research/pull/1155/commits/5e8a3da23b4c4e313102109e520c603a4d886860) clarifies things.

aflaxman: Engineers are wondering if perhaps SE Review is *not* required for this PR, @zmbc can you check in with @rmudambi to see?

PR #1164, opened 2023-03-31
Revise simulant and household ID columns

zmbc:
There are several changes here:
* Add a "ground truth" household ID to all observers.
* Add observer-specific ID columns (e.g. WIC family ID, household survey household ID) where they would be truly collected. For example, the WIC program will assign each participant a static WIC participant ID, so they will know who is the same person from one year to the next. Household surveys will not assign any permanent identifier, but they will at least know who is in the same household _at the time of the survey_.
* Revise WIC to require living in the same household. This is necessary in order to implement WIC family ID in a reasonable way.

**@aflaxman are any of these must-haves before CODS? If anything, I would think adding a household ID column to the decennial Census would be the most important. There is currently no household ID column of any kind.**

### DIFF
--- a/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
@@ -1982,6 +1982,8 @@ Census
   :header-rows: 0
 
   * - Unique simulant ID (for PRL tracking)
+  * - Unique household ID consistent between observers (for PRL tracking)
+  * - Census household ID (arbitrary identifier which is the same for simulants in the same household but **different from other observers and between timesteps for the same household**)
   * - First name
   * - Middle initial 
   * - Last name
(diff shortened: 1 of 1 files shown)


aflaxman (approved):

aflaxman:
I guess all of these are "nice to haves" --- there will still be plenty of great functionality without this, but it will definitely enable some interesting investigations into household linking methods.

NathanielBlairStahn (approved):

zmbc:
I realized that this PR does not contain enough info to actually be implementable for the WIC changes (e.g. does not describe the format of a WIC family ID). I'm going to separate the WIC changes and merge this with only the change which has _already_ been implemented, which is adding a ground-truth household ID everywhere except SSA.

zmbc:
Not making a new ticket for this merge; it has already been implemented in [MIC-4036](https://jira.ihme.washington.edu/browse/MIC-4036).

PR #1167, opened 2023-04-04
Adjusting FPG values on lifestyle ramp to correct units

SylLutze:


aflaxman (approved):
Is this information *only* in a figure?  It seems like it should be in text eventually somewhere as well.

PR #1172, opened 2023-04-06
Move old concept models to archive subfolder

alibow:
I was tired of having to pick through the long list of concept models to find the relevant ones. Made an archive subfolder and moved non-active concept models to it.


SylLutze (approved):

pletale (approved):

aflaxman (commented):
Is this going to break links in all of the project-specific repos?  That will be a little annoying, but I guess not such a big deal.

alibow:
> Is this going to break links in all of the project-specific repos? That will be a little annoying, but I guess not such a big deal.

What links? This should not break any links internally in the vivarium research readthedocs, but I guess a copy-and-pasted external links will be changed. I could reconfigure so this would not be the case and just group them separately on the main page instead of moving to a subfolder. Would that be preferred?

aflaxman:
> What links? This should not break any links internally in the vivarium research readthedocs, but I guess a copy-and-pasted external links will be changed.

I was imagining on our model git repos there might be links back to the research docs, e.g. https://github.com/ihmeuw/vivarium_csu_sanofi_multiple_myeloma might have one.  It turns out that it actually does not have the link.

We discussed this in doc sess, I just wanted to echo here that I don't think this is a big deal... just another opportunity to bring some order to the chaos of our docs.

PR #1181, opened 2023-04-17
Document V&V, automated V&V, and fuzzy checking

zmbc:
Things I would particularly like feedback on:

- Some of the sections are very long. Could/should they be broken up into shorter subsections?
- Where would using math notation make things clearer?
- Have I cleanly separated between PRL-specific and potentially generally useful ideas?

Code that implements the statistics described here can be found at https://github.com/ihmeuw/vivarium_census_prl_synth_pop/pull/256.


docs/source/model_design/vivarium_features/automated_v_and_v/index.rst
```diff
@@ -0,0 +1,245 @@
...
+When simulant-level probabilities of an event vary within a group (for example, if there is a continuous risk factor
+of the event), the Bernoulli trials are independent but not identically distributed.
+The number of events observed has a `Poisson binomial <https://en.wikipedia.org/wiki/Poisson_binomial_distribution>`_
+distribution.
+This distribution has the same mean and **lower** variance, relative to a binomial distribution where each trial
+has the mean probability.
+This simple binomial distribution can be used as an approximation, allowing the use of a binomial test in this situation as well;
+due to the variance property, this approximation will increase the specificity of the fuzzy test to be higher than
+necessary to achieve the specified family-wise error rate.
+This increase in specificity will cause a decrease in sensitivity.
```
SylLutze, line 197: Similarly here, is there a way to show how the use of binomial/poisson would change the specificity/sensitivity visually?

SylLutze (approved):
Hey - I think this looks good! As a more visual person I think there is opportunity for even something like this to be included to help explain some of the math. 

image

docs/source/model_design/vivarium_features/automated_v_and_v/index.rst
```diff
@@ -0,0 +1,245 @@
...
+* What sort of sensitivity/specificity tradeoff do we want for our automated V&V?
+  This likely depends on both of the previous questions.
+
+PRL case study: integration tests
++++++++++++++++++++++++++++++++++
+
+We have automated some V&V checks on the PRL project by adding integration tests
+to `the simulation repository <https://github.com/ihmeuw/vivarium_census_prl_synth_pop/>`_.
+These tests run the simulation for a given number of time steps and check certain conditions
+with explicit Python :code:`assert` statements, which fail in unexpected situations.
```
alibow, line 85:
```suggestion
with explicit Python :code:`assert` statements, which fail when the model exhibits unexpected behavior.
```
Changing this because before it could be interpreted that we cannot predict when our tests will fail :)

docs/source/model_design/vivarium_features/automated_v_and_v/index.rst
```diff
@@ -0,0 +1,245 @@
...
+For example, there is a test that checks that each non-GQ household in the simulation has exactly one
+living reference person.
```
alibow, line 87:
```suggestion
living reference person, a condition which should always be true.
```

docs/source/model_design/vivarium_features/automated_v_and_v/index.rst
```diff
@@ -0,0 +1,245 @@
...
+
+The integration tests have access to intermediate simulation state, not only simulation outputs, and can check
```
alibow, line 89:
```suggestion
The integration tests have access to intermediate simulation states, not only simulation outputs, and can check
```

alibow (approved):
I think this looks great and thanks for making the improvements to the existing V&V page. I think at a minimum, linking to wikipedia or other references for Bayesian versus frequentist statistics and bonferroni correction would be a nice addition if we don't want to get too into the math on this page.

Also, I was unfamiliar with the term integration testing, so a brief definition/context may also be nice.

Overall though I think this is clear and I'm not worried about the length. Maybe we can add a to-do note to discuss how to document/implement these tests (something researchers put in concept model doc and engineers implement? something else?) but I don't think we need to address it in this PR

docs/source/model_design/vivarium_features/automated_v_and_v/index.rst
```diff
@@ -0,0 +1,245 @@
...
+"a household *always* has exactly one living reference person."
+It is more difficult to check that a value is correct when the value in the
+simulation is subject to stochastic variation.
+We clearly cannot check that the rate of events in the simulation is **exactly** the expected rate,
+or that a mean value of a continuous risk factor is **exactly** as expected.
+Nor can we use an approximation like "within X% of the true value,"
+since how much a value deviates due to chance depends on our population size.
+The difficulty of this problem is part of why, in the manual V&V process, we usually check such values visually.
+
+By building a proof of concept for the PRL simulation, we have begun to formalize this to enable
```
aflaxman, line 120:
```suggestion
By building a proof-of-concept for the PRL simulation, we have begun to formalize this to enable
```
I don't want you to think i am ant-hyphen...

aflaxman (approved):

docs/source/model_design/vivarium_features/automated_v_and_v/index.rst
```diff
@@ -0,0 +1,278 @@
...
+considers the simulation "Sick."
+The boundary can be chosen arbitrarily, which illustrates the direct tradeoff
+between sensitivity and specificity.
+By moving it left, we reduce the number of false negatives (missed bugs, in the simulation context), increasing sensitivity.
+However, we also increase the number of false positives (false alarms),
+decreasing specificity.
+
+.. figure:: PPV,_NPV,_Sensitivity_and_Specificity.svg
+
+  By Original by Luigi Albert Maria - SVG version of File\:PPV, NPV, Sensitivity and Specificity.pdf, CC BY-SA 4.0, https://commons.wikimedia.org/w/index.php?curid=99283192
```
aflaxman, line 150:
```suggestion
  Original by Luigi Albert Maria - SVG version of File\:PPV, NPV, Sensitivity and Specificity.pdf, CC BY-SA 4.0, https://commons.wikimedia.org/w/index.php?curid=99283192
```

aflaxman (commented):
I did not have time to read carefully, but I caught a suspected typo that I wanted to share.

pletale (approved):

docs/source/model_design/vivarium_features/automated_v_and_v/index.rst
zmbc: Yep I did notice this, but since the typo [is on their end](https://commons.wikimedia.org/w/index.php?curid=99283192) (it literally says "Original by Luigi Albert Maria" in the "Author" field) I figured I would keep as-is.

PR #1182, opened 2023-04-18
Adding Environment and Package Information to Computing Resources

SylLutze:
Hello - thanks for checking this work. Some things that would be especially helpful to check: 
- If there are any missing packages or missing common aliases 
- The "what environments are available to me" section is clear and you can follow it enough 
- If you were new you could find or make an environment


zmbc (approved):
Awesome! I could totally picture this being really helpful to someone starting on our team 😃

pletale (approved):
All looks great to me, Syl!

PR #1185, opened 2023-04-24
Restructure archive concept model organization

alibow:


docs/source/models/concept_models/vivarium_gates_bep_summer2020/BEP_scenarios_26oct2020.drawio
```diff
@@ -0,0 +1 @@
...wtR4SQRk7pxt2yBh2hqMLjNyR25PYNbF7dRcJOsQXP31ksR7YPpBhJwl00YwTZeb4kWZTOguW86NOq7HR+XdDy4abNM+ARrwDsNa7Q2M5yv39A//qrzLNiL+l69rAfPG6+RTyW/UcwMOyHLhI0WHvor5DDZ/6EYKxn5Io5DyA1m93o/fHaCCnMzpjjcElGe0YYrTOjuQWgG+V7xRamxPKX6fF3FYnqR/djGeff9EYkesyKhF5ZK8uaQjAk5JWWpnLSNpQkbRcHo5GVeD8wsNoMhzf6bS+B0AxQreylQZDXFJqoHIVL6kSrtgUrq6CK9eSbKY44H/7xriKXM1gbMtCazkafHX20Ri6noJu7rcsGkf7XoGd4pk7tc0Jryy6thpgVL7pIsBCtT7wiQ6yajjfVVbZ7hPbKlW9CLaWpzMLlndJdNXc/2GzXkdkxffJUx8aS1Bbic+yfz7L7QYjsc/8bJI5B+ZKo7GEio2g6Y3Xk9O8wXJ4BjDSsXJ4yQitTJ6fnU3LsMms4rqRfXLdKFrKo6LW+n2RlrYtWtjtK0nYM9/t1T6yHx/C24Yeds7fPXhipCMJPOeWv691W2cd0pYN2I5+oqMrc9zWF/iXYg291hOg4ZOlykf2oOugUTarzm1P1M0cY4re4WjpWuq1TfBKi3iq5og8U+DpChbmrST8EaqVxwT97BM97LQP1JOxA80sAFjt62qefcn9DZk6tXDy7VDbOhrFvpHqAnYb5zWq/uvTCWPGU60u9Kr/YwRtaGujJMXMZraC0zm6yf4vMWD+8Hf2wP2weJzu6o3T/Sn2nI663KsyPO1Rlwf6irq8l0ddfR+sQLUo4xWH3yUXpfTLX9F4oU29eErGSQEh6ds1JXL67GpTLjzURBOOKWOiFn0uYUxgizG59iCjeU/L9STV9xpxX1dVr6LtbsP2FIBABDvMenTNjSKCZ92K0P74G3b3+cyEM6jDZZGfIi9xrXnvwn2mjCOnfZyehBo1ypRuuxQjjFu6mxFCW60JK0J4LRdT7Sb3h4rT0mbAGBlyWrbh6159XllqggfLjO/UfS9T9XIbmgWvz9pLE7yOd+WMpV72S697vT/XoRiPZ7oO+zKuwz7Pddiv4TpMXViS7qwUeJLgZfdQ3mYeZ4NGddBTbbqr84jwbLvEHw8/Piok6PATLjT7Hw==</diagram></mxfile>
```
zmbc, line 1: It looks like a bunch of files in this directory were added by accident.

docs/source/models/concept_models/index.rst
```diff
@@ -4,10 +4,32 @@
...
-   */concept_model*
+   vivarium_census_synthdata/concept_model
+   vivarium_nutrition_optimization/concept_model
+   vivarium_us_cvd/concept_model
+   vivarium_wasting_paper/concept_model   
+
+Archive concept model documents can be found here:
+
+.. toctree::
+   :maxdepth: 1
```
zmbc, line 20: I'm curious whether it would "just work" if you made this second toctree using the original glob. I believe Sphinx is smart and won't put the same thing in two toctrees, so I _think_ it wouldn't include any of the active ones explicitly listed above?

If this works, it would make this shorter and also prevent us from accidentally forgetting one. If you want to control the order, we could add the glob **in addition** to the ones you have listed here.

SylLutze (approved):

docs/source/models/concept_models/index.rst
alibow: OK, so I tested that and it just duplicated the models listed in the first section. I think this is actually preferable though because this way the only risk is that we don't add a model to the top call-out section rather than leave anything out on accident. Now the two sections are (1) active models, and (2) all models (active and completed). Let me know if you think you preferred it the other way!

docs/source/models/concept_models/vivarium_gates_bep_summer2020/BEP_scenarios_26oct2020.drawio
zmbc: Ah, that makes sense. Sounds like a good plan!

docs/source/models/concept_models/index.rst
zmbc: Yeah that sounds good to me!

zmbc (approved):

PR #1191, opened 2023-05-08
default vivarium como behavior proposal

alibow:


docs/source/model_design/vivarium_features/other_causes_ylds/index.rst
```diff
@@ -210,34 +210,95 @@ background morbidity, including:
...
+exceptions and maintain any regular updates to their process.
+
+However, until this is able to be achieved, an interim solution will be to estimate
+the DW due to background morbidity as the difference between the all-cause YLD rate 
+and the sum of the YLD rate due to all modeled causes of disability at the 
+location/age/sex/year-specific level:
+
+.. math::
+
+  DW_\text{background} = \text{YLD rate}_\text{c294} - \sum_{c=1}^n \text{YLD rate}_c
```
zmbc, line 254: This gives me an idea, but I haven't fully thought it through. What if we defined:

*YLD_modeled = sum of YLD rates for modeled causes*

*YLD_background = YLD_c294 - YLD_modeled*

And then solved this system of two equations with two unknowns:

*YLD_c294 = 1 - (1 - DW_background) * (1 - DW_modeled)*

*DW_background / DW_modeled = YLD_background / YLD_modeled*

At first glance that seems like it would adjust for the issue you note below about modeled YLDs being a large proportion of YLDs, assuming independence. I'm not super confident about the last equation but it seems intuitively valid.

zmbc (approved):

docs/source/model_design/vivarium_features/other_causes_ylds/index.rst
alibow: This is interesting! I think there is potentially something here, but I don't think it works as written due to combining population-level and individual-level equations.

The relationship between YLDs and DWs depends on the prevalence of the condition at hand. We've been bypassing this for background mortality because we are assuming the prevalence of this condition to be 1, so in this "special" case _YLDs_background = DW_background_, but (assuming there is just one modeled cause for simplicity) _YLDs_modeled = DW_modeled * prevalence_modeled_. 

I think we might be able to do the following...

_YLDs_c294 = prevalence_modeled * (1 - (1 - DW_background) * (1 - DW_modeled)) + (1 - prevalence_modeled) * (DW_background)

DW_background / (DW_modeled * prevalence_modeled) = YLD_background / YLD_modeled_

I _think_ this holds for a single cause of disability, but I think it gets more complicated if we are modelling multiple causes... but I believe that under the assumption of independence all causes should have approximately the same "deflation" factor in their adjusted disability weight to their unadjusted disability weight from the COMO adjustment, so I think that the second equation should hold when we consider the prevalence.

Any thoughts?

alibow: I'll try to run some test calculations/adjustments in the meantime to confirm the suspicion!

alibow: Oh wait... actually I think this doesn't really work because this equation _YLDs_modeled = DW_modeled * prevalence_modeled_ that I said was correct is not actually correct. It's really  _YLDs_modeled = ADW_modeled * prevalence_modeled_, where ADW is the adjusted DW, which does not equal the DW.

BUT, DW_background is the only undefined term in this equation, so we might be able to solve it directly and go with that?

_YLDs_c294 = prevalence_modeled * (1 - (1 - DW_background) * (1 - DW_modeled)) + (1 - prevalence_modeled) * (DW_background)_

Then we only need to think through how to make this scale for more than a couple causes at a time...

zmbc: Sorry -- I don't think I was completely clear here. For DW_modeled I don't propose using an actual DW from GBD. Instead, it's just one of the two unknowns we are solving for. We wouldn't actually do anything with the solved value for DW_modeled.

I was thinking about it as if both "background" and "modeled" were conditions with prevalence 100%. I believe that simplifies the issue of multiple modeled conditions as well?

SylLutze (approved):

PR #1192, opened 2023-05-09
Closed cohort pregnancy model doc

alibow:
Added a version of the pregnancy model doc for a closed cohort. Mostly just copied over from existing doc, but major difference for review is that prevalence at initialization=1 for the pregnancy state and zero for all other states. I also set transition rate of not pregnant to pregnant to zero to avoid annoyance of new pregnancies at the very end of our fixed duration closed cohort sim.

Note that I am still including partial term pregnancies, including abortions/miscarriages/ectopic pregnancies.

NOTE: github did not track it well, but there were no changes to the 2019 pregnancy model doc.


aflaxman:
No don’t worry about it.  I’ll use diff locally to see what you’ve changed.

alibow:
@aflaxman any final thoughts?

aflaxman (approved):

PR #1193, opened 2023-05-09
GBD 2021 maternal disorders cause doc

alibow:
There were no major updates except for different decomp steps here! Just basically copying relevant material over.


aflaxman (approved):

SylLutze (approved):
I did not review this in a lot of detail since it was noted that there were not major changes. I can review it again in more depth if needed!

PR #1206, opened 2023-06-01
Document WIC-specific identifiers

zmbc:
Split out from #1164, which originally was very broad.

Also revises WIC to require living in the same household. This is necessary in order to implement WIC family ID in a reasonable way.

There are analogous changes to be made for the decennial and survey observers. Maybe the solution there is very similar, but I didn't spend any time looking into it. @ckkinuthia can we make sure to track this task?


aflaxman (approved):

NathanielBlairStahn (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2414,6 +2416,33 @@ Here is an example:
...
+then assign them consecutive integers.
+Zero-pad these integers to 9 digits before noising.
+
+.. note::
+  The `actual process for generating WIC family IDs <https://web.archive.org/web/20230601182852/https://www.wichands.com/Content/Docs/Help/ClinicUserManual/Module1/index.html?page=family-id-number.html>`_  involves using the year in combination with
+  a consecutive integer.
+  Our approach is simpler, while still capturing that these
+  numbers are likely to be similar and easy to mistake for one another.
+  In fact, we will tend to overstate how much this is the case by not changing
+  the prefix every year.
```
rmudambi, line 2432: Is there a reason we're not including the year? It seems straightforward to do so.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2414,6 +2416,33 @@ Here is an example:
...
+To generate the WIC client ID for simulants observed in WIC,
+sort them by when they were first observed in WIC (it does not matter how ties are broken),
+then assign them consecutive integers.
+Zero-pad these integers to 11 digits before noising.
+
+.. note::
+  The `actual process for generating WIC client IDs <https://web.archive.org/web/20230601183736/https://www.wichands.com/Content/Docs/Help/ClinicUserManual/Module1/index.html?page=client-id-number.html>`_ involves using the clinic number, local agency number, and year in combination with
+  a consecutive integer.
+  Our approach is simpler, while still capturing that these
+  numbers are likely to be similar and easy to mistake for one another.
```
rmudambi, line 2443: Doesn't this make it easier to link a client with the same client in a different year (maybe this isn't a use case that we care about), since in our simulation they will always have the same number, but in reality they might have a different clinic/local agency in different years?

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2533,8 +2562,15 @@ See the final limitation below for more about this approximation.
...
 #. The creation of race/ethnicity specific coverage by participate category is imperfect. We do not have granular data which includes this breakdown, so it is based on the overall coverage by race/ethnicity which is an assumption. Also, for infants this would lead to over 100% selection, so it is not changed (remains at 98.4% for most race/ethnicity groups)
-#. We use a household ID. There will be cases in which a tracked mother/guardian does not live in the same household as their infant, this might be confusing in the resulting data.
 #. Real WIC data includes dates of starting and ending eligibility. We assume that data is then rolled up over some time period, say a year, to include everyone eligible for that year. For our data, we approximate this by having everyone eligible at the first time point in the new year included in the dataset. This will exclude simulants who were eligible for some part of the prior year, but are not at the time of observation. Therefore, data sampled in Jan 2029 is the approximation of data from all of 2028.
+#. We use household as the proxy for WIC family ID, which causes two issues:
```
rmudambi, line 2566: Could we avoid this by tying the household id to the mother rather than the household?

rmudambi (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Client IDs stay constant over the years, so this is realistic. The year in the ID is the _first_ year that client was ever in WIC.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Eh, I think it's not worth it. It's a bit more work, will be weird because we won't have anyone with an (implied) year before our simulation start, and doesn't add much.

PR #1207, opened 2023-06-01
Create updated concept model template

pletale:
This PR adds the following:
- table of default simulation specifications (e.g., # of draws, pop size per draw; included note about age-start initialization vs. age-start observation) 
- updated table for V&V tracking 
- updated table for requested outputs and stratification
- outstanding V&V issues table


aflaxman (approved):

docs/source/models/templates/concept_model_template.rst
```diff
@@ -47,19 +47,14 @@
 
 
 
-.. _concept_model_template:
-
-.. todo::
-
-  Replace the above ".. _concept_model_template:" with ".. _{YOUR_MODEL_NAME}:""
```
NathanielBlairStahn, line 54: I think these lines should remain in place for the template because the `.. _concept_model_template:` target allows other documentation pages to link to the template page specifically. When individual concept model pages are created, they should each get their own label of the form `.. _{YOUR_MODEL_TITLE}:`.

SylLutze (approved):
Looks great!

docs/source/models/templates/concept_model_template.rst
```diff
@@ -72,188 +75,226 @@ Concept model template
...
+  * - Parameter
+    - Definition
+    - Value
+    - Note
+  * - Location(s)
+    - Target/simulated region & GBD ID 
+    - e.g. Ethiopia (ID: 179)
+    -
+  * - Number of draws
+    - Desired number of draws that a given simulation is to be run for. (Generally, this should be a number between 0 and 999.)
```
alibow, line 130:
```suggestion
    - Desired number of draws that a given simulation is to be run for. (Generally, this should be a number between 1 and 999.)
```
I think we need at least one :)

docs/source/models/templates/concept_model_template.rst
```diff
@@ -72,188 +75,226 @@ Concept model template
...
+    - e.g. 50 draws 
+    - Read more about draws in Vivarium :ref:`here <vivarium_best_practices_monte_carlo_uncertainty>`.
+  * - Population size per draw
+    - Desired simulated population size per draw for a given simulation. 
+    - e.g. 100,000 simulants
+    - Read more about how to determine a reasonable population size for your simulation 
+      :ref:`here <vivarium_best_practices_monte_carlo_uncertainty>`.
```
alibow, line 137:
```suggestion
    - Read more about how to determine a reasonable population size for your simulation 
      :ref:`here <vivarium_best_practices_monte_carlo_uncertainty>. Note that engineers will
      decide how many random seeds to use to achieve the specified population size per draw.`.
```

docs/source/models/templates/concept_model_template.rst
```diff
@@ -72,188 +75,226 @@ Concept model template
...
+    -
+  * - Simulation end date
+    - 
+    - e.g. 2026-12-31
+    -
+  * - Timestep
+    - Amount of time that passes in simulation between each observer.
+    - e.g. 4 days
+    - Read more about how to determine a reasonable timestep for your simulation 
+      :ref:`here <vivarium_best_practices_time_steps>`.
```
alibow, line 173:
```suggestion
  * - Timestep
    - Amount of time that passes in simulation between each observer.
    - e.g. 4 days
    - Read more about how to determine a reasonable timestep for your simulation 
      :ref:`here <vivarium_best_practices_time_steps>`.
  * - Randomness key columns
    - Parameters used to identify identical simulants across scenarios
    - e.g. ['entrance_time', 'maternal_id', 'sex']
    - Entrance time is currently required as a randomness key column 
  
```

Let's include randomness key columns in this table too!

docs/source/models/templates/concept_model_template.rst
pletale: Good point! Lol (-:

docs/source/models/templates/concept_model_template.rst
```diff
@@ -49,17 +49,20 @@
 
 .. _concept_model_template:
 
-.. todo::
-
-  Replace the above ".. _concept_model_template:" with ".. _{YOUR_MODEL_NAME}:""
```
zmbc, line 54: It seems like we'd want to keep this.

docs/source/models/templates/concept_model_template.rst
```diff
@@ -72,188 +75,231 @@ Concept model template
...
+  * - Simulation observation start date
+    - 
+    - e.g. 2022-01-01
+    -
+  * - Simulation end date
+    - 
+    - e.g. 2026-12-31
+    -
+  * - Timestep
+    - Amount of time that passes in simulation between each observer.
```
zmbc, line 170: Maybe this was copied from somewhere, but "between each observer" doesn't sound quite right to me -- there are other reasons to have a short time step that don't have to do with observation. Maybe:

```suggestion
    - Amount of time that passes in simulation between each "step," which is when events and observations can happen.
```

zmbc (approved):
Looks great!

docs/source/models/templates/concept_model_template.rst
NathanielBlairStahn: Wouldn't the maximum be 1000 then, not 999?

docs/source/models/templates/concept_model_template.rst
```diff
@@ -72,188 +75,231 @@ Concept model template
...
+   :widths: 3 10 20
+   :header-rows: 1
 
-.. todo::
+   * - Run number
+     - V&V criteria
+     - V&V summary
+   * - e.g. 1.0 
+     - e.g. Confirm that there is no variation in person-time quantity between different observers of same measure.
+     - e.g. V&V notebooks for model 1.0 can be found here [insert Github link]. V&V criteria satisfied.
```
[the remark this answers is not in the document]
pletale: Will add that note!

docs/source/models/templates/concept_model_template.rst
```diff
@@ -72,188 +75,231 @@ Concept model template
...
+  :header-rows: 1
+
+  * - Run number
+    - Run description
+    - Scenarios
+    - Specification modifications
+    - Stratification modifications
+  * - e.g. 1.0
+    - e.g. Baseline concept model updates
+    - e.g. 50 draws; 200,000 pop size
```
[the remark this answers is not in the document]
pletale: Oh yes good catch -  that is supposed to be for an 'Observer modification' column, which I evidently forgot to add. Adding now!

docs/source/models/templates/concept_model_template.rst
```diff
@@ -72,188 +79,242 @@ Concept model template
...
+  * - Number of draws
+    - Desired number of draws that a given simulation is to be run for. (Generally, this should be a number between 1 and 1,000.)
+    - e.g. 50 draws 
+    - Read more about draws in Vivarium :ref:`here <vivarium_best_practices_monte_carlo_uncertainty>`.
+  * - Population size per draw
+    - Desired simulated population size per draw for a given simulation. 
+    - e.g. 100,000 simulants
+    - Read more about how to determine a reasonable population size for your simulation 
+      :ref:`here <vivarium_best_practices_monte_carlo_uncertainty>. Note that engineers will
+      decide how many random seeds to use to achieve the specified population size per draw.`.
```
NathanielBlairStahn, line 142:
```suggestion
      :ref:`here <vivarium_best_practices_monte_carlo_uncertainty>`. Note that engineers will
      decide how many random seeds to use to achieve the specified population size per draw.
```
I think a misplaced backtick got introduced in commit 07c66ad4c59bbad62613d11b9c54879935b2a72f, which is probably what's causing the build to fail.

docs/source/models/templates/concept_model_template.rst
```diff
@@ -72,188 +79,242 @@ Concept model template
...
+Describe the location (country and/or region(s)) of the simulation here.
 
+.. _{YOUR_MODEL_SHORT_NAME}3.0:
 
-.. _{YOUR_MODEL_SHORT_NAME}5.0:
-
-5.0 Vivarium modelling components
-+++++++++++++++++++++++++++++++++
+3.0 Verification & validation (V&V) tracking
+--------------------------------------------
```
NathanielBlairStahn, line 208: Should this section be one level out, i.e., `++++++` like sections 2.0 and 4.0 instead of `------`?

docs/source/models/templates/concept_model_template.rst
```diff
@@ -72,188 +79,242 @@ Concept model template
...
+      decide how many random seeds to use to achieve the specified population size per draw.
+  * - Age start (initialization)
+    - If sim duration > age start for observation, age_start at initialization will be zero and
+      fertility will need to be included. Otherwise, age start at initialization will be age start for 
+      observation minus simulation duration and a closed cohort (no fertility).
+    - e.g. 0 months
+    - Currently vivarium can only have new simulants enter the simulation at age zero, although 
+      changing this could be a potential framework improvement. 
+  * - Age start (observation)
+    - Age at which simulants are included in observer.
```
NathanielBlairStahn, line 151: Which observer? All of them? Should it be "observers" instead of "observer"?

docs/source/models/templates/concept_model_template.rst
```diff
@@ -72,188 +79,242 @@ Concept model template
...
+    - e.g. 6 months
+    -
+  * - Age end
+    - Age at which simulants are no longer included in observer.
```
NathanielBlairStahn, line 155: Same comment: Should it be "observers" instead of "observer"?

PR #1210, opened 2023-06-09
Preg model 0.0 v and v

alibow:


aflaxman (approved):
It would be cool to have the visual proof of calibration from the notebook included in the table of this doc, too.

SylLutze (approved):

alibow:
> Should we make a ticket for the engineers to add the person-time observer?

I believe they are already aware that we want it in the next run, but let's double check with them in the RT/SE check-in tomorrow

PR #1211, opened 2023-06-12
Bash files for computing resources 

SylLutze:
Finally making the PR for bash files in computing resources. Let me know if there's anything I missed from our conversation before or if things aren't clear.


docs/source/onboarding_resources/computing/index.rst
```diff
@@ -280,22 +280,161 @@ lines from your trainings!
...
+
+Your bash rc file is run more frequently and can contain helpful settings and information 
+needed to run certain commands. First, you will need to run the below line of code in 
+your terminal to set up conda. Once you're logged into the cluster on the terminal of 
+your choosing, run: 
+
+.. code-block:: bash 
+  :linenos:
+
+  $ /ihme/code/central_comp/miniconda/bin/conda init
```
zmbc, line 313: This is also mentioned above, in the conda section -- not necessarily a bad thing, but you might want to have some kind of call-out between the two places.

zmbc (approved):
Looks good!

Not sure where, but can you add a note that after you edit bash_profile/bashrc you need to restart the session for the changes to take effect?

docs/source/onboarding_resources/computing/index.rst
SylLutze: Good call out!

pletale (approved):

docs/source/onboarding_resources/computing/index.rst
```diff
@@ -280,22 +280,166 @@ lines from your trainings!
...
+Once you make edits to either of these files, you will need to log out and then back into 
+the cluster before they will take effect. 
 
-  - Confirm and paste in bash files 
-  - Get Zeb's help in explaining Bash files better 
-  - Moving up since this is needed for a lot of things below 
+bash profile
+~~~~~~~~~~~~
+
+The bash profile is only run when you open the cluster, and so is usually a shorter file.
```
NathanielBlairStahn, line 297:
```suggestion
The bash profile is only run when you log in to the cluster, and so is usually a shorter file. 
```
Based on [this reference](https://www.tutorialspoint.com/bashrc-vs-bash-profile-what-is-difference#:~:text=Bash_profile%20is%20another%20configuration%20file,at%20start%20of%20your%20session.) and [this reference](https://apple.stackexchange.com/questions/51036/what-is-the-difference-between-bash-profile-and-bashrc), I think you mean "log in to" when you say "open". Does that sound right?

docs/source/onboarding_resources/computing/index.rst
```diff
@@ -280,22 +280,166 @@ lines from your trainings!
...
+the cluster before they will take effect. 
 
-  - Confirm and paste in bash files 
-  - Get Zeb's help in explaining Bash files better 
-  - Moving up since this is needed for a lot of things below 
+bash profile
+~~~~~~~~~~~~
+
+The bash profile is only run when you open the cluster, and so is usually a shorter file. 
+Below are the lines of code you need in your bash profile.
```
NathanielBlairStahn, line 298: Are these lines of code that you need to manually add to your `.bash_profile`, or should they already be there? I don't remember having to edit anything in my `.bash_profile` when first logging into the cluster. If it's things that are already there, perhaps we should clarify that you don't need to mess with it. Or even just don't go into any details if it's not something the user needs to know -- e.g., say something like "Your bash profile contains a few lines of code to load generic profile settings and load your `.bashrc` file. You shouldn't need to edit these." Or, if the precise lines of code _are_ something the user needs to know, clarify exactly what the user needs to do.

docs/source/onboarding_resources/computing/index.rst
```diff
@@ -280,22 +280,166 @@ lines from your trainings!
...
+interacting with the cluster. When the cluster tries to execute some command 
+line prompts, it will look to your bash configuration files for information 
+or filepaths. 
 
-However, Bash files can be confusing since it is less obvious when the 
-information is being used or what it is used for. Therefore, we have 
-provided a copy-and-paste formatting for information to be added to 
-your Bash files. 
+You should have 2 bash configuration files, your bash profile or :code:`.bash_profile` 
+and your bash rc or :code:`.bashrc`.
```
NathanielBlairStahn, line 289: It might be nice to mention that the "rc" in `.bashrc` stands for "run commands" and originates from the predecessors of Unix (CTSS and Multics), as described here:

https://superuser.com/questions/173165/what-does-the-rc-in-bashrc-etc-mean

I've never seen it written "bash rc" before as opposed to "bashrc", and it made me curious what in the world "rc" actually meant -- I never knew, so I had to look it up.

docs/source/onboarding_resources/computing/index.rst
SylLutze: Yeah - makes sense!

docs/source/onboarding_resources/computing/index.rst
SylLutze: Hmm good point! I edited this a bit to make it more clear that you shouldn't need to do anything for this one.

docs/source/onboarding_resources/computing/index.rst
SylLutze: Neat! I didn't know this either and added it for a reference.

PR #1216, opened 2023-06-14
Pregnancy model versions table

alibow:
I'm thinking we can go over this at the next RT/SE check-in. This is a suggested outline intended to start the conversation and is not set in stone, so we can definitely make changes based on what will make sense for implementation.


SylLutze (approved):
Approving now knowing we need to talk to the engineers about this. My only reaction is that we are asking for a lot of different outputs and stratifications at different times. Will this slow us down with observer creation?

alibow:
> Approving now knowing we need to talk to the engineers about this. My only reaction is that we are asking for a lot of different outputs and stratifications at different times. Will this slow us down with observer creation?

I think we should be alright with this because all of these observers have already been implemented in IV iron, but it would be good to get the engineer's reaction to the strategy too

aflaxman (approved):

PR #1226, opened 2023-06-20
Nutrition optimization project aims and overview

alibow:
Writing down how I think we've been thinking about this so far. Zeb raised an interesting point about some potential limitations of this approach as it relates to heterogeneity and common random numbers. I've (not very clearly, sorry) jotted down the high-level point of this limitation in a "todo" note for now and plan to expand upon it soon - just getting it in words and on others' radar for now!


docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
```diff
@@ -57,14 +57,77 @@ Documents that contain information specific to the subcomponents of this simulat
 2.0 Modeling aims and objectives
 ++++++++++++++++++++++++++++++++
 
+**Objective:** For a given fiscal envelope, determine the optimal funding allocation across select nutrition-related intervention products for maternal and child health. "Optimal" is defined with respect to some specified outcome such as the funding allocation that minimizes DALYs or maximizes the number of alive, non-stunted children.
 
+.. note::
+  
+  We must be careful in optimizing for "minimizing" the prevalence of some risk such as child growth failure. This is because prevalence of such risks could be decreased via decreasing incidence (positive health impact) *or* via increasing risk-related mortality (negative health impact). That is why we choose to *maximize* the number of alive non-stunted children rather than minimize the prevalence of child growth failure.
```
zmbc, line 64: It seems like minimizing **DALYs** due to child growth failure would be more consistent with how we usually model things? Maximizing the number of alive non-stunted children will incentivize our optimization to increase fertility, and will place no importance on improving the mortality rate among those with stunting.

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
```diff
@@ -57,14 +57,77 @@ Documents that contain information specific to the subcomponents of this simulat
...
+
+**In the case of the first situation:**
+
+  There may be important interactions between the *simultaneous* coverage of intervention A and B in the same subpopulation. For example:
+
+  - An infant who was covered by antenatal supplementation may survive the neonatal period due to associated improvements to their LBWSG exposure and therefore survive to benefit from SQLNS supplementation later in life
+  - A child who may have benefitted from SAM treatment alone may no longer need to be treated for SAM if they are covered by MAM treatment as well as SAM treatment
+  - Etc.
+
+  These interactions may be difficult to quantify directly and therefore we will plan to estimate the effect of joint coverage of multiple interventions on maternal and child health from within our simualtion.
```
zmbc, line 89:
```suggestion
  These interactions may be difficult to quantify directly and therefore we will plan to estimate the effect of joint coverage of multiple interventions on maternal and child health from within our simulation.
```

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
```diff
@@ -57,14 +57,77 @@ Documents that contain information specific to the subcomponents of this simulat
...
+
+3. Some combination of 1 and 2
+
+
+**In the case of the first situation:**
+
+  There may be important interactions between the *simultaneous* coverage of intervention A and B in the same subpopulation. For example:
+
+  - An infant who was covered by antenatal supplementation may survive the neonatal period due to associated improvements to their LBWSG exposure and therefore survive to benefit from SQLNS supplementation later in life
+  - A child who may have benefitted from SAM treatment alone may no longer need to be treated for SAM if they are covered by MAM treatment as well as SAM treatment
```
zmbc, line 86: Does this complicate the calculation of cost at all? My understanding is that we have defined "coverage" such that this child is "covered" by SAM treatment, regardless of whether or not they actually need it. But I'd assume that this "coverage" is cheaper when the treatment doesn't actually need to be performed.

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
```diff
@@ -57,14 +57,77 @@ Documents that contain information specific to the subcomponents of this simulat
...
+
+  These interactions may be difficult to quantify directly and therefore we will plan to estimate the effect of joint coverage of multiple interventions on maternal and child health from within our simualtion.
+
+**In the case of the second situation:**
+
+  For our purposes, we assume that the situation described in point #2 above is equivalent to the weighted average of 50% an identical population with 100% coverage of intervention A and 50% an identical population with 100% coverage of intervention B. 
+
+    NOTE: see a discussion of the potential limitations of this assumption as it relates to simulated common random numbers in the `6.0 Limitations`_ section.
+
+  As this assumption requires only a weighted average, we can perform this calculation outside of our simualtion so long as we have quantified the impacts of intervention A and B within our simulation already.
```
zmbc, line 97:
```suggestion
  As this assumption requires only a weighted average, we can perform this calculation outside of our simulation so long as we have quantified the impacts of intervention A and B within our simulation already.
```

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
```diff
@@ -57,14 +57,77 @@ Documents that contain information specific to the subcomponents of this simulat
...
+
+**In the case of the third situation:**
+
+  Then, we can combine our strategies for the first two situations so that we may estimate more complex combinations of joint intervention coverage (for example: 50% of the population covered by A and B, 25% covered by A, 25% covered by neither, for total poulation coverage of 75% for A, 50% for B).
```
zmbc, line 101:
```suggestion
  Then, we can combine our strategies for the first two situations so that we may estimate more complex combinations of joint intervention coverage (for example: 50% of the population covered by A and B, 25% covered by A, 25% covered by neither, for total population coverage of 75% for A, 50% for B).
```

zmbc (approved):

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
alibow: Coverage is really "probability you will receive treatment if you are eligible/need it." All our simulants are subject to that same probability regardless of whether they need the treatment. But yeah, our simulation outputs both the number of treated SAM cases and total SAM cases (which actually will be the same because we'll have 100% coverage), for instance, so we calculate costs directly from how many treatments were actually administered rather than based on the cost of "100% coverage"

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
zmbc: Interesting!

I don't know the details, but given that we have some pregnancy-related interventions, it still seems like we could end up "improving stunting" by increasing the number of live births in the sim. For example, if the budget was large enough that we had already maxed out all the direct stunting treatment. I wonder if the Optima paper had this issue as well?

Once we get there, it might be interesting to look at both (minimizing DALYs due to stunting, maximizing non-stunted children), which would directly show whether this is an issue. Both measures of impact will be in our outputs, I assume, so it should be pretty straightforward to check it.

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
zmbc: Makes sense. I guess I didn't realize until now that cost requires an "emulator" just as complex as impact, because the cost of 100% coverage of A and B is not the same as the cost of 100% coverage of A + the cost of 100% coverage of B.

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
alibow: Ah, I see what you're saying. Yeah I agree that would be an interesting comparison. FWIW, I do expect the fertility impacts to be tiny (stillbirths are like <1% of live births in Ethiopia and we reduce the risk of stillbirth by like 10% or something). I also think we would need to be clever about tracking CGF-attributable DALYs in the sim... typically this rate is estimated at a population rather than individual level, so we'd have to do some extra math to decide whether someone who is stunted and dies of a stunting-related cause would have still died of that cause if they were not stunted at the time of death. We can definitely do so, but I do think it would be some extra observer logic.

This makes me think though that we need to more carefully think through stillbirth dynamics more generally. By minimizing DALYs we may deprioritize interventions that reduce stillbirths (if we keep with GBD practice of not assigning stillbirths any DALYs). In this case, it may actually be "better" to maximize the number of alive non-stunted children (averting a stillbirth is good! But isn't actually directly related to stunting...). Without stillbirths, I think the approaches will be equivalent, so I think starting with how we want to consider stillbirths will be important in guiding our next steps here...

I am leaning towards assigning stillbirths YLLs. @aflaxman, do you have any concerns with that approach?

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
alibow: image
Yeah! That's how we get this cool trend where the cost of full coverage of SAM (and MAM) treatment decreases as coverage of other interventions increases

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
zmbc: Note: "increasing the number of live births" does not (in principle) depend only on stillbirth dynamics. An intervention could also increase the number of live births by averting miscarriages, averting pregnancy-related deaths which prevent future pregnancies, or even directly increasing the fertility rate. I don't know if any of our interventions do these things, however.

SylLutze (approved):

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
alibow: Yes, true! None of our modeled interventions avert miscarriages or directly affect the fertility rate. Also, we have designed our sim to only evaluate single pregnancies/births/childhoods in a closed cohort fashion. In other words our simulated timeframe does not allow for repeat pregnancies and therefore we won't increase fertility due to averting pregnancy-related deaths within our sim, although this could happen in real life - so a definite limitation here although this would be expected to have a small effect because death during childbirth is still a rare outcome.

I'll be sure to document these things that we're able to make simplifying assumptions around for now that we would want to revisit if we ever added interventions that would break these assumptions.

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
alibow: Hmm yeah, good question. The way this is typically handled in other models is that there is a price for an "intervention bundle" if the bundle is just SAM tx, then cost will be screening + the cost of SAM tx. If the bundle is SAM and MAM tx, then cost is screening + SAM tx + MAM tx and the screening cost is not specific to either. This is likely how we will treat it in our scenarios/emulator, but that doesn't answer how we should present it in these sorts of figures.

I suppose we could either (1) spread out screening (or other sorts of shared costs) across all the interventions they apply to equally, or (2) have a separate color in the bar chart for screening costs. I like the second option better.

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
```diff
@@ -57,14 +57,77 @@ Documents that contain information specific to the subcomponents of this simulat
...
+-----------------------------
+
+2.1.1 Conceptualizing "mixed" intervention coverage
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To achieve the general modeling objective of determining the most cost effective combination of interventions for a given budget size. When we think about "*combinations* of interventions," it is important to distinguish between two modes in which coverage of more than one intervention may occur. For example, imagine a scenario in which we have 50% population coverage of intervention A and 50% population coverage of intervention B. Any of the following could be true:
+
+1. Combined intervention coverage among the same individuals: 50% of the population is covered by intervention A and B, 50% of the population is covered by neither intervention A nor B
+
+2. Combined intervention coverage among separate individuals: 50% of the population is covered by intervention A, 50% of the population is covered by intervention B
```
aflaxman, line 76:
```suggestion
2. Combined intervention coverage among separate individuals: 50% of the population is covered by intervention A and not B, 50% of the population is covered by intervention B and not A
```

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
```diff
@@ -57,14 +57,77 @@ Documents that contain information specific to the subcomponents of this simulat
...
+
+    NOTE: see a discussion of the potential limitations of this assumption as it relates to simulated common random numbers in the `6.0 Limitations`_ section.
+
+  As this assumption requires only a weighted average, we can perform this calculation outside of our simulation so long as we have quantified the impacts of intervention A and B within our simulation already.
+
+**In the case of the third situation:**
+
+  Then, we can combine our strategies for the first two situations so that we may estimate more complex combinations of joint intervention coverage (for example: 50% of the population covered by A and B, 25% covered by A, 25% covered by neither, for total population coverage of 75% for A, 50% for B).
+
+    In this case, we would simulate 3 scenarios (1: 100% coverage of A and B, 2: 100% coverage of A, 3: no coverage) and then estimate the overall impact of the described intervention coverage combination as 0.5 * scenario_1 + 0.25 * scenario_2 + 0.25 scenario_3.
```
aflaxman, line 103: Does this work??

aflaxman (approved):

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
```diff
@@ -57,14 +57,77 @@ Documents that contain information specific to the subcomponents of this simulat
...
```
zmbc, line 103: Oops, missed this on my first review:

```suggestion
    In this case, we would simulate 3 scenarios (1: 100% coverage of A and B, 2: 100% coverage of A, 3: no coverage) and then estimate the overall impact of the described intervention coverage combination as 0.5 * scenario_1 + 0.25 * scenario_2 + 0.25 * scenario_3.
```

PR #1231, opened 2023-06-27
Adding to-do list for costing and emulator

SylLutze:
We can review this and discuss task assignment on Thursday in our research side meeting.


docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
```diff
@@ -131,6 +131,80 @@ Step 1 in this process is detailed in the :ref:`pregnancy<2021_concept_model_viv
...
+  * - 3: Clean emulator and remove hardcoded information 
+    - 
+    - Syl 
+    - 
+  * - 4: Allow for optimization to multiple outcomes 
+    - 
+    - TBD
+    - Example: deaths, DALYs, SAM cases 
+  * - 5: Add maternal interventions 
+    -
```
alibow, line 162:
```suggestion
    - Blocked by simulation progress
```

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
```diff
@@ -131,6 +131,80 @@ Step 1 in this process is detailed in the :ref:`pregnancy<2021_concept_model_viv
...
+    - Example: deaths, DALYs, SAM cases 
+  * - 5: Add maternal interventions 
+    - 
+    - TBD
+    - 
+  * - 6: Test for robustness to different initial values 
+    - 
+    - TBD
+    - 
+  * - 7: Add non-linear cost functions
```
alibow, line 169: Should we have this in the wave I or wave II section?

alibow (approved):
Thanks for putting this together! I made a couple comments but I think it looks great and will be good to continue discussing as a team in our RT check-ins.

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
aflaxman: wave I, if possible

aflaxman (approved):

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
alibow: Ok, I've been thinking about this as not-necessary-for-success in the wave I interim deliverable given that Laura and Rahul said they were interested in just product costs only to start, but agreed that it would be cool to add it in if we can do so without blocking wave I finalization otherwise. Is that what you're thinking?

SylLutze: I agree it is not needed for a successful wave 1. I was thinking of this as testing the emulator with dummy non-linear cost data to make sure it works as expected. Also forces us to build the emulator in such a way that this would be possible. But it could move to wave 2 if we think it will block us.

SylLutze: I added a note stating it wasn't technically needed and changed the language a bit. Hopefully this works for everyone now!

aflaxman:
I like that plan

docs/source/models/concept_models/vivarium_nutrition_optimization/concept_model.rst
alibow: Awesome, sounds good to me!

PR #1234, opened 2023-06-27
Add sex to SSA observer in concept model

pletale:
Related JIRA ticket [here](https://jira.ihme.washington.edu/browse/SSCI-1361). 

This PR adds sex to the SSA observer in concept model docs. Once engineers add this, we need to make sure to update the pseudopeople docs accordingly too.


NathanielBlairStahn (approved):

aflaxman (approved):
Do we also need to say what noise types to apply to this column and what there default values should be?

NathanielBlairStahn:
> Do we also need to say what noise types to apply to this column and what there default values should be?

That's in a different section of the concept model: All the observers are described first, then there's a separate section about noise.

aflaxman:
:thumbs up emoji:

NathanielBlairStahn:
Actually, Abie's comment made me realize that we **also** should update the [noise table](https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#id115), but it looks like that table already says that "Sex" should only appear in the Census, Surveys, and WIC, so I think we're good.

NathanielBlairStahn (changes requested):
As per @aflaxman 's comment, we need to edit the noise table to reflect this change. See my other comment.

pletale:
@NathanielBlairStahn just added that change you requested!

NathanielBlairStahn (approved):
Thanks, Alix, looks good!

PR #1235, opened 2023-06-28
PRL: Remove age and DOB from tax forms

NathanielBlairStahn:
Related Jira tickets for [documentation](https://jira.ihme.washington.edu/browse/SSCI-1362) and [engineering](https://jira.ihme.washington.edu/browse/MIC-4134).

**What?** Remove "date of birth" and "age" columns from tax forms (W2/1099 and 1040), both in the observers and in the noise table.

**Why?** Neither tax form asks about age or date of birth, and [Zeb found](https://ihme.slack.com/archives/C02KUQ9LX32/p1686098171564369) this note in the [AR census report](https://www2.census.gov/programs-surveys/decennial/2020/program-management/evaluate-docs/EAE-2020-admin-records-experiment.pdf) (footnote 19 on page 30):
> IRS records do not contain DOB, and many of the records contain only the first four letters of the last name.


zmbc (approved):

albrja (approved):

aflaxman (approved):
should we do the thing about names having four characters, too?

NathanielBlairStahn:
> should we do the thing about names having four characters, too?

@aflaxman I separated that into its own ticket and put it in Orange release. Do you think we should try to get it into Yellow instead? The main problem is that I don't know what percentage of names it means by "many." I guess one way around this would be to make it configurable -- we could add a new noise function called, e.g., `truncate`.

aflaxman:
Good plan, and we can pick an arbitrary value for many, like 20%.  Please keep it in the orange release, that was a good call on your part.

NathanielBlairStahn:
> Good plan, and we can pick an arbitrary value for many, like 20%. Please keep it in the orange release, that was a good call on your part.

Thanks, Abie. I updated the [JIRA ticket](https://jira.ihme.washington.edu/browse/SSCI-1410) with the idea of creating a new noise function for truncating strings, and specified that it should go in orange release.

PR #1244, opened 2023-07-10
Pregnancy model 1.2 V&V

alibow:


aflaxman (approved):
Awesome.  In notebook, cell 77 might have wrong title on right-hand plots, and putting the plots for draws into a pdf flipbook or animated gif will be fun!

SylLutze (approved):

PR #1256, opened 2023-07-12
Add intro paragraph to section 1.0 in BLISS style guide

pletale:
I'm not sure using Github 'Issues' is how we want people to be able to contact us if they have questions - I just copied this from the pseudopeople docs.

Please lmk what you think of the _what this is vs. what this isn't_ table! Coming up with the _what this isn't_ values was a bit challenging so very open to any feedback!


docs/source/model_design/general_reference_material/bliss/index.rst
```diff
@@ -40,7 +40,37 @@ language into our day-to-day work.
...
+well as other teams at IHME!
+
+On this page you will find our guiding principles, a glossary of commonly used terms related
+to BLISS, recommendations for how to address sex/gender-related limitations in external data
+sources, discussion of the implications of imperfect proxy measures with regards to sex/gender,
+examples of how our team has incorporated BLISS into our work process, and more! 
+
+If you have any questions or comments about the content on this page - please reach out! You can
+contact us directly `on GitHub <https://github.com/ihmeuw/vivarium_research/issues?q=is%3Aopen+is%3Aissue+label%3Abliss>`_ or anonymously 
+on our `team HUB page <https://hub.ihme.washington.edu/display/COS/BLISS%3A+Building+Language+Inclusivity+into+Simulation+Science>`_.
```
aflaxman, line 56: I guess this only works for people who have HUB access, such as IHME staff

aflaxman (approved):

SylLutze (approved):
I like the list of is/isn't!

zmbc (approved):
Looks good!

docs/source/model_design/general_reference_material/bliss/index.rst
zmbc: Yes, we link to internal-only pages pretty frequently in Vivarium Research, and usually don't have any kind of special marker on these links. Seems like something that would be good to have more of a policy around, but not unique to this page.

alibow (approved):
Looks great!!

PR #1258, opened 2023-07-12
Timestep scalar update

alibow:
Would love to get someone else's eyes on this! I think this is correct. @aflaxman @rmudambi you two are the most familiar with these equations, I believe


aflaxman (approved):
I'm not sure I understand this, so let's discuss if it is giving anyone trouble

SylLutze (commented):
I am not really understanding what this PR is. Happy to approve if you feel comfortable though.

SylLutze (approved):

PR #1262, opened 2023-07-14
Updating correlation link

SylLutze:


docs/source/models/risk_correlation/sbp_ldlc_fpg_bmi/index.rst
```diff
@@ -54,7 +54,7 @@ would be included in the following way:
 
 The correlation was then recalculated based on the above (age specific or not) 
 and the final values for correlation coefficients are stored in the csv file 
-here: /ihme/costeffectiveness/artifacts/vivarium_nih_us_cvd/raw_data/correlation.csv 
+here: /mnt/team/simulation_science/costeffectiveness/artifacts/vivarium_nih_us_cvd/raw_data/correlation.csv
```
aflaxman, line 57: Should we (research) be providing draws here, to capture uncertainty?

aflaxman (approved):

docs/source/models/risk_correlation/sbp_ldlc_fpg_bmi/index.rst
SylLutze: We could create draws yes. It is based on NHANES so we can just sample a subset each time randomly to make draws. Or we could just generate a confidence interval around the estimates we are providing. Do you think that add value here?

alibow (approved):

PR #1264, opened 2023-07-14
Add a glossary of general terminology for BLISS 

pletale:
Please let me know if you think these definitions should be tweaked or if there are any other definitions that we should add here! I included a to-do note to make a visual for Dos and Don'ts in the future.

### DIFF
--- a/docs/source/model_design/general_reference_material/bliss/index.rst
+++ b/docs/source/model_design/general_reference_material/bliss/index.rst
@@ -53,9 +53,9 @@ examples of how our team has incorporated BLISS into our work process, and more!
 
 If you have any questions or comments about the content on this page - please reach out! You can
 contact us directly `on GitHub <https://github.com/ihmeuw/vivarium_research/issues?q=is%3Aopen+is%3Aissue+label%3Abliss>`_ or anonymously 
-on our `team HUB page <https://hub.ihme.washington.edu/display/COS/BLISS%3A+Building+Language+Inclusivity+into+Simulation+Science>`_.  
+on our `feedback form <https://docs.google.com/forms/d/e/1FAIpQLSeCED9TFQsH-1u4QkFxJvno4WaEDz6h9rhJeyFlAlqyG7MAJg/viewform>`_.  
 
-. list-table:: **What this is vs. What this isn't**
+.. list-table:: **What this is vs. What this isn't**
    :header-rows: 1
 
    * - This is...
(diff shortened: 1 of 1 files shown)


docs/source/model_design/general_reference_material/bliss/index.rst
```diff
@@ -145,9 +145,27 @@ the work and use the terms we find most accurate and inclusive.
...
+   * - Gender
+     - *Gender*, on the other hand, is a sociocultural construct that refers to social norms, behaviors, and institutional structures that shape any individual's experiences, exposures, and access to power and resources, as well as one's internal sense of self as a gendered individual [Ritz_and_Greaves_2022]_. 
+   * - Sexual and gender minority (SGM) 
+     - An umbrella term that includes (but is not limited to) individuals that are two-spirit, intersex, lesbian, gay, bisexual, transgender, non-binary, asexual, or have other gender non-conforming identities.
+   * - Nondifferentiated sex/gender measure
+     - A useful hybrid term that reflects the dynamic relationship that exists between sex and gender and acknowledges the common conflation between sex and gender in public health research and data [Ritz_and_Greaves_2022]_. 
+
```
SylLutze, line 165: I think this is good! But I think we could expand. First, we could add more BLISS terms: intersex, non-binary, transgender, sex assigned at birth, cisgender, gendered language etc. We also could include some more research/simulation terms, but this is probably not needed if the audience is IHME people only, simulation, primary data, proxy measure, etc.

SylLutze (approved):

docs/source/model_design/general_reference_material/bliss/index.rst
```diff
@@ -145,9 +145,27 @@ the work and use the terms we find most accurate and inclusive.
...
+throughout this style guide. 
+
+
+.. list-table:: **Glossary**
+   :header-rows: 1
+
+   * - Term
+     - Definition
+   * - Sex
+     - *Sex* refers to biological attributes and functions associated with chromosomal makeup, endocrine function, reproductive organs, secondary sex characteristics, and other physiological and morphological variations and processes not directly related to reproduction (e.g., body size and composition, metabolic function, or organ function) [Ritz_and_Greaves_2022]_.
```
zmbc, line 158: It seems like these are very close paraphrases of the Ritz and Greaves paper; probably good to simplify a little bit and put in our own words, or just include the full quotes?

docs/source/model_design/general_reference_material/bliss/index.rst
```diff
@@ -145,9 +145,27 @@ the work and use the terms we find most accurate and inclusive.
...
+   * - Gender
+     - *Gender*, on the other hand, is a sociocultural construct that refers to social norms, behaviors, and institutional structures that shape any individual's experiences, exposures, and access to power and resources, as well as one's internal sense of self as a gendered individual [Ritz_and_Greaves_2022]_. 
+   * - Sexual and gender minority (SGM) 
+     - An umbrella term that includes (but is not limited to) individuals that are two-spirit, intersex, lesbian, gay, bisexual, transgender, non-binary, asexual, or have other gender non-conforming identities.
+   * - Nondifferentiated sex/gender measure
+     - A useful hybrid term that reflects the dynamic relationship that exists between sex and gender and acknowledges the common conflation between sex and gender in public health research and data [Ritz_and_Greaves_2022]_.
```
zmbc, line 164: I think we can define this term more precisely using the Bauer paper. Happy to help workshop this definition.

I also just noticed that Bauer calls it "undifferentiated" instead of "nondifferentiated" -- we might want to switch unless we can remember a good reason we changed this.

zmbc (approved):

docs/source/model_design/general_reference_material/bliss/index.rst
pletale: Agreed, will try to simplify before merging

docs/source/model_design/general_reference_material/bliss/index.rst
pletale: Oh yes good point! Sounds great. I like to think this could somehow have an audience beyond just IHME people, but maybe that's out of the scope for now

PR #1270, opened 2023-07-20
N.O. child model version proposal

alibow:
To be discussed at the upcoming concept model meeting - just getting this written down as a proposal for now.


SylLutze (approved):
I think this is reasonable and makes sense so approving + the other PR that followed but would be interested in engineering's perspective.

aflaxman (approved):

PR #1271, opened 2023-07-20
CGF correlation data links

alibow:


aflaxman (approved):
negative PAFs are weird

alibow:
Yeah, I believe GBD will have them too, though! Because the relative risks for incidence are very close to 1, 5% of draws could be less than one and it would still be statistically significant. For a dichotomous risk tractor, all of the draws with a relative risk less than one would also have a negative PAF. The population_rate * (1 - PAF) * RR equation will still return the values it should.

SylLutze (approved):

PR #1290, opened 2023-07-27
Update docs to show that employer_id remains unnoised

pletale:
JIRA ticket: https://jira.ihme.washington.edu/browse/SSCI-1425 
Please lmk if instead of 'for PRL tracking', I should be more clear and say 'employer_id is an un-noised truth-deck column' or something! Or if I misunderstood this ticket - thanks! 

Related SE PR: https://github.com/ihmeuw/pseudopeople/pull/159

### DIFF
--- a/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
@@ -2583,7 +2583,7 @@ W2 and 1099 Forms
   * - Mailing Address ZIP Code
   * - Social Security Number
   * - Wages (income from this job)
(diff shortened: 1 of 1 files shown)


zmbc (approved):
👍 I think we might want to revisit how we communicate the concept of "ground truth" in the future but for now it's good to be consistent with what we already do!

pletale:
> 👍 I think we might want to revisit how we communicate the concept of "ground truth" in the future but for now it's good to be consistent with what we already do!

Ok cool. Going to update the pseudopeople docs now to

NathanielBlairStahn (approved):
Looks good, thanks Alix!

PR #1294, opened 2023-07-31
PRL: Add middle name and event ID to SSA columns

NathanielBlairStahn:
JIRA tickets: [SSCI-1464](https://jira.ihme.washington.edu/browse/SSCI-1464), [SSCI-1395](https://jira.ihme.washington.edu/browse/SSCI-1395)

Makes some changes to the Social Security observer:
- Change middle initial to middle name
- Add an event ID column for comparing noised to unnoised data
- Edit noise table to reflect the change from middle initial to middle name in SSA dataset

The change from middle initial to middle name is based on these [archival NUMIDENT files from NARA](https://aad.archives.gov/aad/series-description.jsp?s=5057).

I haven't yet added these changes to the pseudopeople docs since I don't know whether they'll be included in the yellow release.

### DIFF
--- a/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
@@ -3107,7 +3107,13 @@ for all column based noise include:
     - Missing data, nicknames, fake names, phonetic, OCR, typographic
     -
   * - Middle Initial
-    - Census, Household Surveys, WIC, Taxes (both), SSA
+    - Census, Household Surveys, WIC, Taxes (both)
+    - 0.01
+    - 0.1
+    - Missing data, phonetic, OCR, typographic
+    -
+  * - Middle Name
+    - SSA
     - 0.01
     - 0.1
     - Missing data, phonetic, OCR, typographic
--- a/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
@@ -2956,13 +2956,14 @@ added later (not in the minimum viable model), if desired.
 
   * - Unique simulant ID (for PRL tracking)
   * - First name
-  * - Middle initial
+  * - Middle name
   * - Last name
   * - DOB (stored as a string in YYYYMMDD format, as indicated by [CARRA_SSA]_ Table 1)
   * - Sex (binary; "Male" or "Female")
   * - Social Security Number
   * - Type of event
   * - Date of event (stored as a string in YYYYMMDD format, as indicated by [CARRA_SSA]_ Table 1)
+  * - Event ID (unique integer identifier for each row in the SSA dataset, representing a ground-truth identifier for the event recorded in that row; unaffected by noise functions; to be used for comparing noised and unnoised data)
 
 .. note::
   Unlike the other observers, there is no ground-truth unique household ID for PRL tracking in this observer.
--- a/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
@@ -3116,8 +3116,8 @@ for all column based noise include:
     - SSA
     - 0.01
     - 0.1
-    - Missing data, phonetic, OCR, typographic
-    -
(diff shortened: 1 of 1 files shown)


docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2956,13 +2956,14 @@ added later (not in the minimum viable model), if desired.
...
   * - First name
-  * - Middle initial
+  * - Middle name
   * - Last name
   * - DOB (stored as a string in YYYYMMDD format, as indicated by [CARRA_SSA]_ Table 1)
   * - Sex (binary; "Male" or "Female")
   * - Social Security Number
   * - Type of event
   * - Date of event (stored as a string in YYYYMMDD format, as indicated by [CARRA_SSA]_ Table 1)
+  * - Event ID (unique integer identifier for each row in the SSA dataset, representing a ground-truth identifier for the event recorded in that row; unaffected by noise functions; to be used for comparing noised and unnoised data)
```
zmbc, line 2966: We probably don't need to require that this be an integer; I know for e.g. simulant ID we have an underscore included to make it easier to do this across shards.

zmbc (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
NathanielBlairStahn: Well, I **want** these all to be changed to integers because strings make the data bloated and slow everything down, so I figured I'd pre-empt the problem by specifying what I want in the concept model. It's just as easy to have a shard prefix with integers as with strings.

zmbc: Fair enough! I think this does mean that the engineers will need to hard-code a maximum number of of simulants per shard in order to "zero-pad" the simulant ID part (otherwise shard 12, simulant 84 and shard 1, simulant 284 would have the same simulant_id). Let's see what their reaction to this is.

zmbc: In this case, it would be the event ID not the simulant ID that gets zero-padded, which is even harder to hard-code a maximum for. Maybe they'd flip it and zero-pad the shard in the low-order digits. Or maybe I'm missing some obviously better solution?

pletale (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
NathanielBlairStahn: With 64-bit ints we have a lot of wiggle room. First note that 2**64 > 1.8e19. Suppose we set unreasonably high maximums of 100 million simulants per shard and 10,000 SSA events per simulant. That's a maximum of 1e8 * 1e4 = 1e12 events for each shard of SSA data (that's 1 trillion SSA events per shard!). That still leaves room for more than 1.8e19 / 1e12 = 1.8e7 shards (i.e., over 18 million shards), which should be way more than sufficient. So I think it shouldn't be too hard to set reasonable pad lengths to have well-defined event IDs across shards, regardless of whether we put shards or events in the low-order digits.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3106,11 +3107,17 @@ for all column based noise include:
...
     - 0.01
     - 0.1
     - Missing data, phonetic, OCR, typographic
     -
+  * - Middle Name
+    - SSA
+    - 0.01
+    - 0.1
+    - Missing data, nicknames, fake names, phonetic, OCR, typographic
+    - Use the list of fake first names for middle names as well
```
NathanielBlairStahn, line 3120: @zmbc and @pletale:

@albrja [asked in Slack](https://ihme.slack.com/archives/C02KUQ9LX32/p1690914431727189) whether we wanted to apply all the noise types from first name to middle name as well, namely nicknames and fake names. I thought this made sense, so I included it here. Running it by the two of you since you already approved the PR without the two additional noise types. Any objections?

zmbc: Sounds good to me!

PR #1296, opened 2023-08-02
Add year columns to datasets in concept model

pletale:
Add 'year' or 'tax_year' to the list of column names for each dataset. 
Jira ticket: https://jira.ihme.washington.edu/browse/SSCI-1465

### DIFF
--- a/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
@@ -2005,7 +2005,6 @@ Census
   * - First name
   * - Middle initial
   * - Last name
-  * - Year
   * - Age (floored to integer years **before** noise is applied)
   * - Date of Birth (stored as a string in MM/DD/YYYY format)
   * - Physical Address Street Number
@@ -2020,6 +2019,7 @@ Census
   * - Race/Ethnicity
   * - Tracked Guardian(s) (for noise functions ONLY)
   * - Tracked Guardian Address(es) (for noise functions ONLY)
+  * - Year
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -2005,6 +2005,7 @@ Census
   * - First name
   * - Middle initial
   * - Last name
+  * - Year
```
zmbc, line 2008: As I noted in the pseudopeople PR, let's check the order of the columns here. I think we'd want this at the beginning (maybe before even the unique IDs?)

zmbc (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
NathanielBlairStahn: As I noted in the [pseudopeople PR](https://github.com/ihmeuw/pseudopeople/pull/241), year/tax_year is currently the _last_ column in the datasets in which it appears. Do we like that, or do we want to move it to the beginning?

Are the column orders always consistent between the concept model, pseudopeople docs, and the actual data? I'm not sure...

NathanielBlairStahn: @pletale As noted in the pseudopeople PR, let's list `year` / `tax_year` as the **last** column in the relevant datasets, to be consistent with what's already implemented.

pletale: @zmbc @NathanielBlairStahn Just moved the year columns to the last column in the datasets (Census, WIC, taxes)!

zmbc:
@pletale is there no year column in the survey observer?

zmbc (approved):

NathanielBlairStahn (approved):

PR #1297, opened 2023-08-04
Add note about pregnancy adjustment factor random seed

alibow:


SylLutze (approved):

zmbc (commented):
Does this mean that we won't actually have any parameter uncertainty for this parameter (it won't vary between draws)? If so, you might consider listing the exact numeric value of the parameter, to ensure that engineers actually use the same thing. If not, I don't understand what's going on here 😄

alibow:
@zmbc, no we will still have parameter uncertainty for this parameter! I am requesting that we use a set random seed to generate that parameter uncertainty so that I can predict what the exact value used in a given draw will be (it's not stored in the artifact, so I can't directly observe it), which ends up reducing noise in V&V and let's me calculate some additional relevant values that depend on the pregnancy adjustment factor values rather than asking the engineers to do it.

alibow:
I think ideally we would just store _all_ parameters in the artifact and then we wouldn't need to specify the random seed in this way, but it solves the problem until we make that standard practice.

zmbc:
> I am requesting that we use a set random seed to generate that parameter uncertainty so that I can predict what the exact value used in a given draw will be

Interesting, I didn't think about this random seed for generating draws when we have a non-GBD parameter! I'm assuming that the engineers already have a standard seed that they use for this, so this is more about setting that explicitly for this project so we can rely on it 👍

patricktnast (approved):

PR #1298, opened 2023-08-08
Add description of employer name generation to concept model

pletale:
This PR adds the description that Abie wrote up to describe how he and Jim produced the employer names/business name bigrams to the PRL concept model.


docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -4283,12 +4283,20 @@ Limitations
...
+was treated as a node, and we included special <start> and <end> nodes, as well. 
+We included a directed multi-edge for each occurrence of a word pair in sequence 
+in each location name. To generate simulated employer names, we performed a random
+walk through the bigram graph. Starting from the start node, we traversed 
+directed edges selected uniformly at random until we reached the end node or 
+exceeded a predetermined maximum path length. We then combined the words associated 
+with each node that was encountered along the path to form the simulated employer name.
+This approach resulted in a diverse range of names that maintained a realistic quality.
+Businesses can have the same name in real life but if/when we do have duplicate names 
+we need to prevent duplicates for the names of the largest employers.
```
zmbc, line 4299: I think we need to check with Abie whether we ended up doing deduplication of the business name list or not. If so, we can delete this sentence; if not, we should move and reword it into a limitation.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -4283,12 +4283,20 @@ Limitations
 
 Speaking of the names for employers, we will generate names for
 employers based on a conditional random model that Abie developed and
-Jim refactored into the vivarium model.  It should have more details
-documented at some point.  Jim will ensure that all employer names are
-unique for now, although it would be interesting one day to add some
-duplicates.  Businesses can have the same name in real life but
-if/when we do have duplicate names we need to prevent duplicates for
-the names of the largest employers.
+Jim refactored into the vivarium model. We based our simulated employer names
```
zmbc, line 4286: Not sure if it's out of scope for this PR, but we should move this -- I think the end of the "Initialization" section above would be the right place. We can also edit it to be more like the rest of our documentation (e.g. no need for a transition like "Speaking of..." and we don't need to say _who_ did this part).

There is also a TODO above about "Document how we name businesses" which should be removed.

NathanielBlairStahn: > and we don't need to say _who_ did this part

Is there any harm in saying who did something? Seems like that could potentially be useful information.

zmbc: I guess there's no harm in it; I think this document aspires to be something external facing and in order to fully get there it would have to explain who these people are instead of just using their first names. But we can leave the names as they are for now, since it's certainly not making anything worse.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
aflaxman: I don't know what we ended up doing for this.  Jim should be able to tell us, though.

aflaxman (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -4203,6 +4199,25 @@ military.
...
+in each location name. 
+
+To generate simulated employer names, we performed a random
+walk through the bigram graph. Starting from the start node, we traversed 
+directed edges selected uniformly at random until we reached the end node or 
+exceeded a predetermined maximum path length. We then combined the words associated 
+with each node that was encountered along the path to form the simulated employer name.
+This approach resulted in a diverse range of names that maintained a realistic quality.
+Businesses can have the same name in real life but if/when we do have duplicate names 
+we need to prevent duplicates for the names of the largest employers.
```
zmbc, line 4218:
```suggestion
However, it also resulted in some duplicates.
Businesses can have the same name in real life, but they are more likely to choose unique
names so that customers can tell them apart.
To approximate this, we split the generated business names into 15 groups, and removed
duplicate names within each group; therefore, no single name could be used by more than
15 businesses.

Lastly, we removed business names that actually existed in the SafeGraph dataset
and were used in that dataset fewer than 1,000 times, to avoid businesses in our
simulation having real names by coincidence.
```

zmbc (approved):
I've suggested how to incorporate the additional detail discussed in [this Slack thread](https://ihme.slack.com/archives/C02KUQ9LX32/p1691523392975599?thread_ts=1688659855.753799&cid=C02KUQ9LX32)

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
pletale: Ah this looks great - thanks Zeb!

PR #1302, opened 2023-08-14
Reflect negative ages after perturbation

zmbc:
Move from clipping (in household age perturbation) and truncating (in individual age perturbation) to using reflection in both cases.


aflaxman (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -4406,13 +4406,27 @@ from international immigration in a household move, the following steps are **al
    and standard deviation 1 year.
 #. The age shift is added to the age values of all individuals in the household.
 #. If any age values in the household exceed 99 years, they are set to 99 years.
```
NathanielBlairStahn, line 4408: @zmbc I'm just curious why we are clipping large ages to 99.

NathanielBlairStahn (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: The ACS PUMS clips large ages to 99, so by doing the same we stay within its range of ages. In practice this mattered because we had no names data for birth years corresponding to ages >110. I guess this clipping will lead to some DOB bunching but there aren't many 99 year olds anyway, so it's probably not worth changing this. I'll add a note though.

albrja (approved):

PR #1308, opened 2023-08-16
Update background YLD strategy after discussing with Patrick

alibow:
FYI to interested research team members, we are finally implementing a test version of background morbidity in the nutrition optimization pregnancy model. This version will be implemented as a custom cause model before we approach integrating into default vivarium behavior.

@patricktnast , there is a slight change to what we discussed with regard to the disability weight equation - we're scaling the maternal disorders YLDs to the rate of pregnancies rather than the rate of non-fatal maternal disorders. Let me know if you have any questions/concerns!


docs/source/models/causes/other_causes/index.rst
```diff
@@ -13,27 +13,15 @@ This cause model document is meant to represent residual morbidity and mortality
...
+For a discussion of the importance of modeling background morbidity, see the :ref:`background morbidity page <other_causes_ylds>`. 
 
 Vivarium Modeling Strategy
 --------------------------
 
 Scope
 +++++
 
-This cause model document is reliant on a list of modeled causes to be specified in a concept model when this cause is included in a given vivarium simulation.
+This cause model document existing as an intermim solution to these issues specific discussed in the :ref:`background morbidity page <other_causes_ylds>`, specifically designed for the :ref:`nutrition optimization pregnancy model <2021_concept_model_vivarium_nutrition_optimization_pregnancies>`.
```
aflaxman, line 24:
```suggestion
This cause model document serves as an interim solution to these issues specific discussed in the :ref:`background morbidity page <other_causes_ylds>`, specifically designed for the :ref:`nutrition optimization pregnancy model <2021_concept_model_vivarium_nutrition_optimization_pregnancies>`.
```

docs/source/models/causes/other_causes/index.rst
```diff
@@ -82,29 +70,23 @@ States Data
...
 
-.. todo::
-  
-  Confirm with the engineers that it will be possible for risk exposures to modify the "disability weight" as stated above 
+  Note that we include YLDs due to maternal disorders in the disability weight of "background morbidity" here despite the fact that maternal disorders is included as a cause of disability within the nutrition optimization pregnancy simulation. This is because in the simulation, simulants acquire pre-COMO-adjusted maternal disorders YLDs over the course of a single timestep in which accumulation of YLDs due to all other causes besides maternal disorders is paused. Therefore, since maternal disorders YLDs are not accumulated during the remainder of the simulation when anemia YLDs are accumulated, anemia YLDs would not be adjusted for comorbidity with maternal disorders unless they were included in the disability weight for background morbidity, which will be present during the timesteps for which anemia YLDs are accumulated
```
aflaxman, line 84: I don't understand this!  Let's talk it through.

aflaxman (approved):

docs/source/models/causes/other_causes/index.rst
SylLutze: I am also confused about this and would appreciate chatting through it.

SylLutze (approved):

docs/source/models/causes/other_causes/index.rst
zmbc, line 24:
```suggestion
This cause model document serves as an interim solution to the issues discussed in the :ref:`background morbidity page <other_causes_ylds>`, specifically designed for the :ref:`nutrition optimization pregnancy model <2021_concept_model_vivarium_nutrition_optimization_pregnancies>`.
```

docs/source/models/causes/other_causes/index.rst
```diff
@@ -82,29 +70,23 @@ States Data
...
+     - 
    * - C
      - disabilty weight
-     - :math:`YLDs_{c294} - \sum_{cid}^{n} YLDs_{cid}`
-     - Where :math:`\sum_{cid}^{n} YLDs_{cid}` is the sum of YLDs due to all specified modeled components (may be causes, sequelae, and/or impairments). List of such IDs must be provided in concept model document.
-   * - ALL
-     - cause specific mortality rate
-     - :math:`ACMR - \sum_{cid}^{n} CSMR_{cid}`
-     - Where :math:`\sum_{cid}^{n} CSMR_{cid}` is the sum of the CSMRs of modeled causes. List of modeled cause IDs must be provided in concept model document.
+     - :math:`YLDs_{c294} - YLDs_{r192} - (YLDs_{c366} - YLDs_{s182,s183,s184}) + \frac{YLD_{c366} - YLDs_{\text{s182,s183,s184}}}{\text{preg_rate}}`
```
zmbc, line 77:
```suggestion
     - :math:`YLDs_{c294} - YLDs_{r192} - (YLDs_{c366} - YLDs_{\text{s182,s183,s184}}) * (1 + \frac{1}{\text{preg_rate}})`
```

Note: I'm just mechanically simplifying here. I don't actually understand this equation. I would have guessed that the equation would be this:

```
YLDs_{c294} - YLDs_{r192} - (YLDs_{c366} - YLDs_{\text{s182,s183,s184}}) / preg_rate
```

docs/source/models/causes/other_causes/index.rst
zmbc: I'm interested as well! I think I kind of follow this, but I'm unclear why it doesn't matter that we are double-counting maternal disorders YLDs overall. And also, it seems like you would have the reverse problem as well: that the maternal disorders YLDs accumulated while everything is paused are not adjusted for those other things.

docs/source/models/concept_models/vivarium_nutrition_optimization/pregnancies/concept_model.rst
```diff
@@ -487,7 +487,7 @@ Specific outputs for specific models are specified in the following section.
     - None
     - * Deaths 
       * YLLs
-      * YLDs
+      * YLDs (including YLDs due to "all causes")
```
zmbc, line 490:
```suggestion
      * YLDs (including YLDs due to "other causes")
```

Or maybe "background morbidity" since you changed the title of that page?

patricktnast (approved):

docs/source/models/causes/other_causes/index.rst
alibow: 1. We will double-count maternal disorders in the estimate of the total YLD rate in the simulation. However, we only ever really evaluate YLDs _averted_ between scenarios and we will only avert maternal disorders YLDs that occur during the parturition state and the maternal disorders YLDs in the background morbidity will not change, so we will not double count YLDs averted which is really what we are trying to get right.
2. How we are modeling YLDs due to maternal disorders is by taking the already COMO adjusted YLD rate and converting that to a "disability weight" rather than calculating YLDs from a DW directly. Therefore, the maternal disorders YLDs are already COMO adjusted and don't need to be adjusted additionally.

I'll try to add some of this context to the doc to make this more clear!

docs/source/models/causes/other_causes/index.rst
alibow: I'll break down what's happening here: 
- YLDs_{c294} # start with population all cause YLD rate that we assume to generalize to the pregnant population
- - YLDs_{r192} # subtract anemia YLDs that we assume generalizes to the pregnant population
- - (YLDs_{c366} - YLDs_{\text{s182,s183,s184}}) # subtract off _population level_ maternal disorders YLD rate excluding maternal disorders YLD sequelae that overlap with anemia (we don't want to subtract off these sequelae twice!). We're subtracting the population level maternal disorders YLD rate and will add back the pregnancy-specific maternal disorder YLD rate in the next step
- + \frac{YLD_{c366} - YLDs_{\text{s182,s183,s184}}}{\text{preg_rate}} # now we are adding back in the maternal disorder YLD rate, but adjusted up to be specific to the pregnant population by dividing by the incidence rate of pregnancy 


Does this clarify what's happening here?

docs/source/models/concept_models/vivarium_nutrition_optimization/pregnancies/concept_model.rst
alibow: This is actually intentionally written as "all" causes because of the detail in the vivarium behavior for calculating all cause and cause-specific YLDs differently, as shown in this screenshot from this page: https://vivarium-research.readthedocs.io/en/latest/model_design/vivarium_features/other_causes_ylds/index.html

image

docs/source/models/causes/other_causes/index.rst
alibow: Basically s182,s183,s184 are anemia due to maternal hemorrhage, so these sequelae are included in both YLDs due to anemia impairment and YLDs due to the maternal disorders cause. We decided that we will count them as part of the anemia impairment rather than the maternal disorders cause, so we always subtract them off of YLDs_c366

docs/source/models/causes/other_causes/index.rst
zmbc: 1. Ah, I thought it was something like that. I assume we don't need to report anything like YLDs averted as a proportion of total YLDs? And do we report DALYs in this sim? It seems like we'd slightly change the balance between YLLs averted and YLDs averted here (i.e. averting a death isn't as good if people accumulate more YLDs per unit time). Probably no big deal even if so; just trying to understand the consequences of this double-count.
2. I see! So, we have the same slight limitation going both directions: our YLDs due to anemia aren't taking into account that we have averted some maternal disorders, and our YLDs due to maternal disorders aren't taking into account that we have averted some anemia. Have I got that right?

docs/source/models/causes/other_causes/index.rst
zmbc: Thanks! Your update makes this much more clear 👍

docs/source/models/concept_models/vivarium_nutrition_optimization/pregnancies/concept_model.rst
zmbc: Ah yep, I did not interpret it that way. This is better!

docs/source/models/causes/other_causes/index.rst
alibow: 1. Oh, good point, I hadn't fully though through that impacting the balance between YLLs and YLDs averted that you mention there. I am not sure if the (a) influence of COMO adjusting anemia YLDs for maternal disorders or (b) influence of this issue you just raised will be more influential on our results. For this specific model, simulants only live for 6 weeks past their averted death date (they can only die at birth) after spending ~35 weeks or so before birth while they can experience anemia. So I'm thinking that given there will be a pretty small period to accumulate YLDs following death that I would rather capture the COMO adjustment for anemia appropriately. Let me know if you disagree!
2. Yes, exactly!

I think I merged this prematurely to note these additional nuances! I will follow-up to get them included in the write-up.

PR #1313, opened 2023-08-25
Updates to pregnancy other YLD cause model

alibow:


aflaxman (approved):

SylLutze (approved):

docs/source/models/causes/other_causes/index.rst
```diff
@@ -74,16 +74,21 @@ States Data
...
+  Note that we are excluding YLDs due to maternal disorders in YLDs due to background morbidity, despite that YLDs due to maternal disorders in the nutrition optimization model are modeled separately and are never comorbid with other conditions. This limits us in that we therefore do not capture:
 
-    Note that while this allows us to more accurately estimate YLDs averted between scenarios, it will cause double counting of YLDs due to maternal disorders in the total amount of YLDs accrued within the simulation (due to all modeled causes combined).
+    1. any increase in maternal disorders YLDs due to reductions in anemia YLDs through the COMO adjustment,
```
zmbc, line 86: When you say "COMO adjustment" here, you are including the comorbidity calculations in Vivarium, I think? In my head "COMO" specifically meant the GBD process, so this took me a second.

docs/source/models/causes/other_causes/index.rst
```diff
@@ -74,16 +74,21 @@ States Data
...
+    2. any increase in anemia YLDs due to reductions in maternal disorders YLDs through the COMO adjustment, 
+    3. any increase in background morbidity YLDs due to reductions in maternal disorders, nor
+    4. COMO adjustments in YLDs due to anemia and other causes from comorbidity with maternal disorders.
```
zmbc, line 89: How is this different from (2)?

docs/source/models/causes/other_causes/index.rst
```diff
@@ -74,16 +74,21 @@ States Data
      - 
    * - C
      - disabilty weight
-     - all_cause_yld_rate - anemia_yld_rate - population_maternal_disorders_yld_rate + pregnancy_maternal_disorders_yld_rate
+     - all_cause_yld_rate - anemia_yld_rate - population_maternal_disorders_yld_rate
```
zmbc, line 77: Maybe I should have started here: what is the common-sense interpretation of continuing to subtract the population maternal disorders YLD rate, but not add in the pregnancy maternal disorders YLD rate?

docs/source/models/causes/other_causes/index.rst
```diff
@@ -74,16 +74,21 @@ States Data
      - 
    * - C
      - disabilty weight
-     - all_cause_yld_rate - anemia_yld_rate - population_maternal_disorders_yld_rate + pregnancy_maternal_disorders_yld_rate
+     - all_cause_yld_rate - anemia_yld_rate - population_maternal_disorders_yld_rate 
      - See extended note below
 
 .. note::
 
   Note that we are assuming the YLD rate due to all causes other than maternal disorders (including YLDs due to anemia) do not significantly vary between the pregnant/postpartum not pregnant/postpartum populations. We do rescale the rate of maternal disorders from the general population of women of reproductive age to the rate among the pregnant/postpartum population in this equation.
```
zmbc, line 82: The "We do rescale..." sentence is no longer true, I think.

zmbc (commented):
If Abie and Syl understand this, it is probably okay, but purely from reading these docs it isn't totally clear to me.

docs/source/models/causes/other_causes/index.rst
alibow: From our test runs, I realized that the majority of the impact that the inclusion of background morbidity had on YLDs averted in our sim was through the pathway of anemia YLDs averted causing increases in "other" YLDs by reducing comorbidities between them. This pathway was greater in magnitude than the effect of "other" YLDs reducing the number of anemia YLDs at baseline (only about a 1% reduction) AND greater than the pathway of accumulating YLDs due to other causes after an averted death (we only model 6 weeks after averted deaths). While anemia comorbidity only adjusts the other causes DW by ~1%, the magnitude of YLDs due to other causes is large, so it ends up being sizable relative to total YLDs averted between scenarios (~8%). 

Given that the maternal disorders YLD rate among the pregnant population is a significant portion of YLDs due to other causes as we had it written in the previous version, it would be overly conservative to apply increases associated with comorbid anemia YLD reductions to the full baseline amount of maternal disorders YLDs included in the other causes YLDs because this value should be reduced between scenarios in a way that we would not be modeling (because we only reduce the actual maternal disorders YLDs, not the ones implicitly included in background morbidity). So essentially, there would be a modeled 1% increase due to the reduction in comorbid anemia, but some unmodeled decrease as well. 

Removing maternal disorders YLDs from background morbidity removes this limitation from the largest pathway of influence on YLDs averted due to the inclusion of background morbidity in our model, although remains limited in other ways.

docs/source/models/causes/other_causes/index.rst
alibow: Correct! Thanks for catching.

docs/source/models/causes/other_causes/index.rst
alibow: I have been thinking of two separate but definitely related impacts of our application of the "COMO" adjustment, including:
- Reductions in baseline cause-specific YLD rates due to comorbidities
- Ripple effects of increased cause-specific YLD rates between scenarios when comorbid cause-specific YLDs are reduced between scenarios

I am not sure if this is a helpful distinction even when more clearly articulated. Do you think it's worth keeping these as separate concepts with more explanation or should I condense in the list?

docs/source/models/causes/other_causes/index.rst
alibow: Yeah, I have been using "COMO" loosely here as basically our application of the GBD COMO process. I'll add a note that indicates that.

docs/source/models/causes/other_causes/index.rst
zmbc: Thanks for the explanation! I understand what this equation is saying now. In a way, we've gone back to the "default" approach to background morbidity that we'd be likely to use in other models -- we simply subtract out all the modeled causes from the all cause YLD rate. We're now ignoring the fact that we model maternal disorders in a special, unusual way, which I agree makes sense given your investigations into what is influential.

> This pathway was greater in magnitude than the effect of "other" YLDs reducing the number of anemia YLDs at baseline (only about a 1% reduction)

Does this make intuitive sense to you? I'm confused how this could be, since I assume that background morbidity YLDs are much larger than anemia YLDs, and it seems like we'd be almost multiplying anemia YLDs by (1 - background DW).

docs/source/models/causes/other_causes/index.rst
zmbc: I think I might be starting to see what you're saying, but it's still a little fuzzy. Is this a list of 4 differences between the equation before and after this PR, between the equation after this PR and not including background morbidity at all, between the equation after this PR and using the same equation but with maternal disorders modeled in a more typical way spread out over time, or between the equation after this PR and explicitly modeling every GBD cause (the ideal comorbidity adjustment)? Now that I read it over again, I think it is the third one, but I didn't get that from my initial read of the sentences before this list.

If I'm right about that, I think the distinction you're talking about can be useful, but it would be good to follow each item all the way through to what effect it has on YLDs averted. And, I'm still not sure which is which in the list.

docs/source/models/causes/other_causes/index.rst
alibow: Oddly, I've realized that the COMO adjustment does not depends on the magnitude of DW of both comorbid causes in ways that don't feel particularly intuitive. The COMO adjustment affects the magnitude of all comorbid causes equally and the influence is small because anemia DWs are small and it is fairly insensitive to the magnitude of the other causes DW.

I don't mean to open a can of worms in this comment thread, so let's plan to talk through all of this (maybe in a doc session?) when I'm back from being OOO!

alibow:
Merging, but noting that we will follow-up with additional documentation related to
- potential for a second cause model for maternal disorders YLDs that are comorbid with other causes along with an update not to allow simulants to change age groups so that we can predict how maternal disorders YLDs change at the beginning of the sim
- Update the wording around the different pathways by which background morbidity is affecting our estimates of YLDs averted between scenarios on this document and expand upon them on the main page as well
- Add additional context around age, the importance of the magnitude of the other causes DW (in light of the above point), and hopefully some visual explainers as well

PR #1345, opened 2023-09-26
Add numeric miswriting to noises available for age

pletale:
Jira ticket: https://jira.ihme.washington.edu/browse/SSCI-1396 

Is it worth updating the names in this table to how they've been implemented in pseudopeople (e.g. change 'numeric miswriting' to 'write the wrong digits')?


zmbc:
> Is it worth updating the names in this table to how they've been implemented in pseudopeople (e.g. change 'numeric miswriting' to 'write the wrong digits')?

My vote would be yes, but it doesn't need to be done in this PR, if you think it would muddy things.

zmbc (approved):

NathanielBlairStahn (approved):

NathanielBlairStahn:
> > Is it worth updating the names in this table to how they've been implemented in pseudopeople (e.g. change 'numeric miswriting' to 'write the wrong digits')?
> 
> My vote would be yes, but it doesn't need to be done in this PR, if you think it would muddy things.

Yeah, It's probably a good idea to change everything, but let's do it in a separate PR -- there are quite a few places where all the names would need to be changed. I guess we could update one name at a time to keep it more manageable if that seems easier.

aflaxman (approved):

aflaxman:
Maybe we should just remove these docs from vivarium_research, since they need to be included in pseudopeople, and having them in two places makes it easy to introduce inconsistencies.

pletale:
> Maybe we should just remove these docs from vivarium_research, since they need to be included in pseudopeople, and having them in two places makes it easy to introduce inconsistencies.

I'm happy to make these updates to the concept model docs, or like Abie said, maybe these docs don't need to be on vivarium_research at all. Another option could be to add a note in the concept model docs that some names are outdated and readers can find up-to-date names in the pseudopeople docs. What do you all think would be best?

NathanielBlairStahn:
> > Maybe we should just remove these docs from vivarium_research, since they need to be included in pseudopeople, and having them in two places makes it easy to introduce inconsistencies.
> 
> I'm happy to make these updates to the concept model docs, or like Abie said, maybe these docs don't need to be on vivarium_research at all. Another option could be to add a note in the concept model docs that some names are outdated and readers can find up-to-date names in the pseudopeople docs. What do you all think would be best? (@zmbc @NathanielBlairStahn @aflaxman)

We have a To-do somewhere that says we should move all the noise function documentation over to the pseudopeople docs, but I'm not sure that it's a good idea to remove this documentation completely, for a few reasons:

1. Sometimes we need to provide more details to the engineers than we want to specify in the pseudopeople docs, e.g., here: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#additional-noise-implementation-details
2. Sometimes it has been useful to look back at the concept model to see what our original intention was or to see what exactly we told the engineers to do (e.g., the order of the noise functions, and having a complete noise table with all the noise functions documented in one place, even if they haven't been implemented)
3. It can be useful to have a separate place to document new features before they're implemented, though perhaps we could address that problem with separate git branches in pseudopeople (although this might be hard to do with the noise table as indicated in my previous point)

I agree that it would be nice to have a single source of truth, but I'm wary of wiping out the noise documentation in the concept model entirely unless we spend some time spelling out a plan to address the above issues and other potential challenges.

I think for now, adding a note to the concept model that the noise type names are different in pseudopeople is a good idea. In fact, we should spell out the mapping of concept model names to pseudopeople names.

pletale:
image

This is how the map of noise function names renders in ReadtheDocs! (-:

zmbc (approved):

NathanielBlairStahn (approved):

PR #1346, opened 2023-09-26
Updates to mediation docs

SylLutze:
Updates based on Steve's comments - fixing links, edges cases, some missed language. Note that I am planning to remake the HF deltas to be draw level data in the near future.


docs/source/models/risk_effects/bmi/index.rst
```diff
@@ -106,6 +106,22 @@ BMI will impact the incidence rates of: ischemic heart disease, ischemic
...
+Mediation
+---------
+
+Mediation is included for BMI through SBP, LDL-C and FPG. We have generally 
+followed the GBD approach to mediation, however we use slightly different 
+equations based on math in the word doc below. 
+
+:download:`Please see this word doc for details of the new math included <Mediation Notes.docx>`.
+
+In cases where the RR for BMI is 1 or all mediators have an RR of 1, mediation
```
stevebachmeier, line 118: Sorry for being pedantic, but I want to be clear here. What you've written here is indeed true (if all mediators or the risk itself has RR=1, no adjustment is made). However, this check is made per-RR. That is, if a given mediator has RR=1, then there is no adjustment made due to that mediator - but other mediators may indeed adjust the rates.
- BMI RR = 1 --> no adjustment at all
- all mediator RR = 1 --> no adjustment at all
- RR = 1 for mediator X --> no adjustment _from mediator X_ (but other mediators with RR>1 may adjust the rates)

Is this all what you were thinking?

SylLutze: Yes - added more information and an example for clarity.

aflaxman (approved):

PR #1352, opened 2023-10-03
Use `build.os` for RTD

zmbc:
Deprecation: https://blog.readthedocs.com/use-build-os-config/

Docs: https://docs.readthedocs.io/en/stable/config-file/v2.html


NathanielBlairStahn (approved):
I haven't read enough to know for sure whether this will work, but approving anyway.

aflaxman (approved):

PR #1353, opened 2023-10-04
Revamping CVD therapeutic inertia

SylLutze:


zmbc (commented):
Cool!

Am I correct in thinking that the "churn" parameter is key here? I'm assuming if churn was 0, then the percent medicated wouldn't stabilize until it hit 100%. I'm asking because if it was influential, it would be nice to have a bit more explanation of how you got 5%.

I'm also not sure what the significance is of 5 in the expression `sims['original_sbp'] - (5*sims['meds'])`

This line is also a bit confusing, maybe a typo? `inertia = 0.20 ## Means 25%...`

SylLutze:
Hey @zmbc thanks for looking through this! To answer your questions: 

1. Yes, without "churn" we would always get to 100% over time. In real life, people on medication die and new people become eligible so we have to have something that mimics that here. The 5% is somewhat arbitrary. It's a hard thing to try to quantify so what I did was pick a reasonable ish number and then do some sensitivity testing. I basically found that while having churn is important, we got reasonable nano sim outputs with churn set between 1-10% which I think covers what we might see in the model. I'll add this to the notebook! 
2. The 5 in that line is an average decrease in SBP level by moving up 1 medication category (so if you're at category 3 medication, you would see a 15 point decrease in your SBP level). It's an oversimplification - your decrease in SBP depends on your starting SBP among other things, but this is a reasonable approximation. I'll add this to the model notes as well. 
3. Yes, I again played with inertia values for some sensitivity testing and it ended with the numbers misaligned. But this is confusing as written and I'll fix.

zmbc (approved):
Thanks for explaining! I'm curious how close to the target medication level you will get in the sim -- you'll be able to check, right?

I think in the future it would be interesting to try to use an interactive sim for this kind of task!

SylLutze:
Yep we'll be able to check! I'm also curious how well this works when we implement it in the main simulation. A good check on how accurate nanosims are with the various simplifications and the tradeoffs on doing this in the interactive sim (more difficult to code than a nanosim and slower) vs the simplified version. We'll see how it goes!

alibow (approved):
Looks good to me, but I'm wondering if it would be good to include an explanation of where/why the 80/84/87% values for those who will not start medication came from -- I think I remember these sorts of changes being hard to follow at earlier stages, so just thinking could be nice to have a record of the decision directly in the docs!

docs/source/models/concept_models/vivarium_us_cvd/concept_model.rst
```diff
@@ -508,6 +508,42 @@ Notes:
     - If simulant is eligible, either 8.55%, 50% or 100% depending on scenario 
     - 
 
+.. _uscvd4.3.1:
+
+4.3.1 Therapeutic Inertia Propensity
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Therapeutic inertia is a complex combination of simulant level characteristics and
```
aflaxman, line 516:
```suggestion
Therapeutic inertia is a complex combination of simulant-level characteristics and 
```

docs/source/models/concept_models/vivarium_us_cvd/concept_model.rst
```diff
@@ -508,6 +508,42 @@ Notes:
...
 
+.. _uscvd4.3.1:
+
+4.3.1 Therapeutic Inertia Propensity
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Therapeutic inertia is a complex combination of simulant level characteristics and 
+random variation. Therefore, we want each individual simulant’s propensity 
+to be autocorrelated between healthcare visits, but we don’t want this autocorrelation 
+to be 1. To model this, we use a combination of a random component which is
```
aflaxman, line 519:
```suggestion
to be 100%. To model this, we use a combination of a random component which is 
```

docs/source/models/concept_models/vivarium_us_cvd/concept_model.rst
```diff
@@ -508,6 +508,42 @@ Notes:
...
+The sum of the two components has a standard normal distribution, which means 
+that the standard normal’s CDF of that sum is uniformly distributed between 
+0 and 1. This can be seen visually in the notebook below. 
+
+This final simulant inertia propensity will be compared against the cutoffs in the 
+healthcare modeling section to determine if a simulant overcomes inertia. 
+
+The math for this is modeled after :ref:`income in the PRL work <census_prl_income>`. The variances 
+were found via `nanosim testing <https://github.com/ihmeuw/vivarium_research_nih_us_cvd/blob/main/therapeutic_inertia_nanosim.ipynb>`_. To derive these values, a percent of eligible 
+people on medication was found from NHANES and then different variances were plugged into
```
aflaxman, line 545:
```suggestion
people on medication was found from NHANES and then we plugged different variances into 
```

docs/source/models/concept_models/vivarium_us_cvd/concept_model.rst
```diff
@@ -508,6 +508,42 @@ Notes:
...
+that the standard normal’s CDF of that sum is uniformly distributed between 
+0 and 1. This can be seen visually in the notebook below. 
+
+This final simulant inertia propensity will be compared against the cutoffs in the 
+healthcare modeling section to determine if a simulant overcomes inertia. 
+
+The math for this is modeled after :ref:`income in the PRL work <census_prl_income>`. The variances 
+were found via `nanosim testing <https://github.com/ihmeuw/vivarium_research_nih_us_cvd/blob/main/therapeutic_inertia_nanosim.ipynb>`_. To derive these values, a percent of eligible 
+people on medication was found from NHANES and then different variances were plugged into 
+the nanosim until the desired medication level was achieved in equilibrium.
```
aflaxman, line 546:
```suggestion
the nanosim until we achieved the desired medication level in equilibrium. 
```

aflaxman (approved):

PR #1355, opened 2023-10-04
V&V for Pregnancy Model 10

SylLutze:


hussain-jafari (approved):

aflaxman (approved):
Super!

alibow:
image
Hmm I am confused by universal BEP being different from the others here... also surprised that zero coverage is so different from baseline but the other scenarios are so close to it... I would have expected baseline to be more in the middle. I want to look into this a bit more before moving on!

alibow:
Oh, oops I think I just got confused by the overlapping lines... I think I now see that only the red line is really close to baseline, as expected, and the other scenarios are layered underneath the universal BEP rather than it being all out on it's own.

So never mind!! This is looking as I expected after all and think we're good to move forward :)

alibow (approved):

PR #1356, opened 2023-10-04
Add summary of SQLNS meta-analysis task to intervention page

pletale:
Adding a brief summary of the SQ-LNS meta-analysis task here. Please let me know if I should include any additional details! 

Slack thread for reference: https://ihme.slack.com/archives/C02944X688L/p1693343529972519


docs/source/models/intervention_models/lipid_based_nutrient_supplements/index.rst
```diff
@@ -291,6 +291,12 @@ The SQ-LNS effects on wasting transition rates should apply to simulants covered
     - 1
     - Apply this effect to both r1 and t1 transition rates
 
+.. note::
+  
+  We reviewed the individual studies included in [Dewey-et-al-2021b]_ to check if any evaluated the effects of SQ-LNS on wasting incidence rather than prevalence ([Dewey-et-al-2021b]_ focused on prevalence). We found that the RCTs rarely reported on incidence, but were still able to confirm that those that did had similar findings as our back-calculated incidence values (e.g., [Huybregts-et-al-2019-sqlns]_).
```
zmbc, line 296:
```suggestion
  We reviewed the individual studies included in [Dewey-et-al-2021b]_ to check if any evaluated the effects of SQ-LNS on wasting incidence rather than prevalence ([Dewey-et-al-2021b]_ focused on prevalence). Few RCTs reported on incidence, but those that did had similar findings as our back-calculated incidence values (e.g., [Huybregts-et-al-2019-sqlns]_).
```

Just saving a few words!

zmbc (approved):
Looks good to me, though I haven't been following this very closely. It might be nice to include a link/file path to where we did this work.

alibow (changes requested):
Thanks Alix! I think it would be worth adding a few things:
1) The effect sizes that Huybregts reported. I also just pulled up the paper and it looks like table 6 shows that there is not a significant difference in the length of treatment between the SQ-LNS group and non-SQ-LNS group, which we'll use as a proxy for recovery! Basically this confirms our assumption that SQ-LNS works via reducing incidence rather than speeding recovery, which is great.
2) The fact that Becquey study had lower/null incidence rate effects than ours, but that they also had low/null effects on prevalence and recovery as well, so we believe it not to be representative based on what we know on overall SQLNS effects from Dewey.
3) A link to somewhere where we might have compiled the study PDFs/any other info in case we want to revisit this for any reason in the future we will have a head start

alibow (approved):
Awesome, looks great! Thanks Alix!

PR #1366, opened 2023-10-16
Notes on GBD 2021 data needs

alibow:
As requested by Rajan - let me know if there is anything else you were hoping to see here that I have not included. Thanks!


alibow:
> This looks good! For the ones listed as "ready", is it possible will they undergo further changes or these are set?

It sounds like they are final according to this timeline:  https://hub.ihme.washington.edu/pages/viewpage.action?spaceKey=GBD2020&title=GBD+2021+Computation. It is possible that they undergo edits in the submission process for capstone papers, but I think we'll just have to cross that bridge when we come to it. I just added a date to this PR so that we can update status if/when changes come up.

PR #1370, opened 2023-10-18
Update BLISS glossary definitions

pletale:
Couple questions: 
- Should I add 'secondary sex characteristics' to this glossary? Not sure it's a great phase but I currently use it in the definition of 'sex and gender-related factors'.
- Should we change 'proxy measure' to 'sex/gender proxy measure' to make it specific to this style guide? 
- Are there any other helpful terms that we're missing here that should be added?


alibow (approved):
I think this looks good and that adding secondary characteristics would be valuable, but I think keeping proxy measure as a general definition rather than sex-specific is adequate here too.

docs/source/model_design/general_reference_material/bliss/index.rst
```diff
@@ -163,30 +163,37 @@ throughout this style guide.
...
+       in an external data source (e.g., vague survey question that different respondents may interpret differently) 
+       or from the compilation of many data sources with clear definitions of 'sex' or 'gender' into one 
+       conglomerate sex/gender variable (e.g., GBD "sex" variable).
    * - Sex- and gender-related factors
-     - "Factors associated with sex/gender that are not themselves dimensions of
-       sex or gender (e.g., gene expression, body weight, risk taking, age at
-       sexual debut, structural sexism)" [Bauer_2022]_. 
+     - An umbrella term that refers to any and all characteristics related to sex and/or gender (e.g., 
+       gamete size, hormone production, reproductive or sexual anatomy, or any so-called 'secondary sex
+       characeristics', like body hair, breast tissue, or body mass index).
```
SylLutze, line 176: I liked in the old list having one item that was "gender norm" not biological - risk taking. Could we add that here as well?

SylLutze (approved):
Agree with Ali on secondary sex characteristics and proxy measure!

docs/source/model_design/general_reference_material/bliss/index.rst
```diff
@@ -163,30 +163,37 @@ throughout this style guide.
...
+       or from the compilation of many data sources with clear definitions of 'sex' or 'gender' into one 
+       conglomerate sex/gender variable (e.g., GBD "sex" variable).
    * - Sex- and gender-related factors
-     - "Factors associated with sex/gender that are not themselves dimensions of
-       sex or gender (e.g., gene expression, body weight, risk taking, age at
-       sexual debut, structural sexism)" [Bauer_2022]_. 
+     - An umbrella term that refers to any and all characteristics related to sex and/or gender (e.g., 
+       gamete size, hormone production, reproductive or sexual anatomy, or any so-called 'secondary sex
+       characeristics', like body hair, breast tissue, or body mass index). 
+   * - Proxy measure
```
aflaxman, line 177: Keeping this general as "proxy measure" seems slightly better to me, but I am fine with you making it the more specific  "sex/gender proxy measure" if you prefer.

aflaxman (approved):

docs/source/model_design/general_reference_material/bliss/index.rst
zmbc: Yes, I think most of these would be considered sex-related factors (though the distinction can be a little fuzzy) so it'd be good to include some clearly gender-related ones.

I've made a suggestion here but feel free to rework as seems right to you!

zmbc, line 176:
```suggestion
     - An umbrella term that refers to any and all characteristics related to sex and/or gender (e.g., 
       gamete size, hormone production, reproductive or sexual anatomy, so-called 'secondary sex
       characteristics' such as body hair or breast tissue, gender identity, gender expression, risk-taking behavior, etc). 
```

zmbc (approved):

PR #1374, opened 2023-10-20
Adding outlines for 2 impairment pages

SylLutze:
I struggled with the title for the GBD impairment/cause/health states/sequelae page so feel free to suggest something better!


zmbc (approved):
I don't know too much about this topic, but this seems like a helpful way to frame it!

aflaxman (approved):

alibow (approved):

PR #1377, opened 2023-10-23
Edit Vivarium components index

NathanielBlairStahn:
While working on merging the "Interventions critical features" page into the "Modeling Interventions" page, I thought the index page pointing to all the "Modeling X" pages could be a bit clearer, so here's a pull request with some proposed changes. My main goal is to make it easier for someone to find the pages they're looking for.

1. I changed the page title from "Types of Vivarium model components" to "How to model Vivarium components" to match the fact that all the subpage titles have the form "Modeling X", indicating that they are describing how to model something in Vivarium. However, I see that Syl's PR (https://github.com/ihmeuw/vivarium_research/pull/1374) is adding a new subpage that doesn't fit this pattern. If we don't like the "How to" framing, I suggest we keep the original title but change all the subpage titles from "Modeling X" to just "X".
2. I combined the two existing tables of contents into a single TOC to make the nested structure of the risk factor subpages more immediately clear.
3. I added a new TOC to the `templates` folder, since if you're going to start modeling a component, you'll probably want to start there after you've looked through the relevant "Modeling X" page.


zmbc (commented):
I think this is good overall, but I don't love the title change. "How to model Vivarium components" feels confusing to me, because the Vivarium components _are_ the model. It's almost closer to "How to model GBD components" except that we try to be more general than GBD, and interventions are included.

I think I prefer your suggestion of renaming the pages from "Modeling X" to just "X".

Definitely open to being persuaded, or to other ideas!

alibow:
> I think this is good overall, but I don't love the title change. "How to model Vivarium components" feels confusing to me, because the Vivarium components _are_ the model. It's almost closer to "How to model GBD components" except that we try to be more general than GBD, and interventions are included.
> 
> I think I prefer your suggestion of renaming the pages from "Modeling X" to just "X".
> 
> Definitely open to being persuaded, or to other ideas!

@zmbc what do you think of "How to design vivarium components"? I think I might prefer something wordier like: "Overview of vivarium component types and notes for how to design/model them"

zmbc:
@alibow I think "types" (or similar) is key to include. I do definitely prefer "design" to "model" as the verb!

One stab at shortening your wordier version: "Types of Vivarium components and how to design them"

NathanielBlairStahn:
I finally got around to updating this PR. I used Zeb's suggested title for the index page, and I changed the titles of subpages from "Modeling X" to "X".

My editor replaced a bunch of Windows line endings (CRLF) with Mac/Linux line endings (LF) on all the subpages, so there are a lot of extraneous whitespace changes. To see only the important changes, you should select the 'Hide whitespace' option in the settings on the diff page, as shown here:

https://stackoverflow.com/questions/37007300/how-to-ignore-whitespace-in-github-when-comparing

zmbc (approved):

SylLutze (approved):

NathanielBlairStahn:
JIRA ticket: https://jira.ihme.washington.edu/browse/SSCI-1609

PR #1378, opened 2023-10-25
PRL: Update omit row noise

NathanielBlairStahn:
JIRA ticket: [SSCI-1540](https://jira.ihme.washington.edu/browse/SSCI-1540)

Update PRL concept model to allow applying "omit a row" (simple omission) noise to census and surveys. Previously, we only allowed "do not respond" (targeted omission) noise for these, but there's no reason not to allow both.

I also documented the default row probabilities for "omit a row" noise, and I took the opportunity to change the requested default value for WIC from 0.0% to 0.5% (same as what we have for W2/1099), since it seems plausible that WIC could have some missing records. If anyone has a different preference for the default WIC value (or any of the other default values), let me know.

(Note that there are some extraneous whitespace changes that you may want to hide.)

### DIFF
--- a/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
@@ -3686,28 +3686,21 @@ We will define two types of omission:
    where differential non-response rates are built into the observer
 
 Note that the census and household surveys can have both types of
-ommission applied. The following two sections discuss the two types of
-omission in more detail.
-
-For the census and household surveys, where non-response is built
-into the observer, the default omission rate will the equal to the
+ommission applied. The default rate of "simple omission" for these
+datasets will be 0%, while the default non-response rate will equal the
 expected omission rate of that survey. For the census and ACS survey,
-this is 1.45%. For the CPS survey, this is 29.05%. The other observers’
-default rate is 0%.
+this is 1.45%. For the CPS survey, this is 29.05% (see details below).
 
-The process for omitting simulants from the data will be divided into
-two groups based on observer:
-
-#. Simple omission will be used for: WIC, taxes and SSA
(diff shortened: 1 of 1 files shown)


docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3656,38 +3680,83 @@ into three buckets:
 
 It is assumed that in any dataset, there are people missed. This
 is important in PRL as the user will try to find a match for someone
-that doesn’t exist. For simplicity, we will have pseudo-people end
-users input one value: the overall omission rate. This value can
+that doesn’t exist. For simplicity, we will have pseudopeople end
+users input one value: the overall omission rate (called :code:`row_probability` in pseudopeople). This value can
```
zmbc, line 3684: Should this say "one value for each omission type" or something like that?

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3656,38 +3680,83 @@ into three buckets:
...
-this is 1.45%. For the CPS survey, this is 29.05%. The other observers’
-default rate is 0%.
+#. **Targeted omission** (called :code:`do_not_respond` in
+   pseudopeople), applicable only to the census and household surveys,
+   where differential non-response rates are built into the observer
 
-The process for omitting simulants from the data will be divided into
-two groups based on observer:
+Note that the census and household surveys can have both types of
+ommission applied; in this case targeted omission (do not respond)
```
zmbc, line 3698:
```suggestion
omission applied; in this case targeted omission (do not respond)
```

zmbc (approved):

aflaxman (approved):

PR #1380, opened 2023-10-27
Updating seed count to use for the draw size runs

SylLutze:


stevebachmeier (approved):

alibow (approved):
I think it can be worth showing the plots you made by draw (separate line per for each draw) in addition to summarized across draws. If you are not presenting draw-level statistics though like we do for NO then it's probably less important

PR #1382, opened 2023-10-30
Document a Bayesian approach to automated V&V

zmbc:
Using Bayesian instead of frequentist hypothesis testing.

Code implementing the statistics and applying this method to domestic migration and immigration is here: https://github.com/ihmeuw/vivarium_census_prl_synth_pop/pull/333


docs/source/model_design/vivarium_features/automated_v_and_v/index.rst
```diff
@@ -121,12 +121,13 @@ The difficulty of this problem is part of why, in the manual V&V process, we usu
...
-For example, we could specify that the simulation's true value should be within 10% of the GBD estimate, which means it is within 10% **not including stochastic variation.**
+For validation, we specify a target 95% uncertainty interval (UI), within which we expect the simulation's **true** value (i.e. the value of the simulation result as the simulated population size goes to infinity) should fall 95% of the time.
+For example, we could specify that the UI of the simulation's true value is +/-10% of the GBD estimate, which means it is 95% certain to be within 10% **as the population size goes to infinity.**
 
-We have begun to formalize fuzzy checking using frequentist hypothesis tests,
+We have begun to formalize fuzzy checking using Bayesian hypothesis tests,
 one for each of the values we want to check in the simulation.
-In these hypothesis tests, the null hypothesis is that the simulation value matches the V&V target;
+In these hypothesis tests, the null hypothesis is that the simulation value comes from our V&V target distribution
+and the alternative hypothesis is that it comes from a prior distribution of bugs/errors;
```
aflaxman, line 130: there is something philosophically interesting here... the alternative hypothesis is that the prior has bugs/errors _and they matter_.  It is possible that there is a bug but it is not caught by this test.  But then is it really a bug?

aflaxman (approved):

SylLutze (approved):

docs/source/model_design/vivarium_features/automated_v_and_v/index.rst
zmbc: Abie, that might be (arguably) the alternative hypothesis we _want_, but here I am describing what alternative hypothesis we are actually _testing_. With how I have currently done this, there is a distribution of rates if there is no bug (specified by the V&V target) and a distribution of rates if there is a bug (currently this prior is always the same). The latter _can_ have mass around or at the correct values, which represents the situation you are describing -- a bug that is accidentally right. We still include that as part of the alternative hypothesis.

alibow (approved):

PR #1389, opened 2023-11-06
Add more about how to interpret hypotheses

zmbc:
This is a little bit philosophical but comes out of good practical questions on Friday about how to think about setting V&V targets. It is also along the lines of the thread Abie started [here](https://github.com/ihmeuw/vivarium_research/pull/1382#discussion_r1376853391).

Note that this PR is on top of #1382; don't review this if you haven't already reviewed that!


pletale (approved):
This was an interesting and easy-to-follow philosophical read! (:

SylLutze (approved):
Interesting! I liked the quote you pulled a lot.

zmbc:
Thanks both for reviewing! FYI to anyone else who looks at this PR, I think we will have a chance to talk about it in real time during the RT meeting on Thursday. I'll leave it open at least until then and we can see what feedback comes up.

PR #1391, opened 2023-11-08
Proposal for default rate to probability calculation update in vivarium

alibow:
This topic has come up as we have been discussing how to determine timesteps for the variable simulation clock in the nutrition optimization model and it's something that we'd like to have a full team review on! I have taken the background./rate adjustment section from this existing page: https://vivarium-research.readthedocs.io/en/latest/model_design/designing_vivarium_model/choosing_appropriate_time_step/index.html, but have essentially just reformatted this into a formal proposal rather than some background information.

Notably, we have already successfully tested the equations used here in a simulation by applying the r->r' calculation for the remission rate of diarrheal diseases and LRI into the artifact, which solved the underestimation of these rates in our simulation prior to implementing this adjustment.


alibow:
> I haven't yet reviewed this closely, but it seems significantly more complicated than I would expect. It looks like the proposal here gets a probability that always equals `r * dt` (the rate times the timestep), but _that_ is the approximation that isn't quite right.
> 
> I was expecting us to say that Vivarium should use `probability = 1 - e^{-r \times dt}` which I believe to be exactly correct.
> 
> Am I missing something?

So I believe that vivarium already uses the `probability = 1 - e^{-r \times dt}` equation! See this line in the vivarium code: https://github.com/ihmeuw/vivarium/blob/710b7344661257a898cbaf80b20cee39569bfffe/src/vivarium/framework/utilities.py#L26 (I believe that the rate here gets rescaled to the timestep prior to this being called). 

To be honest, I don't feel like I understand the theory behind this approximation as well as I could beyond the testing that we've implemented, but @NathanielBlairStahn will likely be able to provide additional context here.

alibow:
A quick note -- I was not aware of that 250 cap on the rate until pulling up that line of code. This very well could be why we are underestimating the mortality rates of the high mortality LBWSG categories.. the RRs for some of these categories if as high as 1000, so very likely we would be truncating the rates and likely need to decrease our timestep by a lot to avoid this -- FYI @NathanielBlairStahn, @rmudambi -- it would be great if we could get a warning if we were running into that 250 cap.

https://github.com/ihmeuw/vivarium/blob/710b7344661257a898cbaf80b20cee39569bfffe/src/vivarium/framework/utilities.py#L26

zmbc:
> So I believe that vivarium already uses the `probability = 1 - e^{-r \times dt}` equation!

If you're right about that code, Vivarium already calculates the _events_ in the "correct" cumulative way, that is, adjusting for the fact that they could have happened at any point _during_ the last timestep. If I understand correctly, the problem is that it calculates person-time assuming that the events always happen at the end of the timestep; these are inconsistent with each other and lead to bogus rates in our results.

I'm not sure the path forward here is obvious. This proposal is to make the number of events _more_ different from what we would see in real life/with infinitely small timesteps, to bring it into alignment with the person-time -- and I think is equivalent to saying that the probability is simply `r * dt`. You could also imagine doing the reverse and changing how we record person-time -- essentially formalizing that when an event happens to a simulant in between timestep checkpoints, they enter a limbo state where they are not at risk for any event (in that state diagram). But that might have negative effects on e.g. prevalence estimation.

zmbc:
> A quick note -- I was not aware of that 250 cap on the rate until pulling up that line of code.

Agreed that a warning would be nice, but if I'm interpreting it correctly, the 250 cap is on the rate of events per _person-timestep_ -- if we are hitting that cap, the timestep is way too long and this would cause plenty of other verification issues.

zmbc:
> You could also imagine doing the reverse and changing how we record person-time -- essentially formalizing that when an event happens to a simulant in between timestep checkpoints, they enter a limbo state where they are not at risk for any event (in that state diagram). But that might have negative effects on e.g. prevalence estimation.

To expand a bit on how we could update person-time, imagine if our person-time observers stratified both by state, and by whether or not the simulant was eligible for events. If a simulant did not transition during a timestep, they are recorded with <timestep> person time in their current state & eligible for events. If a simulant did transition during a timestep, we calculate a precise time of event, and assign the time before that to their old state & eligible for events, and the time since that to their new state & ineligible. Rates should verify exactly, if we subset to eligible person-time. Other calculations could be done across all person time (eligible and ineligible). If our timesteps are appropriate, ineligible person-time should be a minority, both overall and within each state.

NathanielBlairStahn:
Here's the PR where we discussed this before: https://github.com/ihmeuw/vivarium_research/pull/793

alibow:
Ok, I agree with Zeb's alternative suggestion and also am noting that we can simplify what we currently have in this PR to just use `r * dt` rather than going through `r'` as written (if we stick with this strategy).

I agree that the person time observers would be a better solution that what is proposed here, but I am unsure about the scope of work and getting that integrated into the nutrition optimization project on our deadline to allow us to get our production runs through. I'm curious to know what @rmudambi thinks.

If person time observer changes are indeed too much for us at the moment, I then wonder if we think this approach, which I am now understanding as a bit of a hack to be worthwhile for use in the nutrition optimization sim or not. 

Basically, if we are to follow the implied necessary ratio of `dt` to `1/r` for the NO sim (1/10 to get within 5% as shown in those figures) without the rate adjustment, we won't really end up increasing our timestep overall by much... maybe from 4 to 5 days (although we will have improvements in accuracy by lowering it for groups that would benefit from that). Maybe that's worth it on it's own, but we would not gain any run time advantages, which was our goal in pursuing this strategy in the first place.

alibow:
@zmbc @NathanielBlairStahn let me know if you agree, but I am thinking that this may end up being worthwhile for us to implement in the nutrition optimization child sim, but not default vivarium behavior and we can follow-up with a proposal to modify the default behavior of the person-time observers in the future.

zmbc:
@alibow That sounds reasonable; I think the biggest question for me is how much this is a V&V issue versus an actual results issue. We know that the discrepancy between the events (numerator) and person-time (denominator) will cause the sim not to verify to the GBD rate, but is this showing up in the substantive findings of the simulation? If not, we may be making our simulation a little bit less accurate to the real world in order to make it pass V&V.

alibow:
> @alibow That sounds reasonable; I think the biggest question for me is how much this is a V&V issue versus an actual results issue. We know that the discrepancy between the events (numerator) and person-time (denominator) will cause the sim not to verify to the GBD rate, but is this showing up in the substantive findings of the simulation? If not, we may be making our simulation a little bit less accurate to the real world in order to make it pass V&V.

Yeah, I've been trying to think through this too. I think in the case of deaths, then we might not want to do an adjustment by that logic -- here we really care about the number of deaths, and if we slightly overestimate person-time for those at risk of death and then therefore slightly underestimate the mortality rate, that's better than overestimating the number of deaths via the adjustment that allows us to get the rate correct.

I think where the results vs. V&V issue is a little more complicated is in the case of wasting, where we care about both the numerator (number of transitions) and the denominator (person-time in each wasting state). If we increase our timestep and do not do a rate transition, we can expect to overestimate person-time in the source state of a given wasting transition, which will make it look like we are underestimating the transition rates and will also potentially cause us to be off in our overall wasting exposure distribution relative to expectations/GBD (we will bias towards overestimating whatever state(s) has the highest rates out of it, I believe). The question then is does that actually matter... I see a couple places where it could:

- Mortality implications: If we overestimate prevalence of the SAM state (for example), will we also overestimate deaths that occur in the SAM state? I think because we assume that only one transition per component can happen in a given timestep that the answer is no. In this case, I think this is more of a "our observers are preventing us from properly seeing that our simulation is functioning correctly" issue rather than an issue with the behavior of the simulation
- YLD implications: This one I think remains a problem. The SAM and MAM states have YLDs associated with them via the protein energy malnutrition PAF-of-one cause. If we increase our timesteps and therefore end up overestimating person time spent in these states, I believe we will also overestimate YLDs due to PEM. This issue would also apply to our cause models.
- Simulation outcomes: which matters more: incident wasting counts or wasting prevalence? I think this question is informative in deciding which path is best. If we were measuring wasting prevalence as an outcome in the emulator, I think we might want to keep the rate adjustment to make sure we get this as accurate as possible. But we care more about incident wasting cases, so I think this is another argument for no rate adjustment and just letting our V&V be "wrong" in some places.

So overall, I think it might be best to increase our timesteps without doing a rate adjustment, with the knowledge that how our person-time observers function currently, it will look like we are underestimating some rates and we may also be off in our prevalence estimates of transition models (hopefully in predictable ways and to a small degree). There are at least two major issues with this:
1. It makes V&V hard. How do we know if it's an expected issue from the timestep or some other bug? I suppose we could test with small timesteps first and then launch bigger runs with larger timesteps if everything looks good with our small timestep runs.
2. We will be biased in our YLDs. Among children YLDS<<YLLs, so this should be a minor limitation, but would be good to investigate/quantify further.

zmbc:
That all makes sense. Unless we ever directly report an event rate (a number that depends on relating # of events to person-time), I believe it is uniformly better not to change how we calculate the probability.

YLDs are basically a person-time result, which means that our YLDs will get more inaccurate as we increase the timestep (this is the case whether or not we change the event probability).

I hear you on this making V&V harder. Not sure I have any good ideas there.

NathanielBlairStahn:
The fundamental problem we're trying to solve is that Vivarium runs a discrete-time simulation that we are using to model continuous-time events. There at least two ways we can approach this:

1. Accept the fact that Vivarium has discrete time-steps, and try to design things so that we get good discrete approximations of the continuous-time quantities we're interested in;
2. Incorporate some continuous-time measures into Vivarium to try to automatically correct for the discrete-time deficiency and get direct estimates of continuous-time quantities.

My proposal is along the lines of (1), while Zeb's proposal is along the lines of (2). I think Zeb's idea would be cool if we (and the engineers) can figure out a good way to implement it, but it is a bit more complicated and will take more time to make sure we get it right. Currently, Vivarium sort of tries to do both (1) and (2), or neither, and ends up making bad approximations in some cases -- it seems to basically ignore the fact that there is a difference between continuous-time and discrete-time.

If we are (at least for now) going with strategy (1) rather than strategy (2), then I'm not sure the following two quotes are accurate assessments:

(@alibow's comment):

> I think in the case of deaths, then we might not want to do an adjustment by that logic -- here we really care about the number of deaths, and if we slightly overestimate person-time for those at risk of death and then therefore slightly underestimate the mortality rate, that's better than overestimating the number of deaths via the adjustment that allows us to get the rate correct.

(@zmbc's comment):

> Unless we ever directly report an event rate (a number that depends on relating # of events to person-time), I believe it is uniformly better not to change how we calculate the event probability. 

Here's a simple counterexample to the claim that we overestimate deaths if $p=r\Delta t$ and the claim that using $p = 1 - e^{-r\Delta t}$ is uniformly better if we're not looking at rates (hey look, now I can directly use some rudimentary $\LaTeX$ on GitHub! Was that always possible, or is it new??):

Suppose the mortality rate is $r$, and we observe a population for $K$ time-steps of length $\Delta t$, and we record the number of deaths $D$. For simplicity, assume that the population stays approximately constant, say $N$ (for example, imagine that $N$ is very large relative to $Kr \Delta t$, or that the population is being replenished at the same rate $r$). Then in continuous time, we have $D_\text{cont}$ ~ ${\rm Poisson}(N Kr \Delta t)$, and the expected number of deaths is $ED_\text{cont} = N Kr \Delta t$. In discrete time, if the probability a simulant dies on a given time step is $p$, then $D_\text{disc}$ ~ ${\rm Binomial}(NK, p)$, and $ED_\text{disc} = N K p$. Thus, if we set $p = r \Delta t$ as I suggested, then the expected number of deaths in the discrete-time approximation is $ED_\text{disc} = N Kr \Delta t = E D_\text{cont}$, matching the number in continuous-time. On the other hand, if we set $p = 1 - e^{-r\Delta t}$ as Vivarium currently does, the expected number of deaths in the discrete-time simulation is $ED_\text{disc} = N K (1 - e^{-r\Delta t}) < N Kr \Delta t = ED_\text{cont}$, so we systematically underestimate the number of deaths.

Does anyone see any flaws in that logic?

If not, I think we should just try making the very simple change of uniformly setting  $p=r\Delta t$ in Vivarium's `rate_to_probability` calculation instead of $p = 1 - e^{-r\Delta t}$, and see what happens compared to the current version. We can always change it back if it doesn't work. Though I guess we'd have to be careful when using the new variable-length time-steps since this only works if $r \Delta t<1$ -- it might make sense to go back to using $p = 1 - e^{-r\Delta t}$ when $r \Delta t$ gets too large, but I'll have to think more carefully about it.

zmbc:
@NathanielBlairStahn I fully agree with your overall split into (1) and (2). The person-time changes I proposed don't seem like a panacea, more like trying to shoehorn some continuous-time simulation into Vivarium, and they definitely have drawbacks. Apologies if that wasn't clear from my comments yesterday.

Re: your example with death rates, I think it gets at something interesting, but it depends quite crucially on the assumption that the population replenishes _exactly_ as it shrinks; in other words, your result is only true if a new simulant is added _exactly_ when each simulant dies. Otherwise, the deaths have an important impact on person-time spent at risk, which is precisely the thing at issue here.

Here's a version that I think illustrates the core of what you're getting at: imagine a state model (e.g. a cause model) with two states, A and B. The population is initialized 50/50 between the two, and the transition rates A -> B and B -> A are both $r$ (so, equal to each other). It's a fully closed cohort with no simulants exiting or entering. In continuous time, the expected total number of transition events is $E[T_{cont}] = NKr \Delta t$. If we use $p = r \Delta t$, $T_{disc} \sim Binomial(NK, r \Delta t)$ and $E[T_{disc}] = NKr \Delta t = E[T_{cont}]$. However if we use $p = 1 - e^{-r \Delta t}$, $T_{disc} \sim Binomial(NK, 1 - e^{-r \Delta t})$ and $E[T_{disc}] = NK(1 - e^{-r \Delta t}) < E[T_{cont}]$.

Intuitively, the interpretation here is: by using $p = 1 - e^{-r \Delta t}$, we are assuming a continuous time of event, but we only "catch up" to the event at the next timestep. As I described above, in the (implied) time between the event and the "catch up," the simulant is not eligible for any event in that state model. Because some of our person-time is therefore risk-less, we end up with fewer events.

However, if we have a fixed (not variable between simulants) timestep, @alibow's quote about overestimating deaths is still correct. This is because in this case the event we are interested in (deaths) is modified by the state in a state model (wasting) that does not have such nice, equal rates. Let's say that A -> B has a transition rate of $10 \times r$ while B -> A has a transition rate of $r$ (and also that we initialize the prevalence to already be at the steady-state, even including mortality, which I won't worry about here). A also has an RR of 10 on mortality. With a fixed timestep and $p = r \Delta t$, we will systematically overestimate the person-time spent in A, since keeping simulants in A until we "catch up" to the A -> B events will be _more_ biased than keeping simulants in B until we catch up to the B -> A events. Therefore, we overestimate the mortality.

That said, if we have a variable timestep between the states, and simulants in A have shorter timesteps by enough that the "catch up" bias is the same between A and B -- which I think means that $p_{A->B} = 10r \Delta t_A = p_{B->A} = r \Delta t_B$ -- we should no longer overestimate events. You can think of this as, relative to real life everything is shifted later (by waiting to catch up) but the amount of person-time in each state is balanced. As long as we start at the steady-state, the shift doesn't matter.

NathanielBlairStahn:
> @NathanielBlairStahn I fully agree with your overall split into (1) and (2). The person-time changes I proposed don't seem like a panacea, more like trying to shoehorn some continuous-time simulation into Vivarium, and they definitely have drawbacks. Apologies if that wasn't clear from my comments yesterday.

No, it was clear, and I like your idea! I'm just trying to make the distinction between (1) and (2) so that we can analyze each approach separately without conflating things.

> Re: your example with death rates, I think it gets at something interesting, but it depends quite crucially on the assumption that the population replenishes _exactly_ as it shrinks; in other words, your result is only true if a new simulant is added _exactly_ when each simulant dies. Otherwise, the deaths have an important impact on person-time spent at risk, which is precisely the thing at issue here.

My example should be exactly correct in the limit as $N\to \infty$, and therefore should be a good approximation for large $N$. If the population is not exactly constant, the math gets harder, but it's possible there is still an exact solution -- for example, maybe the death count is an inhomogeneous Poisson process with intensity measure equal to the total person-time of an interval, rather than being a homogeneous Poisson process with constant rate. I'd have to do more math to know for sure, but my intuition tells me that even without the simplifying assumption of constant population, the important qualitative features of the process should remain the same.

> Here's a version that I think illustrates the core of what you're getting at: imagine a state model (e.g. a cause model) with two states, A and B. The population is initialized 50/50 between the two, and the transition rates A -> B and B -> A are both r (so, equal to each other). It's a fully closed cohort with no simulants exiting or entering. In continuous time, the expected total number of transition events is E[Tcont]=NKrΔt. If we use p=rΔt, Tdisc∼Binomial(NK,rΔt) and E[Tdisc]=NKrΔt=E[Tcont]. However if we use p=1−e−rΔt, Tdisc∼Binomial(NK,1−e−rΔt) and E[Tdisc]=NK(1−e−rΔt)<E[Tcont].

Ok, so that's another example showing that we're currently underestimating event counts, right? At least under some simplifying assumptions. And I'm betting we can relax some of the assumptions and still get the same result.

I think the core of the problem is that in continuous time, in any time interval, there are _infinitely_ many chances for an event to happen (the number of events is Poisson, with unbounded support {0, 1, 2, ...}), whereas in discrete time, there is a maximum fixed number of events that can happen in any fixed time interval (the number of events is Binomial, with bounded support {0, 1, ..., K}). Since the Binomial distribution has zero probability of very large counts, in order to get the same mean as the Poisson, it has to have a higher probability of success on a single time step than the corresponding Poisson ($p = r\Delta t$ instead of $p = 1- e^{-r\Delta t}$). In my [comment in the other PR](https://github.com/ihmeuw/vivarium_research/pull/793#discussion_r818171257), I identified the same problem from another perspective by looking at the waiting time to an event instead of the event count: In the continuous case, the waiting time is exponential, which assigns a large probability near 0, whereas in discrete time, the waiting time is geometric, which assigns _zero_ probability near 0; thus, for the geometric waiting time to have the same mean as the exponential, it needs to assign more probability to small times near 1 and less probability to large times near $\infty$ than the exponential does.

Note that even if we implement Zeb's strategy (2), we still have the discrepancy that the number of events in discrete time is bounded whereas in continuous time it's not. I'm not sure what downstream effects that has, but I don't see a way to fix it within a discrete-time simulation unless we do something even more complicated like sample a Poisson for the number of events that occurred on the time step, and I don't know if that even makes sense.

> Intuitively, the interpretation here is: by using p=1−e−rΔt, we are assuming a continuous time of event, but we only "catch up" to the event at the next timestep. As I described above, in the (implied) time between the event and the "catch up," the simulant is not eligible for any event in that state model. Because some of our person-time is therefore risk-less, we end up with fewer events.

Yeah, I guess that's another way to think about it. Another way of thinking about increasing the event probability to $p= r\Delta t$ is that we're "stealing" probability from later times to make the event more likely to happen soon in order to make up for the fact that it couldn't happen during the interval $[0, \Delta t]$.

> However, if we have a fixed (not variable between simulants) timestep, @alibow's quote about overestimating deaths is still correct. This is because in this case the event we are interested in (deaths) is modified by the state in a state model (wasting) that does not have such nice, equal rates. Let's say that A -> B has a transition rate of 10×r while B -> A has a transition rate of r (and also that we initialize the prevalence to already be at the steady-state, even including mortality, which I won't worry about here). A also has an RR of 10 on mortality. With a fixed timestep and p=rΔt, we will systematically overestimate the person-time spent in A, since keeping simulants in A until we "catch up" to the A -> B events will be _more_ biased than keeping simulants in B until we catch up to the B -> A events. Therefore, we overestimate the mortality.
> 
> That said, if we have a variable timestep between the states, and simulants in A have shorter timesteps by enough that the "catch up" bias is the same between A and B -- which I think means that pA−>B=10rΔtA=pB−>A=rΔtB -- we should no longer overestimate events. You can think of this as, relative to real life everything is shifted later (by waiting to catch up) but the amount of person-time in each state is balanced. As long as we start at the steady-state, the shift doesn't matter.

I'm not convinced that having unequal rates will inherently make us overestimate person-time or deaths depending on state A or B. @zmbc Do you have a more precise argument for why this would happen, and under what assumptions?

What I would expect is that, if we use $p=r \Delta t$ to guarantee unbiased event counts, _and_ $\Delta t$ is chosen to be small enough that we are getting good approximations for the _highest_ rate (I'm not sure how to tell what is "small enough"), then everything should be approximately correct. I mean, we know the discrete-time model approaches the continuous-time model as $\Delta t\to 0$, so it should just be a question of how small the timestep needs to be. And I'm arguing that for a given size timestep, we should get better approximations using $p=r \Delta t$ for all rates instead of $p = 1- e^{-r\Delta t}$. On the other hand, if our goal is to have a larger timestep and still try to get things to validate, as @alibow mentioned in the meeting, then that seems like a slightly different problem, and maybe you'd have to make some specialized tweaks to the model to compensate for a timestep that's too large compared to the rate.

zmbc:
> What I would expect is that, if we use $p=r \Delta t$ to guarantee unbiased event counts, and $\Delta t$ is chosen to be small enough that we are getting good approximations for the highest rate (I'm not sure how to tell what is "small enough"), then everything should be approximately correct.

This is true for $p = 1- e^{-r\Delta t}$ as well -- the problem goes away as the timestep goes to 0, regardless of our rate-to-probability conversion. I'm more interested in:

> I'm arguing that for a given size timestep, we should get better approximations using $p=r \Delta t$ for all rates instead of $p = 1- e^{-r\Delta t}$.

Maybe you just mean we will _generally_ get better approximations, but if you are saying that our simulation results will _always_ be closer to the continuous-time truth, that isn't the case for death events in my example with "fast" state A and "slow" state B (you can think of these as "infected" and "susceptible" in the diarrhea example). Let's say our timestep is way too long (just to make the effect really stark), and the remission (A -> B) probability per person-timestep is $\approx 1$ with either rate-to-probability conversion. The incidence (B -> A) probability is still much less than 1, and it will clearly be higher with $p=r \Delta t$ than with $p = 1- e^{-r\Delta t}$. Since every incident case will lead (almost surely) to 1 timestep of person-time in the A state, we clearly substantially overestimate (relative to small $\Delta t$) the person-time in the A state, and there is even _more_ incidence when $p=r \Delta t$ than when $p = 1- e^{-r\Delta t}$. Person-time in A is time at higher risk for death, and again (as with incidence), using $p=r \Delta t$ makes the rate of death from state A higher per unit of person-time. So, we overestimate mortality with either rate-to-probability conversion, but the overestimate is worse with the discrete-time one.

Note this is arbitrary and if you reversed which state had the higher mortality, it wouldn't necessarily work out like this.

NathanielBlairStahn:
**tl;dr:** Person-time in each state is correct when we use $p=r\Delta t$, regardless of unbalanced rates. Don't believe me? Read the math section below.

Regarding @zmbc's reply:

> Maybe you just mean we will _generally_ get better approximations, but if you are saying that our simulation results will _always_ be closer to the continuous-time truth, that isn't the case for death events in my example with "fast" state A and "slow" state B (you can think of these as "infected" and "susceptible" in the diarrhea example).

No, my conjecture is that it will _always_ be better to use $p= r\Delta t$ instead of $p = 1-e^{-r\Delta t}$. Since our simulations are complicated, I don't have a _proof_, which is why I want to try the experiment of changing it and seeing what happens. But I _do_ have very strong evidence, some of which I provided above, and more of which I will provide below. Regarding your argument about overestimating person-time and therefore deaths:

> Since every incident case will lead (almost surely) to 1 timestep of person-time in the A state, we clearly substantially overestimate (relative to small Δt) the person-time in the A state, and there is even _more_ incidence when p=rΔt than when p=1−e−rΔt.

I don't follow your logic here -- it is definitely _not_ clear to me that we overestimate person-time in the A state. In fact, I'll prove below that we get the correct person time precisely by setting $p=r\Delta t$. Perhaps you are referring to your previous explanation: 

> With a fixed timestep and p=rΔt, we will systematically overestimate the person-time spent in A, since keeping simulants in A until we "catch up" to the A -> B events will be _more_ biased than keeping simulants in B until we catch up to the B -> A events. Therefore, we overestimate the mortality.

It sounds to me like maybe you're still conflating approaches (1) and (2): The above statement might be true if we use $p = 1-e^{-r\Delta t}$ _instead_ of $p=r\Delta t$, but it is not a problem inherent in discrete-time simulations. The problem with using $p = 1-e^{-r\Delta t}$ is precisely that you're always trying to "catch up" with the events you missed in $[0, \Delta t]$, and you never actually get there. If we increase the probability to $p=r\Delta t$ (which is _greater_ than the continuous-time probability of an event in $[0, \Delta t]$), then instead of interpreting an event that occurs at the discrete time $k\Delta t$ as a previous continuous-time event that we had to wait to observe, we can interpret it as a previous _or future_ continuous-time event that we moved to a nearby discrete timestep. Thus, while some simulants will have their person-time in state $A$ overestimated because their event would have happened earlier even though we're counting a full $\Delta t$ of person time, other simulants will have their person-time in state $A$ _underestimated_ because the event actually would have happened in the future. If we're lucky, maybe the overestimations and underestimations will cancel out.

Another way to think about it is that using a geometric random variable with $p = 1-e^{-r\Delta t}$ is like using the left-hand rectangle rule to approximate the CDF of the rate- $r$ exponential waiting time, whereas using $p=r\Delta t$ is more like using the midpoint rule. The left-hand rule is always biased low because the CDF is increasing, while the midpoint rule gives a better approximation because it's sometimes above and sometimes below. (This analogy isn't perfect because it's the _values_ of the CDFs that we're comparing, not the integrals, but it is visually compelling.)

Ok, but this is all just _philosophy_ (as my Real Analysis teacher used to say) and doesn't tell us anything conclusive. What we really need is...

# MATH!

(Cuts through all the bullshit.<sup>TM</sup>)

Let's take @zmbc's toy example above: Suppose we have a population with two states $A$ and $B$ (e.g., $A = \text{infected}$ and $B = \text{susceptible}$), and suppose $A\to B$ has a transition rate of $10r$ while $B\to A$ has a transition rate of $r$. For simplicity, we will assume no birth, death,  or migration, so we have a fixed population size, say $N$. (And don't bother telling me that's an unrealistic assumption, because that's how math works: You make simplifying assumptions until you can prove something, then later you try to relax your assumptions to generalize the result.)

Zeb claims that "With a fixed timestep and $p=r\Delta t$, we will systematically overestimate the person-time spent in $A$." To evaluate this claim, we need to calculate the person-time spent in $A$ in both continuous time and discrete time, and compare the answers. How can we approach this?

With our stated assumptions, we have a two-state [Markov chain](https://en.wikipedia.org/wiki/Markov_chain), and we can consider the [continuous-time chain](https://en.wikipedia.org/wiki/Continuous-time_Markov_chain) as a limit of [discrete-time chains](https://en.wikipedia.org/wiki/Discrete-time_Markov_chain) as $\Delta t\to 0$. This Markov chain has a [stationary distribution](https://en.wikipedia.org/wiki/Discrete-time_Markov_chain#Stationary_distributions), say $\pi^d$ for the the discrete chain and $\pi^c$ for the continuous chain, which can be represented as a row vector $\pi = \pmatrix{\pi_A& \pi_B}$ whose components are the steady-state prevalences in states $A$ and $B$. If we run the chain for a total time of $T = K\Delta t$ (where $K\in\mathbb{N}$) starting from the steady state $\pi^d$ (for discrete time) or $\pi^c$ (for continuous time), then the person-time spent in $A$ and $B$ is computed approximately as $NT\pi^d$ or $NT\pi^c$, respectively, and the approximation gets better as $NT\to\infty$. Thus, to compare the discrete-time person-time with the continuous-time person-time, we need to compute the stationary distributions $\pi^d$ and $\pi^c$.

Since we're arguing about the correct transition probabilities for the discrete-time chain, let's be general and let $p_{AB}$ be the probability of a transition $A\to B$, and let $p_{BA}$ be the probability of a transition $B\to A$. Then the transition matrix for the discrete-time chain is

$$P = \begin{pmatrix}1-p_{AB} & p_{AB}\\\p_{BA} & 1-p_{BA}\end{pmatrix}.$$

Here, the two rows contain the probabilities of transitioning _out_ of $A$ and $B$, respectively, and the two columns contain the probabilities of transitioning _into_ $A$ and $B$, respectively. The stationary distribution $\pi^d = \pmatrix{\pi^d_A& \pi^d_B}$ is the unique row vector satisfying

$$\pi^d = \pi^d P\quad\text{and}\quad \pi^d_A + \pi^d_B = 1.$$

Solving this system of linear equations, we get

$$\pi^d_A = \frac{p_{BA}}{p_{AB}+p_{BA}} \quad\text{and}\quad \pi^d_B = \frac{p_{AB}}{p_{AB}+p_{BA}}.$$

Now, for the continuous-time chain, let's again be general and let $r_{AB}$ and $r_{BA}$ denote the transition rates from $A$ to $B$ and from $B$ to $A$, respectively. We know that the discrete-time chain approaches the continuous time chain as $\Delta t\to 0$, and that when $\Delta t$ is very small, we have $p_{AB} \approx r_{AB} \Delta t$ and $p_{BA} \approx r_{BA} \Delta t$. Plugging in these approximate probabilities to the discrete-time steady state solution as a function of $\Delta t$, we get

$$
\pi^d_A(\Delta t) = \frac{r_{BA} \Delta t}{r_{AB} \Delta t+p_{BA} \Delta t}
= \frac{r_{BA}}{r_{AB}+p_{BA}}
\quad\text{and}\quad
\pi^d_B(\Delta t) = \frac{r_{AB}\Delta t}{r_{AB}\Delta t+r_{BA}\Delta t}
= \frac{r_{AB}}{r_{AB}+r_{BA}},
$$

which **ends up not depending on $\Delta t$ at all.** This tells us two things. First, the continuous-time stationary distribution $\pi^c = \pmatrix{\pi_A^c & \pi_B^c}$ must be given by

$$
\pi_A^c = \lim_{\Delta t\to 0} \pi^d_A(\Delta t) = \frac{r_{BA}}{r_{AB}+p_{BA}}
\quad\text{and}\quad
\pi_B^c = \lim_{\Delta t\to 0} \pi^d_B(\Delta t)  = \frac{r_{AB}}{r_{AB}+r_{BA}}.
$$

Second, for **any** $\Delta t$ such that $r_{AB}\Delta t<1$ and $r_{BA}\Delta t <1$, if we set $p_{AB} = r_{AB} \Delta t$ and $p_{BA} = r_{BA} \Delta t$ in the discrete time chain with time step $\Delta t$, then **the stationary distribution for the discrete-time chain _equals_ the stationary distribution for the continuous-time chain**, i.e., $\pi^d = \pi^c$. In particular, when we choose $p = r\Delta t$ for both rates, we'll have $NT \pi_A^d = NT \pi_A^c$, so the person-time spent in $A$ will be approximately same in the discrete-time chain as it is in the continuous-time chain. Therefore, the claim that "with $p=r\Delta t$, we will systematically overestimate the person-time spent in $A$," is false.

Q.E.D. Mike drop. BOOM!

zmbc:
> I'm not convinced that having unequal rates will inherently make us overestimate person-time or deaths depending on state A or B. Do you have a more precise argument for why this would happen, and under what assumptions?

First, just to reality check my intuition:

```pycon
>>> state_transition_sim(a_to_b_transition_rate=0.01, b_to_a_transition_rate=0.0005, a_mortality_rate=0.0001, b_mortality_rate=0.00001, timestep=1)
3403

>>> state_transition_sim(a_to_b_transition_rate=0.01, b_to_a_transition_rate=0.0005, a_mortality_rate=0.0001, b_mortality_rate=0.00001, timestep=50)
3402

>>> state_transition_sim(a_to_b_transition_rate=0.01, b_to_a_transition_rate=0.0005, a_mortality_rate=0.0001, b_mortality_rate=0.00001, timestep=100)
3436

>>> state_transition_sim(a_to_b_transition_rate=0.01, b_to_a_transition_rate=0.0005, a_mortality_rate=0.0001, b_mortality_rate=0.00001, timestep=200)
4297

>>> state_transition_sim(a_to_b_transition_rate=0.01, b_to_a_transition_rate=0.0005, a_mortality_rate=0.0001, b_mortality_rate=0.00001, timestep=500)
6373
```

<details>

```python
import pandas as pd
import numpy as np

CONTINUOUS_CONVERSION = lambda r, dt: 1 - np.exp(-r * dt)
DISCRETE_CONVERSION = lambda r, dt: r * dt
STARTING_POP_SIZE = 50_000

def state_transition_sim(
    a_to_b_transition_rate,
    b_to_a_transition_rate,
    a_mortality_rate,
    b_mortality_rate,
    timestep,
    rate_to_probability_conversion=DISCRETE_CONVERSION,
    burn_in_period=500,
):
    rng = np.random.default_rng(seed=1234)
    pop_table = pd.DataFrame({
        'state': ["a", "b"] * (STARTING_POP_SIZE // 2)
    })

    for mini_timestep in range(burn_in_period):
        a_index = pop_table[pop_table.state == 'a'].index
        b_index = pop_table[pop_table.state == 'b'].index
        to_b_index = a_index[rng.random(len(a_index)) < DISCRETE_CONVERSION(a_to_b_transition_rate, 1)]
        to_a_index = b_index[rng.random(len(b_index)) < DISCRETE_CONVERSION(b_to_a_transition_rate, 1)]
        to_die_index = a_index[rng.random(len(a_index)) < DISCRETE_CONVERSION(a_mortality_rate, 1)].union(
            b_index[rng.random(len(b_index)) < DISCRETE_CONVERSION(b_mortality_rate, 1)]
        )
        pop_table.loc[to_b_index, 'state'] = 'b'
        pop_table.loc[to_a_index, 'state'] = 'a'
        pop_table.loc[to_die_index, 'state'] = 'dead_burnin' # Don't count deaths during burn-in

    a_to_b_prob = rate_to_probability_conversion(a_to_b_transition_rate, timestep)
    b_to_a_prob = rate_to_probability_conversion(b_to_a_transition_rate, timestep)
    a_mortality_prob = rate_to_probability_conversion(a_mortality_rate, timestep)
    b_mortality_prob = rate_to_probability_conversion(b_mortality_rate, timestep)

    for mini_timestep in range(5_000):
        if (mini_timestep + 1) % timestep == 0:
            a_index = pop_table[pop_table.state == 'a'].index
            b_index = pop_table[pop_table.state == 'b'].index
            to_b_index = a_index[rng.random(len(a_index)) < a_to_b_prob]
            to_a_index = b_index[rng.random(len(b_index)) < b_to_a_prob]
            to_die_index = a_index[rng.random(len(a_index)) < a_mortality_prob].union(
                b_index[rng.random(len(b_index)) < b_mortality_prob]
            )
            pop_table.loc[to_b_index, 'state'] = 'b'
            pop_table.loc[to_a_index, 'state'] = 'a'
            pop_table.loc[to_die_index, 'state'] = 'dead'

    return (pop_table.state == 'dead').sum()
```

</details>

I don't have a more mathematical description handy because this gets into stuff I'm not too comfortable with with Markov chains etc. But intuitively, if you do a discrete-time simulation of a continuous-time phenomenon, you have to decide at time T what to do if an event (would have) occurred between T-1 and T. The $p = 1- e^{-r\Delta t}$ conversion Vivarium uses is equivalent to saying that between the event and T, the simulant was not at risk for _any_ event. The $p = r \Delta t$ alternative is equivalent to saying that between the event and T, the simulant was at risk for the same event again, which doesn't make a ton of sense. You can see this most obviously in the case where $\Delta t$ is long enough that $r \Delta t > 1$.

There is nothing magical about $p = r \Delta t$. It assigns extra risk to simulants who were at risk at the last timestep, and (like the other option) fails to assign risk to simulants who _should_ be at risk but haven't "caught up" yet to their transition into the at-risk state. These two effects happen to balance out when the rate into the state equals the rate out of the state, as in both of our examples; they don't balance in any other case.

alibow:
@NathanielBlairStahn @zmbc thanks to both of you for your thorough analysis! I believe I have followed along enough to believe in the p=r * dt update (or at least a test first!). I do not necessarily feel that I am the best person to update this PR with the appropriate documentation/justification for the change -- perhaps we can discuss this at an upcoming documentation session. 

I still have an outstanding question about how we determine what an appropriately small dt is relative to r. My hunch is that we might be able to get away with anything so long that r*dt < 1? Do either of you have any thoughts on this? If not I am thinking that we can run some tests to confirm if we don't have a solid theoretical/mathematical basis for that conclusion.

zmbc:
> My hunch is that we might be able to get away with anything so long that r*dt < 1?

I believe Nathaniel's result above implies that as long as _everything_ in the model can be seen as a transition event happening at a certain rate, then our results are right in expectation as long as r * dt is never >1. It would be great to prove that for the variable timestep case, otherwise I'm a bit nervous that might mess something up.

I think even keeping within that restriction, the timestep still changes our variance, so using a shorter timestep would increase our "sample size" like increasing the number of simulants does. There is probably a principled way to think about this decision but I'm not sure the best way to tackle it.

NathanielBlairStahn:
> > My hunch is that we might be able to get away with anything so long that r*dt < 1?
> 
> I believe Nathaniel's result above implies that as long as _everything_ in the model can be seen as a transition event happening at a certain rate, then our results are right in expectation as long as r * dt is never >1. It would be great to prove that for the variable timestep case, otherwise I'm a bit nervous that might mess something up.
> 
> I think even keeping within that restriction, the timestep still changes our variance, so using a shorter timestep would increase our "sample size" like increasing the number of simulants does. There is probably a principled way to think about this decision but I'm not sure the best way to tackle it.

I believe Zeb's analysis here is all accurate. With fixed timesteps, as long as all events in the model can be viewed as happening at some rate, and as long as $r\Delta t \le 1$ for _all_ rates in the model, everything should work out _in expectation_ as Zeb says. If $p=r\Delta t = 1$ for some rate $r$, then the events for that rate are no longer random, and they just happen deterministically once per timestep, so the variance is 0 in this case. As $\Delta t \to 0$, this deterministic process becomes "more and more random," with the "maximum randomness" being what we would get in a continuous-time process. Choosing $p=r\Delta t$ guarantees that the expected values of the things we care about match the continuous process, even when the discrete process is not very random (i.e., has relatively large timsteps and hence smaller variance).

I also don't know what exactly will happen when we move to variable timesteps, so yes, we should be careful in that case. I'm not sure I understand how variable timesteps will be implemented well enough to prove anything about it yet, but maybe if I get some clarity from the engineers, I can try to figure out if there will be anything that causes problems.

NathanielBlairStahn:
@alibow If I remember correctly, I believe that when Beatrix and Nicole designed the dynamic wasting model, they also converted rates to probabilities using $p=1-e^{-r\Delta t}$, so we should switch to using $p=r\Delta t$ for the wasting model as well as Vivarium. If you are seeing miscalibrated prevalences between the wasting states, I bet this would help at least somewhat.

alibow:
@NathanielBlairStahn yes, I've been thinking about this too! We've actually totally reworked the wasting transition rates since then, but do still rely on that conversion. Our wasting transition model has been validating fairly well so far, but it will be interesting to see if that changes as we increase the timestep, etc. Thanks for mentioning it and I think we'll plan to update these as well in our test run where we used p=r*dt.

PR #1394, opened 2023-11-13
Expand BLISS guide

Ironholds:
As discussed in the meeting, add brief discussion (and definition of) gender roles, and rearrange so that terms come after the terms they depend on to make sense.


zmbc (approved):
Looks good to me!

pletale (approved):
Thanks, Os!

SylLutze (approved):

PR #1395, opened 2023-11-13
Timestep rules

alibow:


zmbc (approved):
Do you have a rough guess how the fixed timestep vs variable timestep runtime will compare? As long as it doesn't get much longer with this variable timestep, this seems like a good, simple first pass to implement.

alibow:
@zmbc 
> Do you have a rough guess how the fixed timestep vs variable timestep runtime will compare? As long as it doesn't get much longer with this variable timestep, this seems like a good, simple first pass to implement.

I expect that this will be maybe 50% faster than the fixed timestep runs, because most of the person time will be in the 8 day category

SylLutze (approved):
Very interested to see if/how this affects results!

PR #1398, opened 2023-11-16
Adding GBD health and disease definition page

SylLutze:


docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -33,11 +33,117 @@ Understanding GBD's Disease and Health Categories
...
+
+Impairments
++++++++++++
+
+In some cases, a medical condition or sequela might have multiple causes. For example, heart failure. 
+It is caused by both IHD and HHD (and others!) and has sequelae associated with both causes. 
+We even learned above that it has the exact same health state in both causes. Heart failure 
+is an example of an impairment. 
+
+An impairment is a sequelae that shows up in 2 or more causes.
```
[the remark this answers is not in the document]
alibow: Ok yeah here is an example!

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -33,11 +33,117 @@ Understanding GBD's Disease and Health Categories
...
+    - See cause list for more details 
+  * - Cause List 
+    - The list of all causes in GBD. It is a mutually exclusive and collectively exhaustive list of hierarchical categories that does not ignore any cause of death. All death and disability within GBD can be found within the cause list. 
+    - 
+  * - Health State 
+    - The current state of health associated with a given cause and sequela.
+    - 
+  * - Impairment
+    - The symptoms of a disease, such as vision loss as a result of diabetes. Clinically the same as sequelae. 
+    - GBD calculates these very differently than sequelae despite them being clinically the same.
```
alibow, line 56: I think we could improve the definition here. I've been thinking of impairments as groups of sequelae -- is that accurate?

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -33,11 +33,117 @@ Understanding GBD's Disease and Health Categories
...
--------------------------------
+.. list-table:: GBD definitions of terms 
+  :widths: 15 15 15
+  :header-rows: 1
 
-- Showing what is a direct subset, what overlaps, what interacts but is independent 
-- Include some information on modeling implications as a result of this 
+  * - Term 
+    - Definition
+    - Notes
```
alibow, line 44: I think adding an example column to this table would be great!

alibow:
This is great, thanks Syl! I think having some sort of integrating visual diagram in addition to the cause hierarchies that you already have (basically something that shows how that interacts with health states and impairments) would be a really valuable addition.

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -33,11 +33,117 @@ Understanding GBD's Disease and Health Categories
...
+  * - Cause
+    - A cause of death or disability. Includes injuries, diseases, and conditions. 
+    - See cause list for more details 
+  * - Cause List 
+    - The list of all causes in GBD. It is a mutually exclusive and collectively exhaustive list of hierarchical categories that does not ignore any cause of death. All death and disability within GBD can be found within the cause list. 
+    - 
+  * - Health State 
+    - The current state of health associated with a given cause and sequela.
```
alibow, line 52: Are health states specific to sequela? I think I'm stuck on "cause and sequela" here. Also, is it possible to have multiple health states for a given sequela? For instance -- here is a sequela in the uterine fibroids cause that seems like it has multiple health states (mild abdominal pain AND mild anemia): mild_abdominal_pain_due_to_uterine_fibroids_with_mild_anemia

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -33,11 +33,117 @@ Understanding GBD's Disease and Health Categories
...
+    - 
+  * - Impairment
+    - The symptoms of a disease, such as vision loss as a result of diabetes. Clinically the same as sequelae. 
+    - GBD calculates these very differently than sequelae despite them being clinically the same. 
+  * - Sequela or Sequelae 
+    - The medical conditions that can occur among people who contract a disease or suffer an injury. In other words, the negative health effects of a cause that are associated with certain health states. For example, chronic kidney disease can be a sequela of diabetes, neck pain can be a sequela of whiplash, and foot ulcers is a sequela of diabetes.
```
alibow, line 58: I think that GBD tries to make these mutually exclusive and collectively exhaustive within a given cause -- we should check on that and note it here if that is true.

alibow: It may also be worth noting here that there are asymptomatic sequelae.

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -33,11 +33,117 @@ Understanding GBD's Disease and Health Categories
...
+The cause hierarchy diagrams for IHD and HHD are included here as a reference. 
+
+
+**IHD**:
+
+.. image:: cause_hierarchy_ihd.svg
+
+**HHD**: 
+
+.. image:: cause_hierarchy_hhd.svg
```
alibow, line 86: I think it could be worth combining these into a single diagram that includes both causes.

alibow: I think it could also be worth labeling the "levels" of the hierarchy in the diagram and then noting in the text below that causes are only mutually exclusive/collectively exhaustive at a given level.

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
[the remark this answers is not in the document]
alibow: image

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -33,11 +33,117 @@ Understanding GBD's Disease and Health Categories
...
+a patient might be experiencing, or a doctor might identify. 
+
+Sequelae often contain a level of severity for that condition - for example, mild and moderate 
+angina are both sequelae within IHD. 
+
+Health States
++++++++++++++
+
+A health state is the amount of health or disability within a sequelae. This is summarized 
+through the disability weight. More information on this can be found in `this section of the cause model page <https://vivarium-research.readthedocs.io/en/latest/model_design/vivarium_model_components/causes/index.html#disability-weights>`_.
```
alibow, line 116: I think it would be worth noting that you can look up health states and their disability weights in the GBD methods appendix. 

Also, I'm not sure where we would want to handle this, but if we can have multiple health states for a given sequela (like the uterine fibroid example above), I think we should note how to determine the overall DW of that sequela.

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -33,11 +33,117 @@ Understanding GBD's Disease and Health Categories
...
-Understanding the GBD Hierarchy
--------------------------------
+.. list-table:: GBD definitions of terms 
+  :widths: 15 15 15
+  :header-rows: 1
 
-- Showing what is a direct subset, what overlaps, what interacts but is independent 
-- Include some information on modeling implications as a result of this 
+  * - Term 
+    - Definition
```
alibow, line 43: Oh, I now see that all of these definitions come from the IHME Glossary. I guess I don't find their definitions super useful, but maybe it's good to keep them for consistency and then just add extra context/etc. to the notes section.

alibow (approved):
I think this does a great job laying out the information and I left a bunch of comments about potential edge cases and small details that I think would be fine to come back to as future to-dos/notes.

aflaxman (approved):
Nice work!

zmbc:
I'll just chip in that I agree with Ali that the GBD glossary definitions do not help me much. I think the part you wrote is much better, but we should make sure that it is precisely accurate. I plan to review this PR more thoroughly, but we might talk about it in the doc session before I do!

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -30,39 +30,15 @@ Understanding GBD's Disease and Health Categories
 .. contents::
   :local:
 
-GBD Definitions
----------------
+This document seeks to clarify a few different disease and health concepts in 
+GBD. These are: a cause, a sequela, a health state and an impairment. First, we 
+will explain these different concepts conceptual. The official GBD defintions
```
zmbc, line 35:
```suggestion
will explain these different concepts at a conceptual level. The official GBD definitions 
```

Maybe?

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -71,28 +47,26 @@ In particular, we will examine:
...
+disease. At each level (labeled), the cause hierarchy is mutually exclusive and 
 collectively exhaustive - meaning all disease and death can be placed into a group at 
-each level of the diagram. 
+each level of the diagram. A cause is anything that causes death or disability. 
 
 Both IHD and HHD are within the more general cardiovascular disease cause (c_491). IHD and HHD 
-are causes within GBD. 
+are "causes" within GBD. 
 
 Sequelae
```
zmbc, line 71: I think it would be useful to add here what we discussed about the difference between causes and sequelae: that sequelae are almost like the last level of the cause hierarchy, except that they are not modeled as complexly. Specifically, that the data used to model sequelae are generally not differentiated by location, year, age or sex (is this correct)?

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -130,20 +109,58 @@ say that the health state and therefore the disability weight is the same for bo
...
 +++++++++++
 
-In some cases, a medical condition or sequela might have multiple causes. For example, heart failure. 
-It is caused by both IHD and HHD (and others!) and has sequelae associated with both causes. 
-We even learned above that it has the exact same health state in both causes. Heart failure 
-is an example of an impairment. 
+Sometimes, GBD has a lot of information about a health state - such as severe heart failure in the 
+example above - that the modelers want to utilize. However, in the process of normal GBD 
+modeling, this wouldn't be possible. Usually, the cause level information is calculated 
+first and then divded into sequelae.
```
zmbc, line 115:
```suggestion
first and then divided into sequelae. 
```

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -130,20 +109,58 @@ say that the health state and therefore the disability weight is the same for bo
 Impairments
 +++++++++++
 
-In some cases, a medical condition or sequela might have multiple causes. For example, heart failure. 
-It is caused by both IHD and HHD (and others!) and has sequelae associated with both causes. 
-We even learned above that it has the exact same health state in both causes. Heart failure 
-is an example of an impairment. 
+Sometimes, GBD has a lot of information about a health state - such as severe heart failure in the 
+example above - that the modelers want to utilize. However, in the process of normal GBD 
+modeling, this wouldn't be possible. Usually, the cause level information is calculated
```
zmbc, line 114:
```suggestion
modeling, this wouldn't be possible. Usually, the cause is estimated
```

This feels a little less vague

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -130,20 +109,58 @@ say that the health state and therefore the disability weight is the same for bo
...
-We even learned above that it has the exact same health state in both causes. Heart failure 
-is an example of an impairment. 
+Sometimes, GBD has a lot of information about a health state - such as severe heart failure in the 
+example above - that the modelers want to utilize. However, in the process of normal GBD 
+modeling, this wouldn't be possible. Usually, the cause level information is calculated 
+first and then divded into sequelae. 
 
-An impairment is a sequelae that shows up in 2 or more causes. 
+Therefore, to use this valuable information about a health state, GBD created a new 
+category called impairments. An impairment is medically the same as a sequela, but
```
zmbc, line 118: I think of it as more being medically the same as a health state, although I realize that's not precisely correct either (e.g. the anemia impairment contains multiple health states). Maybe just removing this clause is best?

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -130,20 +109,58 @@ say that the health state and therefore the disability weight is the same for bo
...
+category called impairments. An impairment is medically the same as a sequela, but 
+there is more information about it. To include this, GBD modelers first find the total 
+amount of the impairment - heart failure - and then divide it between causes. This 
+creates the sequelae of "heart failure due to IHD" and "heart failure 
+due to HHD" for example. These are included in blue in the diagram above. 
 
-In the GBD machinery, impairments are estimated very differently than sequelae. Instead of 
-estimating a cause and then its sequelae, impairments are estimated first and then attributed 
-to certain causes as their sequelae. 
+These sequelae therfore have the same health state. Therefore, an
```
zmbc, line 124:
```suggestion
These sequelae have the same health state. Therefore, an 
```

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -30,14 +30,137 @@ Understanding GBD's Disease and Health Categories
...
+there is more information about it. To include this, GBD modelers first find the total 
+amount of the impairment - heart failure - and then divide it between causes. This 
+creates the sequelae of "heart failure due to IHD" and "heart failure 
+due to HHD" for example. These are included in blue in the diagram above. 
+
+These sequelae therfore have the same health state. Therefore, an 
+impairment is a grouping of the related health states (mild/moderate/severe heart failure for 
+example) that show up in 2 or more causes. There are many examples that fit this 
+definition but are not GBD impairments, since the main criteria is the amount of 
+information about the health state.
```
zmbc, line 128:
```suggestion
definition but are not GBD impairments, since the main criteria for being a GBD impairment
is the amount of data available about the health state(s). 
```

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -30,14 +30,137 @@ Understanding GBD's Disease and Health Categories
...
+
+These sequelae therfore have the same health state. Therefore, an 
+impairment is a grouping of the related health states (mild/moderate/severe heart failure for 
+example) that show up in 2 or more causes. There are many examples that fit this 
+definition but are not GBD impairments, since the main criteria is the amount of 
+information about the health state. 
+
+Further information the GBD calculations can be found on the `Impairments 101 Hub page <https://hub.ihme.washington.edu/display/GBD2016/Impairments+101>`_. 
+
+To learn more about how to model impairments, check out this :ref:`page on modeling impairments <impairments>`.
```
zmbc, line 132:
```suggestion
To learn more about how to model impairments in Vivarium, check out this :ref:`page on modeling impairments <impairments>`. 
```

zmbc (approved):
This is great, Syl! Super helpful to me and I expect to future researchers :)

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -30,14 +30,137 @@ Understanding GBD's Disease and Health Categories
...
+
+A health state is the amount of health or disability within a sequelae. This is summarized 
+through the disability weight. More information on this can be found in `this section of the cause model page <https://vivarium-research.readthedocs.io/en/latest/model_design/vivarium_model_components/causes/index.html#disability-weights>`_. A health state for a sequelae and its disability weight can be 
+found in the GBD methods appendix. 
+
+Using our above sequelae (acute myocardial infarction and angina) as an example - 
+the amount of health lost while experiencing an acute 
+myocardial infarction is quite high! You're likely hospitalized and in a lot of pain. 
+However, the health lost from angina might be less significant. You're likely at home and 
+might only experience discomfort when under stress. This affects your disability weight value.
```
NathanielBlairStahn, line 97: Do you show the health states and disability weights associated with acute myocardial infarction and angina anywhere? The first sentence here made me think you were referring to specific numbers for the disability weights, but I don't see them. And I'm also curious what the descriptions of the health states for these sequelae are -- are they basically the same as the sequela names?

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -30,14 +30,137 @@ Understanding GBD's Disease and Health Categories
...
+(a heart attack) and angina (chest pain) as associated sequelae. These are the things 
+a patient might be experiencing, or a doctor might identify. 
+
+Sequelae often contain a level of severity for that condition - for example, mild and moderate 
+angina are both sequelae within IHD. 
+
+Health States
++++++++++++++
+
+A health state is the amount of health or disability within a sequelae. This is summarized
```
NathanielBlairStahn, line 89: Well, technically I don't think a health state is an "amount" -- it seems like the health state is better conceptualized as a "state of being" (not sure how to say it without using the word "state"), which is described in terms of your symptoms, etc., and the _amount_ of health/disability for that state is quantified by the disability weight. But maybe this is getting a bit pedantic, so I'm not sure if it's worth trying to make this distinction.

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -30,14 +30,137 @@ Understanding GBD's Disease and Health Categories
...
+symptoms, such as "headache and chest pain", your health state contains all of these losses to 
+health in one "health state". 
+
+One important note is that health states can be shared across sequelae - and even shared 
+between sequelae of different causes. For example, severe heart failure due to IHD and 
+severe heart failure due to HHD have the same amount of health loss associated with them. 
+Severe heart failure feels the same to a patient no matter the cause. Therefore, we would 
+say that the health state and therefore the disability weight is the same for both sequelae. 
+
+Impairments
```
NathanielBlairStahn, line 109: Here's how GBD modelers described impairments for GBD 2015 in the slide deck from the Impairments 101 HUB page:

> What is an impairment? 
•“Impairments in GBD are disorders or specific domains of functional health loss that are spread across many GBD causes as sequelae and for which there are better data to estimate the occurrence of the overall impairment than for each sequela based on the underlying cause” (2015 YLD capstone)
•TLDR: Outcome of multiple causes

Maybe it would be good to use/reference/paraphrase this description somewhere?

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
NathanielBlairStahn: The phrase "the cause is estimated" doesn't exactly make sense to me, because how do you quantify a cause? Do you mean the cause prevalence is estimated? Or DALYs? Or what?

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
```diff
@@ -30,14 +30,137 @@ Understanding GBD's Disease and Health Categories
...
+
+Therefore, to use this valuable information about a health state, GBD created a new 
+category called impairments. An impairment is medically the same as a sequela, but 
+there is more information about it. To include this, GBD modelers first find the total 
+amount of the impairment - heart failure - and then divide it between causes. This 
+creates the sequelae of "heart failure due to IHD" and "heart failure 
+due to HHD" for example. These are included in blue in the diagram above. 
+
+These sequelae therfore have the same health state. Therefore, an 
+impairment is a grouping of the related health states (mild/moderate/severe heart failure for
```
NathanielBlairStahn, line 125: I'm not sure it's accurate to think of an impairment as a grouping of related health states, because some sequelae (and hence health states) for a given cause are the result of combining _multiple_ impairments. The example given in the powerpoint is
> "Severe motor plus cognitive impairment with blindness and epilepsy due to neonatal preterm birth complications 32-<37wks," 

which involves _three_ impairments (blindness, epilepsy, and intellectual disability) due to one cause (neonatal preterm birth complications 32-<37wks). So the health state for this particular sequela is the result of combining three different impairments. The image Ali posted above about chronic kidney disease might be another example, but I'm not clear on what's a cause vs. impairment there.

So I guess we _could_ say that the impairment corresponds to the the individual health states we'd get in the absence of other impairments, but that seems like a confusing way to think about it when impairments can get combined, resulting in brand new health states.

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
NathanielBlairStahn, line 128: This characterization (throughout this section) of the main criteria for an impairment being "the amount of information available about the health state" doesn't quite match my understanding. What the GBD modelers say in the powerpoint is:

> Why model impairments? 
•Data are commonly available for the overall condition (e.g. anemia) but not as commonly for the specific etiology (e.g. anemia due to uterine fibroids)
•Enforces internal consistency

So the distinction for me is that there is information available about the condition itself (e.g., anemia), but not as much information available when focusing on specific causes (etiologies) of the condition. Perhaps this is the same thing you mean above when you say

> An impairment is medically the same as a sequela, but there is more information about it.

but it's not clear to me from that phrasing that what matters is that we have more information about an overall condition with multiple possible causes than we do about the condition resulting from a specific cause.

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
SylLutze: I added some of this but am not 100% sure that there aren't cases where sequela modeling is different by age/sex/etc. so am hedging a bit in language.

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
SylLutze: I edited the language a bit to try and capture this

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
SylLutze: I had not included these but added them here.

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
zmbc: I believe the word "etiology" is used with a different meaning elsewhere in the GBD, so let's avoid using it here!

Regarding your overall point, though, I think we may all be saying the same thing, but this could be an opportunity to clarify. GBD never uses much data about sequelae -- which doesn't necessarily mean that data isn't available, but it just isn't used. My understanding is that they usually have a little bit of data about sequelae, and they then extrapolate what they have to all locations/years/ages/sexes uniformly.

To me, an impairment doesn't represent a data-rich sequela, but a data-rich _set_ of sequelae. I believe an impairment is usually a set of health states, including all the sequelae that belong to those health states.

Personally, I don't find it very convincing that there is a strong philosophical/conceptual distinction between causes and impairments, but would be happy to be convinced. To me, an impairment is a data-rich grouping of sequelae that doesn't fit anywhere in the cause hierarchy, so needs to be modeled outside it.

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
zmbc: > we could say that the impairment corresponds to the the individual health states we'd get in the absence of other impairments

I think that is what we're saying -- that the vision loss impairment includes the health states of "vision loss with epilepsy and intellectual disability," "vision loss with epilepsy," "vision loss with intellectual disability," etc. One of the key features of impairments (see my comment about the cause hierarchy below) is that they are not mutually exclusive nor collectively exhaustive. Perhaps I'm missing something wrong with this framing.

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
zmbc: Fair question, and I'm not exactly sure. My sense is that incidence, prevalence and remission are estimated at the cause level, then _prevalence_ specifically is divided into the sequelae, and finally YLDs are calculated at the sequela level. Would be good to confirm this (that there are no incidence or remission estimates for sequelae).

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
zmbc: That's one of the things in the official definitions I just can't parse -- does it make sense to you? For example, which sequela is the "anemia" impairment clinically the same as? A whole bunch of them, right?

I think if you understand this and can rephrase it, it could be helpful. But as it is, it _adds_ confusion IMO.

SylLutze: Lemme try and rephrase this. Anemia is medically the same as a a bunch of sequelae. I think this is saying that an impairment is also "The medical conditions that can occur among people who contract a disease or suffer an injury." An impairment is medically the same as this definition. If you were a doctor, you'd look at a sequela as the same whether or not it happens to also be an impairment - because it being an impairment is a GBD definition, not a clinical definition.

zmbc: Ah okay, I think I am starting to understand. "An impairment is medically the same as a sequela" means "From a medical perspective, an impairment is similar to a sequela, in that it is a medical condition that can occur among people who contract a disease or suffer an injury." Is that it?

I also don't totally understand what makes a cause different from this. Aren't all three the same from a medical perspective?

zmbc: (Before, I was interpreting it to mean "_each_ impairment is medically the same as _a specific_ sequela".)

SylLutze: Yep! That's exactly it. Clinically these are the same, GBD just has these other definitions. 

The second part of your question is more about cause vs sequela. The official difference is the thing "causing" the deaths/disability and the medical condition you experience. So a clear example is diabetes is a cause - it defines what is wrong with your body - and blindness is something you experience as a result of diabetes - a sequela. 

For some of the cardiovascular examples, I think this is a bit fuzzier/harder to understand if you're not a doctor. But its the same idea.

SylLutze:
Hey @NathanielBlairStahn and @zmbc I didn't address all of your comments directly because I rephrased and edited more generally in response to them. A few high level notes: 
- I do think we are all basically talking about the same thing 
- The main criteria is the amount of information which I have tried to highlight in my new definition 
- Things can overlap which makes it weird. Yes, anemia is the impairment and it can overlap with other stuff. I tried to remove simplifying definitions that can have counter examples. 
- I tried to separate out the medical/clinical vs the GBD definitions more clearly

docs/source/model_design/vivarium_model_components/GBD_disease_health/index.rst
zmbc: Hm -- like I said in the other thread, I'm not convinced. By that logic, I would think anemia is a cause (that is something wrong in the body) and fatigue is a sequela. In the real world, symptoms have causes, and those causes have causes, potentially many layers deep.

I guess I haven't spoken to any cardiologists, though. If they truly would say that IHD and heart failure are totally different things, then I'm probably wrong. But it seems more likely to me that they would say that IHD can cause heart failure which in turn can cause other things.

zmbc: "A medical condition that can occur among people who contract a disease or suffer an injury" seems like it would include anemia, IHD, and heart failure -- I don't see how any of those three fail to meet the definition.

SylLutze: I do think there is overlap in these definitions which I know if frustrating. Also - I'm not a doctor and so would have a really hard time trying to define what exactly causes what and what is sequela in a medical sense. 

I guess cause -> medical condition under that cause (sequela) -> symptoms of that medical condition (sometimes overlapping?) 

But I think you can poke holes in this as well and the delineations I do think are often based on a modelers decision of how to define each disease area to make it fit into the GBD framework. Which makes then defining these broadly difficult. If you can think of a good way to include this ambiguity in the page, I am happy to make edits, but I think there are some things here that are how they are because some doctor working for GBD decided it would be so.

PR #1399, opened 2023-11-17
Add table of contents and audience intro to BLISS page

alibow:
This PR address the following two tickets:
- Introduce target audience of document: https://jira.ihme.washington.edu/browse/SSCI-1594
- And add a table of contents https://jira.ihme.washington.edu/browse/SSCI-1593

There were a couple of points in the description of the second ticket that I am not sure I have met, but figured I would add in the general default table of contents to start! Specifically, there was a link to this resource that I am not necessarily sure how to incorporate and maybe others have ideas to follow-up with (https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style#Retaining_existing_styles). Further, RST automatically links back to the TOC if you click on an individual header in the body of the document, so I don't think we need to make separate links!

Thanks and let me know if anyone has thoughts/concerns!


aflaxman (approved):

zmbc (approved):
I think the other thing we planned to do for the TOC is use more section headings throughout the document so the TOC is more useful, but it makes sense to do that as we go!

SylLutze (approved):

PR #1400, opened 2023-11-17
Expanded nutrition optimization timestep rules

alibow:


SylLutze (approved):

alibow:
@NathanielBlairStahn FYI - to catch you up on where we stand on the rate/probability discussion. We're planning on running some test runs described here and evaluating the results and then if things are working as expected we will follow-up with formal documentation to potentially update default behavior with an explanatory explanation. Let us know if you have any thoughts/feedback on this plan!

NathanielBlairStahn (approved):
Looks good to me!

PR #1401, opened 2023-11-17
Discontinuation language clarification 

SylLutze:


aflaxman (approved):

alibow:
This might be easier for the engineers to implement if we convert it to a rate in person years already.

The equations used are `probability = 1 - exp(-rate)`, so `rate = -ln(1-probability)`

I believe we get an annual discontinuation rate then around 0.377 events per person year.

zmbc:
> I think the statement that this probability is specific to those who survive 1 year post treatment will make this equation/conversion more complicated, though, and might need to rely on mortality rates if that is something you want to enforce.

I don't think this is the case. No matter what the mortality rate is, the probability of a simulant _who does not die_ making it a year without discontinuing is the probability of spending a year at risk without having an event.

Or that was the thinking, anyway -- I suggested including that clause to make things _simpler_. If it makes things more complicated that is definitely unintended 😄

alibow:
> > I think the statement that this probability is specific to those who survive 1 year post treatment will make this equation/conversion more complicated, though, and might need to rely on mortality rates if that is something you want to enforce.
> 
> I don't think this is the case. No matter what the mortality rate is, the probability of a simulant _who does not die_ making it a year without discontinuing is the probability of spending a person-year at risk without having an event.
> 
> Or that was the thinking, anyway -- I suggested including that clause to make things _simpler_. If it makes things more complicated that is definitely unintended 😄

Ah I have such a tough time with conceptualizing the probability to rate conversion. Yeah I agree with you after thinking about it some more -- sounds good, we can disregard both of my comments then!

zmbc (approved):

alibow (approved):

zmbc:
> I have such a tough time with conceptualizing the probability to rate conversion.

Me too! Thanks for raising this question and it is definitely reassuring that the answer makes sense to you 😃

PR #1419, opened 2023-12-06
Model 11.2 v and v summary

alibow:
Bugs addressed in this run were resolved! There may be something else going on with mortality rates than what I have been suspecting though... will follow up when our model versioning catches up


aflaxman (approved):
yay!

hussain-jafari (approved):

PR #1420, opened 2023-12-07
Adding state variation to discontinuation and updated data

SylLutze:
Note: the data might get updated again, but formatting won't change and the adjustments to values would be minimal. We do not need to wait for the updated data to run V&V.


aflaxman (approved):
Just what I was hoping you would find.  Now I hope that Steve has an easy time implementing this!

alibow (approved):

stevebachmeier (approved):

PR #1431, opened 2023-12-16
Document mediation approaches

zmbc:


docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,384 @@ Risk factor mediation
...
+there is no direct effect.
+For example, lead exposure's effect on ischemic heart disease is entirely mediated by
+high systolic blood pressure.
+Draws of the mediation factor are allowed to be negative, which represents the direct pathway
+and the mediated pathway having different effect directions.
+However, if more than 50 of the 1,000 draws were negative, that mediation
+pathway was excluded from GBD.
+
+.. todo::
+  Why does GBD exclude due to negative draws?
```
aflaxman, line 182: I'm guessing that it is because this mediation is not "significant" at the 0.05 level

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,384 @@ Risk factor mediation
...
+.. note::
+  This ignores the concept of a TMREL below which the RR does not apply and risk
+  is flat.
+  This seems to be ignored throughout the GBD's descriptions of its mediation methods, although
+  some risks that are mediators in GBD, such as BMI, clearly have non-zero TMRELs.
+
+.. note::
+  :math:`\Delta_\text{r→m}` only makes sense if both the risk and the mediator are continuous values.
+  Since all of the mediation we have modeled so far has been between continuous
+  risk factors, we haven't yet explored how to make different assumptions (neither has GBD).
```
aflaxman, line 232: I believe that this is the setting that Pearl calls "Linear Wonderland" in _The Book of Why_

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,384 @@ Risk factor mediation
...
+
+Approach 1: consistent with RR application, inconsistent with GBD mediation
```
aflaxman, line 234: Abie ran out of time here, will read more later.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,384 @@ Risk factor mediation
...
+Every mediation pathway that begins with the same primary risk
+and mediator will give us the information we need to calculate such a linear
+risk-risk effect.
+For example, BMI→SBP→IHD and BMI→SBP→Stroke both imply delta values for BMI→SBP.
+However, the calculated values from different mediation pathways
+may be inconsistent, since GBD does not enforce consistency.
+We will have to choose one or aggregate them, depending on the situation:
+it could be that we trust the GBD modeling of one of the mediation pathways
+more than the others, or that the values are all similar so it doesn't matter
+much what we do.
```
aflaxman, line 288: Average them?

aflaxman (approved):

pletale (approved):

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
zmbc: I didn't want to write something we hadn't actually done, and I believe in CVD in this situation we just chose one arbitrarily. I agree that the average would be a slightly more principled approach.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,384 @@ Risk factor mediation
...
+  (see GBD section for more on this).
+
+However, a diagram involving mediation can be more complex:
+a risk effect may be mediated through multiple (chains of) mediating risk factors,
+and there may be a non-mediated effect as well.
+
+.. image:: mediation-complex.drawio.svg
+
+At the top of this diagram, we see multi-layer mediation through the chain
+of R1→R2→R3→O.
```
SylLutze, line 58: I know you say above the R is a risk and O is an outcome but I think adding "where R1-R4 are different risks and O is an outcome"

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
SylLutze: Yeah I think in CVD we did just pick one. This was mainly a result of the specific process to get there. Did the math wrong -> deltas looked almost identical -> picked one. Then Zeb figured out the math was wrong -> recalculated -> they did NOT look identical -> kept one similar to the prior values we had to maintain V&V and continuity in modeling. 

I agree averaging might be a good thing to recommend moving forward.

SylLutze (approved):
Overall this is awesome! There's a few clarifications I'd like to talk about in our Docs session today for when we need to include or not include mediation.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
zmbc: Yeah, but the way GBD states it, I think they would exclude even if if 970/1,000 draws were negative (a "significant" effect with a negative direction). I'll clarify this and turn it into a note.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
zmbc: On my to-read list! But, I'm guessing we frequently leave "Linear Wonderland" due to RRs, which make many associations in our models log-linear 😞 It's actually only risk-risk effects that are usually linear!

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
zmbc: Ok given the history and the seeming consensus here, I'll include averaging as a recommendation though note that we haven't actually done it before.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,399 @@ Risk factor mediation
...
+  (direct and mediated pathways have opposite effect directions) could never occur;
+  therefore GBD never models such mediation situations.
+
+This "mediation factor" can be used to calculate values for each of the causal effects represented by arrows
+in the diagram in one of two ways; the following sections describe how to do this.
+
+Unfortunately, GBD mediation factors are inconsistent with the multiplicative application of RRs.
+This is because mediation factors are defined as proportions of the *excess* risk -- the amount the RR is
+greater than 1.
+In effect, GBD's mediation methods can split e.g. an RR of 2.2 into an RR of 1.7 and an RR of 1.6 (since 0.7 + 0.6 = 1.2, which
```
NathanielBlairStahn, line 204:
```suggestion
In effect, GBD's mediation methods can split e.g. an RR of 2.3 into an RR of 1.7 and an RR of 1.6 (since 0.7 + 0.6 = 1.3, which
```
Seems like there's an arithmetic error here?

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,399 @@ Risk factor mediation
...
+is the overall RR's excess risk).
+However, if we apply these two RRs, we end up with an RR of of 1.7 * 1.6 = 2.72, which is higher than
```
NathanielBlairStahn, line 206: What is the causal diagram here? Is it R1 -> R2 -> O, without any direct effect? Or is it something else? I think it would be good to include a picture to show what's going on.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,399 @@ Risk factor mediation
...
+  We don't totally understand why GBD models *excess* risk -- while there is a possibility that
+  there is a good reason behind it, we haven't been able to find that reason, so we prefer
+  Approach 1 for now.
+
+If GBD does not report a mediation factor, it may still report one or more of the effects in the diagram.
+The rest can be filled in with external sources.
+
+.. note::
+  Be careful not to interpret GBD RRs as *direct* RRs -- unless there is a mediation factor,
+  the GBD RR will not be adjusted for any mediators.
```
NathanielBlairStahn, line 225: Can you add an example diagram showing what to watch out for?

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,399 @@ Risk factor mediation
...
+
+All our multiplicative math is based on this equation describing how RRs are applied:
+
+.. math::
+
+  RR_\text{r→o} = RR_\text{r→o,direct} * \prod_{m \in M} (RR_\text{m→o})^{\Delta_\text{r→m}}
```
NathanielBlairStahn, line 231:
```suggestion
  RR_\text{r→o} = RR_\text{r→o,direct} \times \prod_{m \in M} (RR_\text{m→o})^{\Delta_\text{r→m}}
```
Or use `\cdot` instead of `\times`. In math, `*` is only ever used for operations that are different from ordinary multiplication, e.g., convolution.

Also, is $RR_\text{m→o}$ actually a relative risk as we usually define it, or is it a log relative risk? I think I've run into this before with GBD terminology, and I find it confusing.

Also, it's not immediately obvious why this equation should make sense. Why do we multiply all the RRs regardless of what the causal diagram looks like? Why is the direct effect singled out as different from the others? Are these all adjusted RRs, or what?

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
NathanielBlairStahn: Also, what are the mediation factors in this example?

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,399 @@ Risk factor mediation
...
+
+:math:`RR_\text{r→o,adjusted for m}` is the underlying data point, but it
+isn't particularly useful to us, as it doesn't necessarily show up
+anywhere in our causal diagram.
+We can use it to calculate a delta, an (assumed linear) increase in the mediator
+per unit increase in the mediated risk:
+
+.. math::
+
+  \Delta_\text{r→m} = \frac{log(RR_\text{r→o,total}) - log(RR_\text{r→o,adjusted for m})}{log(RR_\text{m→o,total})}
```
NathanielBlairStahn, line 291: Where does this equation come from? It looks like maybe it comes from the "equation describing how RRs are applied" above (if so, you could add an equation number and reference to link back to it), with the assumption that there is only a single mediator $m$, and $RR_\text{r→o,direct} = RR_\text{r→o,adjusted for m}$. Is that right? What do you do if there are multiple mediators?

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,399 @@ Risk factor mediation
...
+
+to solve for :math:`RR_\text{r→o,adjusted for m}`:
+
+.. math::
+
+  RR_\text{r→o,adjusted for m} = RR_\text{r→o,total} - MF * (RR_\text{r→o,total} - 1)
+
+:math:`RR_\text{r→o,adjusted for m}` is the underlying data point, but it
+isn't particularly useful to us, as it doesn't necessarily show up
+anywhere in our causal diagram.
```
NathanielBlairStahn, line 285: What causal diagram? Should we add an example diagram to illustrate what's going on?

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,399 @@ Risk factor mediation
...
+
+  MF = \frac{RR_\text{r→o,unadjusted for m} - RR_\text{r→o,adjusted for m}}{RR_\text{r→o,unadjusted for m} - 1}
+
+to solve for :math:`RR_\text{r→o,adjusted for m}`:
+
+.. math::
+
+  RR_\text{r→o,adjusted for m} = RR_\text{r→o,total} - MF * (RR_\text{r→o,total} - 1)
+
+:math:`RR_\text{r→o,adjusted for m}` is the underlying data point, but it
```
NathanielBlairStahn, line 283: "data point" sounds a bit odd to me here -- maybe just "data"?

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,399 @@ Risk factor mediation
...
+
+  RR_\text{r→o} = RR_\text{r→o,direct} * \prod_{m \in M} (RR_\text{m→o})^{\Delta_\text{r→m}}
+
+where :math:`M` is the set of mediators that mediate the relationship between risk :math:`r`
+and outcome :math:`o` and :math:`\Delta_\text{r→m}` is the (assumed linear) increase in :math:`m`
+caused by a one-unit increase in :math:`r`.
+
+.. note::
+  This ignores the concept of a TMREL below which the RR does not apply and risk
+  is flat.
```
NathanielBlairStahn, line 239: This sentence confused me when I first read it. Maybe say "constant" instead of "flat"? Or maybe a more substantial rewording to make it clearer.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,399 @@ Risk factor mediation
...
+.. note::
+  We may calculate a direct RR below 1.
+  This is not logically incoherent (a risk factor could have a protective direct effect
+  that is counteracted by harmful indirect/mediated effects), but it may
+  also indicate problems in the GBD mediation estimation, particularly if a
+  protective effect doesn't make sense given domain knowledge about the
+  risk factor.
+
+  We'll deal with these situations on a case-by-case basis.
+
```
NathanielBlairStahn, line 322: Taking a page from Abie's book and marking where I stopped reading.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
zmbc: Yep! Thanks for spotting that 😄

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
zmbc: This was intended to be about a single mediator, where there is both a direct and a mediated effect. I should be more explicit about this. If 1.7 was the RR _through_ the mediator and 1.6 was the _direct_ RR, the mediation factor here would be `0.7 / 1.3 = 0.538`.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
zmbc: 👍 I think a diagram could help here. This is meant to imply a single causal diagram: R -> M{1-n} -> O with a direct arrow R -> O as well.

$RR_\text{m→o}$ is a normal (not log) relative risk per unit increase in m.

We multiply the RRs of each mediation _pathway_ (the RR of R on O _through_ that mediator) with the direct RR, since that is how they get applied. The RR through the pathway is $(RR_\text{m→o})^{\Delta_\text{r→m}}$ because the total RR of the mediator on the outcome, to the power of the number of units that mediator increases for each unit increase in R.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
zmbc: Good questions! I think I forgot to include how I got here. I believe the thinking was that

$$
RR_\text{r→o,adjusted for m} = \frac{RR_\text{r→o,total}}{\prod_{m2 \in (M-\\{m\\})} (RR_\text{m2→o})^{\Delta_\text{r→m2}}}
$$

so this equation would stay the same regardless of the number of mediators.

I'll take a pass at fleshing this out.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,399 @@ Risk factor mediation
...
+of R1→R2→R3→O, where R1-R4 are different risks and O is an outcome.
+Starting from the end of the chain:
+
+* R3 has a direct effect on O
+* R2 has an effect on O mediated through R3
+* R1 has an effect on O mediated through R2 **and** R3
+
+Additionally, R1 has another effect on O mediated through R4,
+and the bottom arrow in the diagram represents another, direct (non-mediated) causal
+effect.
```
alibow, line 67: I think we should be comprehensive and list all of the causal pathways between the risks and O, including the direct effects of R1 and R4.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,399 @@ Risk factor mediation
...
+**and** that path does not include the already-modeled risk factor.
+
+For example, in a hypothetical smoking simulation, if one of our interventions
+could directly remove tar from the lungs after someone had smoked, we would need
+to model lung tar.
+
+.. image:: necessary-mediator.drawio.svg
+
+But if our interventions only affect smoking itself, there is no need to model lung tar;
+it is implicit between smoking and lung cancer.
```
alibow, line 96:
```suggestion
it is implicitly included in a measured association between smoking and lung cancer.
```

Not sure if this makes the reasoning more clear?

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,399 @@ Risk factor mediation
...
+Imagine we found studies describing the effects in this diagram
+by reporting relative risks (RRs).
+To inform the direct BMI→IHD arrow, we'd need the RR of BMI
+on IHD to be adjusted for SBP, in order to isolate the direct effect.
+If we used an RR of BMI on IHD that was not adjusted for SBP, the mediated
+part of the effect would be double-counted in our model: once in the direct
+path and once in the mediated path.
+This logic extends naturally to the situation where
+there are multiple mediated pathways in addition to a direct effect:
+for the direct effect, we need an RR adjusted for **all** mediated pathways.
```
alibow, line 150: I find the concept of a total effect being comprised of a direct/non-mediated and mediated effect can be helpful in describing this concept, where the total effect would be the measured association between BMI->IHD adjusted for confounders; direct would be BMI -> IHD adjusted for confounders and SPB, and the mediated effect would be the magnitude of the pathway through BMI->SBP (adjusted for confounders) and SBP->IHD adjusted for confounders including BMI. 

I think your description is clear and does a great job, but formalizing the terms can be helpful for conversation, etc. if you feel like adding it in!

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,399 @@ Risk factor mediation
...
+
+  Where should someone look to find this out?
+
+Adjusted RR method
+^^^^^^^^^^^^^^^^^^
+
+In this case, GBD's data were a pair of RRs for the risk-outcome pair,
+one adjusted for the mediator and one not.
+We presume that the unadjusted RR is (approximately) equal to the total RR
+GBD reports for the risk-outcome pair.
```
alibow, line 270: Should we note that (presumably) both of these RRs are adjusted for other confounders besides the mediator?

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
alibow: I am also confused by this sentence. I thought :math:`RR_\text{r→o,adjusted for m}` represents the direct effect which we would expect to use in our diagram.model if we are modeling the basic R, M, O structure.

alibow: Ok, I think I see now -- RR_\text{r→o,adjusted for m} does will equal (or approximate) the direct effect when there is only one mediator, but it will not when there are multiple mediators. Is that consistent with your interpretation? 

Either way, I think some additional context here will be helpful for the reader to more easily follow along!

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,399 @@ Risk factor mediation
...
+  This is not logically incoherent (a risk factor could have a protective direct effect
+  that is counteracted by harmful indirect/mediated effects), but it may
+  also indicate problems in the GBD mediation estimation, particularly if a
+  protective effect doesn't make sense given domain knowledge about the
+  risk factor.
+
+  We'll deal with these situations on a case-by-case basis.
+
+Delta method
+^^^^^^^^^^^^
```
alibow, line 324: Just a note if I am remembering correctly -- I think that @SylLutze tested what we would get for delta for various mediation pathways in CVD under each of these methods and found that they were really close to one another. It might be worth noting that somewhere (if that was actually true) given that we so far don't really know how/where to learn about which method GBD used for specific pathway.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -30,4 +30,399 @@ Risk factor mediation
...
+  Find a good citation for GBD's methods here.
+  Currently the best we have is this code file: https://stash.ihme.washington.edu/projects/CCGMOD/repos/paf/browse/mediate_rr.R?until=2af3bdeabcfce96ea48a0d77e3e24e6edd2f287a&untilPath=mediate_rr.R
+
+.. note::
+  This multiplicative combination of multiple mediation factors is hard to interpret -- it seems
+  to imply that mediators *overlap* in what they mediate, which seems to contradict the idea of
+  a causal effect.
+
+Risk deletion (PAFs) in the presence of mediation
+-------------------------------------------------
```
alibow, line 384: I think we should state more directly/highlight here that the direct effect of R->O should be used for PAF calculation rather than the total effect.

alibow (approved):
This is great! Thanks Zeb! I left a few specific comments, but overall, I think explicitly defining the difference between total and direct effects and adding some more explicit causal diagrams in the later sections of the document will do a lot to increase ease for a new reader.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
zmbc: In my new PR this is much more fleshed out, though I don't reference Pearl (yet!).

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
zmbc: @SylLutze I'd be interested to hear more about this! I have not actually tried doing this for real mediation factors reported by GBD so don't have a good sense.

SylLutze: Hey - yeah did calculate the deltas a few times. 
1. In the back calculate then forward calculate math. The deltas were not as internally consistent as you would want actually! For example BMI -> SBP -> IHD and BMI -> SBP -> Stroke. In theory, BMI -> SBP should be the same delta in both pathways, but we did not find this to be true. 
2. We also checked deltas in regards to an intervention. We had an intervention that affected FPG (treat -> FPG) and affected BMI (treat -> BMI -> FPG) and used deltas to see if we needed to include the (treat -> FPG) pathway, or if it was mainly due to BMI mediation. We found that it had a significant direct effect. 

I was having trouble finding where this comment chain started in the text so let me know if this is unclear/irrelevant.

zmbc: Right -- that is all sounding familiar now. Thanks for the reminder.

I'm curious -- did you try back-calculation using both the delta method and the adjusted RR method? The math is slightly different:

$$\Delta_\text{r→m} = \frac{\log(RR_\text{r→o,total}) - \log(RR_\text{r→o,total} - MF \times (RR_\text{r→o,total} - 1))}{\log(RR_\text{m→o,total})}$$

when GBD used adjusted RR;

$$\Delta_\text{r→m} = \frac{MF \times (RR_\text{r→o,total} - 1) + 1}{\log(RR_\text{m→o,total})}$$

when GBD used the delta method.

I think Ali's question is whether we end up getting similar results with either equation for actual values of RRs and MFs.

SylLutze: Oh - thanks for clarifying. I used the bottom equation for testing (although it looks like I took the log of the top part as well). It wouldn't be too difficult in the existing notebook to add the top equation and see how similar or different the results are. Would you like me to add that and see?

zmbc: > although it looks like I took the log of the top part as well

Hmm, let me double check my algebra.

Can you point me to your implementation of this?

SylLutze: Here's the file where I do mediation things. https://github.com/ihmeuw/vivarium_research_nih_us_cvd/blob/main/data_testing_exploration/risk_mediation_2.ipynb

It has some comments but maybe not the best, so happy to go over it together in a coworking or something if you want.

zmbc: Thanks! Yes, the top of the fraction should be inside a log. I'll correct this in the other PR. I've also corrected it in the comment above if that is easier to use as a reference.

I think a comparison of the two equations would be great! Your notebook won't immediately run for me so I will probably need your help with this.

SylLutze: Hey hey - I added the new equation to the notebook. It seems like the top equation gives a delta that is consistently lower than the bottom equation. The average difference is around ~0.1. I tested for measures: BMI -> FPG -> IHD, BMI -> FPG -> stroke. And the same two pathways with SBP instead of FPG. Updated notebook: https://github.com/ihmeuw/vivarium_research_nih_us_cvd/blob/main/data_testing_exploration/risk_mediation_2.ipynb

Also the notebook should run now for the sections you need. Sorry about that - some lines were commented out to have the by draw sections run and it wasn't well documented. 

Lemme know if you want to connect about any of this!

zmbc: Interesting, thanks for doing this!

I think my takeaway from your results is: it does matter a decent amount which equation you use, so it is worth trying to find out which way GBD did the calculation. I currently have a TODO for looking into how to find out which they did, so I think I'll leave that TODO as is. Sound right?

SylLutze: Hmm interesting! We can note this but it doesn't seem worth a model rerun to switch equations. 

The TODO sounds right to me. Sounds like you might have figured out how to see what GBD did though? Curious to hear more at the next docs session.

zmbc: Yeah, definitely not worth a rerun.

Nope -- they give a few examples, and those happen to be some of them. But I haven't yet found a full list.

PR #1434, opened 2023-12-18
Add examples from previous work to 'Advantages of Microsimulation' docs section

pletale:
The sections ready for review in this PR include: 

- Correlation between risk exposures (CVD example)
- Continuous values (IV Iron example) 
- Time-varying rates dependent on multiple factors (MM example - this one is definitely lacking in detail and completeness)

You will also see in this PR that I added a hyperlink in the Multiple Myeloma concept model - I hope that's okay, it allowed me to link to the relevant section!


docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
```diff
@@ -94,61 +94,90 @@ expands on what differentiates IBM, including microsimulation, from other types
...
-  Fill out the following sections under 'Advantages of microsimulation' subheading. 
-
 Correlation between risk exposures
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-At the individual level, risk exposures are frequently correlated. Examples include high body mass index and high fasting plasma glucose, tobacco smoking and alcohol use, and childhood height and weight.
+At the individual level, risk exposures are frequently correlated. Examples include high body mass index and high fasting plasma glucose, tobacco smoking and alcohol use, 
+and childhood height and weight. Microsimulation methods allow us to assign a risk exposure value independently to simulants, such that each one follows the desired population-
+level univariate distribution (the distribution of which comes from the GBD, in our microsimulations). In the common case of a dichotomous exposure, each simulant 
+(in the same age/sex/location group) has the same probability of exposure, which is equal to the prevalence in that population.
```
zmbc, line 103:
```suggestion
level univariate distribution (which frequently comes from the GBD in our microsimulations).
```

The sentence I removed already appears on the risk correlation page, right? I think it may be a bit more in-the-weeds than we need to get here.

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
```diff
@@ -94,61 +94,90 @@ expands on what differentiates IBM, including microsimulation, from other types
...
 
-Usually, Vivarium assigns each risk exposure independently to simulants, such that each one follows the desired population-level univariate distribution, which frequently comes from GBD. 
-In the common case of a dichotomous exposure, this means that each simulant (in the same age/sex/location group) has the same probability of exposure, which is equal to the prevalence in that population.
+Our cardiovascular disease (CVD) microsimulation offers a useful example of how risk exposure correlation typically looks in microsimulations. There are several correlated 
+risk factors that affect different CVD outcomes, including: body mass index (BMI), systolic blood pressure (SBP), low-density lipoprotein cholesterol levels (LDL-C), and fasting
```
zmbc, line 106:
```suggestion
risk factors that affect CVD outcomes, including: body mass index (BMI), systolic blood pressure (SBP), low-density lipoprotein cholesterol levels (LDL-C), and fasting 
```

nit: it's actually when they affect the _same_ outcomes that this is really interesting.

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
```diff
@@ -94,61 +94,90 @@ expands on what differentiates IBM, including microsimulation, from other types
...
+(in the same age/sex/location group) has the same probability of exposure, which is equal to the prevalence in that population.
 
-Usually, Vivarium assigns each risk exposure independently to simulants, such that each one follows the desired population-level univariate distribution, which frequently comes from GBD. 
-In the common case of a dichotomous exposure, this means that each simulant (in the same age/sex/location group) has the same probability of exposure, which is equal to the prevalence in that population.
+Our cardiovascular disease (CVD) microsimulation offers a useful example of how risk exposure correlation typically looks in microsimulations. There are several correlated 
+risk factors that affect different CVD outcomes, including: body mass index (BMI), systolic blood pressure (SBP), low-density lipoprotein cholesterol levels (LDL-C), and fasting 
+plasma glucose (FPG). Using NHANES survey data, we calculated correlation coefficients for these relationships, which could then be used to generate propensity values for all
```
zmbc, line 107:
```suggestion
plasma glucose (FPG). Using NHANES survey data, we calculated correlation coefficients for these relationships, which we then used to generate propensity values for all
```

Have we already defined the term "propensity"? I'm wondering if it may be a bit too technical for this context.

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
```diff
@@ -94,61 +94,90 @@ expands on what differentiates IBM, including microsimulation, from other types
...
+simulants. As such, the correlation coefificients remain constant as simulants age and their risk exposure values change. You can read more about our modeling strategy for
```
zmbc, line 108:
```suggestion
simulants. As such, the correlation coefficients remain constant as simulants age and their risk exposure values change. You can read more about our modeling strategy for 
```

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
```diff
@@ -94,61 +94,90 @@ expands on what differentiates IBM, including microsimulation, from other types
...
+risk factors that affect different CVD outcomes, including: body mass index (BMI), systolic blood pressure (SBP), low-density lipoprotein cholesterol levels (LDL-C), and fasting 
+plasma glucose (FPG). Using NHANES survey data, we calculated correlation coefficients for these relationships, which could then be used to generate propensity values for all
+simulants. As such, the correlation coefificients remain constant as simulants age and their risk exposure values change. You can read more about our modeling strategy for 
+these risk factor correlations :ref:`on this page <2023_sbp_ldlc_fpg_bmi>`.
 
 Continuous values
 ~~~~~~~~~~~~~~~~~
 
-e.g., allows the modeling of joint distribution between blood pressure and blood glucose levels -- this kind of thing is hard with categorical data characteristic of multiplication models
+Similarly, microsimulation allows the modeling of joint distribution between continuous values, something which is not as possible with categorical data characteristic of
```
zmbc, line 114:
```suggestion
Similarly, microsimulation allows the modeling of joint distributions between continuous values, something which is not as straightforward with categorical data characteristic of 
```

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
```diff
@@ -94,61 +94,90 @@ expands on what differentiates IBM, including microsimulation, from other types
...
+plasma glucose (FPG). Using NHANES survey data, we calculated correlation coefficients for these relationships, which could then be used to generate propensity values for all
+simulants. As such, the correlation coefificients remain constant as simulants age and their risk exposure values change. You can read more about our modeling strategy for 
+these risk factor correlations :ref:`on this page <2023_sbp_ldlc_fpg_bmi>`.
 
 Continuous values
 ~~~~~~~~~~~~~~~~~
 
-e.g., allows the modeling of joint distribution between blood pressure and blood glucose levels -- this kind of thing is hard with categorical data characteristic of multiplication models
+Similarly, microsimulation allows the modeling of joint distribution between continuous values, something which is not as possible with categorical data characteristic of 
+multiplication models.
```
zmbc, line 115:
```suggestion
compartmental models.
```

I think multiplication models generally are all continuous, and it is _categorical_ data they struggle with? I could be wrong about this though.

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
zmbc, line 114: This section feels like it is describing the same thing as the previous. I think it would be more distinct to point out that we can model a precise hemoglobin level for each simulant and leave it at that.

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
```diff
@@ -94,61 +94,90 @@ expands on what differentiates IBM, including microsimulation, from other types
...
+.. todo::
+
+    Check what we agreed is a good BLISS alternative for 'maternal BMI'. 
 
 Time-varying rates dependent on multiple factors
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-e.g., Multiple Myeloma model
+Another complexity enabled by microsimulation is the inclusion of time-varying rates dependent on multiple factors. A good example of this is in our microsimulation of Multiple Myeloma, 
+which allow relapse hazard rates to vary over time since treatment line initiation. Read more about how we calculated relapse hazard rates :ref:`here <calc_mortality_relapse_hazard>`.
```
zmbc, line 130:
```suggestion
which allow relapse and mortality hazard rates to vary over time since treatment line initiation, in addition to being impacted by the line of treatment, demographic covariates, and treatment regimen. Read more about how we calculated hazard rates :ref:`here <calc_mortality_relapse_hazard>`.
```

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
```diff
@@ -94,61 +94,90 @@ expands on what differentiates IBM, including microsimulation, from other types
...
+
+.. todo:: 
+
+    Elaborate on example from Multiple Myeloma model
```
zmbc, line 134: Perhaps my suggestion above is enough? I considered adding the word "non-parametrically," as in "vary non-parametrically over time..." but I think we want to make this page approachable and jargon-free.

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
```diff
@@ -94,61 +94,90 @@ expands on what differentiates IBM, including microsimulation, from other types
...
 
 Incorporating uncertainty
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Our microsimulation framework is formulated to routinely propagate forward the uncertainty captured by these draws. Monte Carlo methods allow us to effectively incorporate different kinds of uncertainty into our models. 
+Microsimulation methods provide a useful approach to modeling stochastic uncertainty in a conceptually valid manner because chance – that is, reliance on randomly generated numbers – 
+is used to govern movement between health states for individual simulants from a single computer-generated population, as opposed to comparing the role of chance across multiple
```
zmbc, line 140:
```suggestion
is used to govern movement between states for individual simulants from a single computer-generated population, as opposed to comparing the role of chance across multiple 
```
Since microsimulation is not specific to health.

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
```diff
@@ -94,61 +94,90 @@ expands on what differentiates IBM, including microsimulation, from other types
...
+Simulation methods have the added advantage of reducing confounding variance between different scenarios by ensuring that the simulated population is identical across all scenarios analyzed.
```
zmbc, line 146: I think this, and the first paragraph of this section, are really describing common random numbers (CRN), not Monte Carlo methods in general. If I'm correctly following the purpose of this section, it is to compare microsimulation methods with non-microsimulation methods; and then later, we will compare Vivarium with competing microsimulation tools. Is that right? If so, CRN is a _Vivarium_ feature that distinguishes it from (most) other microsimulation tools, not an aspect of microsimulation that distinguishes it from non-microsimulation methods.

Overall, I'm not sure I'm convinced there is a key advantage to note here: it seems like non-microsim methods can use Monte Carlo approaches to uncertainty as well.

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
```diff
@@ -94,61 +94,90 @@ expands on what differentiates IBM, including microsimulation, from other types
...
+Simulation methods have the added advantage of reducing confounding variance between different scenarios by ensuring that the simulated population is identical across all scenarios analyzed. 
+The only difference between scenarios, then, is the presence or absence of health interventions. By incorporating heterogeneity at the level of the individual simulant, microsimulations are
```
zmbc, line 147: This last sentence is interesting, but I don't think it fits here. Perhaps in the "allowing other complexities." We may want to think through a little more carefully exactly how this contrasts with non-microsim models.

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
```diff
@@ -94,61 +94,90 @@ expands on what differentiates IBM, including microsimulation, from other types
...
+Disadvantages of microsimulation
+++++++++++++++++++++++++++++++++
 
 Resource requirements
 ~~~~~~~~~~~~~~~~~~~~~
 
-Our microsimulation models require a significant amount of detailed data to accurately represent individual behaviors and the relationships between them. Obtaining this data can be challenging and time-consuming, and there may be privacy concerns associated with using such detailed personal data. 
-
-If we're not modeling the sort of complexity that microsimulation enables... perhaps a multiplication model would suit us just as well?
+Our microsimulation models require a significant amount of detailed data to accurately represent individual behaviors and the relationships between them.
```
zmbc, line 167: I think of this as "data requirements," and as almost exactly the flip side of the "allowing other complexities" coin. So, I think we should either explicitly link the two, and list it as both an advantage and a disadvantage, or lean into being salespeople and remove this disadvantage, leaving only the advantage 😄

However, there is a disadvantage I would call "resource requirements," and that is that our simulations take a lot of _computational_ resources to run. That should definitely be listed here.

zmbc (commented):
Great work! I am leaving a bunch of comments, but I don't expect you to single-handedly address them all -- I think we should discuss as a team some of the trickier ones, or maybe 1:1 if you feel like you just need a bit of clarification.

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
aflaxman: agree, technical and also jargon specific to our simulation work.  "which we used to generate values for all" gets the same point across, and I would suggest the change directly if I could figure out how to do that...

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
aflaxman: let's brainstorm if there is a way to make this clearer.

aflaxman (approved):
Thanks for this work Alix, and thanks to Zeb for giving it such a thoughtful read!

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
zmbc:
```suggestion
plasma glucose (FPG). Using NHANES survey data, we calculated correlation coefficients for these relationships, which we used to generate values for all
```

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
SylLutze: I don't think you really need to include this, but I wanted to note that we actually find correlation at an age/sex specific level. For example, older people might have more correlation between risk factors. It's then actually a limitation of the sim that this correlation does NOT change over time.

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
```diff
@@ -94,61 +94,90 @@ expands on what differentiates IBM, including microsimulation, from other types
...
+multiplication models.
+
+For example, in our intravenous (IV) iron intervention microsimulation, we model the joint distribution of maternal BMI and hemoglobin levels, as two risk exposures associated with poor
+health outcomes for both mother/birthing parent and child. We treat hemoglobin as a continuous risk exposure that varies as simulants age and move through the pregancy model states,
+while maternal BMI is a categorical risk exposure conditional on hemoglobin status (i.e., above or below 100 g/L) at the start of the pregnancy. Read more about the modeling strategy
+we used for this joint distribution :ref:`here <2019_risk_exposure_maternal_bmi_hgb>`.
+
+.. todo::
+
+    Check what we agreed is a good BLISS alternative for 'maternal BMI'.
```
SylLutze, line 124: It is pre-pregnancy/first trimester BMI

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
SylLutze: I agree about the "two sides of the same coin" part and connecting them would be cool! But I do think this is important to include. We have definitely run into issues with CVD where the data on how many people take their medications, or go to the doctor, or restart meds after discontinuing is really hard to find.

SylLutze (approved):
Looks awesome!

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
```diff
@@ -94,63 +94,95 @@ expands on what differentiates IBM, including microsimulation, from other types
...
+
+Disadvantages of microsimulation
+++++++++++++++++++++++++++++++++
 
 Resource requirements
 ~~~~~~~~~~~~~~~~~~~~~
 
-Our microsimulation models require a significant amount of detailed data to accurately represent individual behaviors and the relationships between them. Obtaining this data can be challenging and time-consuming, and there may be privacy concerns associated with using such detailed personal data. 
+Our microsimulations typically require a substantial amount of computational resources to run. For example, our microsimulation of a full-scale
+United States population was made possible through the use of a high performance computing cluster and used approximately 55 gigabytes of memory over a runtime of 21.5 hours.
```
zmbc, line 156:
```suggestion
United States population was made possible through the use of a high performance computing cluster and used 334 parallel runs, each requiring approximately 55 gigabytes of memory over a runtime of 21.5 hours.
```

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
```diff
@@ -94,63 +94,95 @@ expands on what differentiates IBM, including microsimulation, from other types
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-e.g., Multiple Myeloma model
+Another complexity enabled by microsimulation is the inclusion of time-varying rates dependent on multiple factors. A good example of this is in our microsimulation of Multiple Myeloma, 
+which allow relapse and mortality hazard rates to vary over time since treatment line initiation, in addition to being impacted by the line of treatment, demographic covariates, and treatment regimen. Read more about how we calculated these rates :ref:`here <calc_mortality_relapse_hazard>`.
 
-Incorporating uncertainty
-~~~~~~~~~~~~~~~~~~~~~~~~~
+The figure below demonstrates how survival rates in multiple myeloma patients differ based on the year of diagnosis and treatment line initiation. This figure was produced by
+researchers at Amgen, Inc, and you can read the paper in which it was published `here <https://www.tandfonline.com/doi/full/10.1080/10428194.2020.1827253>`_. [Braunlin-MM-2020]_
```
zmbc, line 128: I think you can remove this sentence now that we have the citation.

zmbc (approved):

docs/source/model_design/vivarium_overview/vivarium_versus_other_model_types/index.rst
```diff
@@ -94,63 +94,95 @@ expands on what differentiates IBM, including microsimulation, from other types
...
 
-Disadvantages of microsimulation
-++++++++++++++++++++++++++++++++
+.. todo:: 
+  
+  Elaborate on example of this from previous work: examining hemoglobin on curve on an individual level and seeing the discrepancies from real life that would also be present in a 
+  model that doesn't go to the level of detail we do in microsim. 
 
-.. todo::
+  Another example I'll use in this section is the PRL project: family structure (if something like this is available at all in a comparmental model, then it is only reported in a
```
SylLutze, line 144:
```suggestion
  Another example I'll use in this section is the PRL project: family structure (if something like this is available at all in a compartmental model, then it is only reported in a 
```

SylLutze (approved):

PR #1441, opened 2024-01-04
Update guardian based duplication to 2 (not 3) simulant categories

pletale:
JIRA ticket [here](https://jira.ihme.washington.edu/browse/SSCI-1636)
Slack thread [here](https://ihme.slack.com/archives/C02KUQ9LX32/p1703183820206599)

### DIFF
--- a/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
@@ -3819,10 +3819,9 @@ opportunity for duplication. Since this mechanism occurs within the
 simulation, there is a natural maximum that we will impose in the
 noise function.
 
-Guardian-based duplication is applied to three mutually exclusive categories of
+Guardian-based duplication is applied to two mutually exclusive categories of
 simulants based on age and GQ status: Simulants younger than 18  (<18) and not
-in GQ; simulants 18-23 (18 <= age < 24) and not in GQ; and simulants under 24
-(<24) and in GQ.
+in GQ and simulants under 24 (<24) and in GQ (i.e., college).
 
 For each of the three categories of simulants, the maximum duplication rate will
 be calculated based on those who have a guardian living at a different address
@@ -3833,9 +3832,9 @@ other types of GQ, so all three maximum duplication rates will be less than
 
 The user can then pick a rate of duplication between 0 and 100% **for each of
 the three categories of simulants**. A default duplication rate of 5% will be
-selected for each of the three categories of simulants. That is, each simulant
+selected for each of the two categories of simulants. That is, each simulant
 under 24 is duplicated at a guardian's household with default probability 0.05,
-and there should be three user parameters for overriding this probability, one
+and there should be two user parameters for overriding this probability, one
 for each simulant category.
 If the user selects a duplication rate that is higher than the calculated
 maximum rate in the sim, a warning should be issued explaining that the selected


docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3819,23 +3819,22 @@ opportunity for duplication. Since this mechanism occurs within the
...
-(<24) and in GQ.
+in GQ and simulants under 24 (<24) and in college GQ.
 
-For each of the three categories of simulants, the maximum duplication rate will
+For each of the two categories of simulants, the maximum duplication rate will
 be calculated based on those who have a guardian living at a different address
 in the sim. Note that all simulants in *college* GQ are initialized with a
 guardian living at a different address, but this is not true for simulants in
-other types of GQ, so all three maximum duplication rates will be less than
+other types of GQ, so both maximum duplication rates will be less than
```
zmbc, line 3830: Since we no longer duplicate simulants in GQ other than college, this could be reworded a bit.

I suggest changing "but this is not true for simulants in other types of GQ" to "but this is not true for simulants that move into college GQ during the simulation". Also, removing the emphasis from "*college*" in the previous clause.

Couldn't directly suggest these changes on GitHub because those lines weren't edited.

zmbc (approved):

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3819,23 +3819,22 @@ opportunity for duplication. Since this mechanism occurs within the
...
 
 The user can then pick a rate of duplication between 0 and 100% **for each of
-the three categories of simulants**. A default duplication rate of 5% will be
-selected for each of the three categories of simulants. That is, each simulant
+the two categories of simulants**. A default duplication rate of 5% will be
+selected for each of the two categories of simulants. That is, each simulant
 under 24 is duplicated at a guardian's household with default probability 0.05,
-and there should be three user parameters for overriding this probability, one
+and there should be two user parameters for overriding this probability, one
 for each simulant category.
```
NathanielBlairStahn, line 3838: Based on [Zeb's BOE calculation](https://ihme.slack.com/archives/C02KUQ9LX32/p1705083307080909) from the 2020 AR Census report, let's set the default for under 18 to 2%, and for now, keep the 5% default for college students. We'll need to dig into the references further to see if we can get a better estimate for college students.

Here's a suggested replacement for the highlighted lines (GitHub wouldn't let me do an actual suggested change because of the mix of edited and unedited lines), but feel free to say it differently if you think there's a way to make it more clear:

> the two categories of simulants**. We will set a default duplication rate of 2% for
simulants under 18 and not in GQ, and a default duplication rate of 5% for simulants
under 24 and in college GQ.
> That is, each simulant under 18 and not in GQ is duplicated at a guardian's household with default probability 0.02, and each simulant under 24 in college GQ is duplicated at a guardian's household with default probability 0.05.
> There should be two user parameters for overriding these probabilities, one
for each simulant category.

NathanielBlairStahn (approved):
Looks good. I left one comment suggesting an edit to our default probabilities based on Zeb's comment in Slack.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
zmbc: Let's describe the reasoning behind 2% this way:

> This percentage was estimated from Figure 1 of "An Analysis of Person Duplication in Census 2000" [insert citation to http://www.asasrms.org/Proceedings/y2004/files/Jsm2004-000730.pdf] and roughly corroborated by Table 4 of "Real-Time 2020 Administrative Record Census Simulation" [cite https://www2.census.gov/programs-surveys/decennial/2020/program-management/evaluate-docs/EAE-2020-admin-records-experiment.pdf] combined with the totals from Table 1.

The first report more directly gives us the number we want, while the more recent report is more of a fudge but in the same ballpark.

PR #1444, opened 2024-01-29
Mediation version 2

zmbc:
This is a really substantial rewrite in response to comments on #1431, so I made a separate PR. **Let me know if you think it would be easier to review the whole diff at once, in which case I can just merge this PR without review and we can move back to that one.**


docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -44,214 +45,505 @@ but that effect is (completely) mediated by another risk factor R2.
...
+If the total effect is known, we can split it;
+if it is not known, we can infer what it is by combining
+the effects of each pathway.
+
+Another case in which combining and splitting are straightforward
+is when all the pathways have additive, or "linear," effects.
+These can simply be added to obtain the total effect.
+For example, if a 1 unit increase in R1 in the diagram above caused a 2.2 unit increase in O
+through R2, and also caused a 1.3 unit increase in O directly, the total
+effect would be :math:`2.2 + 1.3 = 3.5` units of increase in O per unit increase in R1.
```
alibow, line 238: Do you think it's worth noting that this is the case as long as there is no interaction between R1 and R2 and their effects on O?

zmbc: I called out above that this assumption is made throughout the document, but I think it's worth noting it again in this section about splitting! The whole idea of "splitting" doesn't really work when there is effect modification.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -44,214 +45,504 @@ but that effect is (completely) mediated by another risk factor R2.
...
+operation, just in reverse.)
+
+We will discuss two common and easy cases here, and leave the more general principles
+to be fleshed out in the future.
+
+Case 1
+++++++
+
+.. image:: composition-case-1.drawio.svg
+
```
SylLutze, line 338: I don't have a great solution, and the text is clear, but these diagrams look to me like you just keep the x effect and ignore the + effect. 

Maybe we could put x^+

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -44,214 +45,504 @@ but that effect is (completely) mediated by another risk factor R2.
...
-.. math::
-
-  RR_\text{r→o} = RR_\text{r→o,direct} * \prod_{m \in M} (RR_\text{m→o})^{\Delta_\text{r→m}}
+.. todo::
 
-where :math:`M` is the set of mediators that mediate the relationship between risk :math:`r`
-and outcome :math:`o` and :math:`\Delta_\text{r→m}` is the (assumed linear) increase in :math:`m`
-caused by a one-unit increase in :math:`r`.
+  We have yet to make a simulation that uses a combination of risks where GBD implies
+  multi-layer mediation.
```
SylLutze, line 427: Well I guess we did in CVD but we didn't really consider it in depth and don't have a recommended approach here.

SylLutze (approved):
Overall great job! Clear and understandable

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
zmbc: Oops -- did not know this when I wrote this sentence. I'll tweak it a bit.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
zmbc: Yeah, that's a good point.

The intention here is to diagram the effect _type_ (linear or log-linear) but not the effect _size_. I think the notation gets a little unwieldy if we try to include the effect size -- for example, in your suggestion, if I put both an `x` and a `+` on the same arrow, it might not be clear anymore what type of effect it is (`x`). Doubly so in the next diagram, where I would need to use a times symbol to show how to calculate the _size_ of a `+` effect.

Definitely open to ideas about either how to make the effect _type_ notation clearer, or more ideas for an unambiguous effect _size_ notation!

SylLutze: Hmm okay, I see your point. Could we have a different symbol for "to the power of"? So then three total symbols: +, x, and maybe e or pow? 

+ and + turn to x. x and + turn to pow. So on? 

Otherwise we can leave as is and not overcomplicate it. As I said - the text is very clear once you read it and see what's happening.

zmbc: I think there are two separate attributes for each arrow in the diagram: what _type_ of effect is it, and how you _calculate_ the effect size (RR for a log-linear `x` effect, delta for a linear `+` effect).

The composition rules here describe both the resulting effect type and its size:
* A `+` effect _type_ with an effect size of $a$ and a `x` effect type with an effect size of $b$ compose into a `x` effect with effect size $b^a$.
* A `+` effect type with an effect size of $a$ and a `+` effect type with an effect size of $b$ compose into a `+` effect type with an effect size of $a \times b$.

Maybe I am misunderstanding what you mean, but I don't think there are three types (+, x, pow) -- there are only two. It is just that the power operation is used in the calculation of the effect _size_ in the first bullet.

Also, as I called out in the doc session -- I have yet to do a very thorough search for other people's work/terminology on this! I should probably do that before we go too deep on it!

SylLutze: Okay - makes sense. I see what you're going for here and how you got there. I think you can leave it as I don't have a significantly better suggestion.

alibow (approved):

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -44,214 +45,504 @@ but that effect is (completely) mediated by another risk factor R2.
...
-part of the effect would be double-counted in our model: once in the direct
-path and once in the mediated path.
-This logic extends naturally to the situation where
-there are multiple mediated pathways in addition to a direct effect:
-for the direct effect, we need an RR adjusted for **all** mediated pathways.
+we can say that
+
+.. math::
+
+  RR_\text{r→o,total} = RR_\text{r→o,direct} \times RR_\text{r1→r2→o}
```
NathanielBlairStahn, line 225: Your labels in this equation don't quite match the diagram
```suggestion
  RR_\text{R1→O,total} = RR_\text{R1→O,direct} \times RR_\text{R1→R2→O}
```

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -44,214 +45,504 @@ but that effect is (completely) mediated by another risk factor R2.
   This diagram is so simple it may not always be described as "mediation."
   For example, on our :ref:`Causal Diagrams page <general_dags>` page we call this an
   :ref:`"intermediate." <general_dags_intermediates>`
-  However, it can be useful to analyze this situation as mediation; for example, this
+  However, it can be useful to analyze this situation as a special case of mediation;
+  for example, this
   causal structure of GBD risks would report a mediation factor
   (see GBD section for more on this).
```
NathanielBlairStahn, line 51: You could link to the GBD section here.

docs/source/model_design/vivarium_model_components/risk_factors/risk_mediation/index.rst
```diff
@@ -44,214 +45,504 @@ but that effect is (completely) mediated by another risk factor R2.
...
 
-However, a diagram involving mediation can be more complex:
-a risk effect may be mediated through multiple (chains of) mediating risk factors,
-and there may be a non-mediated effect as well.
+.. note::
+
+  In our work, we typically run into mediation with risk factors.
+  Throughout this page, we will use risk factor examples.
+  Most of this page would equally apply to other exposure/outcome pairs;
+  only the GBD section is specific to risk factors.
```
NathanielBlairStahn, line 58: You could also link to the GBD section here.

PR #1445, opened 2024-01-30
Adding modeling impairments page

SylLutze:


docs/source/model_design/vivarium_model_components/impairments/index.rst
```diff
@@ -33,17 +33,224 @@ Modeling Impairments
...
+Disease Models
+++++++++++++++
+
+The GBD data availability differs by impairment significantly. For 
+example, heart failure has an incidence, prevalence, excess 
+mortality rate, and disability weights. Therefore, you could create a heart 
+failure model more of less exactly as you might make a cause model. 
+
+On the other hand, anemia does NOT have an incidence or any mortality 
+associated with it. It does have a prevalence and disability weight.
```
zmbc, line 88: Fascinating!

docs/source/model_design/vivarium_model_components/impairments/index.rst
```diff
@@ -33,17 +33,224 @@ Modeling Impairments
...
+anemia model. You can see one example of this based on simulant's hemoglobin 
+levels in the :ref:`hemoglobin, anemia and iron deficiency model <2019_hemoglobin_anemia_and_iron_deficiency>`.
+
+Risk Models
++++++++++++
+
+GBD does NOT include relative risks or PAFs connecting risk factors 
+to impairments. Therefore, if you need to model the impact of a risk 
+factor to an impairment you will need to use literature values for 
+realtive risks. These values
```
zmbc, line 99:
```suggestion
relative risks. These values 
```

docs/source/model_design/vivarium_model_components/impairments/index.rst
```diff
@@ -33,17 +33,224 @@ Modeling Impairments
...
+Therefore, other methods might be needed in order to have a dynamic 
+anemia model. You can see one example of this based on simulant's hemoglobin 
+levels in the :ref:`hemoglobin, anemia and iron deficiency model <2019_hemoglobin_anemia_and_iron_deficiency>`.
+
+Risk Models
++++++++++++
+
+GBD does NOT include relative risks or PAFs connecting risk factors 
+to impairments. Therefore, if you need to model the impact of a risk 
+factor to an impairment you will need to use literature values for
```
zmbc, line 98: It seems like you _could_ in theory use GBD RRs if you assume that the risk factor has the same effect on all sequelae within a given cause. (I think this is basically what GBD assumes?) But, I can definitely imagine situations where that is far from the truth.

docs/source/model_design/vivarium_model_components/impairments/index.rst
```diff
@@ -33,17 +33,224 @@ Modeling Impairments
...
++++++++++++++++++++++++++++++++++++++++++
+
+Another option is to include a single cause model that includes both the cause 
+and impairment. One example of this is in the CVD model where IHD and Heart Failure 
+are included in the same :ref:`cause model <2019_cause_ihd>`.
+
+By including both the cause and impairment in a single modeling component, 
+they are no longer independent which solves our prior problems of overlap. 
+In addition, you can define part of the model to overlap - for example states 
+could inlcude diabetes, blindness due to diabetes, and blindness due to other
```
zmbc, line 193:
```suggestion
could include diabetes, blindness due to diabetes, and blindness due to other 
```

docs/source/model_design/vivarium_model_components/impairments/index.rst
```diff
@@ -33,17 +33,224 @@ Modeling Impairments
...
+
+You could imagine a similar pathway from the blindness impairment to 
+the diabetes cause model. This essentially enforces the idea that a simulant 
+can only go blind by 1 cause. This could in theory scale up for 
+an increasing number of cause and/or impairment models. 
+
+Another way to utilize causes acting as risks would be to have cause model 
+states dictate the options available for the impairment model. 
+
+Imagine a blindness impairment where simulants move from suseptible to blind
```
zmbc, line 234:
```suggestion
Imagine a blindness impairment where simulants move from susceptible to blind 
```

docs/source/model_design/vivarium_model_components/impairments/index.rst
```diff
@@ -33,17 +33,224 @@ Modeling Impairments
...
+
+For example, if a simulant is in the state within the diabetes cause 
+for "blindness due to diabetes", it could prevent them from transitioning 
+from "susceptible" to "blindness due to other causes" in the blindness impairment. 
+Hence, a cause (diabetes) acts on a transition in the impairment component. 
+
+You could imagine a similar pathway from the blindness impairment to 
+the diabetes cause model. This essentially enforces the idea that a simulant 
+can only go blind by 1 cause. This could in theory scale up for 
+an increasing number of cause and/or impairment models.
```
zmbc, line 229: I'm not sure we should recommend this way. It seems like in order to avoid bias you would need to adjust the blindness impairment incidence for the change in denominator, and I'm not sure it is any easier than the next approach discussed.

docs/source/model_design/vivarium_model_components/impairments/index.rst
```diff
@@ -33,17 +33,224 @@ Modeling Impairments
...
+(e.g., including diabetes, meningitis, glaucoma, and vitamin A deficiency, all of 
+which cause blindness) 
+using this approach would make an extremely complex single "cause" model that 
+likely wouldn't be feasible to create or V&V. 
+
+Therefore, considering your modeling situation and the requirements is important 
+in choosing your modeling approach. 
+
+Have Causes Act as "Risks"
+++++++++++++++++++++++++++
```
zmbc, line 213: Overall, I think this section is pretty speculative -- I think I'd like to add a TODO here noting that we don't yet know exactly how to do this.

docs/source/model_design/vivarium_model_components/impairments/index.rst
```diff
@@ -33,17 +33,224 @@ Modeling Impairments
...
+
+The GBD data availability differs by impairment significantly. For 
+example, heart failure has an incidence, prevalence, excess 
+mortality rate, and disability weights. Therefore, you could create a heart 
+failure model more of less exactly as you might make a cause model. 
+
+On the other hand, anemia does NOT have an incidence or any mortality 
+associated with it. It does have a prevalence and disability weight. 
+Therefore, other methods might be needed in order to have a dynamic 
+anemia model. You can see one example of this based on simulant's hemoglobin
```
zmbc, line 90: I feel like there could be a simpler way to do this, with standard GBD assumptions: just aggregate incidence, remission, etc from all the causes, weighted by prevalence of their sequelae in the impairment. Does that sound reasonable to you? If so, should we include it as a possibility here?

zmbc (approved):
This is great! I'd be interested in discussing some of the more speculative/forward-looking ideas further in a future doc session.

docs/source/model_design/vivarium_model_components/impairments/index.rst
SylLutze: Hmm interesting. Are you thinking that we would assume that the RR for say BMI -> Ischemic Heart Disease == BMI -> Heart Failure from IHD? And then even possibly using this or some average across the causes to say the RR for BMI -> HF in general is X. 

Or were you thinking we could take the general impairment bucket (assuming there is one) like blindness from other causes and just use that for the whole impairment? 

This is an interesting idea either way. For some impairments like HF something like this could work well. But, you could also imagine that blindness due to diabetes is tied to say, sugar consumption but blindness due to meningitis is not tied to sugar at all. So this might not work as well. 

Maybe we can discuss this more in our next meeting - it seems like the RR might depend on how dispirit the causes of an impairment are.

docs/source/model_design/vivarium_model_components/impairments/index.rst
SylLutze: Hmm I can definitely include it. And it is an option. I think you're a bit underestimating how many causes and sequelae would be involved in this though! We use some of this data for HF and got it from the CVD team - there are a lot of causes and things involved.

docs/source/model_design/vivarium_model_components/impairments/index.rst
zmbc: Yes, I was thinking that you would take the _weighted average_ of the RRs of BMI -> Ischemic Heart Disease and BMI -> Hypertensive Heart Disease, weighted by the prevalence of Heart Failure from IHD and Heart Failure from HHD respectively (pretending there are only two causes -- it would actually be more).

And I agree I think this may depend on the impairment. The diabetes example is a good one (it would be strange for sugar consumption to be a risk factor for blindness due to meningitis, especially if we were reporting that outcome separately).

I was imagining something even weirder, though I don't know of an example: imagine that there was a risk factor that (in real life) put someone at risk for mild disease but not for severe disease. If only severe disease is included in the impairment, with this approach we would say that risk factor acts on the impairment, even if they are intuitively not linked at all.

docs/source/model_design/vivarium_model_components/impairments/index.rst
zmbc: Fair enough. I didn't mean to say it would be simple! Just seems simpler than having to seek external data sources. And, most of the complexity would be on the research/data prep side so more of a 1-time cost than adding complexity to the sim itself.

alibow (approved):

PR #1447, opened 2024-02-13
Updating the causes as risks section in impairments

SylLutze:


docs/source/model_design/vivarium_model_components/impairments/index.rst
```diff
@@ -218,56 +218,34 @@ in choosing your modeling approach.
...
-  We have not implemented causes as risks in practice and so this section is 
-  speculative. We will further build this out as we implement these ideas in 
-  practice. 
+  While we think this methodology will work well, it has not been tested in a 
+  model up to this point. Therefore the implemenation remains theoretical, and 
+  some details to the approach would need to be finalized before it could be 
+  used. 
+
+Another option is to have cause models that act as "risks" for other 
+causes. In effect, we would adjust the incidence of the impairment
```
zmbc, line 229:
```suggestion
Another option is to have cause models that act as "risks" on the
impairment model. In effect, we would adjust the incidence of the impairment
```

docs/source/model_design/vivarium_model_components/impairments/index.rst
```diff
@@ -218,56 +218,34 @@ in choosing your modeling approach.
...
-This approach shouldn't create bias in the model and is a good option 
-if you need to include more cause models in the simulation. 
-
-.. todo::
+.. note::
   
-  We have not implemented causes as risks in practice and so this section is 
-  speculative. We will further build this out as we implement these ideas in 
-  practice. 
+  While we think this methodology will work well, it has not been tested in a
```
zmbc, line 223: I think we should note somewhere that this idea and the analogy with mortality only work well if the impairment has no remission.

zmbc (approved):

alibow (approved):

PR #1450, opened 2024-02-26
Adding to-dos for SQLNS targeting updates

SylLutze:
Hey - I am trying to make more, small PRs for this instead of one giant PR. This is basically just outlining the doc updates that will be needed and how I plan to organize them. Expect more PRs to follow shortly, each changing one to-do into a to-dah! 

A few things to look out for: 
1. I am calling this "Wave 3", although it overlaps with what we wanted to accomplish in Wave 2 (GBD 2021) 
2. Is there any other sections that you think need updates which are NOT marked? I don't think the other sub-model pages do (outside of link updates for new data generation for wasting, etc.) or the pregnancy model page but let me know if that's incorrect. 
3. Are all updates needed to document the SQLNS targeting approach included as to-do items? That is the goal so let me know if things are missing from the current to-dos.

### DIFF
--- a/docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
@@ -93,12 +93,12 @@ Documents that contain information specific to the overall model and the pregnan
 2.2 Waves, GBD Rounds, and age groups
 -------------------------------------
 
-We will separate the implementation of the child model into two waves of updates. 
+We will separate the implementation of the child model into waves of updates. 
 In addition to other differences detailed in the next section:
 
 - Wave I will use GBD 2019 data, with the exception of using GBD 2021 data for child growth failure risk exposure and risk effects.
 
-- Wave II will use GBD 2021 data for the entire model.
+- Wave II and later will use GBD 2021 data for the entire model.
 
 Notably, GBD 2021 uses different age groups than GBD 2019 (as summarized in the 
 tables below). Therefore, the Wave I implementation that uses data from both GBD 
@@ -156,6 +156,11 @@ in the simulation will be informed using data specific to the post neonatal age 
 
 2.3 Submodels
 -------------
+
+.. todo::
+
+  Update the below tables as needed for a "wave 3" with SQ-LNS targeting and subnational data included. 
+
 
 .. list-table:: Risk exposure subcomponents
   :header-rows: 1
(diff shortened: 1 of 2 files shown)


albrja (approved):

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
```diff
@@ -93,12 +93,12 @@ Documents that contain information specific to the overall model and the pregnan
...
 -------------------------------------
 
-We will separate the implementation of the child model into two waves of updates. 
+We will separate the implementation of the child model into waves of updates. 
 In addition to other differences detailed in the next section:
 
 - Wave I will use GBD 2019 data, with the exception of using GBD 2021 data for child growth failure risk exposure and risk effects.
 
-- Wave II will use GBD 2021 data for the entire model.
+- Wave II and later will use GBD 2021 data for the entire model.
```
alibow, line 101: We didn't actually get 2021 data into the wave II results that we presented in January, so maybe we should just update this to say that the 2021 update is coming in wave III?

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
```diff
@@ -486,6 +499,10 @@ Wave II:
 
 - It is possible we decide to exclude scenarios 13-18 from wave II, reducing the number of child scenarios from 19 to 13 and the total number of scenarios to 66/location for **198 scenarios**
 
+.. todo::
+
+  Update the below table to include targeted SQ-LNS scenarios. Then we will need to define a strategy for running the simulation since we will have multiple SQ-LNS targeting strategies and doing 3-4 complete runs might be infeasible.
+
```
alibow, line 505: I think we can just "rerun" the scenarios that have coverage of targeted SQLNS with the different targeting algorithms rather than rerun _all_ scenarios, right? Sorry if I am misinterpreting your todo note here!

alibow (approved):
One thing that I think is missing is re-evaluating the SQ-LNS intervention effects. I think it would be great to include the effect modification and we will also want to test to make sure that our transition effects are reasonably calibrated to the observed prevalence effects in the sublocations we are modeling. Even if we end up thinking of these things as "nice to haves" I think it's worth putting a todo note to remember to consider them.

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
SylLutze: Yep! Totally agree we don't need to. This will get replaced soon with the actual section so no changes but aligned.

alibow: Great, sounds good

PR #1452, opened 2024-02-27
Adding newest, cleaned emulator version

SylLutze:
Realized that this link hadn't been added in forever so put the latest. I used permalinks which means that even if the emulator is edited in the other open PR, the link will still work as long it has the same name. 

Here is the PR for this emulator version: https://github.com/ihmeuw/vivarium_research_nutrition_optimization/pull/132


zmbc (approved):
> the link will still work as long it has the same name

It will even work if the file name changes!

alibow (approved):

PR #1453, opened 2024-02-27
Adding information about SQ-LNS scenario plans

SylLutze:
Outlining a plan for the targeted SQ-LNS scenarios. 

Highlights: 
- I think we skipped scenario 4 in the table before? Fixed that, but let me know if this was intentional 
- Too many scenarios so plan to address that 
- Updated the computational resources section to reflect scenario plan


docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
```diff
@@ -513,29 +513,36 @@ be the same for all subnational locations.
 2.5 Simulation scenarios
 ------------------------
 
-As of June, 2023, there are a total of 5 scenarios in the pregnancy simulation, :ref:`which can be found here <nutritionoptimizationpreg4.0>`. With the exception of the baseline scenario, all of the following child scenarios should be run on the outputs for each pregnancy scenario.
+As of June, 2023, there are a total of 5 scenarios in the pregnancy simulation, :ref:`which can be found here <nutritionoptimizationpreg4.0>`. With the exception of the baseline scenario, all of the following child scenarios should be run on the outputs for each pregnancy scenario unless otherwise noted, particularly for Wave III.
```
alibow, line 516: The number of pregnancy scenarios has gotten a bit confusing, but we actually ran wave 2 with just 4...

0. Zero
1. IFA
2. MMS
3. MMS/BEP

There are two others we were considering: IFA/BEP, nothing/BEP.

We'll probably want to think through whether we want to include these extras or not, while will influence the total number of child scenarios scoped out here.

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
```diff
@@ -655,6 +662,40 @@ Baseline values for :ref:`wasting treatment <intervention_wasting_tx_combined_pr
...
+From our prior work on this model, we have seen that the difference between antenatal 
+products and child products is much more significant that between different 
+child product combinations. For this reason, we think it makes more sense to 
+limit the pregnancy scenarios included in all child scenario runs. 
+
+We will test this implement plan with the following steps: 
+
+1. Do a run of all child scenarios for a single pregnancy scenario and determine the optimal combination of child products. Note that yes, this could change based on the antenatal scenario and this is a limitation. 
+2. For the optimal child scenario indentified, run all pregnancy scenarios. Ensure that in all cases, MMS and BEP are both scaled up prior to any child products. If this is not true, other scenarios might be needed, such as zero coverage of antenatal products or baseline. 
+3. Assuming the assumption that MMS and BEP always scale up first holds, run the child scenarios with only the MMS+BEP pregnancy scenario. Another option would be to run the child scenarios for both MMS alone and MMS+BEP.
```
alibow, line 683: Hmm, I'm not sure I'm entirely following this and would love to talk through in more detail -- I may not be tracking fully, but I am concerned about how restricting scenarios could affect how we've structured the ifa/child-deleted scenarios in the emulator. I think it will probably still work as long as we guarantee that there are some scenarios/combinations of scenarios that will always be present, but I think it's something we might have to keep an eye on to make sure this works

alibow: Basically, in the scheme that you've laid out, I think that we will also always need:
- whatever antenatal product is identified as optimal + zero child coverage
- zero antenatal + zero child
- baseline + baseline
- more thinking on which IFA scenarios would be required if we want optimizer to be relative to "baseline IFA" as we currently have it

alibow (approved):
So we should definitely discuss with the engineers, but I am not sure that it's a foregone conclusion that we _can't_ run the full set of scenarios here due to computational resources... I think I am seeing that it would be like 40 cluster-hours for all locations to do 4 targeted SQLNS implementations at 20 draws. Some alternatives to scenario reductions I think we should consider/discuss would also include:
- reducing number of draws - mean draw?
- not running targeted SQLNS in all locations? At least to start?
- revisiting variable timesteps?
- excluding the universal MAM treatment intervention? 

I think that what you've proposed here is probably the best path to reducing computational resources required, but just want to make sure we're aware of alternatives when we discuss too!

alibow:
also, I do think the scenario numbers were mislabeled before, but I wonder if maybe we should keep them mislabeled because it will make all of the old tables reference the wrong scenarios now?? I am not sure what is less annoying haha

albrja (approved):

alibow:
What you've added looks good to me!! I also think that fixing the mislabeled numbers as you've done looks good, but maybe we should add a note that we updated it and when we did so so that a future reader of the page can make sense of the old tables the reference the now outdated numeric scenario labels.

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
```diff
@@ -691,20 +718,20 @@ All possible observers and their default stratifications are outlined below. Req
...
 - 4 day timestep in the child simulation if no "timestep inrease strategy" (such as variable timesteps or YLD/YLL-only modeling strategy) is implemented and 28 day timestep if we do implement one of these strategies
 - Simulation takes 32 seconds per timestep. This assumption was informed by the "emulator test runs" of the wasting paper simulation that output only the necessary measures with no stratifications by year, age, or sex
 - Assume 15,000 threads available on all.q
 
-Under these assumptions, a full run of wave I will take 3.8 cluster-hours with 4-day timesteps and 0.6 cluster-hours with 28-day timesteps. A full run of wave II will take 23.5 cluster-hours with 4-day timesteps and 3.5 cluster-hours with 28-day timesteps.
+Under these assumptions, a full run of wave I will take 3.8 cluster-hours with 4-day timesteps and 0.6 cluster-hours with 28-day timesteps. A full run of wave II will take 15.0 cluster-hours with 4-day timesteps and 2.2 cluster-hours with 28-day timesteps. A full run of wave III assuming the higher 219 scenarios will take 17.9 cluster-hours with 4-day timesteps and 2.7 cluster-hours with 28-day timesteps.
```
aflaxman, line 726: I am surprised that wave II is 15 hours and wave III is not much more than that.

aflaxman (approved):

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
SylLutze: Good catch! This was assuming we cut down the number of scenarios significantly, I edited to be the full scenario list.

PR #1454, opened 2024-02-28
Adding targeted SQ-LNS wave 3 model run specs

SylLutze:
Ali - planning to discuss the ordering of these in our meeting today. Capturing the general model specs but can shuffle task order later.

### DIFF
--- a/docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
@@ -1407,8 +1407,8 @@ Wave III
     - Update to GBD 2021 data 
     - Baseline
     - Baseline
-    - 
-    - 
+    - Ethiopia location ONLY
+    - Should use 2021 GBD pregnancy model
   * - 15.0
     - Change child data to subnational
     - Baseline
@@ -1416,32 +1416,26 @@ Wave III
     - 
     - 
   * - 16.0
-    - Add targeted SQ-LNS intervention for wasting prevalence > 10%
+    - Add SQ-LNS intervention targeting for WAZ > 20%
     - Baseline
     - 8
     - 
     - Targeted SQ-LNS only
   * - 17.0
-    - Add other targeted SQ-LNS criteria
+    - Add additional SQ-LNS geographic targeting criteria
     - Baseline
     - 8
     - 
     - Targeted SQ-LNS only
   * - 18.0
-    - Determining scenarios needed part 1
+    - Add targeted SQ-LNS effect modification
     - Baseline
-    - All
+    - 8
     - 
-    - 
+    - Targeted SQ-LNS only
   * - 19.0
-    - Determining scenarios needed part 2
-    - All
-    - TBD - best child scenario from run 18
-    - 
-    - 
-  * - 20.0
     - Production runs
-    - TBD, likely only MMS alone and MMS+BEP
+    - All
     - All
     - 
     - 
@@ -1459,8 +1453,10 @@ Wave III
       2. YLDs (cause-specific)
       3. Cause state person time
       4. Cause state transition counts
-      5. Stunting state person time, stratified by antenatal intervention coverage
-      6. Wasting state person time, stratified by antenatal intervention coverage
+      5. Stunting state person time
+      6. Wasting state person time
+      7. Underweight state person time
+      8. Wasting transition counts 
     - * Age group
       * Sex
   * - 15.0
@@ -1468,8 +1464,10 @@ Wave III
       2. YLDs (cause-specific)
       3. Cause state person time
       4. Cause state transition counts
-      5. Stunting state person time, stratified by antenatal intervention coverage
-      6. Wasting state person time, stratified by antenatal intervention coverage
+      5. Stunting state person time
+      6. Wasting state person time
+      7. Underweight state person time
+      8. Wasting transition counts 
     - * Age group
       * Sex
       * Subnational location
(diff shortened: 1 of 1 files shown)


docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
```diff
@@ -1408,10 +1404,46 @@ Wave III
...
+    - 
+    - 
+  * - 15.0
+    - Change child data to subnational
+    - Baseline
+    - Baseline
+    - 
+    - 
+  * - 16.0
+    - Add targeted SQ-LNS intervention for wasting prevalence > 10%
```
alibow, line 1419: Does this mean effect modification? Or targeting algorithm for sublocations with wasting prevalence > 10%? We might want a separate model to implement the effect modification worked in here if we want to include it

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
```diff
@@ -1408,10 +1404,46 @@ Wave III
...
+    - Baseline
+    - 8
+    - 
+    - Targeted SQ-LNS only
+  * - 17.0
+    - Add other targeted SQ-LNS criteria
```
alibow, line 1425: Does this mean additional geography-based targeting algorithms we might be interested in?

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
```diff
@@ -1422,9 +1454,60 @@ Wave III
...
-  * - 14
-    - 
-    - 
+  * - 14.0
+    - 1. Deaths and YLLs (cause-specific)
+      2. YLDs (cause-specific)
+      3. Cause state person time
+      4. Cause state transition counts
+      5. Stunting state person time, stratified by antenatal intervention coverage
+      6. Wasting state person time, stratified by antenatal intervention coverage
```
alibow, line 1463: We'll probably want underweight state person time too to make sure that's still looking good! I also think we can probably drop the antenatal intervention coverage strata here?

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
```diff
@@ -1408,10 +1404,46 @@ Wave III
     - Spec. mods
     - Note
   * - 14.0
+    - Update to GBD 2021 data 
+    - Baseline
+    - Baseline
+    - 
+    -
```
alibow, line 1411: We should note that we'll want to have this be run on results from the pregnancy sim that have been updated to 2021... so we should probably plan to do the pregnancy sim update before the child sim since the child sim uses the pregnancy sim outputs

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
```diff
@@ -1422,9 +1454,60 @@ Wave III
...
+      6. Wasting state person time, stratified by antenatal intervention coverage
+    - * Age group
+      * Sex
+  * - 15.0
+    - 1. Deaths and YLLs (cause-specific)
+      2. YLDs (cause-specific)
+      3. Cause state person time
+      4. Cause state transition counts
+      5. Stunting state person time, stratified by antenatal intervention coverage
+      6. Wasting state person time, stratified by antenatal intervention coverage
```
alibow, line 1472: Same thing -- let's add underweight and wasting transition observers and I think we can drop antenatal coverage stratification

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
alibow: Oh let's also add wasting transition count observers

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
```diff
@@ -1422,9 +1454,60 @@ Wave III
...
+    - * Age group
+      * Sex
+      * Subnational location
+  * - 16.0
+    - 1. Deaths
+      2. Wasting state person time, stratified by SQ-LNS treatment coverage
+      3. Stunting state person time, stratified by SQ-LNS treatment coverage
+    - * Age group
+      * Sex
+      * Subnational location
```
alibow, line 1482: I was thinking that we could eventually stratify just by "targeted or not" rather than subnational location, which will hopefully make our observers run faster - I guess that's equivalent to what you included as "SQLNS treatment coverage", right? Maybe we can verify in this model that the coverage is as we expect it to be by subnational location and then we can drop the subnational stratification moving forward to increase speed/efficiency

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
```diff
@@ -1422,9 +1454,60 @@ Wave III
...
+  * - 18.0
+    - 1. Deaths and YLLs (non-cause-specific)
+      2. YLDs (all-cause observer only)
+      3. Count of incident SAM cases stratified by SAM treatment coverage
+      4. Count of incident MAM cases stratified by MAM treatment coverage
+      5. Stunting state person time stratified by SQ-LNS utilization
```
alibow, line 1495: SQLNS utilization versus treatment coverage used differently throughout this table... I think we do have different definitions of these terms in our SQLNS intervention page, so we may want to be consistent for clarity here

alibow (approved):

albrja (approved):

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
```diff
@@ -1422,9 +1448,57 @@ Wave III
...
+    - * Age group
+      * Sex
+      * Targeted SQ-LNS location vs not
+  * - 18.0
+    - 1. Deaths
+      2. Wasting state person time, stratified by SQ-LNS treatment coverage
+      3. Stunting state person time, stratified by SQ-LNS treatment coverage
+    - * Age group
+      * Sex
+      * Targeted SQ-LNS location vs not
```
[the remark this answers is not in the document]
alibow: Otherwise all the changes look good to me!

PR #1456, opened 2024-02-29
Update SQ-LNS targeting docs

SylLutze:
Still a lot of to-dos but this should reflect our conversation

### DIFF
--- a/docs/source/models/intervention_models/lipid_based_nutrient_supplements/index.rst
+++ b/docs/source/models/intervention_models/lipid_based_nutrient_supplements/index.rst
@@ -229,9 +229,33 @@ SQ-LNS effects on wasting will persist until 18 months of age and effects on stu
 Targeted SQ-LNS
 ~~~~~~~~~~~~~~~
 
+There is not a gold standard currently for how to target SQ-LNS. This 
+is a relatively new intervention and so the best metric for targeting 
+is still being debated. Simulation is especially well suited for these 
+types of questions, and so we will try several different SQ-LNS targeting 
+criteria. 
+
+For all criteria, a csv file with the subnational locations and whether or 
+not they are included in targeting will be provided.
+
+.. list-table:: SQ-LNS Targeting Criteria
+  :widths: 5 15 15 15
+  :header-rows: 1
+
+  * - Targeting Number 
+    - Targeting Criteria
+    - Description
+    - Notes
+  * - 1 
+    - WAZ > 20%
+    - Percent of the population who is underwight in each subnational location
+    - Based on suggestion from Indi Trehan that underweight data is sometimes more available than height based data
+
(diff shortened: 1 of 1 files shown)


docs/source/models/intervention_models/lipid_based_nutrient_supplements/index.rst
```diff
@@ -229,17 +229,43 @@ SQ-LNS effects on wasting will persist until 18 months of age and effects on stu
 Targeted SQ-LNS
 ~~~~~~~~~~~~~~~
 
+There is not a gold standard currently for how to target SQ-LNS. This
```
alibow, line 232: Let's cite and maybe summarize the current WHO guidelines on targeting SQLNS! There is definitely a recommended practice, it's just a little vague!

docs/source/models/intervention_models/lipid_based_nutrient_supplements/index.rst
```diff
@@ -229,17 +229,43 @@ SQ-LNS effects on wasting will persist until 18 months of age and effects on stu
...
+.. list-table:: SQ-LNS Targeting Criteria
+  :widths: 5 15 15 15
+  :header-rows: 1
+
+  * - Targeting Number 
+    - Targeting Criteria
+    - Description
+    - Notes
+  * - 1 
+    - WAZ > 20%
```
alibow, line 250: Sorry I was unclear when I was talking about this target! I'm thinking the best first step will be a WAZ-based threshold that results in ~20% of the population being targeted (X percent of the population with WAZ < -2) -- I haven't figured out what threshold approximately aligns with that yet though! The 20% of the population figure comes from this thread: https://ihme.slack.com/archives/C02944X688L/p1708123785649299

Basically, WHO recommends giving SQLNS to those who are severely food insecure, which is ~20% of the population based on the FAO data. Therefore, we will use underweight burden as a proxy for severe food insecurity because we have WAZ exposure but not food insecurity in our model ready to go. And we're choosing underweight rather than stunting or wasting because it is a better representation of _concurrent_ stunting and wasting, which is more dangerous than stunting or wasting alone in addition to the ease of measurement argument that you've already included here.

I think it will be great to have this reasoning written down as we move forward so we have ready to explain these decisions :)

alibow: ^ Something else that we will want to specify is among what age group we are assessing WAZ exposure for targeting.  We could do all kids under 5 (simple), 6-59 months (those who will benefit from SQLNS), or 6-24 months (those who are actually eligible for SQLNS). I think all kids under 5 is best for simplicity, but definitely worth considering other opinions too! @aflaxman fyi

aflaxman (approved):

PR #1464, opened 2024-03-08
Update the order that noises are applied in PRL concept model

pletale:
[JIRA ticket](https://jira.ihme.washington.edu/browse/SSCI-1665)

### DIFF
--- a/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
@@ -3102,10 +3102,8 @@ Noise should be added in the order below.
 #. Duplications
 
    a. Guardian-based dupilcation
-#. Omissions
-
-   a. Do not respond (targeted omission)
-   b. Omit a row (simple omission)
+#. Do not respond (targeted omission)
+#. Omit a row (simple omission)
 #. Missing Data
 #. Incorrect Selection
 #. Copy from Within Household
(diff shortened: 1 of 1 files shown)


docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3140,32 +3140,32 @@ Noise should be added in the order below.
       - Pseudopeople name
     * - "Borrowed" SSN
       - Borrow a social security number
+    * - Guardian-based duplication
+      - Duplicate with guardian
+    * - (Simple) Duplication
+      - (not yet implemented)
```
NathanielBlairStahn, line 3146:
```suggestion
```
Let's omit this noise type for now since it's unclear what it means. If we include something like this, it should probably happen around the same time as "simple omission," in particular, after "Do not respond." But we haven't decided how we want this to work, so let's not even mention it.

docs/source/models/concept_models/vivarium_census_synthdata/concept_model.rst
```diff
@@ -3099,22 +3099,22 @@ noise added to the data. We currently divide noise into two types:
 Noise should be added in the order below.
 
 #. "Borrowed" SSN (happens in simulation NOT noise functions)
+#. Duplications
+
+   a. Guardian-based dupilcation
```
NathanielBlairStahn, line 3104:
```suggestion
#. Guardian-based dupilcation
```
Let's not bother trying to group all duplication noise types together. If we come up with a different kind of duplication, we should consider it a separate noise type, and it may happen in a different order than implied here.

aflaxman (approved):

NathanielBlairStahn (approved):

PR #1471, opened 2024-03-20
Creating page on engineering files

SylLutze:
I also included the engineers as reviews to check some of my work here. This is not intended to be enough information to actually do a lot of anything - just higher level background and reference information. I will also create pages for running the sim, etc. that have more depth and instructions.

### DIFF
--- a/docs/source/model_design/programming_vivarium/overview_engineering_files/index.rst
+++ b/docs/source/model_design/programming_vivarium/overview_engineering_files/index.rst
@@ -29,12 +29,159 @@ Overview of Engineering Files
 
 .. contents::
    :local:
-   :depth: 1
-
-.. todo::
-
-  Add to this page. Include explanations for the relevant engineering files (model specs, branches file, loader.py, etc.) that includes when the functions are called, when you might need to look at them, what information is stored where, and any other relevant information for RT based tasks. 
-
-  Integrate links to: `YAML basics <https://vivarium.readthedocs.io/en/latest/concepts/model_specification/yaml_basics.html#model-specification-yaml-concept>`_, `model specs file <https://vivarium.readthedocs.io/en/latest/concepts/model_specification/index.html>`_, and the `branches file <https://vivarium-cluster-tools.readthedocs.io/en/latest/branch.html>`_. 
+   :depth: 2
 
 
+This document is designed to provide an overview of the files found in the 
+engineering repository and some examples of when you might need to use 
+each file. This is NOT inclusive of files as some are less relevant 
+for research team members, and the files included change between models. 
+
+This page is organized in sections by the different folders found in engineering 
+repositories and then the pages within folders, as relevant. 
+
+Components
+----------
+
+The :code:`components` folder contains files on the information for each 
+modeling "component". This includes things like disease models, 
+risk factors, pregnancies, the health care system, or 
+interventions. 
+
+The files included will be specific to your project rather than generalizable. 
+Additionally, these files will mostly be maintained and written by the 
+engineering team. For those reasons, we won't elaborate on these files here. 
+
+Constants
+---------
+
+The :code:`constants` folder contains input information for the model. This includes 
+things like data values, locations, file paths, draw counts, column 
+names, and outputs needed. 
+
+There are a few reasons you might use files from this folder - changing data inputs, 
+adjusting information for the artifact, or validating file paths. The below files 
+are not all inclusive, but represent most of what you should be aware of. 
+
+data_keys
++++++++++
+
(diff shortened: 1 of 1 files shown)


aflaxman (approved):
Thanks for documenting all of this!  I wonder if we can convince the engineers to make it simpler...

zmbc (approved):
LGTM. I can see this eventually migrating to be documentation of the repository template itself.

alibow:
This is great! It makes me realize that there are many more things that I am unclear on how they fit into the task shifting plan -- first, the results_processing file... is this something that researchers are expected to write/maintain or will that stay with engineering? Additionally, I don't see anything about observers on this page... is that something that we would want called out? That's another thing that I am unclear on who will own (cc @aflaxman @rmudambi)

alibow:
Also, Rajan brought up the "changelog" file in our meeting yesterday -- is that something that we want to call out here? I think I personally prefer to keep that sort of documentation in our concept model doc as we've already been maintaining on the RT and then not have duplicate records of it, but maybe we'd like to discuss pros/cons as a team?

alibow (approved):

docs/source/model_design/programming_vivarium/overview_engineering_files/index.rst
```diff
@@ -29,12 +29,159 @@ Overview of Engineering Files
...
+names, and outputs needed. 
+
+There are a few reasons you might use files from this folder - changing data inputs, 
+adjusting information for the artifact, or validating file paths. The below files 
+are not all inclusive, but represent most of what you should be aware of. 
+
+data_keys
++++++++++
+
+The :code:`data_keys.py` file dictates which keys are run to make the artifact. Usually, the
```
hussain-jafari, line 69: I think it would be more correct to say: Any keys not included in the MAKE_ARTIFACTS_KEY_GROUPS list at the bottom of the file will be ignored when making artifacts.

docs/source/model_design/programming_vivarium/overview_engineering_files/index.rst
```diff
@@ -29,12 +29,159 @@ Overview of Engineering Files
...
+interventions. 
+
+The files included will be specific to your project rather than generalizable. 
+Additionally, these files will mostly be maintained and written by the 
+engineering team. For those reasons, we won't elaborate on these files here. 
+
+Constants
+---------
+
+The :code:`constants` folder contains input information for the model. This includes
```
hussain-jafari, line 58: The results.py file in this folder is also where we tell process_results what we expect the output to look like (ie observers and stratifications). This file needs to be updated if our outputs change (but we still need to decide who will be doing this).

docs/source/model_design/programming_vivarium/overview_engineering_files/index.rst
```diff
@@ -29,12 +29,159 @@ Overview of Engineering Files
...
++++++++++
+
+The :code:`data_keys.py` file dictates which keys are run to make the artifact. Usually, the 
+list of data keys is at the bottom of the file. More information on how and when to 
+edit this can be found on the :ref:`artifact building page <artifact_building_rt>`. 
+
+metadata
+++++++++
+
+The :code:`metadata.py` file will mainly need to be edited to change the locations for artifact
```
hussain-jafari, line 76: It might be obvious, but should we say that this a file for storing metadata related constants in addition to the location information? The locations in this file also define which locations we _can_ make artifacts for. We specify which locations we'd like to make artifacts for from the command line, but artifact building will fail if we specify a location that isn't contained in this file.

docs/source/model_design/programming_vivarium/overview_engineering_files/index.rst
```diff
@@ -29,12 +29,159 @@ Overview of Engineering Files
...
+This page is organized in sections by the different folders found in engineering 
+repositories and then the pages within folders, as relevant. 
 
+Components
+----------
 
+The :code:`components` folder contains files on the information for each 
+modeling "component". This includes things like disease models, 
+risk factors, pregnancies, the health care system, or 
+interventions.
```
hussain-jafari, line 49: Related to what Ali said, I think it'd be useful if we included observers in this list.

hussain-jafari:
Very cool! I left some suggestions but looks great otherwise. @alibow I'm also unsure about ownership of results processing and observer changes, but I think that walking RT through how to make updates to results processing and how to check what outputs existing observers are creating would be relatively fast and easy. I think it'd also be worth documenting what VPH observers are recording somewhere.

hussain-jafari (approved):

PR #1480, opened 2024-03-29
Links to age aggregated BMI/hgb data

alibow:
Linked to age aggregated data and also:
- Removed age aggregated line for hemoglobin < 70 because that parameter is actually not used in the child sim
- Added a line to eventually link the 2021 location-specific SQ-LNS effects


SylLutze (approved):
Nice

aflaxman (approved):

PR #1490, opened 2024-04-24
Updating the Child 2021 V&V and requesting PEM runs formally

SylLutze:


alibow (approved):
Thanks for putting this together! You might have seen in slack, but to fill you in on the end of our conversation in the RT/SE check-in yesterday, we decided that I will look into the PEM issues a bit more (via interactive sim and more investigation of DW magnitudes) before the engineers continue investigating on their end, so we may have more detail to add to our description of the PEM issue before launching 13.2

SylLutze:
> Thanks for putting this together! You might have seen in slack, but to fill you in on the end of our conversation in the RT/SE check-in yesterday, we decided that I will look into the PEM issues a bit more (via interactive sim and more investigation of DW magnitudes) before the engineers continue investigating on their end, so we may have more detail to add to our description of the PEM issue before launching 13.2

Sounds good! I'll leave this PR open for now then so we can add more information as we figure it out.

PR #1491, opened 2024-04-26
Updating the MAM subcategory exposure info

SylLutze:
Includes both: 
- Subnational approach details 
- Updates to the MAM subcategory data and RR's with new links to csv's and notebooks + an explanation

### DIFF
--- a/docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
@@ -527,13 +527,18 @@ was in `this PR <https://github.com/ihmeuw/vivarium_research_nutrition_optimizat
 Similarly, MAM subcategory exposure, the percent of children in "worse" MAM 
 and "better" MAM, will be national only. We analyzed the rate of change 
 between regions in worse MAM fraction and found that it was generally 
-consistent between regions, while pooling for age and sex. 
+consistent between regions, while pooling for age and sex. The youngest age 
+group, neonatal, was excluded prior to pooling the data. 
 
 There were two regions where this was not true - Addis Ababa in Ethiopia and 
-Gilgit-Baltistan in Pakistan. However, these regions had many 
-NaNs in processed data. This is a result of too few data points with MAM children to 
-get results from the function. Since both regions made up a small percent of their national 
-children under 5 (about 1% in each country), we decided this was likely data noise. 
+Gilgit-Baltistan in Pakistan. These regions had lower calculated exposure fractions 
+for worse MAM as the population WHZ distributions for these locations as modeled
+had very low density below z-scores of -2.5 (note that the 
+:code:`risk_distributions.EnsembleDistribution` functions used for modeling these curves 
+do not return values below the 0.1st percentile). Since both regions made up a small 
+percent of their national children under 5 (about 1% in each country), we decided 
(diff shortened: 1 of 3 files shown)


aflaxman (approved):

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
```diff
@@ -524,6 +524,21 @@ noted in the location data. Lastly, many location/age/sex groups had fewer than
...
+Similarly, MAM subcategory exposure, the percent of children in "worse" MAM 
+and "better" MAM, will be national only. We analyzed the rate of change 
+between regions in worse MAM fraction and found that it was generally 
+consistent between regions, while pooling for age and sex. 
+
+There were two regions where this was not true - Addis Ababa in Ethiopia and 
+Gilgit-Baltistan in Pakistan. However, these regions had many 
+NaNs in processed data. This is a result of too few data points with MAM children to 
+get results from the function. Since both regions made up a small percent of their national 
+children under 5 (about 1% in each country), we decided this was likely data noise.
```
alibow, line 536:
```suggestion
There were two regions where this was not true - Addis Ababa in Ethiopia and 
Gilgit-Baltistan in Pakistan. These regions had lower calculated exposure fractions 
for worse MAM as the population WHZ distributions for these locations as modeled
had very low density below z-scores of -2.5 (note that the 
:code:`risk_distributions.EnsembleDistribution` functions used for modeling these curves 
do not return values below the 0.1st percentile). Since both regions made up a small 
percent of their national children under 5 (about 1% in each country), we decided 
the variation could be noise and, either way, that not including subnational variation 
in this parameter was an acceptable limitation. 
```

So even if the EnsembleDistribution function returned non-missing values below 0.001, the values would still be less than 0.001 and not make much of a difference in the overall worse frac calculated value -- basically I don't think the NaNs caused the low worse frac values, but were just a byproduct of the higher WHZ mean values that caused low density in the worse frac exposure region (and then the NaNs will cause rounding errors) - I tried to highlight that while still acknowledging the limitations of the function. I also tried to convey that even if the outliers were not noise that we still think it's an acceptable limitation not to model it.

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
```diff
@@ -524,6 +524,21 @@ noted in the location data. Lastly, many location/age/sex groups had fewer than
 50 children, leading to lack of confidence in results. The analysis of this 
 was in `this PR <https://github.com/ihmeuw/vivarium_research_nutrition_optimization/pull/134>`_. 
 
+Similarly, MAM subcategory exposure, the percent of children in "worse" MAM 
+and "better" MAM, will be national only. We analyzed the rate of change 
+between regions in worse MAM fraction and found that it was generally 
+consistent between regions, while pooling for age and sex.
```
alibow, line 530: Is this pooling among all children under 5 or did we end up excluding the neonates and/or 1-5 month olds?

alibow (approved):

docs/source/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.rst
SylLutze: We removed the neonatal age group, but not 1-5 month olds. I'll update the language to include that.

PR #1496, opened 2024-05-16
Start Docs for MNCNH Portfolio Project

SylLutze:
- Start concept model docs page 
- Includes rough overview of the project in 4 waves, I'd like to get people's reaction's to this conception of the stages of the work. Would we like this to be organized in another way? 
- More PRs will follow building out the other sections of the docs, trying to keep PRs manageable


alibow (approved):
Looks like you've alluded to this, but just noting that it may make sense to go through these waves for a subset of risks/causes/interventions first rather than complete each wave fully before moving on to the next. This would allow us to get a complete result for a given intervention(s) earlier on in the process, which may be desirable for providing something to that Gates Foundation in a timely manner as well as prepping for output processing, etc. etc.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -0,0 +1,150 @@
...
+
+.. contents::
+  :local:
+  :depth: 1
+
+1.0 Overview
+++++++++++++
+
+This document is the overall page for the MNCNH Portfolio simulation and 
+contains information that relates to all modeled subcomonents included in
```
NathanielBlairStahn, line 46:
```suggestion
contains information that relates to all modeled subcomponents included in 
```

aflaxman (approved):

PR #1504, opened 2024-06-11
Non-log-linear risk exposure docs

aflaxman:
Additions to the risk factors docs regarding how to use the new non-log-linear relative risks released in GBD 2021


docs/source/model_design/vivarium_model_components/risk_factors/risk_effects/index.rst
```diff
@@ -86,6 +86,55 @@ to discuss reasonable assumptions with subject matter experts to determine the
 most appropriate measure to which to apply the GBD risk effects in our vivarium
 simulations.
 
+In GBD 2021, some continuous risk exposures were modeled with general
```
zmbc, line 89:
```suggestion
Starting in GBD 2021, some continuous risk exposures were modeled with general
```

docs/source/model_design/vivarium_model_components/risk_factors/risk_effects/index.rst
```diff
@@ -86,6 +86,55 @@ to discuss reasonable assumptions with subject matter experts to determine the
...
+relationships between the exposure level and the relative risk (going
+beyond the log-linear relationship assumed for previous iterations).
+Interpreting the GBD estimates is straightforward, once you have
+chased down all of the necessary definitions.  The relevant estimates
+include a column for exposure level, as well as columns for 500 draws
```
zmbc, line 94:
```suggestion
include a column for point exposure level, as well as columns for 500 draws
```

docs/source/model_design/vivarium_model_components/risk_factors/risk_effects/index.rst
```diff
@@ -86,6 +86,55 @@ to discuss reasonable assumptions with subject matter experts to determine the
...
+the TMREL as well.  The precise calculation to go from exposure levels
+and GBD-recorded risks to a function suitable for use as
+:math:`f_{rr}` as defined below are perhaps most clearly represented
+as python code:
+
+.. code-block:: python
+
+    def get_rr(draw_id):
+        # select rows for this sex and age_group
+        t1 = df[
```
zmbc, line 107: Let's rename these `t1`, `t2` etc variables.

docs/source/model_design/vivarium_model_components/risk_factors/risk_effects/index.rst
```diff
@@ -86,6 +86,55 @@ to discuss reasonable assumptions with subject matter experts to determine the
...
+        import scipy.interpolate
+        x = np.array(t3.index) # df.exposure
+        y = np.array(t3.values)  # df.draw_n
+        f_rr = scipy.interpolate.interp1d(x, y, kind='linear', bounds_error=False, fill_value=(min(y), max(y)))
+
+        # pick a tmrel between tmred.min and tmred.max --- for certain risk factors, the modeling team uploads a model for this, too --- shiny tool has information about this
+        # and calculate relative risk at tmrel
+        tmrel = np.random.uniform(tmred.min, tmred.max)
+        rr_at_tmrel = f_rr(tmrel)
+        t4 = t3 / rr_at_tmrel
```
zmbc, line 129: FWIW, it's not at all obvious to me that this rescaling is necessary for our purposes. I think if you didn't do it (and clipped to `rr_at_tmrel` instead of 1), you'd get a different PAF from GBD, but the derived quantities we actually use ($i$ in the equations below) would be the same. Does that seem right to you?

I see why we might prefer to still do it, to have PAFs compatible with GBD, but I'd like to note it somewhere. And when we are calculating our own PAFs it just adds an extra step that may be confusing. I think this is a broader conversation that we should continue beyond this PR.

zmbc (commented):
Nice!

I'd like to make this code easier to follow -- let me know if you'd like me to take a stab at it.

aflaxman:
> Nice!
> 
> I'd like to make this code easier to follow -- let me know if you'd like me to take a stab at it.

Please do!

docs/source/model_design/vivarium_model_components/risk_factors/risk_effects/index.rst
aflaxman: what do you mean by point?

zmbc: May not be the right word. I originally thought that this would be categorical, i.e. there would be _ranges_ of exposure. But it seems there are just exact values of exposure, which you interpolate.

docs/source/model_design/vivarium_model_components/risk_factors/risk_effects/index.rst
zmbc: That is fair. I think this is a longer-term thing to improve.

Basically, I'd like that if someone isn't using GBD data, they don't need to have a PAF at all -- it is an extra step that is only introduced by the fact that GBD doesn't give us exactly what we need for simulation.

aflaxman: My thinking is that there will always be a need for "risk deleting" the rate to be modified --- we might as well call it a PAF.

zmbc: But for it to _really_ be a PAF we need to have a TMREL, which might not be relevant to our simulation's research question.

docs/source/model_design/vivarium_model_components/risk_factors/risk_effects/index.rst
```diff
@@ -145,6 +199,8 @@ Where,
    a desired unit
  - :math:`p(e)` is probability density function used to calculate the probability 
    of given risk exposure level e
+ - :math:`f_{rr}(e)` is function capturing the relationship between the exposure level and the relative risk at that exposure level (for log-linear relative risks, :math:`f_{rr}(e) = rr^{max(e-tmrel,0)/scalar}.`)
```
NathanielBlairStahn, line 202:
```suggestion
 - :math:`f_{rr}(e)` is function capturing the relationship between the exposure level and the relative risk at that exposure level (for log-linear relative risks, :math:`f_{rr}(e) = rr^{max(e-tmrel,0)/scalar}`)
```

docs/source/model_design/vivarium_model_components/risk_factors/risk_effects/index.rst
```diff
@@ -86,6 +86,55 @@ to discuss reasonable assumptions with subject matter experts to determine the
...
+        # pick a tmrel between tmred.min and tmred.max --- for certain risk factors, the modeling team uploads a model for this, too --- shiny tool has information about this
+        # and calculate relative risk at tmrel
+        tmrel = np.random.uniform(tmred.min, tmred.max)
+        rr_at_tmrel = f_rr(tmrel)
+        t4 = t3 / rr_at_tmrel
+
+        # TODO: see if GBD would also benefit from clipping these RRs
+        t5 = np.clip(t4, 1.0, np.inf)
+
+        y = np.array(t5.values)
```
SylLutze, line 134: I think adding some comments in the code here as well would be helpful.

SylLutze (approved):

docs/source/model_design/vivarium_model_components/risk_factors/risk_effects/index.rst
```diff
@@ -126,9 +175,14 @@ The mathematical expressions are mainly fall into two categories:
      - :math:`PAF = \frac{E(RR_e)-1}{E(RR_e)}`
      - :math:`E(RR_e) = p \times RR + (1-p)`
  - risk exposure is continuous distributed:
-     - :math:`i = i \times (1-PAF) \times rr^{max(e−tmrel,0)/scalar}`
-     - :math:`PAF = \frac{E(RR_e)-1}{E(RR_e)}`
-     - :math:`E(RR_e) = \int_{lower}^{upper}rr^{max(e−tmrel,0)/scalar}p(e)de`
+     - risk effect has a log-linear "dose-response" relationship with exposure:
+         - :math:`i = i \times (1-PAF) \times rr^{max(e-tmrel,0)/scalar}`
```
NathanielBlairStahn, line 179: This line existed previously, so perhaps this is out of scope for this PR, but the notation here is confusing because the variable $i$ is being used for two different things: On the LHS, $i$ is the individual incidence hazard for a single simulant, whereas on the RHS, $i$ is the population-level incidence rate.

docs/source/model_design/vivarium_model_components/risk_factors/risk_effects/index.rst
```diff
@@ -126,9 +175,14 @@ The mathematical expressions are mainly fall into two categories:
...
+         - :math:`PAF = \frac{E(RR_e)-1}{E(RR_e)}`
+         - :math:`E(RR_e) = \int_{lower}^{upper}rr^{max(e-tmrel,0)/scalar}p(e)de`
+     - risk effect has a non-log-linear relationship with exposure:
+         - :math:`i = i \times (1-PAF) \times f_{rr}(e)`
```
NathanielBlairStahn, line 183: Same comment here: The variable $i$ is being used for two different things, so there should actually be a second variable defined below.

docs/source/model_design/vivarium_model_components/risk_factors/risk_effects/index.rst
```diff
@@ -86,6 +86,55 @@ to discuss reasonable assumptions with subject matter experts to determine the
...
+Interpreting the GBD estimates is straightforward, once you have
+chased down all of the necessary definitions.  The relevant estimates
+include a column for exposure level, as well as columns for 500 draws
+of relative risk values at each exposure level.  The GBD 2021 PAF
+calculator often selected a TMREL for each draw from a uniform
+distribution, but for some risk factors, analysts provided draws for
+the TMREL as well.  The precise calculation to go from exposure levels
+and GBD-recorded risks to a function suitable for use as
+:math:`f_{rr}` as defined below are perhaps most clearly represented
+as python code:
```
NathanielBlairStahn, line 101: Can we add a link to the Central Comp code that you looked at to figure out what they did?

zmbc (approved):

PR #1506, opened 2024-06-14
Cleanup code for Python demonstration of non-log-linear risk curve data

zmbc:
Makes this more complete, so it does not depend on variables not defined; adds comments; removes the use of vivarium_inputs which only worked with a custom code change; substitutes more meaningful variable names.


zmbc:
In case interesting, here's what the draws of the resulting risk curve look like for SBP -> IHD:

image

aflaxman (approved):
Great improvement, thanks!

NathanielBlairStahn:
@zmbc Can we include that picture of the risk curve in the docs?

SylLutze (approved):

docs/source/model_design/vivarium_model_components/risk_factors/risk_effects/index.rst
```diff
@@ -102,38 +102,76 @@ as python code:
...
+    # for certain risk factors, the modeling team uploads a model for this with TMREL draws --
+    # those should be used instead of this, when available!
+    tmrel = np.random.uniform(risk.tmred.min, risk.tmred.max)
+    rr_at_tmrel = raw_relative_risk_function(tmrel)
+    normalized_relative_risk_draw = relative_risk_draw / rr_at_tmrel
+
+    # Clip relative risk to 1, since the TMREL is supposed to represent minimum risk
+    # NOTE: It is not clear if this is always appropriate, since TMRELs are across all
+    # risk-cause pairs involving this risk, not only the pair we are modeling
+    # TODO: see if GBD would also benefit from clipping these RRs
```
NathanielBlairStahn, line 160: I'm not sure what this comment means... But along with all the other discussion, it makes me think the answer to my question in the doc session ("Is it unclear how GBD computes the PAF?") should have been "Yes." Let me make my question more precise:

1. Does GBD clip the RRs to 1 when computing the PAF? (My understanding is that, based on the code, it looks like the answer is "no.")

2. When we clip to 1, do we get the same PAF as GBD?

   - If the answer is "yes", then why are we getting the same result as GBD with a different RR curve (assuming the answer to Question 1 is "no")?
   - If the answer is "no", then we should note that somewhere (but it seems like the answer is "yes" since Abie's goal was to match GBD's PAF).

3. If we don't clip to 1, do we know for sure that we'd get a negative PAF sometimes, or is that just a hypothetical possibility?

Maybe I'm misunderstanding something, but it seemed like we still had some outstanding questions about this issue.

With regard to this PR, is it possible to clarify what this comment means?

docs/source/model_design/vivarium_model_components/risk_factors/risk_effects/index.rst
```diff
@@ -102,38 +102,76 @@ as python code:
...
+    # pick a tmrel between tmred.min and tmred.max and calculate relative risk at tmrel
+    # for certain risk factors, the modeling team uploads a model for this with TMREL draws --
+    # those should be used instead of this, when available!
+    tmrel = np.random.uniform(risk.tmred.min, risk.tmred.max)
+    rr_at_tmrel = raw_relative_risk_function(tmrel)
+    normalized_relative_risk_draw = relative_risk_draw / rr_at_tmrel
+
+    # Clip relative risk to 1, since the TMREL is supposed to represent minimum risk
+    # NOTE: It is not clear if this is always appropriate, since TMRELs are across all
+    # risk-cause pairs involving this risk, not only the pair we are modeling
```
NathanielBlairStahn, line 159: @zmbc I didn't really know what this comment meant until you explained it in the doc session today. It would be good to put that explanation somewhere in the docs, but that should probably be a different PR.

NathanielBlairStahn: And again, if it's not clear whether this is always appropriate, then why do you think it's clear what GBD actually did?

docs/source/model_design/vivarium_model_components/risk_factors/risk_effects/index.rst
zmbc: Yes I see your point -- I don't think it is totally clear what GBD did. IIUC @aflaxman's conclusion was that he got similar numbers to GBD when he _did_ clip, but not when he didn't? That is confusing if we don't see GBD doing clipping.

aflaxman: > Ok, looking back at [the Slack thread](https://ihme.slack.com/archives/CQ8MGENNA/p1717510424955569?thread_ts=1717448354.571769&cid=CQ8MGENNA) it looks like we matched GBD even without clipping. Given that and the conceptual problems with clipping, I'm not sure I see any reason to clip.

The PAF of high sodium on stomach cancer matches up better with clipping --- the median is negative without clipping, but positive in GBD.

zmbc: Ok let me take a look at that. I wonder if it has to do with how sodium is "weird"?

I can confirm that GBD does not clip, at least not uniformly, because it does estimate negative PAFs:

```python
import pandas as pd
import gbd_mapping, vivarium_gbd_access.gbd
sudan_alcohol_pafs = vivarium_gbd_access.gbd.get_paf(gbd_mapping.risk_factors.high_alcohol_use.gbd_id, 522, 2021)
pd.Series(
    sudan_alcohol_pafs[(sudan_alcohol_pafs.sex_id == 2) & (sudan_alcohol_pafs.cause_id == gbd_mapping.causes.ischemic_heart_disease.gbd_id)]
        .filter(like='draw').values.reshape(-1)
).hist(bins=100, log=True)
```

image

aflaxman: Is this showing a PAF of -350% for one draw??

zmbc: I'd assume -3.5%, though I'm not sure exactly what the units are.

docs/source/model_design/vivarium_model_components/risk_factors/risk_effects/index.rst
zmbc: I've added a better comment here.

docs/source/model_design/vivarium_model_components/risk_factors/risk_effects/index.rst
zmbc: For posterity: answers to both questions 1 & 2 were "yes."

PR #1519, opened 2024-08-02
Minimum viable model page

SylLutze:
Very similar to what was reviewed on Monday together. Step 6 is new and I added some about what "minimum" is. Let me know if this is a good place for this to live, or if another place would be more appropriate!


aflaxman (approved):
This is a great addition!

zmbc (approved):
This is great!

PR #1524, opened 2024-08-27
Adding updated pregnancy model

SylLutze:
Noting that a lot of this is a copy and paste from the nutrition optimization pregnancy model page, with some edits to remove things that will be covered in the intrapartum model instead.


docs/source/models/other_models/pregnancy/gbd_2021_mncnh/index.rst
```diff
@@ -0,0 +1,335 @@
...
+
+We will model pregnancy as a characteristic of women of reproductive age in our simulations. We will inform the *incidence* of pregnancy using the age-specific fertility and stillbirth to live birth ratio covariates from GBD. We will inform the *duration* of pregnancy using the GBD 2019 exposure distribution of gestational age.
+
+.. image:: diagram.svg
+
+.. note::
+
+  Maternal disorders are entirely handled in the intrapartum model for wave 1 of this project. At current there are no risk factors (other than age) included either. 
+
+  We will add risk factors to this model (BMI, blood pressure, hemoglobin) and may also add some maternal disorders, such as preeclampsia and anemia. Though we have not determined where we will have simulants accumulate YLDs/deaths (might all be in the intrapartum model).
```
aflaxman, line 119: blood glucose level for gestational diabetes, too, probably

docs/source/models/other_models/pregnancy/gbd_2021_mncnh/index.rst
```diff
@@ -0,0 +1,335 @@
...
+  * - p
+    - Pregnant
+    - 
+  * - output
+    - Output data 
+    - Recording of simulant data for use in the next phase of the simulation, the intrpartum model.
+
+.. note::
+
+  For this model, we will not be using "time" or "time steps" in the traditional sense. Rather, simulants will be initialized and then go through the decision tree outlined on the :ref:`MNCNH Portfolio project <2024_concept_model_vivarium_mncnh_portfolio>` page. Following that, data will be recorded for use in the intrapartum model.
```
aflaxman, line 140: Do we researchers expect that this will use the "individual clock" affordance that the engineers worked on last year?

docs/source/models/other_models/pregnancy/gbd_2021_mncnh/index.rst
```diff
@@ -0,0 +1,335 @@
...
+    - (incidence_c995 + incidence_c374) / (ASFR + ASFR * SBR + incidence_c995 + incidence_c374)
+    - 
+  * - Full term
+    - 1 - probability_partial_term
+    - 
+
+Sex of infant
+~~~~~~~~~~~~~~~
+
+For pregnancies that result in live birth or stillbirth outcomes, infant sex should be determined and recorded acording to the probability of male sex shown in the table below (probability of female birth is equal to 1 minus the probability of male birth). This should be performed at the start of pregnancy (transition from np to p states) or upon initialization into the p state. These sex ratios were calculated using the live births by sex 2020 GBD covariate (ID 1106), `shown here <https://github.com/ihmeuw/vivarium_research_nutrition_optimization/blob/data_prep/data_prep/Live%20births%20by%20sex.ipynb>`_. Note that there is no variation by draw in this parameter.
```
aflaxman, line 274: Limitation to note, but not to address: only singleton births (no twins, triplets, etc)

aflaxman: oh, i see you have noted it below already :)

aflaxman (approved):

docs/source/models/other_models/pregnancy/gbd_2021_mncnh/index.rst
SylLutze: That isn't quite how I was thinking about this. With the individual clocks time still exists in the model, just some simulants can take different size time steps than other simulants. Here, we don't really have "time" in the model, you just jump to "birth" and then we record separately the gestational age and calculate YLDs.

docs/source/models/other_models/pregnancy/gbd_2021_mncnh/index.rst
```diff
@@ -0,0 +1,335 @@
...
+
+  * - State
+    - Value
+    - Note
+  * - np
+    - 0
+    - 
+  * - p
+    - 1
+    -
```
NathanielBlairStahn, line 154: I don't understand what information this table is conveying...

NathanielBlairStahn: Oh, never mind, these are prevalences -- for some reason it was confusing when I first looked at it. Maybe it would be clearer if the column were named "Prevalence" rather than "Value"?

docs/source/models/other_models/pregnancy/gbd_2021_mncnh/index.rst
```diff
@@ -0,0 +1,335 @@
...
+#. Assign pregnancy state according to state prevalence values
+#. Assign partial or full term duration according to table in `Pregnancy term lengths`_ section
+#. Assign sex of infant if pregnancy is full term (stillbirth or live birth)
+#. Assign duration of pregnancy depending on term length and, if applicable, sex of the infant. Note that this is the same value as "gestational age" in other parts of the documentation.
+#. Assign birthweight of simulant child 
+#. Assign propensity values for ANC and ultrasound 
+#. Begin simulation
+
+
+See below for more detailed information on how to assign several of these items.
```
NathanielBlairStahn, line 238: I think it would be clearer if some version of this comment were moved to in front of the list of steps. I started reading the list and was like, "Wait, how are you supposed to do that?" before realizing that all the details are below instead of above. If we want to summarize all the steps before giving all the details, I think we should start the section with a comment like, "Here's a summary of the algorithm for the pregnancy model. Details for each step are provided below," or something like that.

docs/source/models/other_models/pregnancy/gbd_2021_mncnh/index.rst
```diff
@@ -0,0 +1,335 @@
...
+     -
+   * - Age group start
+     - 10 to 14 years
+     - ID=7
+   * - Age group end
+     - 50 to 54 years
+     - ID=15
+
+
+**Pregnancy modeling steps:**
```
NathanielBlairStahn, line 225: Why not make this an actual subsection so it shows up in the TOC?

docs/source/models/other_models/pregnancy/gbd_2021_mncnh/index.rst
```diff
@@ -0,0 +1,335 @@
...
+    *   - Ethiopia
+        - 179
+        - 0.514271  
+
+Birthweight and Gestational Age
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A duration of pregnancy value will need to be assigned to all pregnancies regardless of the pregnancy outcome. This value will inform the duration that the simulant remains in the pregnancy state prior to transitioning to the postpartum state.
+
+For partial term pregnancies (that result in abortion/miscarriage/ectopic pregnancy), assign a duration of pregnancy sampled from a uniform distribution beween 6 and 24 weeks (individual heterogeneity with no parameter uncertainty). For these simulants, the birthweight can be assigned as N/A since they will not be going through the intrapartum model.
```
NathanielBlairStahn, line 299: Will we be using the duration of partial term pregnancies for anything? Maybe YLDs of maternal disorders? I'm not sure what the implications of the uniform distribution are since I don't know how we're using this value.

docs/source/models/other_models/pregnancy/gbd_2021_mncnh/index.rst
```diff
@@ -0,0 +1,335 @@
...
+
+Assumptions and limitations
+++++++++++++++++++++++++++++
+
+- We assume that the gestational age distribution of stillbirths is equal to the gestational age distribution of live births. This is a limitation of our analysis given the lack of data on the distribution of gestational ages for which these outcomes occur. Given that the gestation for these outcomes is likely shorter than gestation for live births on average, we are likely overestimating the average duration of pregnancy for outcomes other than live births.
+- We assume that all abortions, miscarriages and ectopic pregnancies occur uniformly between six and 24 weeks gestatation. Six weeks was chosen as a reasonable earliest possible time of pregnancy detection (prior to which miscarriages would be undiagnosed) and 24 weeks was chosen as the threshold between miscarriage and stillbirth. 
+- We assume that abortions that occur after 24 weeks are not considered stillbirths for estimation of the stillbirth to livebirth ratio. We may overestimate the incidence rate of pregnancy due to this assumption.
```
NathanielBlairStahn, line 318: I don't understand the logic here. Is "incidence rate of pregnancy" actually the thing we're overestimating, given that we defined this to be 0 above?

docs/source/models/other_models/pregnancy/gbd_2021_mncnh/index.rst
```diff
@@ -0,0 +1,335 @@
...
+- We are limited in the assumption that the stillbirth to livebirth ratio does not vary by maternal age and does not incorporate an uncertainty distribution.
+- We do not distiguish between intended and unintended pregnancies.
+- We do not consider the impact of birth interval timing or family size in our model of pregnancy.
+- We do not consider the impact of singleton versus non-singleton pregnancies.
+- We are limited in that we apply the age-specific fertility rate to the *beginning* of pregnancy when it is actually measured at the *end* of pregnancy (birth). Therefore, maternal age at birth in our simulation will be overestimated due to simulants becoming pregnant in one age group and giving birth in the next.
```
NathanielBlairStahn, line 323: Is this comment still accurate given that we don't have time in this model?

docs/source/models/other_models/pregnancy/gbd_2021_mncnh/index.rst
```diff
@@ -0,0 +1,335 @@
...
+
+Verification and validation criteria
+++++++++++++++++++++++++++++++++++++++
+
+The following should validate:
+
+- Average duration of pregnancy
```
NathanielBlairStahn, line 330: Should it be not just the average, but the whole distribution of gestational ages (and birthweights)?

docs/source/models/other_models/pregnancy/gbd_2021_mncnh/index.rst
NathanielBlairStahn: I think putting it at the ++++++++++++ level would be the appropriate thing (subsection of Vivarium modeling strategy, super-section of the two details subsections below, same level as Assumptions and limitations)

NathanielBlairStahn (approved):

docs/source/models/other_models/pregnancy/gbd_2021_mncnh/index.rst
SylLutze: So we defined the started prevalence as 0, the incidence is a formula based on the fertility rate (births) and the stillbirth ratio. This comment is saying that we might accidentally log some things that are really abortions as stillbirths, making the stillbirth to live birth ratio off. And that in turn might make the pregnancy incidence rate off. 

Since we have done this and completed V&V previously, I am not too concerned about it, but I think it's still valid!

docs/source/models/other_models/pregnancy/gbd_2021_mncnh/index.rst
SylLutze: Added to do note about this to come back after we assess YLDs as discussed

PR #1525, opened 2024-08-29
Updating the pregnancy concept model page and changing to LBW

SylLutze:


aflaxman (approved):

pletale (approved):

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -94,24 +94,34 @@ all labeled and additional information will be included below.
...
-    - Need to determine if this is random or correlated with other outcomes. Need to find baseline and scenario values.
+    - Ultrasound rate at ANC in baseline scenario
+    - 60.7% 
+    - https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9201203/
+    - Data value is a stand-in as we finalize. We would like this to be location specific, please code accordingly.
   * - 3
-    - All patients are assigned a stated gestational age (GA). Their stated GA will be their real GA +/- an error margin randomly assigned within an interval. For AI assisted ultrasound, the interval is +/- XX weeks, for standard ultrasound XX weeks, for no ultrasound XX weeks. 
-    - Check if distribution for error is random or more likely to be an over/underestimate. 
+    - Stated gestational age (GA) at ANC 
+    - Real GA +/- a value from a uniform distribution of: 0-5 days for AI ultrasound, 10-30 days for standard ultrasound, 10-45.5 days for no ultrasound
```
NathanielBlairStahn, line 117: Are we sure we want a uniform distribution here, rather than something that would make values closer to 0 more likely? Based on a cursory look through the linked paper, it seems like maybe a normal distribution would be appropriate, but I'm not totally sure.

Also, I'm not clear about the intended values here. For example, if the real GA were 280 days, it looks like the reported GA from a standard ultrasound should be uniformly distributed on the union of the two intervals [290, 310] and [250, 270] (so there's probability 0 of getting a value in [270, 290]). Is that the intention, or am I misinterpreting something?

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -94,24 +94,34 @@ all labeled and additional information will be included below.
...
+    - Ultrasound rate at ANC in baseline scenario
+    - 60.7% 
+    - https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9201203/
+    - Data value is a stand-in as we finalize. We would like this to be location specific, please code accordingly.
   * - 3
-    - All patients are assigned a stated gestational age (GA). Their stated GA will be their real GA +/- an error margin randomly assigned within an interval. For AI assisted ultrasound, the interval is +/- XX weeks, for standard ultrasound XX weeks, for no ultrasound XX weeks. 
-    - Check if distribution for error is random or more likely to be an over/underestimate. 
+    - Stated gestational age (GA) at ANC 
+    - Real GA +/- a value from a uniform distribution of: 0-5 days for AI ultrasound, 10-30 days for standard ultrasound, 10-45.5 days for no ultrasound 
+    - Ultrasound values assumed from BMGF inputs. No ultrasound value from this paper: https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0272718#sec007
```
NathanielBlairStahn, line 118: This comment confused me: At first I thought it meant "There are no values for ultrasounds found in this paper," which seems strange because I'm more interested in what _is_ in the paper than what _isn't_ in it. But then I realized that it meant that "the values wer're using for the 'no ultrasound' case were taken from this paper," I think.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -94,24 +94,34 @@ all labeled and additional information will be included below.
 .. image:: pregnancy_decision_tree_vr.svg
 
 .. list-table:: Pregnancy Decision Tree
-  :widths: 3 15 15
+  :widths: 3 5 10 10 15
```
NathanielBlairStahn, line 97: Is there a way to make this table narrower? It's really wide on my screen, making the Notes column difficult to read. If the URLs are making the Source column too wide, you could maybe replace them with hyperlinked text that can be shorter and contain spaces, instead of the raw URL.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
NathanielBlairStahn: Also, it occurred to me that maybe we're not making the right comparisons here. Are we sure that AI ultrasound is supposed to be more accurate than regular ultrasound? Or is it just that the AI allows people to do ultrasounds without having special training, and the accuracy isn't any different (or maybe even _worse_ for all we know)? In which case we should just be comparing "any ultrasound" to "no ultrasound," and the intervention is that more people are able to obtain an ultrasound because of the AI.

Here's what makes me think this. Chris's response about the TPP for AI ultrasound says:

> The assumptions we are using are that point-of-care ultrasound (AI Ultrasound) reduces mean absolute error in gestational age dating in any trimester from 10-30 days to 5 days.

Meanwhile, the linked paper says that (standard) ultrasound dating of GA is the "gold standard," and, for example (emphasis mine),

> **The precision error in the SFH GA estimate is reflected in the SD of the mean difference.** In pooled analysis (n = 2,447 pregnancies), the mean difference between GA estimated by a single SFH measurement (1cm SFH = 1wk GA assumption) and ultrasound-confirmed GA was -14.0 days, with a **pooled SD of 21.4 days (95% CI of ±42.8 days).** This negative bias towards underestimation of GA was likely strongly influenced by one study in Bangladesh, in which rates of SGA were high [[44](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0272718#pone.0272718.ref044)]. In sensitivity analysis excluding this study, the pooled mean bias was +1 day and was not significant.

This makes me think that the 10-30 days number from Chris looks way too high for standard ultrasound, and tracks more with the SD of 21.4 days for _no_ ultrasound.

SylLutze: As discussed, changed this to a normal distribution with a mean of zero and different standard deviations. Changed note to indicate that we had outstanding questions that we were waiting on inputs from BMGF. Linked to this PR so reference to comment thread.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
SylLutze: Lol fair, rephrased.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
SylLutze: Replaced all links with hyperlinks. Not sure how else to make it shorter...

NathanielBlairStahn (approved):

PR #1529, opened 2024-09-04
Add country-specific ANC utilization rates from GBD to pregnancy concept model 

pletale:


SylLutze (approved):

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -104,7 +104,7 @@ all labeled and additional information will be included below.
     - Notes
   * - 1
     - ANC1 rates
-    - get_covariate_estimates(location_id=location_id, gbd_round_id=7, year_id=2021, decomp_step='iterative', covariate_id=7)
+    - Ethiopia: 75.7%, Nigeria: 74.3%, Pakistan: 90.8%
```
NathanielBlairStahn, line 107: Do we want the engineers to pull the covariates themselves, or copy the numbers from here? I think we need to make it clearer which is intended.

NathanielBlairStahn: My vote would be: I think we should have the engineers pull the numbers themselves to reduce the possibility of typos, miscommunication, outdated docs, etc., though I also like having them shown here. So if there's some way to show the numbers here but also make it clear that the engineers should pull them directly, that would be ideal.

pletale: Yeah that sounds good to me! I'll make those changes now

NathanielBlairStahn (approved):

PR #1531, opened 2024-09-10
Update ultrasound utilization rates to be country-specific

pletale:
This PR updates the numbers for baseline coverage of standard ultrasounds in our 3 countries of interest. Unfortunately, I was unable to find data for ultrasound use in Pakistan, so here I'm using values from India's 2016 DHS, which I'm not sure is valid (but they do have very similar ANC utilization rates in GBD)! 

Links to the literature:  
- Ethiopia ultrasound rate: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8905208/
- Nigeria ultrasound rate: https://www.researchgate.net/publication/51782476_Awareness_of_information_expectations_and_experiences_among_women_for_obstetric_sonography_in_a_south_east_Nigeria_population 
- India ultrasound rate (proxy for Pakistan): https://dhsprogram.com/pubs/pdf/FR339/FR339.pdf

### DIFF
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
@@ -109,8 +109,8 @@ all labeled and additional information will be included below.
     - This is location specific, but not age specific. Currently assume that there is no correlation of ANC with other factors. Engineers, you can pull these value straight from GBD, but expected values are as follows - Ethiopia: 75.7%, Nigeria: 74.3%, Pakistan: 90.8%
   * - 2
     - Ultrasound rate at ANC in baseline scenario
-    - Ethiopia: 60.7%, Nigeria: 58.7%, Pakistan: 61.4%
-    - `India ultrasound rate <https://dhsprogram.com/pubs/pdf/FR339/FR339.pdf>`_, `Ethiopia ultrasound rate <https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8905208/>`_ , `Nigeria ultrasound rate <https://www.researchgate.net/publication/51782476_Awareness_of_information_expectations_and_experiences_among_women_for_obstetric_sonography_in_a_south_east_Nigeria_population>`_  
+    - Ethiopia: 60.7%, Nigeria: 58.7%, Pakistan: 55.7%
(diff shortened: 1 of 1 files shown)


aflaxman (approved):
If there is no systematic review that reports ultrasound rates around the world, should we do one?  If a lot of DHS surveys have asked about it, then it might be straightforward... I wonder if Annie H's team is doing this.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -109,9 +109,9 @@ all labeled and additional information will be included below.
     - This is location specific, but not age specific. Currently assume that there is no correlation of ANC with other factors. Engineers, you can pull these value straight from GBD, but expected values are as follows - Ethiopia: 75.7%, Nigeria: 74.3%, Pakistan: 90.8%
   * - 2
     - Ultrasound rate at ANC in baseline scenario
-    - 60.7% 
-    - `Linked ultrasound rate <https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9201203/>`_ 
-    - Data value is a stand-in as we finalize. We would like this to be location specific, please code accordingly.
+    - Ethiopia: 60.7%, Nigeria: 58.7%, Pakistan: 61.4%
```
aflaxman, line 112: Is this e.g. 60.7% of the 75.5 pregnant people who attended ANC (out of 100) in Ethiopia?  Or is it 60.7 pregnant people out of 100?

pletale: I just double-checked, and all 3 sources are among pregnant people only (rather than all WBPRA). However, the studies in Ethiopia and Nigeria are specifically about pregnant women attending an ANC facility. Because we have data regarding ANC visits for all 3 locations, maybe we can adjust the India value to only include pregnant people attending an ANC? 
cc @NathanielBlairStahn

NathanielBlairStahn: Good to know! I think a single number for pregnant people attending ANC any number of times is good for now.

NathanielBlairStahn (approved):

PR #1532, opened 2024-09-11
Update intrapartum concept model

pletale:
This PR provides placeholder values of a few data inputs for our intrapartum concept model decision tree: 
- Incidence of ectopic pregnancies, abortion, or miscarriage (using GBD; confirmed that this is incidence across total population) 
- % of simulants attending each facility type (we will do a more in-depth DHS analysis for the country-specific numbers, but for now using a study that used Ethiopia DHS) 
- % of each delivery facility type that has c-section capabilities (using Ethiopia EmONC Assessment; we will need to do more data seeking to find similar data sources for Pakistan and Nigeria) 

There are more data values to update in this table, but in order to keep this PR from getting too big, I will make those updates in a separate PR!

### DIFF
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
@@ -216,8 +216,8 @@ V&V Checks:
     - Notes
   * - 0
     - Incidence of ectopic pregnancies, abortion or miscarriage
-    - get_draws(gbd_round_id=7, location_id=location_id, gbd_id_type='cause_id', gbd_id=[995,374], source='como', measure_id=6, metric_id=3, age_group_id=24, sex_id=2, year_id=2021, decomp_step='iterative')
     - incidence_c374 + incidence_c995
+    - get_draws(gbd_round_id=7, location_id=location_id, gbd_id_type='cause_id', gbd_id=[995,374], source='como', measure_id=6, metric_id=3, age_group_id=24, sex_id=2, year_id=2021, decomp_step='iterative')
     - These simulants will NOT continue in the model. Use the `total population incidence <Total Population Incidence Rate>`_ rate directly from GBD and do not rescale this parameter to susceptible-population incidence rate using condition prevalence.
   * - 1
     - % of simulants to attend each delivery facility type, based on their propensity
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
@@ -232,7 +232,7 @@ V&V Checks:
   * - 3
     - % of each facility type have cesarian section capabilities
     - Public/governmental health facility (62%), private and/or NGO health facility (73%), at home  (0%), other (50%) 
-    - EmONC (Ethiopia; saved at J:\Project\simulation_science\optimization_model_data\EmONC 2016_Master Dataset_Final\ETH_EmONC Assessment 2016_Final Report Jan11 2018.pdf) 
+    - EmONC (Ethiopia; filepath saved `on SharePoint <https://uwnetid.sharepoint.com/:w:/r/sites/ihme_simulation_science_team/_layouts/15/Doc.aspx?sourcedoc=%7B63F98143-C6C3-4CF7-BF62-969344726A87%7D&file=ethiopia_data_received_notes.docx&action=default&mobileredirect=true>`_.) 
     - We want these to be location specific, please code accordingly. These are placeholder values for now (extracted from the EmONC Final Report, link in 'Source' column; the 'other' value is made-up), hopefully we will be able to find similar data available for Pakistan and Nigeria.   
   * - 4
     - XX relative risk on incidence of hemorrhage and obstructed labor 
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
@@ -221,7 +221,7 @@ V&V Checks:
     - These simulants will NOT continue in the model. Use the `total population incidence <Total Population Incidence Rate>`_ rate directly from GBD and do not rescale this parameter to susceptible-population incidence rate using condition prevalence.
   * - 1
     - % of simulants to attend each delivery facility type, based on their propensity
-    - At home (68.3%), in public/governmental health facility (26.6%), in private/NGO health facility (3.3%), and other (4.8%) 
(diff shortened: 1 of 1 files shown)


docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -206,35 +206,53 @@ V&V Checks:
...
   * - 2
     - Need to figure out how we will determine which simulants need a c-section
+    -
     - 
+    -    
   * - 3
-    - XX% of each facility type have cesarian section capabilities
-    -  
+    - % of each facility type have cesarian section capabilities
+    - Public/governmental health facility (62%), private and/or NGO health facility (73%), at home  (0%), other (50%)
```
aflaxman, line 234: How should I think about "other"?  Is this facility type common?

aflaxman (approved):

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
pletale: Yes, I should have made a comment about this also! 'Other' signifies any health facility that doesn't fall into one of the four standard categories (public hospital, private for-profit, NGO, or at home -- these four categories were updated in [my second PR](https://github.com/ihmeuw/vivarium_research/pull/1533) after I had a better sense of data availability) and includes what is coded as 'country-specific government health facility' in DHS. For Ethiopia, for example, I think this includes [health posts or Health Extension Workers](https://l10k.jsi.com/Health/hep.htm#:~:text=Health%20posts%20are%20located%20at,health%20extension%20workers%20(HEWs).). I still need to do more investigating as to what this would entail for Pakistan and Nigeria.  I can clarify the wording in this to make this more clear: e.g., change 'Public/governmental health facility' to 'Public/governmental hospital' and 'other' to 'country-specific government health facility'.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -206,35 +206,53 @@ V&V Checks:
...
     - Need to figure out how we will determine which simulants need a c-section
+    -
     - 
+    -    
   * - 3
-    - XX% of each facility type have cesarian section capabilities
-    -  
+    - % of each facility type have cesarian section capabilities
+    - Public/governmental health facility (62%), private and/or NGO health facility (73%), at home  (0%), other (50%) 
+    - EmONC (Ethiopia; saved at J:\Project\simulation_science\optimization_model_data\EmONC 2016_Master Dataset_Final\ETH_EmONC Assessment 2016_Final Report Jan11 2018.pdf)
```
NathanielBlairStahn, line 235: It looks like backslashes aren't rendering because they're a special character, so you'll have to [escape them using a double backslash](https://stackoverflow.com/questions/57265111/escape-special-characters-in-restructuredtext-sphinx#comment131155294_57266065) `\\`.

Also, this single long filepath string with no spaces is making this table column very wide... Not sure what the best solution is -- would it look ok to put a space after each backslash?

pletale: Ah, I see. I also could delete the file path from here, as it's saved in Sharepoint also and I don't think it really needs to be here!

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -206,35 +206,53 @@ V&V Checks:
...
     - Decision Information 
+    - Data Value 
+    - Source
     - Notes
   * - 0
-    - XX% of simulants have ectopic pregnancies, abortion or miscarriage
-    - These simulants will NOT continue in the model
+    - Incidence of ectopic pregnancies, abortion or miscarriage
+    - get_draws(gbd_round_id=7, location_id=location_id, gbd_id_type='cause_id', gbd_id=[995,374], source='como', measure_id=6, metric_id=3, age_group_id=24, sex_id=2, year_id=2021, decomp_step='iterative')
+    - incidence_c374 + incidence_c995
```
NathanielBlairStahn, line 220:
```suggestion
    - incidence_c374 + incidence_c995
    - get_draws(gbd_round_id=7, location_id=location_id, gbd_id_type='cause_id', gbd_id=[995,374], source='como', measure_id=6, metric_id=3, age_group_id=24, sex_id=2, year_id=2021, decomp_step='iterative')
```
I think these columns should be switched -- i.e., put `incidence_c374 + incidence_c995` in the "Data Value" column, and put the `get_draws` call in the "Source" column.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
NathanielBlairStahn: Seems reasonable -- you could add a link to the Sharepoint file instead.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -206,35 +206,53 @@ V&V Checks:
...
+    - These simulants will NOT continue in the model. Use the `total population incidence <Total Population Incidence Rate>`_ rate directly from GBD and do not rescale this parameter to susceptible-population incidence rate using condition prevalence.
```
NathanielBlairStahn, line 221: Do you know why we need the total population incidence rate here?  I think it probably has to do with [how we'll calculate the probabilities for full term v. partial term](https://vivarium-research.readthedocs.io/en/latest/models/other_models/pregnancy/gbd_2021_closed_cohort/index.html#pregnancy-term-lengths), but I'm not entirely sure.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
NathanielBlairStahn: > I can clarify the wording in this to make this more clear: e.g., change 'Public/governmental health facility' to 'Public/governmental hospital' and 'other' to 'country-specific government health facility'.

Do you think this wording would be more accurate? It doesn't look like you made this change in your other PR.

NathanielBlairStahn (approved):
Looks good! Just a few minor suggestions.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
pletale: Yeah, I'm not entirely sure either, but the page you linked (specifically [this table's "Notes" column](https://vivarium-research.readthedocs.io/en/latest/models/other_models/pregnancy/gbd_2021_closed_cohort/index.html#id8)) is where I pulled that assumption from.

PR #1533, opened 2024-09-12
Update intrapartum concept model (part 2)

pletale:
This PR fills in the rest of the intrapartum decision tree inputs with placeholder values. Data points include:
- % of each type of facility with azithromycin (for Ethiopia only; but we do not have data on % of pregnancies receiving azithromycin by facility type) 
- RR of azithromycin on sepsis (from a BMGF-funded RCT) 
- % of pregnancies receiving antenatal corticosteroids by facility type (for Ethiopia only) 
- RR of c-section delivery on PPH

I added a row for the RR of c-sections on obstructed labor, but I'm not sure using an RR here makes sense, if we expect c-sections to completely remove the possibility of OL? Or do we expect some c-sections to still lead to some degree of OL? Also, I'm wondering if we should be grouping OL and uterine rupture together as we currently are. In doing some brief research, it looks like previous c-section surgery is a pretty significant risk factor for uterine rupture.

### DIFF
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
@@ -231,7 +231,7 @@ V&V Checks:
     -    
   * - 3
     - % of each facility type have cesarian section capabilities
-    - Public/governmental hospital (62%), private and/or NGO health facility (73%), at home  (0%), country-specific governmental health facility (50%) 
+    - Public/governmental health facility (62%), private for-profit health facility (60%), at home (0%), private not-for-profit health facility (86%), other (10%)
     - EmONC (Ethiopia; filepath saved `on SharePoint <https://uwnetid.sharepoint.com/:w:/r/sites/ihme_simulation_science_team/_layouts/15/Doc.aspx?sourcedoc=%7B63F98143-C6C3-4CF7-BF62-969344726A87%7D&file=ethiopia_data_received_notes.docx&action=default&mobileredirect=true>`_.) 
     - We want these to be location specific, please code accordingly. These are placeholder values for now (extracted from the EmONC Final Report, link in 'Source' column; the 'other' value is made-up), hopefully we will be able to find similar data available for Pakistan and Nigeria.   
   * - 4a
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
@@ -221,7 +221,7 @@ V&V Checks:
     - These simulants will NOT continue in the model. Use the `total population incidence <Total Population Incidence Rate>`_ rate directly from GBD and do not rescale this parameter to susceptible-population incidence rate using condition prevalence.
   * - 1
     - % of simulants to attend each delivery facility type, based on their propensity
-    - At home (68.3%), in public/governmental health facility (26.6%), in private/NGO health facility (3.3%), and other (4.8%) 
+    - At home (68.3%), in public/governmental health facility (26.6%), in private and/or NGO health facility (3.3%), and other (4.8%) 
     - DHS for each location; placeholder values are from `this Ethiopia paper <https://link.springer.com/article/10.1186/s12884-020-03002-x#Tab2>`_.
     - Denominator in DHS is all births (live and stillbirths) to interviewed women in the 2 years preceding the survey. The above values are placeholders until we do a more in-depth analysis. We would like this to be location specific, please code accordingly. 
   * - 2
@@ -230,10 +230,10 @@ V&V Checks:
     - 
     -    
   * - 3
-    - XX% of each facility type have cesarian section capabilities
-    - 
-    - 
-    -  
+    - % of each facility type have cesarian section capabilities
+    - Public/governmental health facility (62%), private and/or NGO health facility (73%), at home  (0%), other (50%) 
+    - EmONC (Ethiopia; saved at J:\Project\simulation_science\optimization_model_data\EmONC 2016_Master Dataset_Final\ETH_EmONC Assessment 2016_Final Report Jan11 2018.pdf) 
+    - We want these to be location specific, please code accordingly. These are placeholder values for now (extracted from the EmONC Final Report, link in 'Source' column; the 'other' value is made-up), hopefully we will be able to find similar data available for Pakistan and Nigeria.   
   * - 4
     - XX relative risk on incidence of hemorrhage and obstructed labor 
     - 
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
@@ -221,7 +221,7 @@ V&V Checks:
     - These simulants will NOT continue in the model. Use the `total population incidence <Total Population Incidence Rate>`_ rate directly from GBD and do not rescale this parameter to susceptible-population incidence rate using condition prevalence. 
   * - 1
     - % of simulants to attend each delivery facility type, based on their propensity
-    - At home (68.3%), in public/governmental health facility (26.6%), in private for-profit health facility (2.5%), in private not-for-profit health facility (0.8%), and other (1.8) 
+    - At home (68.3%), in public/governmental health facility (26.6%), in private for-profit health facility (2.5%), in private not-for-profit health facility (0.8%), and other (1.8%) 
     - DHS for each location; placeholder values are from `this Ethiopia paper <https://link.springer.com/article/10.1186/s12884-020-03002-x#Tab2>`_.
     - Denominator in DHS is all births (live and stillbirths) to interviewed women in the 2 years preceding the survey. The above values are placeholders until we do a more in-depth analysis. We would like this to be location specific, please code accordingly. 
   * - 2
@@ -231,7 +231,7 @@ V&V Checks:
     -    
   * - 3
     - % of each facility type have cesarian section capabilities
-    - Public/governmental health facility (62%), private for-profit health facility (60%), at home  (0%), private not-for-profit health facility (86%) 
+    - Public/governmental health facility (62%), private for-profit health facility (60%), at home  (0%), private not-for-profit health facility (86%), other (10%)  
     - EmONC (Ethiopia; saved at J:\Project\simulation_science\optimization_model_data\EmONC 2016_Master Dataset_Final\ETH_EmONC Assessment 2016_Final Report Jan11 2018.pdf) 
     - We want these to be location specific, please code accordingly. These are placeholder values for now (extracted from the EmONC Final Report, link in 'Source' column; the 'other' value is made-up), hopefully we will be able to find similar data available for Pakistan and Nigeria.   
   * - 4
@@ -241,7 +241,7 @@ V&V Checks:
     - Outstanding items: how does c-section need overlap with hemorrhage/OL, what is the RR, how will we implement this with overlaps in total MD impact of facility type 
   * - 5
     - % of pregnancy receive azithromycin in each delivery facility type
-    - At home (10%), in public/governmental health facility (41%), in private for-profit health facility (10%), and private not-for-profit health facility (10%) 
+    - At home (10%), in public/governmental health facility (41%), in private for-profit health facility (10%), private not-for-profit health facility (10%), other (10%)  
     - SARA (Ethiopia; Table 3.8.2)
     - These are placeholder values and will be updated with further analysis. We want these to be location specific, please code accordingly.  
   * - 6
@@ -251,7 +251,7 @@ V&V Checks:
     - Outstanding items: how will we implement this with overlaps in total MD impact of facility type 
   * - 7
     - % of pre-term or known LBW pregnancies will receive antenatal corticosteroids, split by delivery facility type
-    - At home (10%), in public/governmental health facility (2%), in private for-profit health facility (22%), and private not-for-profit health facility (4%) 
+    - At home (10%), in public/governmental health facility (2%), in private for-profit health facility (22%), and private not-for-profit health facility (4%), other (10%)  
     - EmONC (Ethiopia; Table 10.5.4A)
     - Outstanding items: is this for preterm, LBW, or both/combination; believe this only affected neonatal outcomes, confirm with BMGF
 
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
@@ -235,7 +235,7 @@ V&V Checks:
     - EmONC (Ethiopia; saved at J:\Project\simulation_science\optimization_model_data\EmONC 2016_Master Dataset_Final\ETH_EmONC Assessment 2016_Final Report Jan11 2018.pdf) 
     - We want these to be location specific, please code accordingly. These are placeholder values for now (extracted from the EmONC Final Report, link in 'Source' column; the 'other' value is made-up), hopefully we will be able to find similar data available for Pakistan and Nigeria.   
   * - 4
-    - XX relative risk on incidence of hemorrhage and obstructed labor 
+    - XX relative risk of c-section on incidence of hemorrhage and obstructed labor 
     - 
     - 
     - Outstanding items: how does c-section need overlap with hemorrhage/OL, what is the RR, how will we implement this with overlaps in total MD impact of facility type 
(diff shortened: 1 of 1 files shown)


docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -231,29 +231,29 @@ V&V Checks:
...
+    - XX relative risk of c-section on incidence of hemorrhage and obstructed labor 
     - 
     - 
     - Outstanding items: how does c-section need overlap with hemorrhage/OL, what is the RR, how will we implement this with overlaps in total MD impact of facility type 
   * - 5
-    - XX% of pregnancy receive in each delivery facility type
-    - 
-    - 
-    - Confirm understanding that all pregnancies can/should receive this
+    - % of pregnancy receive azithromycin in each delivery facility type
```
pletale, line 243: The placeholder values I've included here are actually the percentage of each facility type that have azithromycin, not the percentage of pregnancies that receive it. So we might need to do further data seeking if we want facility types to be more or less likely to actually give out the azithromycin to pregnant people.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -231,29 +231,29 @@ V&V Checks:
...
+    - At home (10%), in public/governmental health facility (41%), in private for-profit health facility (10%), private not-for-profit health facility (10%), other (10%)  
+    - SARA (Ethiopia; Table 3.8.2)
+    - These are placeholder values and will be updated with further analysis. We want these to be location specific, please code accordingly.  
   * - 6
-    - XX relative risk of incidence of sepsis and other infections
-    - 
-    - 
-    - Outstanding items: what is the RR, how will we implement this with overlaps in total MD impact of facility type 
+    - Relative risk of azithromycin on incidence of sepsis and other infections
+    - 0.65 (0.55, 0.77)
```
pletale, line 249: I'm not sure if it makes sense to include upper and lower uncertainty bounds for this number, but the data source provided them, so I included them - I can remove though.

aflaxman: definitely include --- we will want to propagate this uncertainty through to our results, by using "input draws" sampled from an appropriate distribution calibrated to have the mean and UI you find.

aflaxman (approved):

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
NathanielBlairStahn: Could you add this information to the notes column so we know what the numbers actually represent and so we don't forget about possibly doing more data seeking?

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -231,29 +231,34 @@ V&V Checks:
...
-    - Public/governmental health facility (62%), private and/or NGO health facility (73%), at home  (0%), other (50%) 
+    - Public/governmental health facility (62%), private for-profit health facility (60%), at home  (0%), private not-for-profit health facility (86%), other (10%)  
     - EmONC (Ethiopia; saved at J:\Project\simulation_science\optimization_model_data\EmONC 2016_Master Dataset_Final\ETH_EmONC Assessment 2016_Final Report Jan11 2018.pdf) 
     - We want these to be location specific, please code accordingly. These are placeholder values for now (extracted from the EmONC Final Report, link in 'Source' column; the 'other' value is made-up), hopefully we will be able to find similar data available for Pakistan and Nigeria.   
-  * - 4
-    - XX relative risk on incidence of hemorrhage and obstructed labor 
+  * - 4a
+    - Relative risk of c-section on incidence of hemorrhage
+    - 2.05 (1.84-2.29)
+    - https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7887994/
```
NathanielBlairStahn, line 240: The hyperlinks are making this table column extra wide because they don't have spaces. One way to fix that would be to change it from a raw URL to hyperlinked text so that the link could be a shorter word or phrase that fits more nicely in the table.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -231,29 +231,34 @@ V&V Checks:
...
+    - SARA (Ethiopia; Table 3.8.2)
+    - These are placeholder values and will be updated with further analysis. We want these to be location specific, please code accordingly.  
   * - 6
-    - XX relative risk of incidence of sepsis and other infections
-    - 
-    - 
-    - Outstanding items: what is the RR, how will we implement this with overlaps in total MD impact of facility type 
+    - Relative risk of azithromycin on incidence of sepsis and other infections
+    - 0.65 (0.55, 0.77)
+    - https://www.ajog.org/article/S0002-9378(22)02210-4/fulltext#undfig1
```
NathanielBlairStahn, line 255: Ditto with the long URL

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
NathanielBlairStahn: Also, grammar:
```suggestion
    - % of pregnancies that receive azithromycin in each delivery facility type
```

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -231,29 +231,34 @@ V&V Checks:
...
+    - Relative risk of c-section on incidence of obstructed labor 
     - 
     - 
-    - Outstanding items: how does c-section need overlap with hemorrhage/OL, what is the RR, how will we implement this with overlaps in total MD impact of facility type 
+    - Outstanding items: how does c-section need overlap with hemorrhage/OL, what is the RR, how will we implement this with overlaps in total MD impact of facility type  
   * - 5
-    - XX% of pregnancy receive in each delivery facility type
-    - 
-    - 
-    - Confirm understanding that all pregnancies can/should receive this
```
NathanielBlairStahn, line 246: Maybe don't delete this so we don't forget, unless you've confirmed it

NathanielBlairStahn:
> I added a row for the RR of c-sections on obstructed labor, but I'm not sure using an RR here makes sense, if we expect c-sections to completely remove the possibility of OL? Or do we expect some c-sections to still lead to some degree of OL? Also, I'm wondering if we should be grouping OL and uterine rupture together as we currently are. In doing some brief research, it looks like previous c-section surgery is a pretty significant risk factor for uterine rupture.

Good questions! Thanks for looking into this, @pletale. Based on your research so far, do you think there could still be some sort of obstructed labor even with a C-section, or would it make more sense to completely remove the possibility of obstructed labor after a C-section? Unless we find something to the contrary, it seems like completely removing it would be reasonable. I guess completely removing OL would correspond to a relative risk of 0.

As for obstructed labor vs. uterine rupture, unfortunately it looks like GBD has a single subcause of Maternal disorders called "Maternal obstructed labor and uterine rupture," according to the [cause hierarchy in our documentation for Maternal disorders: GBD 2021](https://vivarium-research.readthedocs.io/en/latest/models/causes/maternal_disorders/gbd_2021/index.html#cause-hierarchy), probably because untreated obstructed labor can lead to uterine rupture. But currently we're not modeling whether or not simulants have had a previous C-section, since we're only modeling a single pregnancy for each simulant. So I don't think we need to worry about prior C-sections leading to higher risk of uterine rupture -- we'll just use the values from GBD, which should already include the fact that some people have had prior C-sections.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -218,10 +218,10 @@ V&V Checks:
...
     - get_draws(gbd_round_id=7, location_id=location_id, gbd_id_type='cause_id', gbd_id=[995,374], source='como', measure_id=6, metric_id=3, age_group_id=24, sex_id=2, year_id=2021, decomp_step='iterative')
     - incidence_c374 + incidence_c995
-    - These simulants will NOT continue in the model. Use the `total population incidence <Total Population Incidence Rate>`_ rate directly from GBD and do not rescale this parameter to susceptible-population incidence rate using condition prevalence.
+    - These simulants will NOT continue in the model. Use the `total population incidence <Total Population Incidence Rate>`_ rate directly from GBD and do not rescale this parameter to susceptible-population incidence rate using condition prevalence. 
   * - 1
     - % of simulants to attend each delivery facility type, based on their propensity
-    - At home (68.3%), in public/governmental health facility (26.6%), in private and/or NGO health facility (3.3%), and other (4.8%) 
+    - At home (68.3%), in public/governmental health facility (26.6%), in private for-profit health facility (2.5%), in private not-for-profit health facility (0.8%), and other (1.8%)
```
NathanielBlairStahn, line 224: So it looks like you're splitting the private/NGO health facility category into two facility types, "private for-profit" and "private not-for-profit", right? Do you see a particular advantage to adding this further stratification?

pletale:
> > I added a row for the RR of c-sections on obstructed labor, but I'm not sure using an RR here makes sense, if we expect c-sections to completely remove the possibility of OL? Or do we expect some c-sections to still lead to some degree of OL? Also, I'm wondering if we should be grouping OL and uterine rupture together as we currently are. In doing some brief research, it looks like previous c-section surgery is a pretty significant risk factor for uterine rupture.
> 
> Good questions! Thanks for looking into this, @pletale. Based on your research so far, do you think there could still be some sort of obstructed labor even with a C-section, or would it make more sense to completely remove the possibility of obstructed labor after a C-section? Unless we find something to the contrary, it seems like completely removing it would be reasonable. I guess completely removing OL would correspond to a relative risk of 0.
> 
> As for obstructed labor vs. uterine rupture, unfortunately it looks like GBD has a single subcause of Maternal disorders called "Maternal obstructed labor and uterine rupture," according to the [cause hierarchy in our documentation for Maternal disorders: GBD 2021](https://vivarium-research.readthedocs.io/en/latest/models/causes/maternal_disorders/gbd_2021/index.html#cause-hierarchy), probably because untreated obstructed labor can lead to uterine rupture. But currently we're not modeling whether or not simulants have had a previous C-section, since we're only modeling a single pregnancy for each simulant. So I don't think we need to worry about prior C-sections leading to higher risk of uterine rupture -- we'll just use the values from GBD, which should already include the fact that some people have had prior C-sections.

I will do a bit more research into whether we'd want some RR value between c-section and OL/uterine rupture, and get back to you on this! But this context about GBD definitions is very helpful, thank you!

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -231,29 +231,34 @@ V&V Checks:
     -    
   * - 3
     - % of each facility type have cesarian section capabilities
-    - Public/governmental health facility (62%), private and/or NGO health facility (73%), at home  (0%), other (50%) 
+    - Public/governmental health facility (62%), private for-profit health facility (60%), at home  (0%), private not-for-profit health facility (86%), other (10%)
```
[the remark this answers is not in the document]
pletale: For the value here, the 'other' category is entirely made up, I'm expecting we'll update it based on further research. The 4.8% to 1.8% change was because I am no longer including 'country-specific government health facility' as 'other'.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
pletale: Syl confirmed that the understanding is that all pregnancies *should* receive it, but that they currently don't and we need to figure out how much we want to increase coverage in the alternative scenarios. Slack thread for reference: https://ihme.slack.com/archives/C02944X688L/p1726077170940639?thread_ts=1726074424.175769&cid=C02944X688L

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
NathanielBlairStahn: Ok, thanks, that's all good to know. Yes, this seems like a good topic to discuss tomorrow!

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
NathanielBlairStahn: Ok, great, thanks for the clarification. I'll make a note to follow up with BMGF about what the intervention scenarios should look like.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -231,29 +231,34 @@ V&V Checks:
...
+    - Outstanding items: how does c-section need overlap with hemorrhage/OL, what is the RR, how will we implement this with overlaps in total MD impact of facility type  
   * - 5
-    - XX% of pregnancy receive in each delivery facility type
-    - 
-    - 
-    - Confirm understanding that all pregnancies can/should receive this
+    - % of pregnancy receive azithromycin in each delivery facility type
+    - At home (10%), in public/governmental hospital (41%), in private for-profit health facility (10%), private not-for-profit health facility (10%), country-specific governmental health facility (10%)  
+    - SARA (Ethiopia; Table 3.8.2)
+    - These are placeholder values and will be updated with further analysis. We want these to be location specific, please code accordingly.
```
NathanielBlairStahn, line 251:
```suggestion
    - These are placeholder values (percentage of each facility type that have azithromycin, not the percentage of pregnancies that receive it) and will be updated with further analysis. We want these to be location specific, please code accordingly.  
```

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -231,29 +231,34 @@ V&V Checks:
...
+    - `Tita et al 2023 <https://www.ajog.org/article/S0002-9378(22)02210-4/fulltext#undfig1>`_ 
+    - Outstanding items: how will we implement this with overlaps in total MD impact of facility type 
   * - 7
-    - XX% of pre-term or known LBW pregnancies will receive, split by delivery facility type
-    - 
-    - 
-    - Outstanding items: data by delivery facility, is this for preterm, LBW, or both/combination; believe this only affected neonatal outcomes, confirm with BMGF
+    - % of pre-term or known LBW pregnancies will receive antenatal corticosteroids, split by delivery facility type
+    - At home (10%), in public/governmental hospital (2%), in private for-profit health facility (22%), and private not-for-profit health facility (4%), country-specific governmental health facility (10%)  
+    - EmONC (Ethiopia; Table 10.5.4A)
```
NathanielBlairStahn, line 260: What's the denominator in these percentages? E.g., preterm births? All births? Preterm or LBW births?

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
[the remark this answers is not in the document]
NathanielBlairStahn: Ok, thanks for the clarification.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -231,29 +231,34 @@ V&V Checks:
     -    
   * - 3
     - % of each facility type have cesarian section capabilities
-    - Public/governmental health facility (62%), private and/or NGO health facility (73%), at home  (0%), other (50%) 
+    - Public/governmental hospital (62%), private and/or NGO health facility (73%), at home  (0%), country-specific governmental health facility (50%)
```
NathanielBlairStahn, line 234: It looks like here you still have the broader "private and/or NGO health facility" category rather than splitting it into "private for-profit health facility" and "private not-for-profit health facility". Was that intentional? It looks like the original change to this line in this PR was:
> Public/governmental health facility (62%), private for-profit health facility (60%), at home  (0%), private not-for-profit health facility (86%), other (10%)

Maybe there was an accidental reversion when you updated the category names?

NathanielBlairStahn: Ah, it looks like this might have been due to a merge conflict with the other PR.

pletale: Oops, thanks for catching that! Should be fixed now.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
pletale: The denominator is preterm or LBW births, I can add a note of that

pletale: The exact wording from the source is "Percent distribution of cases of preterm and low birth weight babies according to 
client status at birth/admission, treatment, and outcome, by facility type and managing authority"

pletale: I added a note about this and that we want these values to be location-specific, even though currently these are values for Ethiopia only.

NathanielBlairStahn: Ok great, thanks! And the numerator is the number of those that got antenatal corticosteroids?

Why is the public hospital percentage so low (2%)? I would have expected the percentage in any hospital (public or private) to be higher than the percentage for home birth (10%). And 10% for home births seems rather high...

NathanielBlairStahn (approved):

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
pletale: Yes, you're right. I think this is another reason to simplify the stratification of facilities to the 3 options of hospital (high-level), clinic (low-level), and home. In EmONC, hospitals show 12% of preterm/LBW pregnancies receiving ACS, while public/governmental health facilities more generally only showed 2%. I will update these values accordingly, but I can do so in a separate PR.

NathanielBlairStahn: Sounds good, thanks

PR #1534, opened 2024-09-16
Update intrapartum concept model diagram 

pletale:
This PR updates our intrapartum concept model diagram to include intrapartum sensors as an intervention to reduce risk of maternal disorders (hemorrhage and obstructed labor). However, I think this could potentially be better suited to include in the pregnancy model, as a way to identify higher risk pregnancies that in turn get sent to higher level delivery facilities, similar to how the AI ultrasound will function.


NathanielBlairStahn:
@pletale, you are right that the intrapartum sensors are a way to identify high-risk pregnancies. But the intervention happens after labor starts, and whether someone gets it could depend on what facility type they first go to when labor starts. So I think rather than moving the sensors to the pregnancy model, we should reposition the node in the intrapartum model. Specifically, I think intrapartum sensors should be an intermediate node between the facility type (node 1) and C-section (node 2), since it could affect the probability of getting a C-section. Here's what Chris Troeger had to say about it:

> **Nathaniel:** Question 2.1: It sounds like intrapartum sensors could be used to identify high-risk cases during labor, which could potentially lead to referral for C-section and/or emergency transfer to a higher-level facility after labor starts – is that accurate? Are there any other interventions that could increase the probability of better outcomes if a patient is identified as high-risk during labor?
> **Chris:** You are right, C-section delivery or even induction of labor are two interventions. Antenatal corticosteroids are given to mothers in preterm labor to stimulate lung maturation, averting respiratory distress syndrome. Those are the ones most aligned with our portfolio.

It is unclear from Chris's response whether the sensors could lead to transferring to another facility, but let's assume they can. Then we should also have an arrow pointing from the sensor node back to the facility node, making a loop. Chris also mentions induction as a possible intervention -- I'm not sure how we would include that, as it's not in our current plan.

@aflaxman (and @SylLutze) Does this align with your thinking? Any other comments?

SylLutze:
I overall agree with Nathaniel's comment. The intrapartum sensors should go in the intrapartum diagram after labor has begun. It's a second chance opportunity to identify folks that are high risk and should be in a higher level facility. 

I think if we assume that the sensors only impact c-section then the whole comment is correct. I am less clear on if the intrapartum sensors would be able to identify folks needing ACS (pre-term births) and if that would allow enough time to get pregnant folks to a higher level facility and have them receive ACS prior to birth (though, I suspect not). 

One other thing - if they moved to a higher level facility, they would theoretically then receive azithromycin - which is indicated for everyone. 

I think depending on how we include the sensors (move to higher level facility and get all benefits or have this only impact the c-section rate) we would draw the diagram differently. We have a meeting with Gates next Thursday so could get confirmation from them at that time?

pletale:
I've updated the diagram to include intrapartum sensors between 'facility type' and 'c-section received?'. In the intrapartum decision tree table, I added data value 8 (% of pregnancies identified by intrapartum sensors as needing c-section) and some outstanding questions about it that we need to answer (i.e. can a simulant be sent to a higher level delivery facility based on intrapartum sensors, do they affect who receives ACS).

NathanielBlairStahn:
For reference, here is an [updated decision tree](https://play.d2lang.com/?script=lFRPbxo_EL37U4zEJZF-5PDrzYeu0kUKSC2KAolUKRdnPbBWvDb1mKS0ynevxruwf4CkPeGZ9-bNeD2PEVxb84IZjD_DbcC1Uy5mEr4jiVGbYPR6nsODIRMv6HLPEL1kYs3g3sagyG-dbmhHpNuAEUMFVj35AEaji2ZlkPlzL0Rf44OC1OBfCrjDzbWELybEEl6NtfCEsKkLMsb-ouOHpNRFQl56TwhPqddKFcaauOMvK7mY0EUCFQf4o7tw3o01FoaMd-C8RhgDmWprlYuw8tb6VwJlLXyCjYolXbLmFZVqgxJ-bHGLYiHhDgs0L0hgXAxqo0LcVkDoyAfKhPplYinhBnkGPgdf7QrjMrHp3et_ef6iovAhmsJTxOCNpkZukM1EqQiCoWcJU0VQmnU5tviCNiX54tGH7NFd4NX66j9wiJogHxMW0XjHl2t67hqVWTds6knkEqYnpbsjH7r0b8kkp02hImpY-dBvn7Q0Riwiagl370qfF0oyMaBKKulTpajC9Kwh9WknnM_yex5snZ6oK1T5gIlMEr4dzhA9FCUWz5kYQX4r2wrYYFj5UPGTTSXMXPcN9nuXiaWEZVCOVhgCapY7SRvBRMKDWhunLGjkP5DAWz3Zb2CJP9XaO6FVVGkNfdDAQZP53fCKnTVOY3gT4uyKsdFuatd9xEmm4_OiXoRD8NW_toCvULBdUtHx8jfIYIUPFQshFvzbuIfH6sY8Qjpy7shHzD6Pcu0gybw8ISeAI--x_NCRjQJDpxHuOoJFh9dGjO2t2yEMUszK-dzdyrmvc9NmrCkH_f3nfEova41l80KcX57Q69lniHPRezjXt3HnMv1kvZ_1N8nTg-dIKqByMGkWff-RanhyYA95ezk2xuF8itj6ofHNm_gTAAD__w%3D%3D&sketch=1&layout=dagre&) I made that tries to take @SylLutze 's concerns into account.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -259,6 +259,12 @@ V&V Checks:
     - At home (10%), in public/governmental hospital (2%), in private for-profit health facility (22%), and private not-for-profit health facility (4%), country-specific governmental health facility (10%)  
     - EmONC (Ethiopia; Table 10.5.4A)
     - These are placeholder values and will be updated with further analysis. We want these to be location specific, please code accordingly. The denominator for these values is LBW and preterm births. Outstanding items: believe this only affected neonatal outcomes, confirm with BMGF
+  * - 8
+    - % of those who receive intrapartum sensors that get identified as needing a c-section
```
aflaxman, line 263: I think it might be hard to find an evidence-based number to use here --- have you seen this sort of quantity in the literature yet?  Let's make it high-priority to confirm that we will be able to find something for tihs.

aflaxman (approved):
I think these diagrams would benefit from a clearer way to visually signify which nodes represent a "choice" between paths and which paths all happen (in parallel?)

All of the discussion makes sense to me, and we might consider where to include a node for inducing labor, which seems like another important choice that would be influenced by  sensors.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
NathanielBlairStahn: Intrapartum sensors don't even exist yet, right? What sort of data would we look for that would be relevant to this? Also, we still haven't figured out in general how to identify which simulants need a C-section.

NathanielBlairStahn:
> I think these diagrams would benefit from a clearer way to visually signify which nodes represent a "choice" between paths and which paths all happen (in parallel?)

In the [updated decision tree](https://play.d2lang.com/?script=lFRPbxo_EL37U4zEJZF-5PDrzYeu0kUKSC2KAolUKRdnPbBWvDb1mKS0ynevxruwf4CkPeGZ9-bNeD2PEVxb84IZjD_DbcC1Uy5mEr4jiVGbYPR6nsODIRMv6HLPEL1kYs3g3sagyG-dbmhHpNuAEUMFVj35AEaji2ZlkPlzL0Rf44OC1OBfCrjDzbWELybEEl6NtfCEsKkLMsb-ouOHpNRFQl56TwhPqddKFcaauOMvK7mY0EUCFQf4o7tw3o01FoaMd-C8RhgDmWprlYuw8tb6VwJlLXyCjYolXbLmFZVqgxJ-bHGLYiHhDgs0L0hgXAxqo0LcVkDoyAfKhPplYinhBnkGPgdf7QrjMrHp3et_ef6iovAhmsJTxOCNpkZukM1EqQiCoWcJU0VQmnU5tviCNiX54tGH7NFd4NX66j9wiJogHxMW0XjHl2t67hqVWTds6knkEqYnpbsjH7r0b8kkp02hImpY-dBvn7Q0Riwiagl370qfF0oyMaBKKulTpajC9Kwh9WknnM_yex5snZ6oK1T5gIlMEr4dzhA9FCUWz5kYQX4r2wrYYFj5UPGTTSXMXPcN9nuXiaWEZVCOVhgCapY7SRvBRMKDWhunLGjkP5DAWz3Zb2CJP9XaO6FVVGkNfdDAQZP53fCKnTVOY3gT4uyKsdFuatd9xEmm4_OiXoRD8NW_toCvULBdUtHx8jfIYIUPFQshFvzbuIfH6sY8Qjpy7shHzD6Pcu0gybw8ISeAI--x_NCRjQJDpxHuOoJFh9dGjO2t2yEMUszK-dzdyrmvc9NmrCkH_f3nfEova41l80KcX57Q69lniHPRezjXt3HnMv1kvZ_1N8nTg-dIKqByMGkWff-RanhyYA95ezk2xuF8itj6ofHNm_gTAAD__w%3D%3D&sketch=1&layout=dagre&) I linked above, all the nodes are decision nodes, meaning you can only go down one path at a time. It was seeming to me that the diagrams we currently have in the concept model are more like causal diagrams than decision trees. Rajan requested that we draw an actual decision tree to represent what's going on, which is why I started making diagrams using D2.


> All of the discussion makes sense to me, and we might consider where to include a node for inducing labor, which seems like another important choice that would be influenced by sensors.

This could potentially fit into the "High-level risk factor identified?" and "Gets treatment for risk?" nodes in my D2 diagram. Though maybe it would make more sense to create a separate node since induction could happen anywhere, not just at high-level facilities. Or maybe it would be simpler to eliminate the "In high-level facility?" node and just have a single loop for all risks, high-level or not.

NathanielBlairStahn (approved):

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
pletale: Yes, as I understand, this would be more of a hypothesized data point based on what Gates can tell us!

PR #1537, opened 2024-10-03
Remove LBW as a risk factor in pregnancy model

pletale:
I'm not sure I'm updating this correctly - are we still modeling identification of LBW with AI ultrasound? I've removed it here but can add it back in.


NathanielBlairStahn (approved):

SylLutze (approved):

aflaxman:
I had a chance to talk to some experts about this, and my takeaway is that there will be a lot of uncertainty about age vs growth in third-trimester ultrasounds.  In first trimester, determining gestational age will be easier, and if you can do an ultrasound in first and third trimester, that can address some of the ambiguity that a third-trimester-only ultrasound would likely have.

PR #1538, opened 2024-10-04
MNCNH V&V draw and seed count

pletale:
In this PR I'm adding information on model run plans based on what I could find in [Nutrition Optimization docs](https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.html#production-run-specifications), but I am not confident these are the right numbers for V&V runs. 

Syl, I've tagged you in this PR in case you have some insight as to what I should be doing for MNCNH V&V runs (vs. production runs?). And maybe this isn't the format for how we usually save draw and seed count for V&V in the docs? Maybe we can discuss more on Monday - thanks!

### DIFF
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
@@ -489,27 +489,33 @@ Limitations:
     - Population size per draw
     - Note
   * - 1
-    - Wave I Pakistan maternal
+    - Wave I Pakistan maternal V&V
     - Incomplete
     - N/A
-    - 25
-    - 600,000
+    - 20
+    - 20,000
     - 
   * - 2
-    - Wave I Nigeria maternal
+    - Wave I Nigeria maternal V&V
     - Incomplete
     - N/A
-    - 25
-    - 600,000
+    - 20
+    - 20,000
     - 
   * - 3
-    - Wave I Ethiopia maternal
+    - Wave I Ethiopia maternal V&V
     - Incomplete
     - N/A
-    - 25
-    - 600,000
+    - 20
+    - 20,000
     - 
 
+.. note:: 
+  
+  These numbers are based on calculations from the `Nutrition Optimization project <https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.html#production-run-specifications>`_)
+  that found the appropriate seed and draw count for production runs. I'm not sure 
+  if these numbers are correct for V&V runs. This table also doesn't include seed count,
+  but based on the page I linked, NO runs used 10 random seeds. -Alix 
 
 .. _mncnh_portfolio_6.0:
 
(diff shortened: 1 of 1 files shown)


docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -478,10 +478,44 @@ Limitations:
 5.0 Model run requests
```
SylLutze, line 478: This is the appropriate place/format for this information - looks good

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -478,10 +478,44 @@ Limitations:
...
+    - Status
+    - Priority
+    - Number of draws
+    - Population size per draw
+    - Note
+  * - 1
+    - Wave I Pakistan maternal V&V
+    - Incomplete
+    - N/A
+    - 20
```
SylLutze, line 495: 20 was for production runs. I think we could use a better/more standardized way to decide on draws/seeds for v&v runs but half is usually what I would go for, so 10 draws for V&V.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -478,10 +478,44 @@ Limitations:
...
-  Fill in this section as we continue to work
+  * - Number
+    - Run
+    - Status
+    - Priority
+    - Number of draws
+    - Population size per draw
+    - Note
+  * - 1
+    - Wave I Pakistan maternal V&V
```
SylLutze, line 492: You can usually do all the locations as a single model "run". So just call is "Wave 1 Maternal (or pregnancy?) V&V" and then in the note include that we are doing XYZ locations. We could consider adding a column in the table for location if we anticipate changing locations for different runs often! In the past we rarely have done anything other than all locations.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -478,10 +478,44 @@ Limitations:
...
+    - Incomplete
+    - N/A
+    - 20
+    - 20,000
```
SylLutze, line 496: I have always found this confusing - we usually give the engineers a pop size and then they determine the spread over seeds. For example, if we say 100,000 population they can do a run size of 10,000 over 10 seeds or 5,000 pop size with 20 seeds. However, since the RT is now maintaining the model spec files and helping with kicking off runs, maybe that approach isn't appropriate any more. Something we can discuss further later! 

At any rate, I would have this be 10,000 sims at 10 seeds (total pop is then 100,000) for a "standard" V&V approach. Usually we list that as "100,000" in this column and then in the notes include something about 10 seeds. 

The NO docs are confusing but they are doing 20,000 per seed in production runs, so similar to draws we are cutting down the size.

SylLutze (approved):
Let me know if you want to discuss anything here further today. Also happy to recheck things later if you'd like!

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
pletale: Okay, thank you for explaining! I'll summarize this in the note.

NathanielBlairStahn (approved):

PR #1539, opened 2024-10-14
MNCNH V&V and observer output tables

pletale:
For observer output requests, I added 'Overall stratifications' and 'Use case' to the decision tree outputs table. I'd appreciate review particularly of the use case column, because I'm not sure I've captured this accurately. The purpose of this column is to  distinguish between whether the output is a decision tree value only (i.e., to be used in next step) or an export value (i.e., used for V&V or final results) or both.

I've also in this PR added a V&V tracking table.


docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -144,34 +144,45 @@ all labeled and additional information will be included below.
...
 
   * - Input
     - Data Source 
-    - Notes
+    - Overall stratification
+    - Use case 
   * - ANC attendance
     - Decision tree point
-    - 
+    - Age, scenario
```
SylLutze, line 156: Does ANC attendance change with age?

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -144,34 +144,45 @@ all labeled and additional information will be included below.
...
   * - Ultrasound status (AI assisted, standard, none)
     - Decision tree point
-    - 
+    - Scenario
+    - Decision tree value & export value 
   * - Gestational age at birth
     - :ref:`Pregnancy model <other_models_pregnancy_closed_cohort_mncnh>`
-    - 
+    - Decision tree value & export value 
+    - Scenario
```
SylLutze, line 165: These lines are flipped!

SylLutze: I think gestational age (LBWSG) will be age group specific? So we should include that as a stratification. But if LBWSG is not age specific ignore this comment.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -144,34 +144,45 @@ all labeled and additional information will be included below.
...
   * - Gestational age stated
     - Decision tree value
-    - We should track both the real GA at birth and the believed GA
```
SylLutze, line 164: You're using this for V&V right now. I agree we won't use it in the final sim outputs! 

But it's only helpful in V&V if there's individual data so maybe you were thinking this is an interactive sim item only? In which case I agree that makes sense and no next steps needed.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -144,34 +144,45 @@ all labeled and additional information will be included below.
...
     - Decision tree value
-    - We should track both the real GA at birth and the believed GA 
-  * - Birthweight
-    - :ref:`Pregnancy model <other_models_pregnancy_closed_cohort_mncnh>`
     - 
-  * - Low birthweight status
+    - Decision tree value
+  * - Birthweight
     - :ref:`Pregnancy model <other_models_pregnancy_closed_cohort_mncnh>`
     -
```
SylLutze, line 172: Similar to above, I think birthweight (LBWSG) is scenario and age specific and should be checked in a model. If you were thinking intrapartum or if I'm wrong about age specific (see above) then ignore.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -144,34 +144,45 @@ all labeled and additional information will be included below.
...
     - 
-  * - Low birthweight status
+    - Decision tree value
+  * - Birthweight
     - :ref:`Pregnancy model <other_models_pregnancy_closed_cohort_mncnh>`
     - 
+    - Decision tree value & export value 
   * - Pregnancy term (full term or partial term)
     - :ref:`Pregnancy model <other_models_pregnancy_closed_cohort_mncnh>`
     -
```
SylLutze, line 176: Is this needed for removing partial term pregnancies for V&V? I don't know.

SylLutze (commented):
I think Nathaniel and Abie are in a better place to review than me, but I had a few mins so a couple thoughts from my end. Lemme know if you want help implementing anything!

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
NathanielBlairStahn: Do you mean LBWSG exposure would be dependent on the age of the birthing parent? It seems like that would be true in real life, but GBD models LBWSG in isolation, without connecting it to information about the birthing parent, so we don't have that data. Or did you mean something else?

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
NathanielBlairStahn: From the V&V notebook, it looks like the ANC values we have in the artifact don't vary by age, so I'm not sure whether we need it here.

But if we do say to stratify by age, does that mean we want the age group of the pregnant simulant? I think we should be as explicit as possible.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -144,34 +144,45 @@ all labeled and additional information will be included below.
...
   * - Input
     - Data Source 
-    - Notes
+    - Overall stratification
+    - Use case 
   * - ANC attendance
     - Decision tree point
-    - 
+    - Age, scenario 
+    - Decision tree value & export value
```
NathanielBlairStahn, line 157: To make these categories as clear as possible, I think the options should be something like:

- Input for later modeling stage (i.e., intrapartum or neonatal)
- V&V
- Final outputs

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -144,34 +144,45 @@ all labeled and additional information will be included below.
...
     - :ref:`Pregnancy model <other_models_pregnancy_closed_cohort_mncnh>`
-    - 
+    - Decision tree value & export value 
+    - Scenario
   * - Gestational age stated
     - Decision tree value
-    - We should track both the real GA at birth and the believed GA 
-  * - Birthweight
-    - :ref:`Pregnancy model <other_models_pregnancy_closed_cohort_mncnh>`
     -
```
NathanielBlairStahn, line 168: Is there a reason we're _not_ including scenario in the stratification for this or the later outputs? On the other hand, can we assume that scenario will _always_ be included in the stratification, so we don't need to explicitly include it at all?

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -144,34 +144,45 @@ all labeled and additional information will be included below.
...
-  :widths: 3 15 15
+  :widths: 3 15 15 15
   :header-rows: 1
 
   * - Input
     - Data Source 
-    - Notes
+    - Overall stratification
+    - Use case 
   * - ANC attendance
```
NathanielBlairStahn, line 154: This is not specific enough -- we need to specify how to measure ANC attendance. For example, we could say "ANC attendance (attended or did not attend)".

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -144,34 +144,45 @@ all labeled and additional information will be included below.
 
 
 .. list-table:: Outputs from Pregnancy Decision Tree
-  :widths: 3 15 15
+  :widths: 3 15 15 15
   :header-rows: 1
 
   * - Input
```
NathanielBlairStahn, line 150: Should this be "Output" instead of "Input"?

NathanielBlairStahn:
Based on Jim's feedback, I think it's still not totally clear from this table what format we want the output in. Here are some questions:

1. Do we want the output to simply be the population table, and the columns should include everything we've listed in the "Input" column (which should probably be called "Output")? If so, we should specify this explicitly, maybe in a sentence before the table.
2. Do we want any _other_ output besides the population table? For example, do we explicitly want the engineers to output the fraction of simulants who attended ANC, or the number of simulants with each type of ultrasound, or anything like that? Or is the population table enough? It seems like the population table is probably enough, but let me know if I'm missing something. If the population table is enough, maybe we should explicitly say that as well, e.g., "We do not need any additional outputs to be observed besides the columns in the population table listed below."

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -144,34 +144,45 @@ all labeled and additional information will be included below.
...
     - Decision tree point
-    - 
+    - Scenario
+    - Decision tree value & export for V&V
   * - Gestational age at birth
     - :ref:`Pregnancy model <other_models_pregnancy_closed_cohort_mncnh>`
-    - 
+    - Decision tree value & export for V&V 
+    - Scenario
   * - Gestational age stated
```
NathanielBlairStahn, line 166: I think we should rename this "Gestational age estimate" or "Estimated gestational age" to be more clear. But perhaps that's out of the scope of this PR since this term appears in other places as well.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -144,34 +144,45 @@ all labeled and additional information will be included below.
...
+    - Decision tree value and & export for V&V 
+  * - Birthweight
     - :ref:`Pregnancy model <other_models_pregnancy_closed_cohort_mncnh>`
     - 
+    - Decision tree value & & export for V&V  
   * - Pregnancy term (full term or partial term)
     - :ref:`Pregnancy model <other_models_pregnancy_closed_cohort_mncnh>`
     - 
+    - Decision tree value
+  * - Age of pregnant simulant
```
NathanielBlairStahn, line 178: When we say "Age of pregnant simulant," do we want their exact age, their GBD age group, or both?

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -490,6 +501,23 @@ Limitations:
...
+    - V&V summary
+    - Link to notebook
+  * - 1.0
+    - 
+      - Confirm ANC visit rate matches expectations
+      - Confirm ultrasound rates matches inputs for all scenarios
+      - Confirm stated gestational age and real gestational age have the correct margin of error based on ultrasound type
+      - Confirm pregnancy population is within expected WRA age group (15-49 years) 
+    - All checks passed except last one; RT is updating our observer output requests to add an observer for pregnant person age.
+    - https://github.com/ihmeuw/vivarium_research_mncnh_portfolio/blob/main/verification_and_validation/pregnancy_model.ipynb
```
NathanielBlairStahn, line 518: This would probably look better as hyperlinked text to avoid the long url in the table.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
NathanielBlairStahn: Seems like probably yes.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
NathanielBlairStahn: And whatever terms we use, we should probably define them outside the table.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
pletale: Agree that we should make it more specific, but should I request a counts table rather than a state table? I thought that maybe specifying stratifications implied counts but maybe I should be more explicit?

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
pletale: Yeah I added a row for age of pregnant simulant at the end of the table, but maybe I should change it to GBD age group, as you said in your other comment!

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
pletale: I'm happy to make this change and all the instances elsewhere on this page, in this PR! But lmk if you'd rather it be in a separate PR

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
pletale: Good point, I think we do want scenario stratification for all the outputs

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
NathanielBlairStahn: Hmm, yeah, this is related to my comment below. I'm not sure whether it's better to have the full population table or whether we want only some specific aggregate outputs from observers. Perhaps we can discuss at the MNCNH RT meeting tomorrow.

aflaxman (approved):
I was surprised by how high the baseline ultrasound rates are, let's make sure to call attention to the values and data sources in our next meeting with funder.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
NathanielBlairStahn:
```suggestion
  * - ANC attendance (true if attended ANC, false if not)
```

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -142,36 +142,44 @@ all labeled and additional information will be included below.
...
     - :ref:`Pregnancy model <other_models_pregnancy_closed_cohort_mncnh>`
-    - 
-  * - Low birthweight status
-    - :ref:`Pregnancy model <other_models_pregnancy_closed_cohort_mncnh>`
-    - 
+    - Input for later modeling stage & export for V&V  
   * - Pregnancy term (full term or partial term)
     - :ref:`Pregnancy model <other_models_pregnancy_closed_cohort_mncnh>`
-    - 
+    - Input for later modeling stage
```
NathanielBlairStahn, line 175:
```suggestion
    - Input for later modeling stage & export for V&V
```

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -142,36 +142,44 @@ all labeled and additional information will be included below.
...
+    - Use case 
   * - ANC attendance
     - Decision tree point
-    - 
+    - Input for later modeling stage & export for V&V
   * - Ultrasound status (AI assisted, standard, none)
     - Decision tree point
-    - 
+    - Input for later modeling stage & export for V&V
   * - Gestational age at birth
```
NathanielBlairStahn, line 164:
```suggestion
  * - Gestational age at birth (weeks)
```

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -142,36 +142,44 @@ all labeled and additional information will be included below.
...
     - :ref:`Pregnancy model <other_models_pregnancy_closed_cohort_mncnh>`
-    - 
-  * - Gestational age stated
-    - Decision tree value
-    - We should track both the real GA at birth and the believed GA 
+    - Input for later modeling stage & export for V&V
+  * - Gestational age estimate
```
NathanielBlairStahn, line 167:
```suggestion
  * - Gestational age estimate (weeks)
```

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -142,36 +142,44 @@ all labeled and additional information will be included below.
...
+    - Decision tree point
+    - Input for later modeling stage and & export for V&V 
   * - Birthweight
```
NathanielBlairStahn, line 170:
```suggestion
  * - Birthweight (grams)
```

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -142,36 +142,44 @@ all labeled and additional information will be included below.
     - :ref:`Pregnancy model <other_models_pregnancy_closed_cohort_mncnh>`
     - 
 
+The following table details outputs from the pregnancy model. Each row in this table should be a column in the population
+state table outputted by the model. RT will tabulate the population table with the stratifications needed for V&V (e.g., age group and 
+scenario). The 'Use case' column in the table denotes what we will be using the output for: either as an input for a later modeling stage
```
NathanielBlairStahn, line 147:
```suggestion
state table outputted by the model. RT will tabulate the population table with the stratifications needed for V&V (e.g., age group, 
scenario, and input draw). The 'Use case' column in the table denotes what we will be using the output for: either as an input for a later modeling stage
```

NathanielBlairStahn (approved):
Added a few suggestions to make things more explicit. Otherwise looks good.

PR #1544, opened 2024-11-02
Add graphviz extension

NathanielBlairStahn:
Add the [Sphinx Graphviz extension](https://www.sphinx-doc.org/en/master/usage/extensions/graphviz.html) so we can use a `.. graphviz::` directive to write code to generate Graphviz graphs directly in the docs.

I added a sample graph to the MNCNH maternal sepsis page to verify that the extension works.

If this gets merged, everyone will need to [install Graphviz](https://graphviz.org/download/) on their computer in order to render the docs without warnings/errors.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,6 +31,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -r requirements.txt
+          apt install graphviz
       - name: Doc build
         run: |
           make html -C docs/ SPHINXOPTS="-W --keep-going -n"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -r requirements.txt
-          apt install graphviz
+          sudo apt-get install graphviz
       - name: Doc build
         run: |
           make html -C docs/ SPHINXOPTS="-W --keep-going -n"
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -8,6 +8,8 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.8"
+  apt_packages:
+    - graphviz
 
 python:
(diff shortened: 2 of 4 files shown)


NathanielBlairStahn:
Looks like the build failed because Graphviz is not installed in the build environment on GitHub:
```
WARNING: dot command 'dot' cannot be run (needed for graphviz output), check the graphviz_dot setting
```
I have no idea how to install stuff for the GitHub checks or readthedocs. @zmbc or @rmudambi, do either of you know how to get this to work?

zmbc:
For readthedocs: https://stackoverflow.com/a/77369269

For the GitHub checks we should add an `apt install graphviz` step by copying [this one](https://github.com/ihmeuw/vivarium_research/blob/d2dd7480ebf0499b8cf4f52c93dd210346f6f51b/.github/workflows/build.yml#L31-L33).

NathanielBlairStahn:
Thanks @zmbc! For the GitHub checks, do you think we should add a line to the "Install dependencies" step, like this:
```yaml
- name: Install dependencies
        run: |
          pip install -r requirements.txt
          apt install graphviz
```
Or do you think we should add a completely new step, like this:
```yaml
- name: Install graphviz
        run: |
          apt install graphviz
```
Does it matter?

zmbc:
I think same step is fine! It doesn't matter much.

SylLutze (approved):

zmbc (approved):

NathanielBlairStahn:
**Note:** Instead of installing Graphviz via `sudo apt-get graphviz` as is done here, we may want to use `conda` to install Graphviz so that the same method will work across platforms, and we could include a script in the `vivarium_research` repo that would do the installation for you instead of everyone having to install Graphviz on their own. This method would be more flexible if we need to add any more dependencies in the future. Here's the JIRA ticket for adding an environment creation script: https://jira.ihme.washington.edu/browse/SSCI-2000

PR #1548, opened 2024-11-13
MNCNH: Maternal sepsis burden

NathanielBlairStahn:
Describes how to calculate YLLs and YLDs for maternal sepsis and other maternal infections in the MNCNH Portfolio sim.

I included a note about a limitation in our strategy: It won't correctly capture YLDs from infertility due to maternal sepsis. I'm not entirely confident in my logic arguing why I think our strategy will undercount averted YLDs, so it would be good to get some additional eyes on this.


albrja (approved):

NathanielBlairStahn:
I just snuck in an edit to the name of one of the nodes in the decision graph, which is unrelated to the rest of this PR.

PR #1549, opened 2024-11-14
MNCNH: Maternal sepsis cause hierarchy & restrictions

NathanielBlairStahn:
Add GBD cause hierarchy, restrictions, and descriptions of sequelae for maternal sepsis and other maternal infections.


pletale (approved):

zmbc (approved):
The fact that infertility causes DALYs but stillbirths don't is wild 👀

albrja (approved):

PR #1551, opened 2024-11-19
Adding backlog ticket links to to-dos

SylLutze:
Only the ones we moved from backlog yesterday. I suspect this will be the start of a longer project adding links between tickets and to-dos!


zmbc (approved):
Thank you Syl!

PR #1554, opened 2024-11-21
Add data tables from sepsis to OL page

pletale:
image
This PR adds the data tables from sepsis to the OL page, and also updates the wording in our decision graph (lmk if some other wording would be better!)

### DIFF
--- a/docs/source/models/causes/maternal_disorders/gbd_2021_mncnh/obstructed_labor.rst
+++ b/docs/source/models/causes/maternal_disorders/gbd_2021_mncnh/obstructed_labor.rst
@@ -161,8 +161,8 @@ while dashed lines indicate pieces of the underlying pregnancy model.
         rankdir = LR;
         ftp [label="full term\npregnancy, post\nintrapartum", style=dashed]
         ftb [label="full term\nbirth", style=dashed]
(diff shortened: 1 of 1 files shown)


docs/source/models/causes/maternal_disorders/gbd_2021_mncnh/obstructed_labor.rst
```diff
@@ -161,8 +161,8 @@ while dashed lines indicate pieces of the underlying pregnancy model.
         rankdir = LR;
         ftp [label="full term\npregnancy, post\nintrapartum", style=dashed]
         ftb [label="full term\nbirth", style=dashed]
-        alive [label="parent alive"]
-        dead [label="parent dead"]
+        alive [label="parent did not die of cause"]
+        dead [label="parent died of cause"]
```
NathanielBlairStahn, line 165: Instead of being generic and saying "cause", I think it would be better to be specific and say "obstructed labor" or "OL". Might be worth changing the variable names of the "alive" and "dead" nodes as well, for clarity, though I'm not sure what would be clear, short names.

aflaxman (approved):

PR #1572, opened 2024-12-11
Details on LBWSG Effect for MNCNH Neonatal Mortality Model

aflaxman:


docs/source/models/causes/neonatal/index.rst
```diff
@@ -158,32 +175,57 @@ unit time.
...
+where :math:`\text{mr}_i` is the probability of mortality for simulant :math:`i` during the early or late neonatal period, :math:`\text{ACMR}_i` is the all-cause mortality rate for the early or late neonatal period, and :math:`\Delta t` is length of that period in years.
+
+The calculation of :math:`\text{ACMR}_i` is a bit complicated, however. We begin with a population ACMR and use the LBWSG PAF to derive a risk-deleted ACMR to which we can then apply the relative risk of LBWSG matching each risk exposure category.  Mathematically this is achieved by the following formula:
+
+.. math::
+    \begin{align*}
+    \text{ACMR}_j &= \text{ACMR} \times (1 - \text{PAF}_{\text{LBWSG}}) \times \text{RR}_{\text{LBWSG},j},
+    \end{align*}
+
+where :math:`\text{ACMR}_j` is the all-cause mortality rate for the population with LBWSG exposure category :math:`j`, :math:`\text{ACMR}` is the all-cause mortality rate for the total population, :math:`\text{PAF}_{\text{LBWSG}}` is the population attributable fraction for LBWSG, and :math:`\text{RR}_{\text{LBWSG}}` is the relative mortality rate for LBWSG.
```
NathanielBlairStahn, line 205: This implies that we're using the categorical RRs from GBD. Is that what we want, or do we want to use our interpolated RRs, in which case each simulant would have an ACMR defined by this formula?

Also, the last sentence seems to be missing a subscript j on the RR (which would change to a subscript for the simulant if we use the interpolated RRs).

docs/source/models/causes/neonatal/index.rst
```diff
@@ -158,32 +175,57 @@ unit time.
...
+To obtain the ACMR for a specific simulant, we subtract off the *population* CSMRs for each modeled subcause for the LBWSG exposure level of the simulant, and then add back in the (potentially pipeline-modified) *individual* CSMRs for the specific simulant, which might differ from baseline due to intervention coverage.
 
 .. math::
     \begin{align*}
-    \text{mr} &= \frac{\text{neontal deaths}}{\text{live births}}\\
-        &= \frac{\text{(neonatal mortality rate)} \cdot  \text{(neonatal person-time)}}
-            {\sum_{\text{age groups}} \text{ASFR} \cdot (\text{age-specific person-time})
-            }.
+    \text{ACMR}_i &= \text{ACMR}_j - \sum_k \text{CSMR}_{k,j}^{\text{population}}
+    + \sum_k \text{CSMR}_{k,i},^{\text{individual}}
```
NathanielBlairStahn, line 212: Ah, so this is what you meant when you said we could modify the EMRs (or CSMRs) instead of having the intervention revive the simulant after we decide they're going to die, right?

docs/source/models/causes/neonatal/index.rst
```diff
@@ -158,32 +175,57 @@ unit time.
...
     \end{align*}
 
-The following table shows the data needed from GBD for these
-calculations.
+where :math:`\text{CSMR}_{k,j}^{\text{population}}` is the cause-specific mortality rate for subcause :math:`k` for the population with LBWSG exposure category :math:`j`, and :math:`\text{CSMR}_{k,i}^{\text{individual}}` is the cause-specific mortality rate for subcause :math:`k` for simulant :math:`i` (both detailed in the subcause models linked from this page).
```
NathanielBlairStahn, line 215: Somewhere before or after the above equation, you should specify that the simulant is assumed to be in LBWSG exposure category j.

docs/source/models/causes/neonatal/index.rst
```diff
@@ -158,32 +175,57 @@ unit time.
...
+
+In addition to determining which simulants die due to any cause, we also need to determine which subcause is underlying the death.  This is done by sampling from a multinomial distribution obtained by renormalizing the CSMRs:
```
NathanielBlairStahn, line 217: I generally prefer *categeorical distribution* for this instead of multinomial.

docs/source/models/causes/neonatal/index.rst
```diff
@@ -158,32 +175,57 @@ unit time.
...
+
+.. math::
+    \begin{align*}
+    \text{Pr}[\text{subcause} = k\;|\;\text{neonate died}] &= \frac{\text{CSMR}_{k,i}^{\text{individual}}}
+    {\text{ACMR}_i}.
+    \end{align*}
```
NathanielBlairStahn, line 223: Seems like somewhere we should say that one of these CSMR_k's should be the "other causes" CSMR, obtained by subtracting out the modeled causes.

docs/source/models/causes/neonatal/index.rst
```diff
@@ -128,28 +137,36 @@ unit time.
...
-        lb -> nn_dead [label = "mr"]
+        enn_alive [label="neonate survived\nfirst 7 days"]
+        enn_dead [label="neonate died\nduring first 7 days"]
+        lnn_alive [label="neonate survived\nfirst 28 days"]
+        lnn_dead [label="neonate died\nduring first 28 days"]
+
+        lb -> enn_alive  [label = "1 - enn_mr"]
+        lb -> enn_dead [label = "enn_mr"]
+        enn_alive -> lnn_alive [label = "1 - lnn_mr"]
+        enn_alive -> lnn_dead [label = "lnn_mr"]
```
NathanielBlairStahn, line 148: This graph seems excessively wide -- not sure if there's a good way to fix it

docs/source/models/causes/neonatal/index.rst
```diff
@@ -192,29 +234,26 @@ calculations.
...
-      - get_population
-      - Specific to age/sex/location/year demographic group. Numerically
-        equal to person-time for the year.
+    * - ACMR
+      - all-cause mortality rate (per person year)
+      - GBD
+      -
+    * - PAF_LBWSG
+      - population attributable fraction of all-cause mortality for low birth weight and short gestation
+      - GBD
```
NathanielBlairStahn, line 243: We've had problems using the GBD PAF for LBWSG before. I'm not clear on exactly what the problems were, but we may need to compute our own PAF. If we use interpolated RRs, we'll definitely need to compute our own PAF.

pletale (approved):

NathanielBlairStahn (approved):

docs/source/models/causes/neonatal/index.rst
aflaxman: I found some hack but there must be a better way
```
    digraph NN_decisions {
        graph [size="7,5"]
        rankdir = LR;
        ...
```

docs/source/models/causes/neonatal/index.rst
aflaxman: As discussed, I will merge this PR with the categorical approach and then follow-up with a PR to switch to continuous.

PR #1573, opened 2024-12-17
Switch to continuous LBWSG relative risk

aflaxman:


docs/source/models/causes/neonatal/index.rst
```diff
@@ -8,24 +8,22 @@ Neonatal all-cause mortality: GBD 2021, MNCNH
 
     This page is adapted from the :ref:`Maternal disorders: GBD 2021, MNCNH <2021_cause_maternal_disorders_mncnh>` page and is also part of the :ref:`MNCNH Portfolio project
     <2024_concept_model_vivarium_mncnh_portfolio>`.  In this work we are modeling
-    several neonatal subcauses (see Modeled
-    Subcauses) which is complicated because the LBWSG risk factor in GBD acts on all-cause mortality during the neonatal period.
+    several neonatal subcauses (see `Modeled Subcauses`_) which is complicated because the :ref:`Low Birth Weight and Short Gestation (LBWSG) <2019_risk_effect_lbwsg>` in GBD acts on all-cause mortality during the neonatal period.
```
NathanielBlairStahn, line 11:
```suggestion
    several neonatal subcauses (see `Modeled Subcauses`_) which is complicated because the :ref:`Low Birth Weight and Short Gestation (LBWSG) <2019_risk_effect_lbwsg>` risk factor in GBD acts on all-cause mortality during the neonatal period.
```
Also, is it complicated... or is it actually quite simple? :)

NathanielBlairStahn:
Why is this PR showing a bunch of commits that should have already been merged into main...?

docs/source/models/causes/neonatal/index.rst
```diff
@@ -126,30 +133,39 @@ unit time.
 .. graphviz::
 
     digraph NN_decisions {
+        graph [size="7,5"]
```
NathanielBlairStahn, line 136: This change seems to have [made the problem worse](https://vivarium-research--1573.org.readthedocs.build/en/1573/models/causes/neonatal/index.html#cause-model-decision-graph) (i.e., now you can barely see any of the graph).

Also... again, why are these changes showing up in this PR when [they have already been merged](https://vivarium-research.readthedocs.io/en/latest/models/causes/neonatal/index.html#cause-model-decision-graph)?

docs/source/models/causes/neonatal/index.rst
```diff
@@ -158,32 +174,62 @@ unit time.
...
+      - mortality risk during the late neonatal period
+      - The probability that a simulant who was born alive dies between day 8 to 28 of life
 
 
 Data Tables
 +++++++++++
 
-The neonatal death model requires only the probability of death (aka "mortality risk") for use
-in the decision graph. This will be computed as
+The neonatal death model requires only the probability of death (aka "mortality risk") for the early and late neonatal time periods.  But computing this for an individual simulant is a bit complicated.  It will follow the pattern from the general mortality component in :code:`vivarium_public_health`, and work in rate space to make the math simpler.  The final step will be converting from rates to risks.
```
NathanielBlairStahn, line 188:
```suggestion
The neonatal death model requires only the probability of death (aka "mortality risk") for the early and late neonatal time periods.  But computing this for an individual simulant is a bit complicated.  It will follow the pattern from the general mortality component in :code:`vivarium_public_health`, and work in rate space to make the math simpler.  The final step will be converting from rates to risks:
```

Also, is it really *that* complicated?

docs/source/models/causes/neonatal/index.rst
```diff
@@ -158,32 +174,62 @@ unit time.
...
-            {\sum_{\text{age groups}} \text{ASFR} \cdot (\text{age-specific person-time})
-            }.
+    \text{ACMR}_{\text{BW},\text{GA}} &= \text{ACMR} \times (1 - \text{PAF}_{\text{LBWSG}}) \times \text{RR}_{\text{BW},\text{GA}},
     \end{align*}
 
-The following table shows the data needed from GBD for these
-calculations.
+where :math:`\text{ACMR}_{\text{BW},\text{GA}}` is the all-cause mortality rate for a population with birth weight :math:`\text{BW}` and gestational age :math:`\text{GA}`, :math:`\text{ACMR}` is the all-cause mortality rate for the total population, :math:`\text{PAF}_{\text{LBWSG}}` is the population attributable fraction for LBWSG, and :math:`\text{RR}_{\text{BW},\text{GA}}` is the relative mortality rate for a specific birth weight :math:`\text{BW}` and gestational age :math:`\text{GA}`.
+
+To obtain the ACMR for a specific simulant, we subtract off the *population* CSMRs for each modeled subcause for the birth weight and gestational age of the simulant, and then add back in the (potentially pipeline-modified) *individual* CSMRs for the specific simulant, which might differ from baseline due to intervention coverage.
```
NathanielBlairStahn, line 206:
```suggestion
To obtain the ACMR for a specific simulant, we subtract off the *population* CSMRs for each modeled subcause for the birth weight and gestational age of the simulant, and then add back in the (potentially pipeline-modified) *individual* CSMRs for the specific simulant, which might differ from baseline due to intervention coverage:
```

docs/source/models/causes/neonatal/index.rst
```diff
@@ -239,7 +282,9 @@ For simplicity, we will not include YLDs in this model.
 Validation Criteria
 +++++++++++++++++++
 
-Neonatal mortality counts and rate in simulation should match GBD estimates.
+Neonatal deaths per live birth in simulation should match same ratio derived from GBD.
+
+Relative Risk of neonatal death at specific categories of LBWSG exposure should match GBD estimates.
```
NathanielBlairStahn, line 287: Ideally yes, we'd like this, but we don't currently expect these to match exactly because of (1) our interpolation of the RRs, and (2) we use a constant mortality hazard at each BW,GA, whereas I think GBD doesn't assume a constant mortality hazard in each LBWSG category.

docs/source/models/causes/neonatal/preterm_birth.rst
```diff
@@ -72,7 +72,7 @@ and sex of simulants to which different aspects of the cause model apply.
 Vivarium Modeling Strategy
 --------------------------
 
-Preterm birth is a PAF-of-one cause, meaning it is 100% attributable to the Low Birth Weight and Short Gestation (LBWSG) risk factor.  In models where this risk factor is included, it is important that only simulants with a LBWSG exposure category corresponding to preterm birth are able to acrrue DALYs from this cause.
+Preterm birth is a PAF-of-one cause, meaning it is 100% attributable to the :ref:`Low Birth Weight and Short Gestation (LBWSG) <2019_risk_effect_lbwsg>` risk factor.  It is important that only simulants with a gestational age of less than 38 weeks are able to accrue DALYs from this cause.
```
NathanielBlairStahn, line 75:
```suggestion
Preterm birth is a PAF-of-one cause, meaning it is 100% attributable to the :ref:`Low Birth Weight and Short Gestation (LBWSG) <2019_risk_effect_lbwsg>` risk factor.  It is important that only simulants with a gestational age of less than 37 weeks are able to accrue DALYs from this cause.
```
Doesn't preterm mean <37 weeks?

docs/source/models/causes/neonatal/preterm_birth.rst
```diff
@@ -126,8 +126,8 @@ unit time.
         is determined in the
         intrapartum step of the :ref:`pregnancy model
         <other_models_pregnancy_closed_cohort_mncnh>`)
-    * - neonate survived
-      - The child simulant survived for the first 28 days of life
+    * - neonate did not die of preterm
+      - The child simulant did not die of preterm birth complications during the first 28 days of life
```
NathanielBlairStahn, line 130: Don't we need to split this into 0-7 days and 7-28 days just like as described on the ACMR page?

docs/source/models/causes/neonatal/preterm_birth.rst
```diff
@@ -154,17 +154,31 @@ Data Tables
 The Preterm Birth model requires only the probability of death (aka "mortality risk") for use
 in the decision graph. This will be computed from the overall neonatal mortality risk and the cause-specific mortality fraction.
 
-Since this is a PAF-of-one cause, the calculation must take into account the "structural zeros" representing no mortality risk for simulants withe a gestational age above 37 weeks.
+Since this is a PAF-of-one cause, the calculation must take into account the "structural zeros" representing no mortality risk for simulants with a gestational age of 38 or more weeks.
```
NathanielBlairStahn, line 157:
```suggestion
Since this is a PAF-of-one cause, the calculation must take into account the "structural zeros" representing no mortality risk for simulants with a gestational age of 37 or more weeks.
```
Again, I thought "term" means $\ge 37$ weeks...

docs/source/models/causes/neonatal/preterm_birth.rst
```diff
@@ -154,17 +154,31 @@ Data Tables
...
 
-The details of this calculation require information about other subcauses as well as preterm, and therefore are included in the :ref:`Overall Neonatal Disorders Model <2021_cause_neonatal_disorders_mncnh>` page.
+The details of this calculation require information about other subcauses as well as preterm, and therefore are included in the :ref:`Overall Neonatal Disorders Model <2021_cause_neonatal_disorders_mncnh>` page.  This page describes the birth-weight- and gestational-age-specific cause specific mortality rates that are used for this cause on that page, :math:`\text{CSMR}^{\text{preterm with RDS}}_{\text{BW},\text{GA}}` and :math:`\text{CSMR}^{\text{preterm without RDS}}_{\text{BW},\text{GA}}`. In both cases, the formula is:
```
NathanielBlairStahn, line 159: > The details of this calculation require information about other subcauses as well as preterm

I find this confusing. Do you mean the calculation of the mortality risk? We never explicitly compute that, and you don't mention it anywhere after the first sentence of this section. And the calculations you do have below do *not* require any information about other subcauses.

docs/source/models/causes/neonatal/preterm_birth.rst
```diff
@@ -154,17 +154,31 @@ Data Tables
...
+    &=
+    \begin{cases}
+    \text{CSMR}\cdot f_k \cdot \text{RR}_{\text{BW},\text{GA}} \cdot Z, & \text{if BW} < 38; \\
+    0, & \text{if BW} \geq 38;
+    \end{cases}
+    \end{align*}
+
+where :math:`k` is the subcause of interest (preterm birth with or without RDS),
+:math:`\text{CSMR}` is the cause-specific mortality rate for preterm birth complications,
+:math:`f_k` is the fraction of preterm deaths due to subsubcause :math:`k` (with or without RDS), :math:`\text{RR}_{\text{BW},\text{GA}}` is the relative risk of all-cause mortality for a birth weight of :math:`\text{BW}` and gestational age of :math:`\text{GA}`, and :math:`Z` is a normalizing constant selected so that :math:`\int_{\text{BW<38}} \int_{\text{GA}} \text{RR}_{\text{BW},\text{GA}} \cdot Z = 1`.
```
NathanielBlairStahn, line 173: I think it's important to mention somewhere that the choice to use $\text{RR}\_{\text{BW,GA}}$ in this equation is essentially arbitrary, and it could be replaced by any other "weight function" $w(\text{BW,GA})$ as long it doesn't lead to a negative "other causes" mortality hazard.

I think it's important to mention this because, since there are a lot of pieces that go into this, it's useful to know which pieces are essential to getting it right and which ones aren't. Also, if we get more specific data about RDS death rates stratified by gestational age, we may want to change these weights to match that.

Also, I believe it is possible that the choice of $\text{RR}\_{\text{BW,GA}}$ might not work for every subcause. Since we're moving all the preterm mortality into the preterm categories, there is less room there for mortality from other causes, so depending on the hazards involved, we may need to shift mortality from some other causes into the non-preterm categories in order to avoid making things negative.

In fact, it's conceivable that there's no way to make this work at all (i.e., any choice of weight function leads to negative mortality hazards) if the GBD LBWSG data is inconsistent with the GBD preterm birth data. Hopefully that won't be an issue, but we haven't actually tried it with the real data yet.

docs/source/models/causes/neonatal/index.rst
```diff
@@ -192,29 +238,26 @@ calculations.
...
-      - Assume lognormal distribution of uncertainty. Units in GBD are
-        live births per person, or equivalently, per person-year.
-    * - population
-      - average population in a given year
-      - get_population
-      - Specific to age/sex/location/year demographic group. Numerically
-        equal to person-time for the year.
+    * - ACMR
+      - all-cause mortality rate (per person year)
+      - GBD
```
NathanielBlairStahn, line 243: Usually we're more specific about how to get the data from. E.g., to get ACMR, you probably have to pull deaths (with source = 'codcorrect'?) using `get_draws`, and pull population using `get_population`, then divide them. Perhaps the engineers have done this enough by now that we don't need to be so explicit anymore, but that's traditionally how we lay out these tables.

docs/source/models/causes/neonatal/index.rst
```diff
@@ -192,29 +238,26 @@ calculations.
...
+      -
+    * - PAF_LBWSG
+      - population attributable fraction of all-cause mortality for low birth weight and short gestation
+      - Computed to be consistent with GBD
```
NathanielBlairStahn, line 247: What does this mean? I.e., in what way are we trying to be consistent with GBD? Presumably we want to match GBD's ACMR, but we don't really need to know that to compute the LBWSG PAF. We just need to compute PAF = 1 - 1 / E(RR) from the interpolated relative risk function.

docs/source/models/causes/neonatal/index.rst
```diff
@@ -192,29 +238,26 @@ calculations.
...
+      -
+    * - RR_LBWSG
+      - relative risk of all-cause mortality for low birth weight and short gestation
+      - interpolated from GBD values, as described in :ref:`Low Birth Weight and Short Gestation (LBWSG) <2019_risk_effect_lbwsg>` docs
+      -
+    * - CSMR_k,BW,GA
```
NathanielBlairStahn, line 253: It could be nice to use the `:math:` role to match the formatting in the equations above and make it more clear that these are the same variables, but it's probably not essential. (Same goes for the other variables in the data table.)

docs/source/models/causes/neonatal/neonatal_encephalopathy.rst
```diff
@@ -134,10 +134,10 @@ unit time.
         is determined in the
         intrapartum step of the :ref:`pregnancy model
         <other_models_pregnancy_closed_cohort_mncnh>`)
-    * - neonate didn't die of encephalopathy
-      - The child simulant survived for the first 28 days of life
+    * - neonate did not die of encephalopathy
+      - The child simulant did not die of encephalopathy within the first 28 days of life
     * - neonate died of encephalopathy
-      - The child simulant died within the first 28 days of life
+      - The child simulant died of encephalopathy within the first 28 days of life
```
NathanielBlairStahn, line 140: Same comment as on the preterm page, that we probably need to split this into early neonatal and late neonatal instead of lumping all 28 days together.

docs/source/models/causes/neonatal/preterm_birth.rst
```diff
@@ -154,17 +154,31 @@ Data Tables
 The Preterm Birth model requires only the probability of death (aka "mortality risk") for use
```
NathanielBlairStahn, line 154: As I mentioned in the meeting, we're not explicitly using the mortality risk in our modeling strategy, which could make the diagram a bit confusing. Perhaps we should change this wording. And as I suggested, maybe we should add a note below the tables in the graph section, explaining that the graph is more conceptual and is not meant to model the control flow of the program. And perhaps we could write down a formula for calculating the mortality risk if you want it.

NathanielBlairStahn: Also, I think maybe the entire description of the equations should go in a section called something other than "Data Tables", e.g., "Modeling strategy" or something, and the "Data Tables" section should just contain the table showing where to get the values. The order in which everything appears could remain the same, just add a section title and move the Data Tables section title.

docs/source/models/causes/neonatal/index.rst
```diff
@@ -158,32 +174,62 @@ unit time.
...
-      - The probability that a simulant who was born alive dies during the neonatal period
+    * - enn_mr
+      - mortality risk during the early neonatal period
+      - The probability that a simulant who was born alive dies during the first 7 days
+    * - lnn_mr
+      - mortality risk during the late neonatal period
+      - The probability that a simulant who was born alive dies between day 8 to 28 of life
 
 
 Data Tables
```
NathanielBlairStahn, line 185: As I just noted on the preterm page, I think the "Data Tables" section should be reserved for the actual data tables, and we should add a new section (e.g., "Modeling strategy") to contain the modeling strategy. Also, it would be nice to add a table of contents to all the pages to make them easier to navigate.

docs/source/models/causes/neonatal/index.rst
```diff
@@ -158,32 +174,62 @@ unit time.
...
 +++++++++++
 
-The neonatal death model requires only the probability of death (aka "mortality risk") for use
-in the decision graph. This will be computed as
+The neonatal death model requires only the probability of death (aka "mortality risk") for the early and late neonatal time periods.  But computing this for an individual simulant is a bit complicated.  It will follow the pattern from the general mortality component in :code:`vivarium_public_health`, and work in rate space to make the math simpler.  The final step will be converting from rates to risks.
+
+.. math::
+    \begin{align*}
+    \text{mr}_i &= 1-\exp(-\text{ACMR}_i \cdot \Delta t),
+    \end{align*}
```
NathanielBlairStahn, line 193: If we compute the mortality risk this way, do you think it will match deaths per live birth as computed from GBD?

docs/source/models/causes/neonatal/index.rst
```diff
@@ -73,7 +80,7 @@ The following table describes the restrictions from GBD 2021 and our intended us
      - **Note:** GBD estimates both YLLs and YLDs for neonatal disorders, but
        **we are ignoring YLDs for this model**.
    * - YLD only
-     - True
+     - False
```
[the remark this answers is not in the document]
aflaxman: Maybe I forgot to push before I merged my last PR?

docs/source/models/causes/neonatal/index.rst
aflaxman: I will revert

docs/source/models/causes/neonatal/index.rst
aflaxman: I think it will, but Zeb just reminded us of an epic discussion about whether the `exp` is perhaps worse than a simpler version.

docs/source/models/causes/neonatal/index.rst
aflaxman: Can I use vivarium_inputs and gbd_mapping calls?  I don't want to look up all of the magic numbers for get_draws.

docs/source/models/causes/neonatal/index.rst
aflaxman: I will make that clearer

docs/source/models/causes/neonatal/neonatal_encephalopathy.rst
aflaxman: I'm going to leave these for now, and see if the engineers have a preferred change.  If they are confusing people too much, maybe I'll just delete them.

docs/source/models/causes/neonatal/preterm_birth.rst
aflaxman: as above, I will leave this for now

docs/source/models/causes/neonatal/preterm_birth.rst
aflaxman: I see.  I'll edit.

docs/source/models/causes/neonatal/preterm_birth.rst
aflaxman: I will add this as a note in the docs

docs/source/models/causes/neonatal/index.rst
NathanielBlairStahn: I guess we can leave it as is for now and see if the engineers complain.

PR #1575, opened 2024-12-19
ENH: add info to programming_vivarium page

aflaxman:
(And make package order match cookiecutter script from viavrium_research_template)

This will make it easier to use the cookiecutter script, although I also have a better solution on the script side...


zmbc (approved):

zmbc:
I'm guessing this is the "better solution on the script side" but it really should just default to the latest release...

aflaxman:
> I'm guessing this is the "better solution on the script side" but it really should just default to the latest release...

I'm pretty pleased with how I added defaults to the script: https://github.com/ihmeuw/vivarium_research_template/pull/92

NathanielBlairStahn (approved):

PR #1545, opened 2024-11-04
Add instructions to install Graphviz

NathanielBlairStahn:
Add instructions to install [Graphviz](https://graphviz.org/) on the "Contributing New Documentation" page. This PR is a follow-up to https://github.com/ihmeuw/vivarium_research/pull/1544, which, once merged, will require Graphviz to be installed when building the docs.

**Note:** Instead of having everyone install Graphviz on their own in a platform-dependent way as described here, it would probably be better to use `conda` to install it. Then the same method would work across platforms, and we could include an installation script in the repo that could do the installations automatically. Here's the JIRA ticket for creating such a script: https://jira.ihme.washington.edu/browse/SSCI-2000

Until that's accomplished, people should follow the instructions in this PR to manually install Graphviz on their own.

PR #1576, opened 2024-12-24
Upgrade Python and Sphinx

zmbc:
Upgrades us to latest stable version of Python and Sphinx. Build is passing and I manually inspected a smattering of pages on the local and RTD build and didn't see anything wacky. Verified that our CSS overrides were still working.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,9 +5,9 @@ version: 2
 # Configure the python version and environment construction run before
 # docs are built.
 build:
-  os: ubuntu-22.04
+  os: ubuntu-24.04
   tools:
-    python: "3.8"
+    python: "3.12"
   apt_packages:
     - graphviz
 
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -7,7 +7,7 @@ version: 2
 build:
   os: ubuntu-24.04
   tools:
-    python: "3.12"
+    python: "3.13"
   apt_packages:
     - graphviz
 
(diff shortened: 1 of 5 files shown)


rmudambi (approved):

zmbc:
Also supersedes #1545

pletale (approved):

.readthedocs.yml
```diff
@@ -5,9 +5,9 @@ version: 2
 # Configure the python version and environment construction run before
 # docs are built.
 build:
-  os: ubuntu-22.04
+  os: ubuntu-24.04
```
stevebachmeier, line 8: I think the sphinx defaults are ubuntu 22.04 and py 3.11. Did you go higher than that just to be super up to date?

requirements.txt
```diff
@@ -1,3 +1,4 @@
-sphinx==4.0.2
-sphinx_rtd_theme==0.5.2
+sphinx
```
stevebachmeier, line 1: We have pins on sphinx and sphinx-autobuild, but I honestly don't know why. I'm assuming this was the case here?

stevebachmeier (approved):

.readthedocs.yml
zmbc: What do you mean by the sphinx defaults? I just updated to the the most recent major releases of both.

requirements.txt
zmbc: This introduces sphinx-autobuild for the first time "officially" on the research side, though I have been using it since 2022. The pin to 4.0.2 was probably pretty current at the time it was made, but I don't know why it was pinned.

PR #1577, opened 2025-01-08
Add instructions to README about building docs

aflaxman:
I decided this would help me and maybe others, instead of having to find the instructions in the `onboarding_resources/contributing/index.html` subfolder.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,3 +12,13 @@ research projects.
 The hosted documentation can be found at https://vivarium-research.readthedocs.io/en/latest/.
 
 Information on how to contribute to vivarium research docs can be found at https://vivarium-research.readthedocs.io/en/latest/onboarding_resources/contributing/index.html. 
+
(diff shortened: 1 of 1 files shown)


zmbc:
I think we used to have this, and then found it frustrating that they could get out of sync. However with [Steve's `include` method](https://github.com/ihmeuw/easylink/blob/33c429f15f1c72d66201844692b3c4f42ead140c/docs/source/index.rst?plain=1#L19-L21) we could do this without that problem!

aflaxman:
> I think we used to have this, and then found it frustrating that they could get out of sync. However with [Steve's `include` method](https://github.com/ihmeuw/easylink/blob/33c429f15f1c72d66201844692b3c4f42ead140c/docs/source/index.rst?plain=1#L19-L21) we could do this without that problem!

Yes, let's do that!  Way better.

PR #1584, opened 2025-01-16
Update neonatal decision tree table with RDS data inputs 

pletale:
I am still looking for sources for CPAP capabilities by facility type for Nigeria and Pakistan, so using placeholder values from EmONC 2016 Final Report (Ethiopia) for now. The data here doesn't include NICU capabilities or relative risk on RDS, only CPAP, but I think we could find NICU data if it's needed!


docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -380,15 +380,26 @@ V&V Checks:
     - XX relative risk on incidence of encephalopathy if birthing parent experiences obstructed labor
     - Need to determine how neonatal mortality in general will be modeled and how we will handle overlaps with preterm and LBWSG RR's on all cause mortality
   * - 7
-    - XX% of each type of facility have CPAP or NICU capabilities
-    - 
+    - Percentage of each facility type that have CPAP capabilities: CMONC - 39.3% and BMONC - 7.5% 
+    - These values are from the 2016 EmONC Final Report and are therefore only reflective of Ethiopian health systems. Please use these as a placeholder for now while we
```
aflaxman, line 384:
```suggestion
    - These values are from the 2016 EmONC Final Report and are therefore only reflective of Ethiopian health system a decade ago. Please use these as a placeholder for now while we 
```

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -380,15 +380,26 @@ V&V Checks:
...
+    - 0.64 (95% CI 0.50-0.82) relative risk on RDS mortality of facility having CPAP capabilities 
+    - Need to confirm this will impact mortality not incidence. Also need to determine how neonatal mortality in general will be modeled 
+      and how we will handle overlaps with preterm and LBWSG RR's on all cause mortality. Note that we might want RR for NICU, but this value 
+      is for CPAP capabilities. Source: `2020 Cochrane review <https://pmc.ncbi.nlm.nih.gov/articles/PMC8094155/>`_
   * - 9
-    - XX relative risk for RDS incidence based on birthing parent receiving antenatal corticosteroids
-    - Need to confirm this will impact incidence not mortality. Also need to determine how neonatal mortality in general will be modeled and how we will handle overlaps with preterm and LBWSG RR's on all cause mortality
-
+    - 0.66 (95% CI 0.56-0.77) relative risk for RDS incidence based on birthing parent receiving antenatal corticosteroids
+    - This value is for incidence of RDS, however there is also an RR on RDS mortality (0.69, 95% CI 0.59-0.81). Also need to determine how neonatal mortality
```
aflaxman, line 393: We probably want 0.69 (0.59-0.81) as our RR, since we are only modeling mortality.  It is reassuring that these values are similar.  At this point, we have a strategy for how to model NN mortality in general and how to handle the overlaps.

aflaxman (approved):
Nice work, Alix.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -380,15 +380,26 @@ V&V Checks:
...
+    - Percentage of each facility type that have CPAP capabilities: CMONC - 39.3% and BMONC - 7.5% 
+    - These values are from the 2016 EmONC Final Report and are therefore only reflective of Ethiopian health systems. Please use these as a placeholder for now while we 
+      try to find reliable values for Nigeria and Pakistan. 
   * - 8
-    - XX relative risk on RDS mortality of facility having CPAP or NICU capabilities vs. not
-    - Need to confirm this will impact mortality not incidence. Also need to determine how neonatal mortality in general will be modeled and how we will handle overlaps with preterm and LBWSG RR's on all cause mortality
+    - 0.64 (95% CI 0.50-0.82) relative risk on RDS mortality of facility having CPAP capabilities 
+    - Need to confirm this will impact mortality not incidence. Also need to determine how neonatal mortality in general will be modeled 
+      and how we will handle overlaps with preterm and LBWSG RR's on all cause mortality. Note that we might want RR for NICU, but this value 
+      is for CPAP capabilities. Source: `2020 Cochrane review <https://pmc.ncbi.nlm.nih.gov/articles/PMC8094155/>`_
```
NathanielBlairStahn, line 390:
```suggestion
    - Note that we might want RR for NICU, but this value 
      is for CPAP capabilities. Source: `2020 Cochrane review <https://pmc.ncbi.nlm.nih.gov/articles/PMC8094155/>`_
```
The number you cite does in fact impact mortality, not incidence, right? And we've figured out how to model neonatal mortality, preterm, etc., so that sentence can be deleted.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -380,15 +380,26 @@ V&V Checks:
...
   * - 7
-    - XX% of each type of facility have CPAP or NICU capabilities
-    - 
+    - Percentage of each facility type that have CPAP capabilities: CMONC - 39.3% and BMONC - 7.5% 
+    - These values are from the 2016 EmONC Final Report and are therefore only reflective of Ethiopian health systems. Please use these as a placeholder for now while we 
+      try to find reliable values for Nigeria and Pakistan. 
   * - 8
-    - XX relative risk on RDS mortality of facility having CPAP or NICU capabilities vs. not
-    - Need to confirm this will impact mortality not incidence. Also need to determine how neonatal mortality in general will be modeled and how we will handle overlaps with preterm and LBWSG RR's on all cause mortality
+    - 0.64 (95% CI 0.50-0.82) relative risk on RDS mortality of facility having CPAP capabilities
```
NathanielBlairStahn, line 387: Are you sure 0.64 is the right number? Checking the source, it looks like this was the effect size for reducing "death or use of additional ventilatory support," whereas the RR for "mortality" was 0.53 (95% CI 0.34 to 0.83).

Also, is this the relative risk of "the facility having CPAP capabilities," or is it the relative risk of the neonate actually getting the CPAP intervention? My reading of the Cochrane review was that it's the latter (actually getting the intervention) -- does that seem right? We could potentially model these as different things in our simulation (i.e. a neonate born in a facility with CPAP, but they didn't actually get it), so we want to make sure we know what the number is measuring.

Also, it would be good to know what population this effect size applies to. E.g., was CPAP given to all preterm babies, or just those with RDS? According to the methods section, it says the type of participants were "preterm infants (less than 37 weeks of gestation) with respiratory failure becoming evident soon after birth," so I think it's the latter. We should probably note that somewhere, maybe in the notes column.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
NathanielBlairStahn: We can delete the sentence "Also need to determine how neonatal mortality in general will be modeled and how we will handle overlaps with preterm and LBWSG RR's on all cause mortality," because we've figured that out.

Also, it would be helpful to note who the intervention is given to, to make sure we model the intervention correctly. The Cochrane review says:
> Types of participants
Women, with a singleton or multiple pregnancy, expected to deliver preterm as a result of either spontaneous preterm labour, preterm prelabour rupture of the membranes or planned preterm delivery.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -380,15 +380,26 @@ V&V Checks:
...
   * - 9
-    - XX relative risk for RDS incidence based on birthing parent receiving antenatal corticosteroids
-    - Need to confirm this will impact incidence not mortality. Also need to determine how neonatal mortality in general will be modeled and how we will handle overlaps with preterm and LBWSG RR's on all cause mortality
-
+    - 0.66 (95% CI 0.56-0.77) relative risk for RDS incidence based on birthing parent receiving antenatal corticosteroids
+    - This value is for incidence of RDS, however there is also an RR on RDS mortality (0.69, 95% CI 0.59-0.81). Also need to determine how neonatal mortality 
+      in general will be modeled and how we will handle overlaps with preterm and LBWSG RR's on all cause mortality. 
+      Source: `2017 Cochrane review <https://pubmed.ncbi.nlm.nih.gov/28321847/>`_
+  * - 10
+    - Percentage of preterm deaths caused by RDS: Ethiopia - 45.3%; Nigeria - 52.4%; Pakistan - 35.6%
```
NathanielBlairStahn, line 397: I think we have to be careful with these numbers because of how GBD models things. Specifically, I think the percentages listed here tell us
> Among all babies who were born preterm and died, what fraction died of RDS?

But what we need for the model is 
> Among all babies who were born preterm **and died of preterm birth complications as defined by GBD**, what fraction died of RDS?"

For example, in Table 4 in the Ethiopia paper, I divided the causes of death into two groups:

- **Causes explicitly modeled by GBD (% of deaths):**

   - sepsis (26.1%)
   - pneumonia (3.0%, presumed to be captured in the cause "Lower respiratory infections")
   - meningitis (0.8%)
   - congenital anomalies (3.4%)
   - asphyxia (13.6%, presumed to be captured in the cause "Neonatal encephalopathy due to birth asphyxia and trauma)
   - Intraventricular haemorrhage (1.1%, presumed to be captured in GBD's intracerebral and subarachnoid hemorrhage causes, but we should double check that)
 
- **Causes not explicitly modeled by GBD (% of deaths):**
   - respiratory distress syndrome (45.3%)
   - apnoea (1.4%)
   - necrotising enterocolitis (0.8%)
   - others (4.5%)

Assuming the causes in the second category are included in GBD's "Neonatal preterm birth" cause (which we should double-check) and the ones in the first category aren't (which we should also double-check), the fraction of RDS deaths among GBD's preterm birth deaths would be 45.3 / (45.3 + 1.4 + 0.8 + 4.5) = 87.1%. (I think previously I had included intraventricular hemorrhage in the preterm birth cause, which would give 85.3%.)

@aflaxman and @pletale does that seem right?

If so, I think we'll have to do something similar for the other sources.

NathanielBlairStahn: Also, would it be better to put these numbers on the preterm birth cause page instead, so that we don't have to duplicate the data?

pletale: I'm struggling to find confirmation of what is included in GBD's neonatal preterm birth complications cause. The neonatal disorders methods appendix doesn't go into those details. Do either of you know where we might be able to confirm what is included? For now, I'll add a note that we need to confirm this.

NathanielBlairStahn (approved):
Looks good. After reviewing Abie's CPAP PR (and preterm birth PR), it seems like a lot of these numbers actually belong on the appropriate intervention or cause page, even though we had place holders for them on the concept model page. We should think about how to get these numbers onto the pages where they're needed while still having a single source of truth. But for now, here is good.

PR #1585, opened 2025-01-21
CPAP for RDS treatment intervention model

aflaxman:


NathanielBlairStahn:
I think interventions should generally go in the [intervention models section](https://vivarium-research.readthedocs.io/en/latest/models/intervention_models/index.html) of the docs, not in a separate subfolder for each simulation, and then you just link to it from the appropriate place in the concept model. Do others agree?

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
```diff
@@ -0,0 +1,130 @@
...
+
+Baseline Coverage Data
+++++++++++++++++++++++++
+
+The :ref:`MNCNH Portfolio model <2024_concept_model_vivarium_mncnh_portfolio>` includes data on current coverage levels for the specific locations we are focused on modeling.  We might wish to enhance this by predicting coverage in additional locations, in correlation with other key intervention technologies, based on sources such as existing SPA and SARA data.
+
+Vivarium Modeling Strategy
+--------------------------
+
+We will use the enhanced decision tree below to find the probability of RDS mortality with and without access to CPAP that are logically consistent with the baseline delivery facility rates and baseline CPAP coverage (and population level RDS mortality probability). Since the neonatal model include timesteps for both the early and late neonatal periods, this  alculation must be calculated for both time periods.
```
pletale, line 53:
```suggestion
We will use the enhanced decision tree below to find the probability of RDS mortality with and without access to CPAP that are logically consistent with the baseline delivery facility rates and baseline CPAP coverage (and population level RDS mortality probability). Since the neonatal model includes timesteps for both the early and late neonatal periods, this calculation must be calculated for both time periods.
```

pletale (approved):

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
```diff
@@ -0,0 +1,130 @@
...
+    - Continuous positive airway pressure
+    - 
+  * - RDS
+    - respiratory distress syndrome
+    - 
+
+Intervention Overview
+-----------------------
+
+CPAP is used for the prevention and treatment of respiratory distress syndrome. A recent (2020)Cochrane Review found that CPAP was associated with lower risk of treatment failure (death or use of assisted ventilation) and lower overall mortality with moderate‐certainty evidence.
```
zmbc, line 28:
```suggestion
CPAP is used for the prevention and treatment of respiratory distress syndrome. A recent (2020) Cochrane Review found that CPAP was associated with lower risk of treatment failure (death or use of assisted ventilation) and lower overall mortality with moderate‐certainty evidence.
```

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
```diff
@@ -0,0 +1,130 @@
...
+    - Note (ex: is this relationship direct or mediated?)
+  * - Preterm with RDS Mortality Probability
+    - Decrease multiplicatively
+    - Yes
+    - 
+
+Baseline Coverage Data
+++++++++++++++++++++++++
+
+The :ref:`MNCNH Portfolio model <2024_concept_model_vivarium_mncnh_portfolio>` includes data on current coverage levels for the specific locations we are focused on modeling.  We might wish to enhance this by predicting coverage in additional locations, in correlation with other key intervention technologies, based on sources such as existing SPA and SARA data.
```
zmbc, line 48: What are SPA and SARA?

zmbc, line 48: Should the baseline coverage data be here instead of in the concept model?

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
zmbc: It should say *probabilities in the first sentence

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
```diff
@@ -0,0 +1,130 @@
...
+
+We will use the enhanced decision tree below to find the probability of RDS mortality with and without access to CPAP that are logically consistent with the baseline delivery facility rates and baseline CPAP coverage (and population level RDS mortality probability). Since the neonatal model include timesteps for both the early and late neonatal periods, this  alculation must be calculated for both time periods.
+
+We will then add an attribute to each simulant indicating whether the birth occurs at a facility with access to CPAP (which will be dependent on the facility choice, i.e. no access for home births and lower access for BEmONC than CEmONC facilities).
+
+We will then use the conditional probabilities for simulants with and without access to determine which simulants die from RDS.
+
+.. todo::
+
+  include the relative risk of RDS mortality with CPAP here and in the concept model document, without duplicating it (see recent PR to add quickstart docs in the readme for an example of how)
```
zmbc, line 61: Is having it on both pages really what we want?

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
```diff
@@ -0,0 +1,130 @@
...
+        &= \sum_{\text{paths without CPAP}} p(\text{path})\cdot p_\text{RDS_wo}\\
+        &+ \sum_{\text{paths with CPAP}} p(\text{path})\cdot p_\text{RDS_w}\\[.1in]
+        p_\text{RDS_w} &= RR_\text{CPAP} \cdot p_\text{RDS_wo}
+    \end{align*}
+
+
+
+When we fill in the location-specific values for delivery facility rates, CPAP coverage, relative risk of mortality with CPAP access, and mortality probability (which is also age-specific), this becomes a system of two linear equations have two unknown (p_RDS_w and p_RDS_wo), which we can solve analytically.
+
+Scenarios
```
zmbc, line 107: I'm not sure I understand the purpose of this section

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
```diff
@@ -0,0 +1,130 @@
...
+
+- We assume that CPAP availability captures actual use, and not simply the machine being in the facility 
+- We assume that the delivery facility is the final facility in the care continum for deliveries that are transferred due to complications
+- We assume that the relative risk of RDS mortality with CPAP in practice is similar to that found in the Cochrane Review meta-analysis
+
+Validation and Verification Criteria
+------------------------------------
+
+- Population-level mortality rate should be the same as when this intervention is not included in the model
+- The ratio of RDS mortality rate among those without CPAP access divided by those with CPAP access should equal the relative risk from the Cochrane Review
```
zmbc, line 124: This seems unlikely to be true due to confounding with facility choice?

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
SylLutze: We have baseline coverage in other interventions I checked (SQLNS, SAM and MAM).

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
SylLutze: It should definitely be here, I agree we maybe don't need it on the concept model page.

SylLutze (approved):

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
NathanielBlairStahn: Yes, what are those? Also, this sentence is rather cryptic to me -- why would we want to expand to other locations besides the ones we're modeling, and what other intervention technologies are you talking about?

NathanielBlairStahn: Agree, baseline coverage should be on the intervention page.

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
NathanielBlairStahn: Whether early and late neonatal periods are modeled as separate "timesteps" is a technical detail that's not really the same as the relevant modeling decision -- the important point is that we're modeling these age groups separately because that's how the GBD data is organized.

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
NathanielBlairStahn: > I'm not sure I understand the purpose of this section

Is this section where we would describe how to model the CPAP intervention in different scenarios, or would that go in the concept model?

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
```diff
@@ -0,0 +1,130 @@
...
+
+.. todo::
+
+  include the relative risk of RDS mortality with CPAP here and in the concept model document, without duplicating it (see recent PR to add quickstart docs in the readme for an example of how)
+
+
+Calibration Strategy
+--------------------
+
+The following decision tree shows all of the paths from delivery facility choice to CPAP availability.  Each path corresponds to a disjoint event that we can sum over to find the population probability of RDS mortality.
```
NathanielBlairStahn, line 67:
```suggestion
The following decision tree shows all of the paths from delivery facility choice to CPAP availability.  Distinct paths in the tree correspond to disjoint events, which we can sum over to find the population probability of RDS mortality.
```

The original version sounded like the sum was over a single path or event, which is confusing.

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
```diff
@@ -0,0 +1,130 @@
...
+    \begin{align*}
+        p_\text{RDS} 
+        &= \sum_{\text{paths without CPAP}} p(\text{path})\cdot p_\text{RDS_wo}\\
+        &+ \sum_{\text{paths with CPAP}} p(\text{path})\cdot p_\text{RDS_w}\\[.1in]
+        p_\text{RDS_w} &= RR_\text{CPAP} \cdot p_\text{RDS_wo}
+    \end{align*}
+
+
+
+When we fill in the location-specific values for delivery facility rates, CPAP coverage, relative risk of mortality with CPAP access, and mortality probability (which is also age-specific), this becomes a system of two linear equations have two unknown (p_RDS_w and p_RDS_wo), which we can solve analytically.
```
NathanielBlairStahn, line 105: I'm curious how this decision-tree approach to baseline calibration compares to our usual approach of computing a PAF from the relative risks when we're modeling time and rates. They must be equivalent somehow, but it's not immediately obvious to me...

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
NathanielBlairStahn: @zmbc What do you mean? Do you mean that P(die of RDS | have RDS) is likely to be lower in facilities than in home births, even if the facility doesn't have a CPAP machine?

@aflaxman Can you be more specific when you talk about "mortality rate"? Do you mean rates per person-time, or probabilities (i.e., mortality risk)? Does the Cochrane review tell us the relative risk for the case fatality rate (i.e., the conditional probability I wrote above), or the overall mortality rate (or risk) of RDS among preterm babies? Or among all newborns? What population is in the denominator?

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
```diff
@@ -0,0 +1,130 @@
...
+
+Scenarios
+---------
+
+The :ref:`MNCNH Portfolio model <2024_concept_model_vivarium_mncnh_portfolio>` includes specific scenarios that might change the importance of this intervention. This includes interventions that might lower the prevalence of RDS (like ACS), and might also include interventions that increase the coverage of CPAP or increase the rate of preterm delivery with access CPAP.
```
NathanielBlairStahn, line 110: If you're going to mention ACS, you should define the abbreviation.

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
```diff
@@ -0,0 +1,130 @@
...
+        CEmONC -> CEmONC_wo  [label = "CPAP unavailable"]
+    }
+
+.. math::
+    \begin{align*}
+        p_\text{RDS} 
+        &= \sum_{\text{paths without CPAP}} p(\text{path})\cdot p_\text{RDS_wo}\\
+        &+ \sum_{\text{paths with CPAP}} p(\text{path})\cdot p_\text{RDS_w}\\[.1in]
+        p_\text{RDS_w} &= RR_\text{CPAP} \cdot p_\text{RDS_wo}
+    \end{align*}
```
NathanielBlairStahn, line 101: You should define your variables somewhere. Does $p_\text{RDS}$ mean the probability of dying from RDS? Is $p_\text{RDS-wo}$ the probability of dying from RDS if you don't get CPAP and $p_\text{RDS-w}$ the probability of dying of RDS if you do get CPAP? Are those probabilities conditional on having RDS or not?

The notation is a bit sloppy since it doesn't match the graph exactly, but maybe that's not a big deal. But `p_\text{RDS_w}` is malformed $LaTeX$ since underscores can only be used in math mode (GitHub gave me an error when I tried to render it).

Also, it would probably be good to explain how to compute $p(\text{path})$. It looks like the edges of your graph are events, not probabilities, right?

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
```diff
@@ -0,0 +1,130 @@
...
+
+.. graphviz::
+
+    digraph CPAP {
+        rankdir = LR;
+        start [color="none"]
+        facility [label="1"]
+        home [label="p_RDS_without_CPAP"]
+        BEmONC [label="7"]
+        CEmONC [label="7"]
```
NathanielBlairStahn, line 77: Why are the nodes labeled "1" and "7"? It seems like the "1" node should be something like "choose facility", and the "7" node should be something like "baby is born". What's the point of the `start` node and the `start -> facility [label = "delivery facility type"]` edge?

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
```diff
@@ -0,0 +1,130 @@
...
+We will then add an attribute to each simulant indicating whether the birth occurs at a facility with access to CPAP (which will be dependent on the facility choice, i.e. no access for home births and lower access for BEmONC than CEmONC facilities).
+
+We will then use the conditional probabilities for simulants with and without access to determine which simulants die from RDS.
+
+.. todo::
+
+  include the relative risk of RDS mortality with CPAP here and in the concept model document, without duplicating it (see recent PR to add quickstart docs in the readme for an example of how)
+
+
+Calibration Strategy
```
NathanielBlairStahn, line 64: I think it would be useful to add an explanation at the beginning of this section of what the goal is here and why we need to do it. Is this calibrating for baseline coverage, or is it slightly different?

NathanielBlairStahn, line 64: Is this analogous to what we usually do to calibrate for baseline coverage, or is the goal a bit different?

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
NathanielBlairStahn: Also, what makes this an "enhanced" decision tree?

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
```diff
@@ -0,0 +1,130 @@
...
+    - Decrease multiplicatively
+    - Yes
+    - 
+
+Baseline Coverage Data
+++++++++++++++++++++++++
+
+The :ref:`MNCNH Portfolio model <2024_concept_model_vivarium_mncnh_portfolio>` includes data on current coverage levels for the specific locations we are focused on modeling.  We might wish to enhance this by predicting coverage in additional locations, in correlation with other key intervention technologies, based on sources such as existing SPA and SARA data.
+
+Vivarium Modeling Strategy
```
NathanielBlairStahn, line 50: Where are we going to specify who gets the intervention and how to apply it? The RR from the Cochrane review is for giving CPAP to preterm babies who appear to have RDS, right? Since we're not modeling RDS incidence, what exactly do we do with that?

I think we have to modify the CSMR_i of RDS using p(CPAP | facility) (the probability of having a CPAP machine available given the facility you're born in). Is that right? We need to spell out those details somewhere. Is that planned for a later PR?

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
aflaxman: I will duplicate it here for now, and we can do a separate PR to remove it from the concept model or make it insert this there after #1584 is merged

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
aflaxman: ok

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
aflaxman: I was thinking that this would include the general approach to scenarios, e.g. increase coverage to different levels in different types of health facilities, and then the specific values would go in the concept model.

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
aflaxman: I think it cancels out, since it is a ratio, but I will write it out precisely

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
aflaxman: I think they are following the same approach, but there is not a PAF for us to pull from GBD so we are just skipping that part.  Maybe we should compute a PAF for lack of access to CPAP, just as an interesting intermediate value.

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
aflaxman: Good call, Antenatal corticosteroids (ACS) will eventually have its own intervention model page, which we can link to from here as well.

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
aflaxman: I'll add more detail, thanks for calling this out.  It looks ok in my local html, though:
image

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
aflaxman: I was trying to follow the labeling from the Neonatal Model Decision Tree in the Concept Model Document.  I guess that is not very helpful when they are on different page

1 is "Delivery facility type" (home, BEmONC, CEmONC) 
7 is "Availability to manage RDS (CPAP, NICU)" 

I was trying to make the decision tree show the decision question before each node, and the possible answers after, but it did not come out clearly. Maybe I'll try making the nodes the decision question, like instead of "1" say "Facility type?" and instead of "7" say "CPAP?"

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
aflaxman: Analogous; the goal is to match the population rates in the baseline scenario, in a way that we can tweak for intervention scenarios.

docs/source/models/concept_models/vivarium_mncnh_portfolio/mncnh_interventions.rst
aflaxman: I was thinking we would model "CPAP available" and then reduce the CSMR_i for the individuals who had CPAP available.  it sounds like that is what you were thinking, too.  I'll try to write it out more clearly.

PR #1590, opened 2025-02-05
sketch code to derive PAF and RR

aflaxman:


pletale (approved):

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
```diff
@@ -112,6 +112,24 @@ where :math:`p(\text{RDS})` is the probability of dying from Preterm with RDS in
...
+  p_BEmONC = .1
+  p_CEmONC = .4
+  p_CPAP_BEmONC = 0.075
+  p_CPAP_CEmONC = 0.393
+  RR_CPAP = 0.53
+
+  p_RDS_w = ... # solve system of equations from previous section
+  p_RDS_wo = ...
+
+  p_RDS_w = (1 - PAF_no_CPAP) * p_RDS
```
zmbc, line 130: `PAF_no_CPAP` is referenced before it is defined.

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
```diff
@@ -112,6 +112,24 @@ where :math:`p(\text{RDS})` is the probability of dying from Preterm with RDS in
...
+  # with appropriate artifact draws
+  p_RDS = .1
+  p_home = .5
+  p_BEmONC = .1
+  p_CEmONC = .4
+  p_CPAP_BEmONC = 0.075
+  p_CPAP_CEmONC = 0.393
+  RR_CPAP = 0.53
+
+  p_RDS_w = ... # solve system of equations from previous section
```
zmbc, line 127: This is overwritten immediately on line 130?

zmbc (commented):
I feel like I could use some more context here. Is the implicit framing here that the engineers want to model this component using a VPH `Risk` and `RiskEffect`? And why is this a code block? It seems like it would be more concise to simply state that for the purposes of modeling with `Risk` components, you can invert the RR of CPAP to get an RR of no CPAP. Though I'm not even sure we need to do this; why not just leave the RR as is?

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
aflaxman: I was doing more of a constraint programming than procedural programming thing here...  line 130 is really the definition of PAF_no_CPAP, but I thought it would be clearer to write out the way I did, instead of solving to put PAF_no_CPAP on the left hand side.

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
aflaxman: maybe it would be clearer if I had commented this line out
```suggestion
  # p_RDS_w = (1 - PAF_no_CPAP) * p_RDS
```

zmbc: Ah yes, that is clearer

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
zmbc: This feels discordant with your response to the other comment -- once you comment out 130 this can be seen as procedural again, unless I am missing something

aflaxman:
> I feel like I could use some more context here. Is the implicit framing here that the engineers want to model this component using a VPH `Risk` and `RiskEffect`? And why is this a code block? It seems like it would be more concise to simply state that for the purposes of modeling with `Risk` components, you can invert the RR of CPAP to get an RR of no CPAP. Though I'm not even sure we need to do this; why not just leave the RR as is?

Yes, exactly this is a way to implement the intervention effect using standard VPH components.  I was going to write out the full code for how to derive the PAF but (1) I ran out of time and (2) I thought it would be clearer this way.  The PAF construct does not work well with the RR for CPAP, and I think it will help the engineers to write out how to compute the PAF in addition to the RR transformation.

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
aflaxman: I'm going to merge the change commenting out line 130, and you can ignore my thinking out loud about constraint programing if it is not helpful.

zmbc:
>  The PAF construct does not work well with the RR for CPAP

Maybe bikeshedding here, but I don't agree, as there are negative PAFs even if you pull directly from GBD.

I think the context should be added to the docs itself, and also some text should be added before we jump into (pseudo-)code. There is really no way for the reader to understand why you would want any of this, without knowing about standard VPH components.

I think it makes sense to flip the RR just so the "risk" sounds like something "risky," but again this is purely because of how standard VPH components are named.

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
```diff
@@ -112,6 +112,24 @@ where :math:`p(\text{RDS})` is the probability of dying from Preterm with RDS in
 
 When we fill in the location-specific values for delivery facility rates, CPAP coverage, relative risk of mortality with CPAP access, and mortality probability (which is also age-specific), this becomes a system of two linear equations have two unknown (p_RDS_w and p_RDS_wo), which we can solve analytically.
 
+We can derive the PAF and RR of lack of access to CPAP from these quantities as follows::
```
aflaxman, line 115: @zmbc : Does this give enough context?  Maybe you can suggest an edit if it is not clear.

```suggestion
As mentioned above, we can calculate a PAF for "lack of access to CPAP".  This is convenient because Vivarium Public Health includes RiskExposure and RiskEffect components that can represent the way CPAP changes RDS mortality without requiring additional code.

We can derive the PAF and RR of "lack of access to CPAP" from these quantities as follows::
```

zmbc: Hm I want to suggest an edit but I'd like to edit line 65 as well and can't because you haven't touched it in this PR 🤔

OK I make a sub-PR into this branch?

aflaxman: I'll merge this suggestion and you can make a PR onto my PR

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
```diff
@@ -112,6 +112,26 @@ where :math:`p(\text{RDS})` is the probability of dying from Preterm with RDS in
 
 When we fill in the location-specific values for delivery facility rates, CPAP coverage, relative risk of mortality with CPAP access, and mortality probability (which is also age-specific), this becomes a system of two linear equations have two unknown (p_RDS_w and p_RDS_wo), which we can solve analytically.
 
+As mentioned above, we can calculate a PAF for "lack of access to CPAP".  This is convenient because Vivarium Public Health includes RiskExposure and RiskEffect components that can represent the way CPAP changes RDS mortality without requiring additional code.
+
+We can derive the PAF and RR of "lack of access to CPAP" from these quantities as follows::
+
+  # TODO: replace these stand-in values
+  # with appropriate artifact draws
+  p_RDS = .1
```
NathanielBlairStahn, line 121: Where is the real value for p(die from RDS) going to come from? We have a CSMR for RDS, and the conditional probability p(die from RDS | neonate died) = CSMR_RDS / ACMR. Do we also need to compute the probability p(neonate died)? I guess this would be:

$p(\text{neonate died}) = 1 - \exp(-\text{ACMR} \cdot \Delta t)$,

where $\Delta t$ is 7 days for the early neonatal period and 21 days for the late neonatal period. This is similar to the equation for the mortality risk $\text{mr}_i$ we have on the "Neonatal all-cause mortality" page, except there we're computing an "individualized" ACMR and mortality rate, whereas I think here we want the population level averages. Once we compute the overall mortality risk, then we should have

p(die from RDS) = p(die from RDS | neonate died) * p(neonate died)

Does that all look right?

I think we should spell this all out explicitly, as it wasn't obvious to me at first.

NathanielBlairStahn: @pletale This method of computing the probability of death from a specific cause could be relevant to V&V, as it gives a different way to compare to GBD. Namely, instead of computing (deaths from sepsis per live birth) = (deaths from sepsis) / (live births) as we discussed (where the numerator and denominator should be things you can pull directly from GBD), you could instead compute:

(deaths from sepsis per live birth)
= p(neonate dies from sepsis)
= p(neonate dies from sepsis | neonate dies) * p(neonate dies)
= [(CSMR_sepsis) / ACMR] * (1 - exp(-ACMR * $\Delta t$)
~= [(CSMR_sepsis) / ACMR] * ACMR * $\Delta t$ (this is an approximation for small $\Delta t$)
= CSMR_sepsis * $\Delta t$,

where $\Delta t$ corresponds to either the early or late neonatal period. This method more closely matches what's actually happening in the simulation, although theoretically I think it should give the same the same thing as the direct calculation above... (hmm, except for the late neonatal period it seems like the denominator for this method shouldn't be live births but the population at the start of the late neonatal period...)

It would probably be good to check how close these different calculations are.

NathanielBlairStahn:
This page still looks like it's missing an explicit description of how to apply the intervention to the simulants. It's implicitly described in the sentences, "Instead the model will have different cause-specific mortality rates for RDS for individuals born with and without access to CPAP," (line 61) and, "This is convenient because Vivarium Public Health includes RiskExposure and RiskEffect components that can represent the way CPAP changes RDS mortality without requiring additional code," (line 115), but since model is not exactly the same as our standard Vivarium models, I think it would be better to explicitly show how this relates to the quantities on the "Neonatal all-cause mortality" page.

Namely, in the absence of other interventions, I think we should have:

$\text{CSMR}\_{i}^\text{RDS} =  \text{CSMR}\_{\text{BW}\_i,\text{GA}\_i}^\text{RDS} \cdot (1 - \text{PAF}\_\text{no CPAP}) \cdot \text{RR}\_i^\text{no CPAP}$,

where $\text{RR}\_i^\text{no CPAP}$ is simulant $i$'s individual relative risk for "no CAAP." That is, $\text{RR}\_i^\text{no CPAP} = \text{RR}\_\text{no CPAP}$ if simulant $i$ is born in a facility without CPAP, and $\text{RR}\_i^\text{no CPAP} = 1$ if simulant $i$ is born in a facility with CPAP.

And I guess if there are other interventions affecting the CSMR of RDS, it would be something more like

$\text{CSMR}\_{i, \text{updated}}^\text{RDS} =  \text{CSMR}\_{i, \text{original}}^\text{RDS} \cdot (1 - \text{PAF}\_\text{no CPAP}) \cdot \text{RR}\_i^\text{no CPAP}$.

Is that correct?

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
aflaxman: I tried to put this on the neonatal mortality cause model page, and I don't want to repeat it here.  Maybe I should add a link.  Also, can you see if this material should be added there or if you have additional suggestions for how to clarify it, while this is on the top of your mind?  https://vivarium-research.readthedocs.io/en/latest/models/causes/neonatal/index.html

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
```diff
@@ -112,6 +112,26 @@ where :math:`p(\text{RDS})` is the probability of dying from Preterm with RDS in
...
+  p_home = .5
+  p_BEmONC = .1
+  p_CEmONC = .4
```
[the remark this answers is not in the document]
aflaxman: I will put them in for now, because I think the concreteness is helpful
> At home (68.3%), in hospital (26.6%), in clinic/low-level facility (5.1%)

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
```diff
@@ -112,6 +112,26 @@ where :math:`p(\text{RDS})` is the probability of dying from Preterm with RDS in
...
+  p_CPAP_BEmONC = 0.075
+  p_CPAP_CEmONC = 0.393
+  RR_CPAP = 0.53
+
+  p_RDS_w = ... # solve system of equations from previous section
+  p_RDS_wo = ...
+
+  # p_RDS_w = (1 - PAF_no_CPAP) * p_RDS
+  PAF_no_CPAP = ... # rearange equation from line above to solve for this
```
NathanielBlairStahn, line 133: I'm left with a rather nebulous feeling after reading this. I feel like we might as well be more explicit here. Namely, if we note that

$p(\text{no CPAP}) = \sum\_{\text{paths without CPAP}} p(\text{path})$,

then we have, as always,

$\text{PAF}\_\text{no CPAP} = 1 - \frac{1}{E(\text{RR})}$,

where, since our risk is binary,

$E(\text{RR}) = p(\text{no CPAP}) \cdot \text{RR}\_\text{no CPAP} + (1 - p(\text{no CPAP})) \cdot 1$.

And I think that should be all we need. I don't think we actually need the values of $p(\text{RDS} \mid \text{no CPAP})$ and $p(\text{RDS} \mid \text{CPAP})$, right?

I think if you combine the two equations you started with, you get

$p(\text{RDS} \mid \text{CPAP}) = \frac{1}{E(\text{RR})} \cdot p(\text{RDS})$,

so if you _define_ the PAF by the equation $p(\text{RDS} \mid \text{CPAP}) = (1 - \text{PAF})\cdot p(\text{RDS})$, then you get the formula for the PAF above. But it's just our standard PAF formula, so I don't see much point in re-deriving it...

aflaxman:
You are correct, and you have made me think it would be helpful to add text like this (but with better typesetting)

> To be precise, in the absence of other interventions, we have:
> 
> CSMR i RDS = CSMR BW i , GA i RDS ⋅ ( 1 − PAF no CPAP ) ⋅ RR i no CPAP ,
> 
> where RR i no CPAP is simulant i 's individual relative risk for "no CAAP." That is, RR i no CPAP = RR no CPAP if simulant i is born in a facility without CPAP, and RR i no CPAP = 1 if simulant i is born in a facility with CPAP.
>

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
NathanielBlairStahn: If you go with my suggestion in [this comment below](https://github.com/ihmeuw/vivarium_research/pull/1590#discussion_r1945621410), then you don't even need to compute p(die of RDS)...

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
aflaxman: I see the logic of your approach. The way I wrote it is the way I tend to think about PAFs, but maybe your way is simpler.  In either way, we have to sum over all the paths without CPAP.  If the engineers find this way easier to think about, let's go with it.

NathanielBlairStahn:
> This page still looks like it's missing an explicit description of how to apply the intervention to the simulants.
> 
> It's implicitly described in the sentences, "Instead the model will have different cause-specific mortality rates for RDS for individuals born with and without access to CPAP," (line 61) and, "This is convenient because Vivarium Public Health includes RiskExposure and RiskEffect components that can represent the way CPAP changes RDS mortality without requiring additional code," (line 115). But since this model is not exactly the same as our standard Vivarium models, I think it would be better to explicitly show how this relates to the quantities on the "Neonatal all-cause mortality" page.

Oh, I guess this is a bit more explicit than I thought -- I had missed the description in the Affected Outcomes table (line 46).

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
zmbc: I strongly agree that $p(\text{RDS} \mid \text{CPAP}) = (1 - \text{PAF})\cdot p(\text{RDS})$ is not the _definition_ of a PAF, but the equation we use to _apply_ PAFs in Vivarium; I've labeled it as such in my PR.

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
NathanielBlairStahn: > I tried to put this on the neonatal mortality cause model page, and I don't want to repeat it here.

We don't actually explain this on the neonatal mortality page -- we explain how to get the "individual all-cause mortality risk" $\text{mr}_i$, which is different from $p(\text{RDS})$ = "the average cause-specific mortality risk for RDS," even though the calculations are similar.

> Also, can you see if this material should be added there or if you have additional suggestions for how to clarify it, while this is on the top of your mind?

I'm still not convinced it's worth the effort to do this computation and solve the system of equations, so I don't know where the best location would be...

NathanielBlairStahn (approved):
Approving so we can get this merged, along with https://github.com/ihmeuw/vivarium_research/pull/1591

PR #1591, opened 2025-02-06
Explain use of risk components for CPAP intervention more explicitly

zmbc:
Suggestion of how to address https://github.com/ihmeuw/vivarium_research/pull/1590#issuecomment-2640640518

### DIFF
--- a/docs/source/models/intervention_models/neonatal/cpap_intervention.rst
+++ b/docs/source/models/intervention_models/neonatal/cpap_intervention.rst
@@ -110,7 +110,7 @@ where :math:`p(\text{RDS})` is the probability of dying from Preterm with RDS in
 
 When we fill in the location-specific values for delivery facility rates, CPAP coverage by facility type, relative risk of mortality with CPAP access, and mortality probability (which is also age-specific), this becomes a system of two linear equations with two unknowns (p_RDS_w and p_RDS_wo), which we can solve analytically.
 
-As mentioned above, it is convenient to model this like a dichotomous risk factor, so that we can reuse the
+As mentioned above, it is convenient to model this intervention like a dichotomous risk factor, so that we can reuse the
 :class:`Risk<vivarium_public_health.risks.base_risk.Risk>`
 and :class:`RiskEffect<vivarium_public_health.risks.effect.RiskEffect>` components in Vivarium Public Health,
 rather than having to write new components from scratch.
--- a/docs/source/models/intervention_models/neonatal/cpap_intervention.rst
+++ b/docs/source/models/intervention_models/neonatal/cpap_intervention.rst
@@ -44,16 +44,16 @@ This section describes how a CPAP intervention can be implemented and calibrated
     - Modeled?
     - Note (ex: is this relationship direct or mediated?)
   * - Preterm with RDS Mortality Probability (:math:`\text{CSMR}_i^\text{RDS}`)
-    - Adjust multiplicatively using PAF and RR, just like a dichotomous risk factor
+    - Adjust multiplicatively using RR
     - Yes
-    - The approach to calculating the PAF is complicated and described below
+    - For convenience, we will model this like a dichotomous risk factor; more details below
 
 Baseline Coverage Data
 ++++++++++++++++++++++++
 
 39.3% of CEmONC facilities and 7.5% of BEmONC facilities have CPAP, according to the 2016 Ethiopia EmONC Final Report.  Please use these as a placeholder for now while we try to find reliable values for Nigeria and Pakistan. 
 
-We might be able to borrow strength from other locations and times by predicting coverage more country-years simultaneously, perhaps even in combination with other key intervention technologies, based on sources such as existing `Service Provision Assessment (SPA) <https://www.dhsprogram.com/methodology/Survey-Types/SPA.cfm>`_ and `Service Availability and Readiness Assessment (SARA) <https://www.who.int/data/data-collection-tools/service-availability-and-readiness-assessment-(sara)>`_ data.
+We might be able to borrow strength from other locations and times by predicting coverage for more country-years simultaneously, perhaps even in combination with other key intervention technologies, based on sources such as existing `Service Provision Assessment (SPA) <https://www.dhsprogram.com/methodology/Survey-Types/SPA.cfm>`_ and `Service Availability and Readiness Assessment (SARA) <https://www.who.int/data/data-collection-tools/service-availability-and-readiness-assessment-(sara)>`_ data.
 
 Vivarium Modeling Strategy
 --------------------------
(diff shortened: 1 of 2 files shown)


docs/source/models/intervention_models/neonatal/cpap_intervention.rst
```diff
@@ -110,11 +108,31 @@ The following decision tree shows all of the paths from delivery facility choice
...
+Instead, we flip it so the risk factor is "*lack* of access to CPAP."
+
+The :code:`RiskEffect` component expects a relative risk (RR) and a population-attributable fraction (PAF).
+Because we are flipping the direction of the risk factor, we need to use the inverse of our original RR, so:
+
+.. math::
+    \text{RR}_{\text{no CPAP}} = \frac{1}{\text{RR}_{\text{CPAP}}}
+
+The PAF is the proportion of deaths due to preterm with RDS that would not occur if all births had access to CPAP.
+Since we use the equation :math:`p(\text{RDS}|\text{CPAP}) = (1 - \text{PAF}_\text{no CPAP}) \cdot p(\text{RDS})`
```
NathanielBlairStahn, line 127: I'm not familiar with the code in the `RiskEffect` component, but based on how this intervention should work, where would this equation get used?

I think the simulation is never using the probabilities $p(\text{RDS}|\text{CPAP})$ and $p(\text{RDS})$ directly, because it is operating on CSMRs, not probabilities. I.e., the "treatment-deleted" $\text{CSMR}\_i^\text{RDS}$ would be $(1 - \text{PAF})\cdot \text{CSMR}\_i^\text{RDS}$, which then gets multiplied by the simulant's individual RR. It would be interesting to explain why the PAF for the CSMRs should be the same as the PAF for the probabilities, which is not immediately obvious to me...

zmbc: You're right and I guess I was hoping to get away with this imprecision. Once we have `p = r * dt` in Vivarium, the PAFs will be the same. But right now they (slightly) aren't, but should be really close if our timesteps are short

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
```diff
@@ -110,11 +108,31 @@ The following decision tree shows all of the paths from delivery facility choice
...
+in the :code:`RiskEffect` component, we solve for :math:`\text{PAF}_\text{no CPAP}` as follows:
+
+.. math::
+    \text{PAF}_{\text{no CPAP}} = 1 - \frac{p(\text{RDS}|\text{CPAP})}{p(\text{RDS})}
```
NathanielBlairStahn, line 131: Ok, this equation explains why it's helpful to solve for the probability $p(\text{RDS}|\text{CPAP})$ in order to compute the PAF, which wasn't immediately obvious to me. It would also be nice to note that this is actually the same as our usual PAF equation, i.e.,

$PAF = 1 - \frac{1}{E(\text{RR})}$,

where $E(\text{RR}) = p(\text{no CPAP}) \cdot \text{RR}_\text{no CPAP} + (1 - p(\text{no CPAP})) \cdot 1$, because when you solve, you get $\frac{p(\text{RDS}|\text{CPAP})}{p(\text{RDS})} = \frac{1}{E(\text{RR})}$. It wasn't obvious to me why solving for $p(\text{RDS}|\text{CPAP})$ was helpful because using the formula above, you only need to compute $p(\text{no CPAP})$, which is the sum of the probabilities of the paths without CPAP. I guess technically you _don't_ need to solve for $p(\text{RDS}|\text{CPAP})$, you just need to solve for the ratio $\frac{p(\text{RDS}|\text{CPAP})}{p(\text{RDS})}$.

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
```diff
@@ -110,11 +108,31 @@ The following decision tree shows all of the paths from delivery facility choice
...
 
-As mentioned above, we can calculate a PAF for "lack of access to CPAP".  This is convenient because Vivarium Public Health includes RiskExposure and RiskEffect components that can represent the way CPAP changes RDS mortality without requiring additional code.
+where the terms on the right hand side can be obtained by solving the system of equations above.
 
-We can derive the PAF and RR of "lack of access to CPAP" from these quantities as follows::
+Here is some pseudocode for deriving the PAF and RR of "lack of access to CPAP" from data::
```
NathanielBlairStahn, line 135: Seems like this hardly counts as pseudocode, since all the interesting parts are left as ellipses and comments...

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
```diff
@@ -110,11 +108,31 @@ The following decision tree shows all of the paths from delivery facility choice
...
 
-When we fill in the location-specific values for delivery facility rates, CPAP coverage, relative risk of mortality with CPAP access, and mortality probability (which is also age-specific), this becomes a system of two linear equations have two unknown (p_RDS_w and p_RDS_wo), which we can solve analytically.
+When we fill in the location-specific values for delivery facility rates, CPAP coverage by facility type, relative risk of mortality with CPAP access, and mortality probability (which is also age-specific), this becomes a system of two linear equations with two unknowns (p_RDS_w and p_RDS_wo), which we can solve analytically.
```
NathanielBlairStahn, line 111: It is still not obvious to me how we are supposed to get the value of $p(\text{RDS})$, as we don't explain it anywhere (see [this comment](https://github.com/ihmeuw/vivarium_research/pull/1590#discussion_r1945539294)).

NathanielBlairStahn: It still seems to me like computing $p(\text{RDS})$ and going through all the trouble of solving these equations is overkill for computing the PAF, when all we need is $p(\text{no CPAP})$.

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
zmbc: Yeah, I feel like all this really adds is the order of operations: first solve the system of equations, then calculate PAF and RR. Maybe this should just be prose and say that. Plus, I'm not sure if others agree, but as I mentioned this morning I think we should solve the system of equations explicitly here.

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
NathanielBlairStahn: This simulation has no timesteps... we're using $p(\text{neonate dies}) = 1 - \exp(-\text{ACMR} \cdot \Delta t)$, where $\Delta t$ is either 7 or 21 days.

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
NathanielBlairStahn: I agree, I think we should solve it explicitly.

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
zmbc: Oh right! Yeah, in that case it does technically matter. I suppose we could change this whole page to deal in rates instead of probabilities? Or is there a simpler path forward?

NathanielBlairStahn: > You're right and I guess I was hoping to get away with this imprecision.

I think you can get away with the imprecision -- maybe just don't say the `RiskEffect` component uses that equation, since I don't think that's true in this case, which could be confusing.

NathanielBlairStahn:
I think overall this looks like a good explanation. But really we should have a separate page somewhere that explains that in order to leverage the existing Vivarium framework, we often handle interventions by treating "lack of intervention" as a risk factor. Maybe that should go on the page explaining about [how to model interventions](https://vivarium-research.readthedocs.io/en/latest/model_design/vivarium_model_components/interventions/index.html)?

zmbc:
> Maybe that should go on the page explaining about [how to model interventions](https://vivarium-research.readthedocs.io/en/latest/model_design/vivarium_model_components/interventions/index.html)?

Made a ticket for this: https://jira.ihme.washington.edu/browse/SSCI-2116

PR #1592, opened 2025-02-11
Document RR and PAF affected measures, current NO fix

zmbc:

PR #1594, opened 2025-02-13
Proprosal for neonatal mortality equation to account for causes unaffected by LBWSG

alibow:
This isn't formatted or detailed in a manner worthy for final documentation, but before I stop working for the week I just wanted to outline what I was thinking could work for this approach that will remain consistent with prior approaches and GBD assumptions around potential LBWSG reverse causality concerns. In case anyone would like to take a look at existing documentation about the difference between affected and unaffected causes, it is discussed on this page: https://vivarium-research.readthedocs.io/en/latest/models/risk_effects/low_birthweight_short_gestation/index.html

Essentially, I want to satisfy the following criteria:
- mortality due to unaffected causes should be higher among LBW/preterm simulants (HIV and malaria can _cause_ LBW/preterm, so there will be an observed association with HIV and malaria mortality), so we apply LBWSG RRs to these unaffected causes at baseline
- improvements in LBWSG due to interventions like MMS should *not* reduce mortality rates from causes unaffected by LBWSG like malaria and HIV (GBD says LBWSG PAF for these causes is zero), so we continue to calculate mortality due to unaffected causes according to unchanged baseline LBWSG exposure (even if there are intervention-associated changes)
- improvements in LBWSG due to interventions like MMS *should* reduce mortality rates from causes affected by LBWSG like LRI, so we use RRs specific intervention-shifted LBWSG exposure to calculate affected mortality rates

(I didn't specify it in the :math:`\sum_k \text{CSMR}_{i}^{k}` term, but we should follow the same baseline versus current LBWSG exposure pattern here too for affected versus unaffected causes).

This approach is still limited by: 
(1) the fact that the RRs are not cause-specific, so we are assuming they are the same for affected and unaffected causes (GBD is subject to this same limitation)
(2) we don't allow for any possibility that despite HIV/malaria _causing_ poor LBWSG exposure that a baby who has HIV/malaria and an _improved_ LBWSG exposure through interventions like MMS may still have a better outlook/lower chance of mortality. Again, GBD is still subject to this same limitation due to data issues and not being able to disentangle reverse causality from the overall association

So this is obviously not the perfect approach, but at least it is (a) consistent with GBD assumptions and (b) more consistent with how we have most recently modeled the impact of interventions that act through LBWSG.

### DIFF
--- a/docs/source/models/causes/neonatal/index.rst
+++ b/docs/source/models/causes/neonatal/index.rst
@@ -221,12 +221,19 @@ To obtain the ACMR for a specific simulant, we subtract off the *population* CSM
 
 .. math::
     \begin{align*}
-    \text{ACMR}_i &= \text{ACMR}_{\text{BW}_i,\text{GA}_i} - \sum_k \text{CSMR}_{\text{BW}_i,\text{GA}_i}^{k}
+    \text{ACMR}_i &= \text{UUMR}_{\text{baseline BW}_i,\text{baseline GA}_i} + \text{UAMR}_{\text{current BW}_i,\text{current GA}_i} - \text{MUMR}_{\text{baseline BW}_i,\text{baseline GA}_i} - \text{MAMR}_{\text{current BW}_i,\text{current GA}_i}
     + \sum_k \text{CSMR}_{i}^{k},
     \end{align*}
 
-where :math:`\text{BW}_i` and :math:`\text{GA}_i` are the birth weight and gestational age for simulant :math:`i`,
-:math:`\text{CSMR}_{\text{BW}_i,\text{GA}_i}^{k}` is the cause-specific mortality rate for subcause :math:`k` for a population with the same gestational age and birth weight as this simulant, 
+Where:
+
+  - :math:`UUMR` is the "unmodeled unaffected mortality rate," equal to the sum of cause-specific mortality rates for all **unmodeled** causes that are **unaffected** by the LBWSG
+  - :math:`UAMR` is the "unmodeled affected mortalty rate," equal to the sum of cause-specific mortality rates for all **unmodeled** causes that are **affected** by the LBWSG risk factor
+  - :math:`MUMR` is the "modeled unaffected mortalty rate," equal to the sum of cause-specific mortality rates for all **modeled** causes that are **unaffected** by the LBWSG risk factor
+  - :math:`MAMR` is the "modeled affected mortalty rate," equal to the sum of cause-specific mortality rates for all **modeled** causes that are **affected** by the LBWSG risk factor
+
+where :math:`\text{BW}_i` and :math:`\text{GA}_i` are the birth weight and gestational age for simulant :math:`i` (:math:`\text{baseline BW}_i` refers to simulant birthweight in the baseline scenario prior to any changes due to interventions such as antenatal interventions. :math:`\text{current BW}_i` refers to simulant birthweight following any changes due to interventions),
+:math:`\text{CSMR}_{\text{BW}_i,\text{GA}_i}^{k}` is the cause-specific mortality rate for modeled subcause :math:`k` for a population with the same gestational age and birth weight as this simulant, 
 and :math:`\text{CSMR}_{i}^{k}` is the cause-specific mortality rate for subcause :math:`k` for simulant :math:`i` (both detailed in the `Modeled Subcauses`_
 linked from this page).
 


NathanielBlairStahn:
I'll need to read through this more carefully to digest all of it, but I have a clarifying question:

Do you think that the LBWSG RRs reported by GBD can be interpreted as causal RRs on all-cause mortality, or do you think that GBD's RRs are _not_ causal RRs on all-cause mortality, and should only be interpreted as "associational"?

To put it another way, regardless of which cause-specific mortality rates are causally affected by LBWSG, do the RRs measure the causal association with _all-cause_ mortality, or not (i.e., if we ignore which cause-specific mortality rates are increasing vs. not when all-cause mortality goes up)?

alibow:
> I'll need to read through this more carefully to digest all of it, but I have a clarifying question:
> 
> Do you think that the LBWSG RRs reported by GBD can be interpreted as causal RRs on all-cause mortality, or do you think that GBD's RRs are _not_ causal RRs on all-cause mortality, and should only be interpreted as "associational"?
> 
> To put it another way, regardless of which cause-specific mortality rates are causally affected by LBWSG, do the RRs measure the causal association with _all-cause_ mortality, or not (i.e., if we ignore which cause-specific mortality rates are increasing vs. not when all-cause mortality goes up)?

Well they are from observational data so they can't be truly causal measures, but we are using them as causal measures anyway because they are as close as we can get with the data available to us. But at the end of the day they still are associational and so we should be mindful of the potential biases present when we make the assumption that they are causal. As far as I recall (we should double check the exact wording), GBD does not adjust for any potential confounding in the generation of its RRs. But they do specifically try to account for reverse causality bias through their differentiation of affected/unaffected causes by only applying the RRs to a subset of causes they don't think are subject to reverse causality (even though those reverse causality causes are still skewing the actual RR estimates that they apply, but they think this will still get them closer to the truth). Additionally complicating, as far as I recall the RRs are derived from data in high income settings, so they are likely attenuated relative to what they may be in other settings due to differences in health care access (like CPAP stuff). The fact that the data are from high income settings though may mean that HIV and malaria (the unaffected causes) are likely low burden in these populations, so hopefully the RR estimates weren't skewed too much due to reverse causality bias because of this.

zmbc:
[This](https://github.com/ihmeuw/vivarium_research/pull/1592#discussion_r1951631612) (potentially more controversial than I realized) feels relevant to this conversation.

Simply put, it feels unreasonable to expect us to backtrack the GBD's work to re-evaluate the causality of their RRs. They are given to us with an explicitly causal interpretation. And we can only use them for counterfactual scenarios by assuming they are causal.

> by only applying the RRs to a subset of causes they don't think are subject to reverse causality

This is the internal inconsistency within GBD, right? Because the RRs are explicitly calculated as being on ACMR.

zmbc:
It sounds like the question is which causal interpretation of GBD's RRs is more accurate and/or true to GBD's own use of them:

- These RRs represent the causal effect of LBWSG on all-cause mortality rates
- These RRs represent the causal effect of LBWSG on cause-specific mortality rates of each affected cause (after multiplying the RRs by $\frac{\text{ACMR}}{\sum_{k \in K} \text{CSMR}_k}$ where $K$ is the set of affected causes??)

alibow:
> It sounds like the question is which causal interpretation of GBD's RRs is more accurate and/or true to GBD's own use of them:
> 
> 
> 
> - These RRs represent the causal effect of LBWSG on all-cause mortality rates
> 
> - These RRs (multiplied by $\frac{\text{ACMR}}{\sum_{k \in K} \text{CSMR}_k}$ where $K$ is the set of affected causes??) represent the causal effect of LBWSG on cause-specific mortality rates of each affected cause 

Yes, agreed! And I am advocating for the latter.

zmbc:
Was the part in parentheses with the "??" correct, or is it a different adjustment?

NathanielBlairStahn:
> Was the part in parentheses with the "??" correct, or is it a different adjustment?

I started writing it out with the "??" adjustment, and I'm skeptical that that's correct, but I'm not sure.

alibow:
@zmbc @NathanielBlairStahn oh, sorry I misread the unformatted version of the equation in your ?? parenthetical when responding earlier. To be most consistent with GBD and our prior approaches, we would not adjust the RRs up like you suggested and rather just apply the RRs causally as they are to the affected causes directly.

zmbc:
@alibow I don't understand how this could possibly be consistent with the data without an adjustment, though. We/GBD have data that allow us to approximate effect on ACMR. It makes sense to impose the assumption that this effect must be entirely mediated by known-affected causes, but if you don't adjust your RRs, you are simply deleting part of the effect you measured, no?

zmbc:
I suppose you did say "to be consistent with GBD and our prior approaches," not "to be consistent with the observed data," but unless affected causes make up the vast majority of the mortality in these age groups, this bias seems like it would be large.

alibow:
@zmbc So yes, we are simply deleting part of the effect that was measured as you said, but the idea is that the part they are deleting is (hopefully) the part that is due to reverse causality bias. I say hopefully because (in my opinion) what they actually did is a very weird and limited approach to try to get around the fact that they weren't actually able to do the logical thing which is to tease apart the reverse causal and directly causal effects in the RRs they estimated. My high-level interpretation of what they did is say "we know that some portion of the magnitude of the association we see with LWBSG and ACMR is driven by reverse causality, so the RRs we estimated will **overestimate** the magnitude of the _causal_ impact LBWSG on neonatal mortality" so then... "to counteract this overestimation, we will apply these RRs (the closest thing we have to any measure of the effect of LBWSG on mortality) to a subset of causes only in our estimation of the causal impact of LBWSG on neonatal mortality." It's weird in that this is an imperfect adjustment for a known bias and I don't think there's really any way to say exactly what the direction or magnitude of the remaining bias is, but basically: they knew they were biased up by some unknown degree (by reverse causality), so they did something to bring that bias back down (by applying to a subset of causes only).

And luckily, as you mentioned, the affected causes do make up a majority of ACMR in the neonatal ages, which should minimize the magnitude of bias
image

alibow:
> @alibow That is clear, and I hope explaining this has generated a solid start toward writing it down in such a way that future readers will understand it too.
> 
> In particular, what strikes me about this is that **these RRs are not like other GBD RRs**. Other GBD RRs have _already_ been through all of the processing GBD decided to do to try to isolate the causal effect. Here, the RRs are saved, for whatever reason, in a sort of intermediate state, before one last bias correction the GBD modelers decided to make. Our job is to re-apply that last bias correction to the saved values before interpreting them in the usual way we interpret GBD RRs (as causal effects).
> 
> Does that match your understanding?

Yes I agree with your summary, but I also think it could be put another way too: The GBD RRs derived from ACMR are considered "close enough" to be causal (despite the concerns about reverse causality) and so they're saved in the GBD system. Then they are applied to the GBD causes thought to be affected by LBWSG (in the same way that most causes are only applied to the causes they affect). Maybe that's the way GBD teams would interpret it?

I feel like we almost know too much by knowing how the RRs were derived from ACMRs. It makes it more confusing, but does allow us to take advantage of using the (raw) RRs as an associational measure with ACMR in our sim and then apply them causally only to the subset of affected causes (what I hope my proposal in this PR will achieve!).

zmbc:
> The GBD RRs derived from ACMR are considered "close enough" to be causal (despite the concerns about reverse causality) and so they're saved in the GBD system. Then they are applied to the GBD causes thought to be affected by LBWSG (in the same way that most causes are only applied to the causes they affect). Maybe that's the way GBD teams would interpret it?

Are the GBD RRs for LBWSG saved as applying to each of the affected causes? I thought they were saved explicitly as RRs _on ACMR_. If they're not, then what I wrote above was wrong; they are interpreted exactly the same way as other GBD RRs (the causal effect of a risk factor on a single cause, only it happens to be the same for all affected causes).

> allow us to take advantage of using the (raw) RRs as an associational measure with ACMR in our sim

Ah, I had not realized this. Do we think that GBD hasn't done any adjustments to make them "more causal" by the time we get them?

alibow:
> Are the GBD RRs for LBWSG saved as applying to each of the affected causes? I thought they were saved explicitly as RRs _on ACMR_. If they're not, then what I wrote above was wrong; they are interpreted exactly the same way as other GBD RRs (the causal effect of a risk factor on a single cause, only it happens to be the same for all affected causes).

Yeah, in the GBD system they apply directly to the affected causes only -- they are not saved as having an impact on ACMR. It's just that in the methods appendix (relevant info starting on pdf page 336 of appendix 1 [here](https://www.thelancet.com/journals/lancet/article/PIIS0140-6736(24)00933-4/fulltext)) they describe how they derived the RRs using ACMRs because there wasn't enough data to do it at the cause-specific level. So while the risk is structurally consistent with other GBD risks, again we're able to utilize that knowledge to treat it differently when convenient as has been done from this modeling strategy.

> Ah, I had not realized this. Do we think that GBD hasn't done any adjustments to make them "more causal" by the time we get them?

I just re-read the methods appendix and it looks like it's a pretty crude calculation based on how they describe it there. They calculate smoothed mortality risk stratified by LBWSG category and then divide by the TMREL. They pool over data years and sexes (except for one location that was sex-specific) and then meta-analyze across locations. I'm not quite sure how they reconcile sex specificity later on, but they don't mention any specific adjustments for covariates or anything that would be a red flag.

zmbc:
Ah, okay! In that case, the unusual thing we're doing here is taking GBD RRs of LBWSG -> CSMR (which happen to be the same across all affected causes) and using them to inform a _correlation_ between LBWSG and ACMR in our model. We're comfortable doing this because GBD didn't adjust for any confounding, therefore we can take their values as being (meta-analyzed) observed correlations.

The tricky part is that we don't typically model both correlation and causal effect between the same two factors, which is why we need to get creative to pull this off.

If I've understood the goal correctly, I can now read the content of this PR 😄

alibow:
Yes, exactly!! This has been helpful back and forth :). And it sounds like you're caught up, but I will also add that as I understand it, the prior modeling strategy would use the RRs causally on ACMR across the board, which I am worried would overestimate the effect. So this PR is proposing to tweak the strategy so that we can maintain the ACMR association but only apply the RRs causally to the affected causes. 

It did sound like @aflaxman had another idea for how we could address the issue, by using the normalizing constant Z (discussed on these pages https://vivarium-research.readthedocs.io/en/latest/models/causes/neonatal/neonatal_sepsis.html#cause-neonatal-sepsis-mncnh) to effectively create no association between LBWSG and the cause fraction of unaffected causes. But if I am understanding it correctly (which I am not convinced I am) then I am worried that it would still have the same issue of applying the LBWSG RRs causally to ACMR. CC @NathanielBlairStahn here

NathanielBlairStahn:
> but does allow us to take advantage of using the (raw) RRs as an associational measure with ACMR in our sim

> the unusual thing we're doing here is taking GBD RRs of LBWSG -> CSMR (which happen to be the same across all affected causes) and using them to inform a correlation between LBWSG and ACMR in our model.

@zmbc @alibow I don't understand what the two of you mean by these statements. Can you explain further?

> It did sound like @aflaxman had another idea for how we could address the issue, by using the normalizing constant Z (discussed on these pages https://vivarium-research.readthedocs.io/en/latest/models/causes/neonatal/neonatal_sepsis.html#cause-neonatal-sepsis-mncnh) to effectively create no association between LBWSG and the cause fraction of unaffected causes. But if I am understanding it correctly (which I am not convinced I am) then I am worried that it would still have the same issue of applying the LBWSG RRs causally to ACMR.

This is why I asked my original question of whether we believe these are actually causal RRs on ACMR. If they are, then there's a clear way to use my existing strategy to specify which causes we think they affect. If not, then it's not clear what to do.

zmbc:
@NathanielBlairStahn here's my understanding:

- These RRs are _not_ on ACMR, they are on CSMR of the affected causes, and we should apply them as causal effects like any other RRs we get from GBD.
- It just so happens that the way these RRs were calculated (for every one of the affected causes) was from a meta-analysis of the correlation between LBWSG and ACMR. So we can use them to inform an excess _correlation_ between LBWSG and ACMR **in addition to** our causal effect.

NathanielBlairStahn:
> @NathanielBlairStahn here's my understanding:
> 
> * These RRs are _not_ on ACMR, they are on CSMR of the affected causes, and we should apply them as causal effects on CSMR like any other mortality RRs we get from GBD.
> * It just so happens that the way these RRs were calculated (for every one of the affected causes) was from a meta-analysis of the correlation between LBWSG and ACMR. So we can use them to inform an excess _correlation_ between LBWSG and ACMR **in addition to** our causal effect.

But how can they measure the causal effect on CSMRs if they were calculated from all-cause mortality risk? Won't the RRs just be confounded and not actually measure the causal effect on CSMRs? When you say "calculated from a meta-analysis of the correlation between LBWSG and ACMR," by "correlation," do you mean "observed relationship" or something (as opposed to, e.g., the linear relationship measured by a correlation coefficient)?

GBD _says_ they're supposed to be the causal RRs on the affected CSMRs, but what they actually did was this:

> We analysed the relative risk of all-cause mortality across all available sources and
selected outcomes based on criteria of biological plausibility. Some causes, most notably congenital
birth defects, haemoglobinopathies, malaria, and HIV/AIDS, were excluded based on the criteria that
reverse causality could not be excluded.

Based on how they were calculated, what are the RRs actually measuring, and how should we use them?

zmbc:
> by "correlation," do you mean "observed relationship" or something (as opposed to, e.g., the linear relationship measured by a correlation coefficient)?

Yes, I mean "observed association" -- I'm using "correlation" here loosely (and possibly confusingly) to emphasize that causal-inference-wise this is the same idea as [risk correlation](https://vivarium-research.readthedocs.io/en/latest/model_design/vivarium_model_components/risk_factors/risk_correlation/index.html) -- which I think we call "correlation" because we've only done it on continuous risks.

zmbc:
> But how can they measure the causal effect on CSMRs if they were calculated from all-cause mortality risk? Won't the RRs just be confounded and not actually measure the causal effect on CSMRs?

The idea is that GBD believes there are no confounders, but there is a reverse-causality problem. So the only adjustment they do to the association in order to arrive at what they claim are causal effects is to zero out effects on causes where reverse causality is plausible. We can definitely wonder whether that is the best way to estimate these CSMR RRs, but I think at that point we are re-hashing the decisions that GBD already made.

NathanielBlairStahn:
> Yes, I mean "observed association" -- I'm using "correlation" here loosely (and possibly confusingly) to emphasize that causal-inference-wise this is the same idea as [risk correlation](https://vivarium-research.readthedocs.io/en/latest/model_design/vivarium_model_components/risk_factors/risk_correlation/index.html) -- which I think we call "correlation" because we've only done it on continuous risks.

> Just like we might seek out a meta-analysis about the correlation of two modeled risk factors when we think that correlation is strong and impactful on our results, we want to model the correlation of LBWSG and ACMR.

I was confused about what @zmbc meant by these statements --- namely, _how_ exactly we use the LBWSG RRs to model the correlation/observed association. But I think maybe I've figured it out, so let me see if my interpretation aligns with with yours and Ali's.

First, I think you're saying that the first equation defining ACMR in the [docs](https://vivarium-research.readthedocs.io/en/latest/models/causes/neonatal/index.html#modeling-strategy),

(eq. 1) $\text{ACMR}\_{\text{BW},\text{GA}} = \text{ACMR} \times (1 - \text{PAF}\_{\text{LBWSG}}) \times \text{RR}\_{\text{BW},\text{GA}}$,

is capturing the "correlation" or "observed association" between ACMR and LBWSG category because of how the RRs were calculated, but we can't interpret the RRs in this equation as causal. (I was confused by Zeb's reference to our risk correlation page because this is completely different from how we usually capture correlation, but in both cases we're initializing the values of two correlated variables to align with observed data.) So Ali is saying that this equation is fine in baseline because it's about what's actually observed, but it needs to be modified in a scenario where we intervene on birthweight or gestational age, because the reverse causality problem would mean that we'd be reducing ACMR by too much when we improved LBWSG exposure if we used these RRs causally on ACMR. So instead, we want to apply the RRs causally to the CSMRs of the _affected_ causes, but keep the _observational_ relationship of LBWSG with the CSMRs of the _unaffected_ causes by applying the RRs at the birthweight and gestational age we would have seen in baseline instead of applying the RRs at the improved exposure in the intervention scenario (and instead of applying no RR at all, which is what we've done before).

Assuming that's correct, here's a different proposal for how to write Ali's equation (which is also a bit messy, but matches more closely with what's already there). Instead of modifying the second equation defining $\text{ACMR}\_i$, we can modify the first equation, i.e., (eq. 1) above. Namely, in scenarios with an intervention on BW or GA, we can change (eq. 1) to

(eq. 1') $\text{ACMR}\_{\text{BW},\text{GA}}^{\text{BW}^0,\text{GA}^0} = \sum\_{c\in\text{unaffected}} \text{CSMR}\_c \times (1 - \text{PAF}\_{\text{LBWSG}}) \times \text{RR}\_{\text{BW}^0,\text{GA}^0} + \sum\_{c\in\text{affected}} \text{CSMR}\_c\times (1 - \text{PAF}\_{\text{LBWSG}}) \times \text{RR}\_{\text{BW},\text{GA}}$,

where $(\text{BW}^0,\text{GA}^0)$ is the LBWSG exposure in the baseline scenario and $(\text{BW},\text{GA})$ is the LBWSG exposure in the intervention scenario (I'm not married to this notation if someone has a different idea). Then equation 2 would remain almost the same:

(eq. 2') $\text{ACMR}\_i = \text{ACMR}\_{\text{BW}\_i,\text{GA}\_i}^{\text{BW}\_i^0,\text{GA}\_i^0} - \sum\_k \text{CSMR}\_{\text{BW}\_i,\text{GA}\_i}^{\text{BW}\_i^0,\text{GA}\_i^0,k} + \sum\_k \text{CSMR}\_{i}^{k}$,

and the various CSMRs could be defined appropriately on the individual cause pages, so we could handle affected vs. unaffected causes on a case-by-case basis.

Does that all match your thinking, @alibow and @zmbc ?

For the record, note that (eq. 1') is similar to what we've done before, [here](https://vivarium-research.readthedocs.io/en/latest/models/risk_exposures/low_birthweight_short_gestation/gbd_2017/index.html#equation-mortality-hazard), except now we're proposing using the LBWSG RRs to also model the correlation of the CSMRs of the unaffected causes with LBWSG, and we don't haveEMRs because we're not modeling prevalence. The equation that would match our previous approach is:

(eq. 1'') $\text{ACMR}\_{\text{BW},\text{GA}} = \sum\_{c\in\text{unaffected}} \text{CSMR}\_c + \sum\_{c\in\text{affected}} \text{CSMR}\_c\times (1 - \text{PAF}\_{\text{LBWSG}}) \times \text{RR}\_{\text{BW},\text{GA}}$.

Also for the record, here's a link to an editable version of Ali's [GBD Compare plot](http://ihmeuw.org/6v4u) above.

alibow:
Yes, @NathanielBlairStahn I agree with your summary and like your proposal of how to restructure the equations :)

zmbc:
Agreed, I think that is a good way to express it! Let's make sure that we summarize the _why_ we've discussed here and that we are sufficiently clear in the docs about how weird the thing we are doing here is. I might even suggest that we call the RRs something else, without changing them numerically, when we use them for the unaffected causes in these equations, to emphasize that these are not causal.

alibow:
Closing this PR as it has been replaced with https://github.com/ihmeuw/vivarium_research/pull/1760

PR #1609, opened 2025-03-21
Add scenarios section to MNCNH concept model doc

pletale:
This adds a section of alternative scenarios for our MNCNH project, so far just with neonatal interventions going from baseline -> 100%. I'll post another PR shortly that will include in the updated Model Run Requests table a request for a run with the CPAP alternative scenarios!

(Note: I used https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_iv_iron/concept_model.html#simulation-scenarios as my template for how I formatted this section).


aflaxman (approved):

SylLutze (approved):
This looks good! We can probe BMGF if they want all interventions to be independently scaled up or down or if we can group some. But no changes needed now.

PR #1610, opened 2025-03-21
Document proposed improvement to CRN for CoD assignment

zmbc:
As discussed on Slack and in coworking today!


docs/source/model_design/vivarium_features/cause_of_death_selection/index.rst
```diff
@@ -0,0 +1,216 @@
...
+So that counterfactual can only delay their death by some finite amount of time.
+Perhaps they wouldn't have died that *minute* but they still would have died within
+the hour, or perhaps they would have lived 50 more years!
+
+Cause of death in GBD
+---------------------
+
+The Global Burden of Disease study (GBD), informed primarily by death certificates
+and by methods that seek to approximate them such as verbal autopsy, **assigns every
+death a single cause.**
```
aflaxman, line 67:
```suggestion
death a single cause.** (Or, very occasionally, when the interplay between causes is too important to ignore, GBD constructs a new cause to get at this, such as cause A.1.1.2, "HIV and Multidrug-resistant Tuberculosis without extensive drug resistance".)
```

docs/source/model_design/vivarium_features/cause_of_death_selection/index.rst
```diff
@@ -0,0 +1,216 @@
...
+
+Cause of death is a different sort of causal claim than we typically make in
+our microsimulations.
+Usually, we do causal inference by comparing two scenarios, where we change one thing;
+any differences in outcomes are therefore caused by the one thing we changed.
+With cause of death, we don't actually run a scenario in which we delete that cause;
+the cause of death is assigned within the scenario.
+
+To be internally consistent, though, we **want the deaths we assign cause A on a given timestep
+to be the same deaths that would not occur if we did run a no-cause-A scenario.**
```
aflaxman, line 108: confusing

docs/source/model_design/vivarium_features/cause_of_death_selection/index.rst
```diff
@@ -0,0 +1,216 @@
...
+cause's probability.
+As long as this happens rarely, we will still get good variance reduction in aggregate.
+If this is common, it may indicate that our timestep is too long.
+
+.. note::
+
+  I believe we could get slightly better variance reduction by starting each cause's probability
+  at the midpoint of its reservation, and growing symmetrically outward toward the edges of the
+  reservation, then overflowing the reservation in a random direction (e.g. right).
+  But the benefit is probably small and not worth the implementation effort.
```
aflaxman, line 216: I bet we will get most of the benefit from moving "all other causes" to be furthest left

aflaxman (approved):

docs/source/model_design/vivarium_features/cause_of_death_selection/index.rst
```diff
@@ -0,0 +1,216 @@
...
+.. image:: cod_issue_diagram.drawio.svg
+
+Instead of some simulants who would have died of cause A not dying at all,
+some simulants who would have died of cause A die of cause B instead,
+some who would have died of cause B die of cause C instead,
+and some who would have died of cause C don't die at all.
+Cause B and C don't change size -- so in expectation, the number of deaths is the same --
+but depending on where our finite population of simulants actually fall on this
+graph, we may get nonzero (positive or negative) numbers of deaths averted
+for both cause B and C.
```
SylLutze, line 164: Can we add a bit more about when this is more of an issue and when we're most likely to see it? E.g., having multiple chronic conditions and having lots of simulants who have high comorbidity rates.

SylLutze (approved):

docs/source/model_design/vivarium_features/cause_of_death_selection/index.rst
```diff
@@ -0,0 +1,216 @@
...
+
+Instead of the above, we want to transfer probability *directly* from the intervened-on
+cause to "does not die."
+In order to do this, we reserve sections of the probability space for each
+potential cause of death in the model.
+**These reservations are the same between scenarios.**
+The probability section for cause B never starts until after the *reserved* section
+for cause A, regardless of the scenario-specific probability of cause A.
+
+.. image:: cod_solution_diagram.drawio.svg
```
zmbc, line 177: Add note about numerical precision

docs/source/model_design/vivarium_features/cause_of_death_selection/index.rst
zmbc: Only because/when "actual" comorbidity is rare, and assuming we don't intervene on "all other causes"

docs/source/model_design/vivarium_features/cause_of_death_selection/index.rst
zmbc: I'm using "cause" in this sentence to refer to GBD causes already, not causes on death certificates. Maybe I should just clarify that.

docs/source/model_design/vivarium_features/cause_of_death_selection/index.rst
```diff
@@ -0,0 +1,216 @@
...
+In this case, we revert to the status quo: an overflowing probability will shift over the next
+cause's probability.
+As long as this happens rarely, we will still get good variance reduction in aggregate.
+If this is common, it may indicate that our timestep is too long.
+
+.. note::
+
+  I believe we could get slightly better variance reduction by starting each cause's probability
+  at the midpoint of its reservation, and growing symmetrically outward toward the edges of the
+  reservation, then overflowing the reservation in a random direction (e.g. right).
```
NathanielBlairStahn, line 215: Do you mean "random" or "arbitrary"? I.e., would we randomly pick the direction each time, or do we arbitrarily pick either left or right, and always use that?

Also, why exactly would this lead to more variance reduction?

zmbc: Best would be random, at the cause level, I think.

I don't want to go too far down this rabbit hole, but I got to this suggestion by thinking about how we could _minimize_ the probability that one cause's probability allocation "bumps into" (and therefore shifts) another. Starting everything at the left side of its reservation, as I've actually proposed here, means that anytime a cause overflows and the next is non-zero, shifting occurs. The midpoint expansion approach means that a shift would require overflow > (next_cause_reservation - next_cause_probability) / 2. But it's possible I've totally missed something

tylerdy (approved):

docs/source/model_design/vivarium_features/cause_of_death_selection/index.rst
```diff
@@ -59,32 +59,50 @@ So that counterfactual can only delay their death by some finite amount of time.
 Perhaps they wouldn't have died that *minute* but they still would have died within
 the hour, or perhaps they would have lived 50 more years!
 
+It isn't necessarily the case that every death even has *any* disease that "caused" it,
+unless the when question is answered with "wouldn't have died *that instant*."
+Consider someone who has two separate types of terminal cancer.
+For any non-instantaneous period of time, they could be sick enough with both cancers
+that even in a counterfactual where one type of cancer was cured,
+they would die of the other, and vice versa.
+
```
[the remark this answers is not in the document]
zmbc: As described in the previous paragraph, for a disease to "cause" a death, it must be the case that the death wouldn't have occurred (in some timeframe) if the person didn't have the disease. In the counterfactual where they didn't have cancer A, they still would have died; in the counterfactual where they didn't have cancer B, they also still would have died. Therefore neither cancer A nor cancer B caused their death.

I may be cheating a little bit here by saying "die _of_ the other."

Does that help? Any ideas how to clarify?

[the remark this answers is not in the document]
tylerdy: I'm not sure I'm understanding the claim of your last sentence beginning with "therefore". Are you saying that in those respective counterfactuals cancer A and cancer B did not cause their death (but that in each counterfactual, one of the cancers did cause the death)? Or are you making a claim that if you can think of a set of counterfactuals such that there exists a counterfactual where each possible cause of death was not the cause of death in that counterfactual, then you can say that in that case no disease caused that death? I'm guessing the second one based on "It isn't necessarily the case that every death even has *any* disease that "caused" it" in the original edit text, but I'm not following the logic of how thinking of this set of counterfactuals allows you to say that, or why being able to say that is relevant to the discussion.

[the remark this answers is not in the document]
zmbc: I don't think I'm precisely saying either of those two things, but I definitely see why this is confusing.

The key point is that saying that a disease caused a death is _not_ a claim you can make looking at only one state of affairs. You have to compare _two_ "alternate universes" to make that claim.

Let me try to write out this example more explicitly. Here are some possible states of affairs:
1. The person has cancer A and cancer B, and they die (that month).
2. The person has cancer A but not cancer B, and they die (that month).
3. The person has cancer B but not cancer A, and they die (that month).
4. The person has neither cancer A nor cancer B, and they do not die (that month).

If (1) is what actually happened, we can say that:
- Cancer A was not the cause of their death, because in counterfactual 3 they also died.
- Cancer B was not the cause of their death, because in counterfactual 2 they also died.

However if we imagine that scenario 3 is what _actually_ happened, we would say that cancer B _was_ the cause of that person's death, because in counterfactual 4 they _didn't_ die. Or seen another way: what actually happened is important. We only consider what actually happened (the factual) and counter-factuals created by making _one_ change to the factual. Simply having the list of 4 options above, without knowing which actually occurred, is not sufficient to say what disease caused the person's death.

This is how I arrived at such an odd statement as "neither A nor B _caused_ their death because even without B they would have died _of_ A." The "of A" there really meant that _if_ the hypothetical ("without B") were the actual, we could use an _additional_ hypothetical to say that their death in the "without B" scenario was caused by A. However that doesn't mean their death in the actual world was caused by A.

This is all very weird and I'd be interested to get more perspectives/reactions here!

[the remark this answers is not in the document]
SylLutze: I think what you have here is correct, but maybe a bit confusing. Specifically you say that there doesn't need to be a cause and then say unless we are talking about an instant. I think the point makes more sense in my brain to think of it as a limit. With a long period of time, the counterfactual of CoD could be a lot of things and is fuzzy, but as time goes to zero, there gets to be a single cause. 

It's easy to imagine a very old person who is quite likely to die this year and even if I cured their heart disease (or insert other cause here), they would probably still die. I make the time step a month it's harder but still possible to imagine a scenario where this makes sense (2 cancers, sure). A week we have to be getting to overlapping acute infections, a day even harder, etc. Until we arrive at instantaneous where we can say there is one cause only because it actually happened.

SylLutze (approved):
I think all the changes look good! Added a comment to try and be helpful but please ignore if it is not.

docs/source/model_design/vivarium_features/cause_of_death_selection/index.rst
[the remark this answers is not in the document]
tylerdy: That explanation is helpful for me trying to understand how we think about causes, thanks - I think it pretty much sense. Esp the part about a counterfactual where you change one thing was very helpful to understanding the claims we're talking about.

As far as how to improve the text based on my confusion, I think my biggest takeaway is that if theres one thing this whole conversation has helped me understand its that its difficult to assign cause without being very precise about your definition. So I guess this paragraph was pretty successful even if it I had trouble understanding it!

As for the stuff about needing the instantaneous time period to be able to to really assign cause, I'm still trying to wrap my head around it a bit but I think its starting to make sense, your comment was helpful Syl. I think its needed bc if theres a significant chance of dying from cancer B, or from age, even if we remove cancer A from the scenario, they _may_ still die, so we can't assign the cause of death to cancer A with total certainty. Whereas with instantaneous time period, the chances of dying from B or age approach 0, so we can have total certainty. Is this sort of it?

I guess I'm maybe a bit unclear about how this would work in practice - is it the case that if we remove A in the simulation and the person doesnt die over some larger timestep, then we might say its not good enough evidence to assign CoD to A bc we know if the randomness in the scenario had gone another way theres a decent chance they wouldn't have died? And we're kind of using "does the reservation overflow" as a heuristic of estimating whats a reasonable timestep in your plan?

Anyways sorry to turn this into me just not asking CoD works in simulations, not sure I have much specific feedback on how to make the explanation easier to understand other than to spell what we've been talking about out a little more. But it seems like what you have would be pretty good if I wasn't coming from basically zero experience with this stuff!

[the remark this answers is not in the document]
zmbc: I think @SylLutze's "limit" framing is super helpful and I can use it to improve this text! 🙏

@tylerdy -- for the purposes of this paragraph, let's ignore simulations 😄 This is intended to be a statement about the actual world/the concept of CoD in general, not about how we choose to approximate it in simulation models.

As to this:

> I think its needed bc if theres a significant chance of dying from cancer B, or from age, even if we remove cancer A from the scenario, they may still die, so we can't assign the cause of death to cancer A with total certainty. Whereas with instantaneous time period, the chances of dying from B or age approach 0, so we can have total certainty. Is this sort of it?

You're thinking about things from a probabilistic perspective, but I'm thinking about it from a [potential outcomes](https://imai.fas.harvard.edu/teaching/files/potential_outcomes.pdf) perspective. The key difference is that we assume that whether or not the person would have died is fixed -- they either would have or they wouldn't have. (Of course, in the real world we can't observe what the outcome would have been in a counterfactual scenario, but that is treated as an "us" problem -- we didn't observe something -- rather than a "real" problem -- it didn't exist.)

So the idea is that for a single person and a fixed time period, their death would either be averted by removing cancer B, or it wouldn't. Syl's point is that when the time period is large, there are many people (with cancer B) for whom removing cancer B would not avert their death in that time period. As the time period gets smaller, more and more people's deaths would be averted in that time period by removing cancer B. For very small time periods, only a few people in very unusual situations would _not_ have their death averted in that time period. And as the time period goes to instantaneous, all the deaths would be averted.

This might seem like an academic distinction, but the point is that it's not about "certainty" with respect to CoD. It's about whether there actually _is_ a single disease that is a CoD. Even if we were completely omniscient, we couldn't assign a CoD when there isn't one.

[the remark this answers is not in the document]
zmbc: And thanks so much for asking these excellent questions!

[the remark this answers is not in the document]
zmbc: Notes from a real-time discussion (edits to the PR to come):

- I do mean the fraction, not the number!
- There are three layers here: the data available, how GBD uses them, how we use GBD
  - Data available: not super explicit, but based on assigning a single cause to each death, it seems clear that death certificates can best be interpreted as instantaneous CoD
  - How GBD uses them: for mathematical simplicity, assumes instantaneous deaths averted from cause X do not increase the number of deaths from any other cause during that year; at the individual level, this implies invincibility potions
  - How we use them: pretty similar to GBD, but we decrease the duration from a year to a timestep

zmbc:
I think something weird is happening with GitHub right now, when I (or ReadTheDocs) fetch `pull/1610/head` it comes back without the last few commits... Will try again tomorrow.

PR #1611, opened 2025-03-25
Update Model Run Requests table 

pletale:
This adds a Scenarios column to the model run requests table. All runs have been baseline only except for the last request, cpap_6, which will include alternative scenarios 2-4. 

The PR also accordingly updates the VnV tracker table - I went thru slack to get an idea of what all happened while I was out, but if I'm missing or misconstruing any details here, pls lmk!


docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -574,31 +577,84 @@ V&V Checks:
...
+    - Baseline only
     - Locations include Pakistan, Nigeria, and Ethiopia. 10 seeds * 10,000 simulants = 100,000 total population.
   * - 3.3
     - Wave I Neonatal disorders V&V with early NN observer bugfix
     - Complete
     - N/A
     - 10
     - 100,000
+    - Baseline only
+    - Locations include Pakistan, Nigeria, and Ethiopia. 10 seeds * 10,000 simulants = 100,000 total population.
```
aflaxman, line 604: I like to let the engineers decide how to balance seeds and simulants-per-seed --- if 5 seeds * 20,000 simulants is faster, no need to lock in 10 * 10,000.  
```suggestion
    - Locations include Pakistan, Nigeria, and Ethiopia.  100,000 total population (e.g. 10 seeds * 10,000 simulants).
```

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -574,31 +577,84 @@ V&V Checks:
...
+  * - 4.6
+    - Wave I CPAP with bugfix for negative other causes mortality rates
+    - Incomplete
+    - N/A
+    - 10
+    - 100,000
+    - Baseline and alternative scenarios 2, 3, and 4
     - Locations include Pakistan, Nigeria, and Ethiopia. 10 seeds * 10,000 simulants = 100,000 total population.
```
aflaxman, line 653: Oh, but maybe you are documenting what *was* run, not requesting what should be run...

aflaxman (approved):

albrja (approved):

alibow:
For V&V tracking -- I am not exactly sure what model version this corresponded to (I think model 3.3 trying to piece it together? see [this slack thread](https://ihme.slack.com/archives/C02944X688L/p1741890129942359?thread_ts=1741881522.643889&cid=C02944X688L) for confirmation?), but [this notebook](https://github.com/ihmeuw/vivarium_research_mncnh_portfolio/blob/main/verification_and_validation/20250313_lbwsg_effects_interactive_simulation.ipynb) has some additional V&V pieces we should write down: (1) individual RRs vary with LWBSG exposure, (2) individual RRs affect mortality rates appropriately, (3) no non-preterm babies are dying of preterm.

We also identified something odd that we will want to track: LBWSG exposures in the state table do not match LBWSG exposure pipeline values. Details are discussed in the same notebook I already linked.

@aflaxman can you confirm model versioning for Alix based on the slack thread I linked?

alibow:
@pletale, so the issue I flagged earlier of the state table and pipeline values not matching was determined not to be an issue -- it is just that the pipeline values refresh after being recorded in the state table (and then they are not used again). So we can maybe just make a quick note of that in the V&V table that we noticed it but it's not a problem. However, we did uncover that the pregnancy duration for partial term pregnancies is incorrect, so that's a new issue to track that @albrja is aware of and working on implementing.

PR #1613, opened 2025-04-03
Add maternal intervention models to Model Run Requests table

pletale:
I didn't add c-sections because I believe that's wave 2 - not sure if I should add wave 2 interventions too. Also I'm not sure this is the order we're planning on implementing these in.

### DIFF
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
@@ -662,7 +662,7 @@ V&V Checks:
     - Locations include Pakistan, Nigeria, and Ethiopia. 10 seeds * 10,000 simulants = 100,000 total population.
   * - 6.0
     - Wave I azithromycin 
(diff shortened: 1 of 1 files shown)


albrja (approved):

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -652,6 +652,46 @@ V&V Checks:
...
+  * - 6.0
+    - Wave I azithromycin 
+    - Incomplete
+    - N/A
+    - 10
+    - 100,000
+    - Baseline
+    - Locations include Pakistan, Nigeria, and Ethiopia. 10 seeds * 10,000 simulants = 100,000 total population.
+  * - 7.0
+    - Wave I AI ultrasound
```
SylLutze, line 680: I think this should be misoprostol? We already have AI ultrasound but need misoprostol still.

SylLutze (approved):

PR #1615, opened 2025-04-03
MNCNH pregnancy component reorg

alibow:
Everything deleted from the concept model page has been moved elsewhere (this may be more obvious if you look at individual commit history rather than the overall PR if you are interested in tracking at that level).

Note that I have NOT split the pregnancy module into two parts like I proposed in the all team meeting today. That's because I want the docs to reflect what's been implemented so far at this point and then will layer on additional PRs so we have a record of changes necessary for wave II implementation

I would also like someone to check my interpretation of the AI ultrasound module page. Specifically, I am a little unsure about two things:
1. Are partial term pregnancies totally excluded from the AI ultrasound module or do we treat them the same? As written I have them treated the same.
2. Do simulants who do not go to ANC have the same error in gestational age dating as those who go to ANC and have no ultrasound? That's how I've handled it as written.


aflaxman (approved):
> Are partial term pregnancies totally excluded from the AI ultrasound module or do we treat them the same? As written I have them treated the same.

I think it is best to treat them as the same, but I'm not sure what is implemented in the sim currently, it is probably fine to have them totally excluded.


> Do simulants who do not go to ANC have the same error in gestational age dating as those who go to ANC and have no ultrasound? That's how I've handled it as written.

I think that they do, based on last menstrual period (LMP).  @NathanielBlairStahn , is this right?

SylLutze (approved):

albrja:
@alibow Your two assumptions are correct.

albrja (approved):

alibow:
> @alibow Your two assumptions are correct.

@albrja Great! Thank you!

PR #1620, opened 2025-04-08
Add new txt file

albrja:
Adds new text file and updates README


README.rst
```diff
@@ -9,6 +9,7 @@ Vivarium Research
 Here you'll find a variety of useful information about working on ``vivarium``
 research projects.
 
+# This is a url link to the docs
```
SylLutze, line 12: I think this should be phrased differently.

PR #1631, opened 2025-04-16
Add hemoglobin screening accuracy instructions

alibow:


docs/source/models/concept_models/vivarium_mncnh_portfolio/hemoglobin_component/module_document.rst
```diff
@@ -307,6 +322,8 @@ The following pseudocode outlines how this can be done.
 
 - We assume no additional effect of oral iron supplementation when taken following IV iron administration
 
+- We assume a hemoglobin screening sensitivity of 85% and specificity of 80%, as requested by the Gates Foundation
```
SylLutze, line 325: I think we should add a limitation here that the sensitivity/specificity is not tied to how close your value is to 100 (at least in my reading of this). I would expect that someone with a hem value of 95 is more likely to be missed in screening than someone with a value of 50. So you would get the most needed people anyways.

SylLutze (approved):

aflaxman (approved):

docs/source/models/concept_models/vivarium_mncnh_portfolio/hemoglobin_component/module_document.rst
alibow: Agreed, thanks! Added to the limitations section

PR #1640, opened 2025-05-01
MNCNH V&V notebook clean-up

pletale:
This PR updates both the model run requests table and the V&V tracker table to include all prior runs. I added a 'File path' column to the model run requests table and retrospectively filled in those of past runs.

### DIFF
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
@@ -604,7 +604,7 @@ Specific observer outputs and their stratifications may vary by model run as nee
 ++++++++++++++++++++++
 
 .. list-table:: Model run plan as of October 4, 2024
-  :widths: 3 3 3 3 3 3 3 15
+  :widths: 3 3 3 3 3 3 3 5 15
   :header-rows: 1
 
   * - Number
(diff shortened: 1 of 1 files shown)


alibow (approved):
So helpful! It looks like the "Priority" column in this table is never used. Should we remove it?

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -754,6 +807,7 @@ Specific observer outputs and their stratifications may vary by model run as nee
     - 10
     - 100,000
     - Baseline
+    - ``mnt/team/simulation_science/pub/models/vivarium_gates_mncnh/results/azithromycin``
```
alibow, line 810: For the models that have not been run yet, would we like to start using numbering convention for filepaths rather than names?

alibow:
Oh, one more thing. It looks like you've verified that the pregnancy duration for partial term pregnancies looks good in model 4.7, so we should remove it from the outstanding V&V issues table!

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
SylLutze: YES!!!

NathanielBlairStahn (approved):
Looks good, thanks Alix! I agree with Ali's comments (e.g., removing the Priority column since we're not using it).

PR #1642, opened 2025-05-06
WIP: crude birth rate fertility component details

tylerdy:
Making WIP commit from Abie into a PR

### DIFF
--- a/docs/source/models/other_models/fertility/index.rst
+++ b/docs/source/models/other_models/fertility/index.rst
@@ -1,0 +1,128 @@
+.. _fertility_crude_birth_rate:
+
+..
+  Section title decorators for this document:
+
+  ==============
+  Document Title
+  ==============
+
+  Section Level 1 (#.0)
+  ---------------------
+
+  Section Level 2 (#.#)
+  +++++++++++++++++++++
+
+  Section Level 3 (#.#.#)
+  ~~~~~~~~~~~~~~~~~~~~~~~
+
+  Section Level 4
+  ^^^^^^^^^^^^^^^
+
+  Section Level 5
+  '''''''''''''''
+
+  The depth of each section level is determined by the order in which each
+  decorator is encountered below. If you need an even deeper section level, just
+  choose a new decorator symbol from the list here:
+  https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#sections
+  And then add it to the list of decorators above.
+
+===================================================
+Fertility Component that uses Crude Birth Rate Data
+===================================================
+
+.. contents::
+   :local:
+   :depth: 1
+
+.. list-table:: Abbreviations
+  :widths: 15 15 15
+  :header-rows: 1
+
+  * - Abbreviation
+    - Definition
+    - Note
+  * - CBR
+    - Crude Birth Rate
+    - A thorough discussion of fertility and
+      birth rate models can be found on
+      `Wikipedia <https://en.wikipedia.org/wiki/Birth_rate>`_.
+
+Overview
+-------------
+
+This document is intended to represent the fertility model for a simulation that adds new simulants based on a CBR.
+
+GBD Modeling Strategy
+----------------------
+
+Pregnancy and births are not explicit outcomes in the GBD study. However, there are location- and year-specfic GBD covariates related to fertility and births, and we use the "Live births by sex" covariate together with the GBD population estimates to calculate the CBR.  :math:`CBR = \frac{Live\ births}{Population} \times 1000`.
+
+.. list-table:: Covariates
+  :widths: 15 15 15
+  :header-rows: 1
+
+  * - Covariate name
+    - Covariate ID
+    - Note
+  * - Live births by sex
+    - 1106
+    - 
+
+
+Vivarium Modeling Strategy
+----------------------------
+
+The key challenge for using CFR in a fertility component in Vivarium is determining how many simulants to add in each time step.  This quantity would be simple were it not for the variety of ``age_start`` and ``age_end`` parameters that might appear in the sim.
+
+Since the population represented in the sim might be only a fraction of the total population used as the denominator of the CBR, we need to find the fraction of the total population being simulated.
+
+.. math::
+
+   \text{simulated_fraction} = \frac{
+       \int_{a=\text{age_start}}^{\text{age_end}} \text{population}(a) \, da
+   }{
+       \int_{a=0}^{125} \text{total_population}(a) \, da
+   }
+
+With this quantity, the simulants to add per year are
+
+.. math::
+
+   \text{simulants_to_add_rate} = \frac{\text{sim_population}}{\text{simulated\_fraction}} \times \frac{\text{CBR}}{1000}
+
+This can be accomplished elegantly by adding a random number of simulants each time step with Poisson parameter of :math:`\text{simulants_to_add}\times \text{time_step}/365`.
+
+.. list-table:: Parameters
+  :header-rows: 1
+
+  * - Parameter
+    - Value
+    - Note
+  * - 
+    - 
+    - 
+
+.. list-table:: Data values
+  :header-rows: 1
+
+  * - Parameter
+    - Data type  
+    - Data ID
+    - Source
+    - Note
+  * - gbd_population
+    - population estimate
+    - N/A
+    - get_population: decomp_step='iterative'
+    - 
+  * - Live birth by sex
+    - Covariate
+    - 1106
+    - ?
+    - Assume lognormal distribution of uncertainty  ?
+
+  
+
+


docs/source/models/other_models/fertility/index.rst
```diff
@@ -0,0 +1,128 @@
...
+       \int_{a=\text{age_start}}^{\text{age_end}} \text{population}(a) \, da
+   }{
+       \int_{a=0}^{125} \text{total_population}(a) \, da
+   }
+
+With this quantity, the simulants to add per year are
+
+.. math::
+
+   \text{simulants_to_add_rate} = \frac{\text{sim_population}}{\text{simulated\_fraction}} \times \frac{\text{CBR}}{1000}
```
tylerdy, line 93: Copying comments from the commit to here now that I've made a PR:

tylerdyon May 6, 2025Collaborator
@aflaxman is this the right equation? as I'm thinking about it I would have expected it to be just simulated fraction * CBR / 1000, but maybe I'm misremembering our conversation. If it is right, what is sim_population here? Is it the numerator of the above equation?
aflaxman replied on May 6, 2025 aflaxmanon May 6, 2025MemberAuthor
We need to include the number of simulants somewhere, to make the number of simulants to add greater than 1.0.  sim_population is the current number of (living, tracked) simulants.
zmbc replied on May 6, 2025 zmbcon May 6, 2025Collaborator
I'm a bit confused how/why this discussion is happening here without a PR, but: I think the simulated_fraction here is something we haven't discussed before, and the way it is defined seems a bit convoluted since it immediately cancels out with the population term in the CBR.
If the simulated fraction were defined as 
  
    initialized population
    
      
        ∫
        
          a
          =
          age start
        
        
          age end
        
      
      population
      (
      a
      )
      ,
      d
      a
    
  
 it would make Tyler's equation above correct and would be interpretable as the "model scale".
Note that the current number of living, tracked simulants should not change the number of simulants to add. If you want to stick with the framing currently written, sim_population should be the size of the initial population of simulants.

tylerdy: The changes Abie made yesterday discussing how for the current definition of simulated fraction it would change depending on age_end (age_end=125, age_end=1) were helpful to me to understand the framing of simulated fraction. However I think the way it is written is confusing bc the total_population(a) da being in both makes it look like they are both being calculated on the same population. 

Based on the notes "Since the population represented in the sim might be only a fraction of the total population used as the denominator of the CBR, we need to find the fraction of the total population being simulated." and "where $\int_{a=\text{age start}}^{\text{age end}} \text{population}(a) \, da$ is the portion of the total population that is represented initially in the sim.", my understanding is that 

1. the numerator of simulated fraction represents the number of simulants initially in the sim
2. all simulants in the sim are represented by that numerator, not just a subgroup of the simulants falling in a certain age group, in other words whether age_start and age_end are set to 0 and 125, or some actual values, the integral value will be the same, and age_start and age_end are only there as convenience/information to describe the ages of our simulants and help frame the definition of simulated_fraction (this is based on my understanding that we will never be calculating births for a subset of the simulants we are modeling, correct me if I'm wrong)
3. the denominator reflects a different population, not our simulated population but the total GDB population associated with the CBR we are using
4. so, simulated fraction is the fraction of the GDB population we are modeling, even if the age_start and age_end limits to the integral make it look confusing/ like its denoting some subset of our population

If I'm understanding how we want to scale the model correctly, this would mean we would represent simulants_to_add_rate using the equation from my first comment, simulated fraction * CBR / 1000. If we did this I'd propose defining simulated fraction as simulated population / GBD population, and then noting that the simulated population can be thought of to have an age_start and age_end and using that to show the same examples that Abie did with age_end being 125 vs 1. 

But it sounds like Zeb is thinking about it differently, maybe we can discuss in coworking today @aflaxman @zmbc ?

zmbc: > the way it is written is confusing bc the total_population(a) da being in both makes it look like they are both being calculated on the same population.

They _are_ both being calculated on the same population, which is the GBD total population, in Abie's framing. With Abie's framing, the equation written in this PR is correct.

I think you are getting confused because you want it to be interpreted how I said, where "simulated_population" is the "model scale" -- that is **not** what Abie's doing. Maybe we can discuss which framing we want to go with, so we can get on the same page and avoid confusion in the future.

docs/source/models/other_models/fertility/index.rst
```diff
@@ -0,0 +1,138 @@
...
+    - 1106
+    - 
+
+
+Vivarium Modeling Strategy
+----------------------------
+
+The key challenge for using CBR in a fertility component in Vivarium is determining how many simulants to add in each time step.  This quantity would be simple were it not for the variety of ``age_start`` and ``age_end`` parameters that might appear in the sim.
+
+Since the population represented in the sim might be only a fraction of the total population used as the denominator of the CBR, we need to find the fraction of the total population being simulated.
```
zmbc, line 79: Start by defining in words the "initially represented population" -- which is the GBD population corresponding to the age and sex restrictions on simulants to initialize.

Then, say there are two ways to get to simulants to add from here: "epidemiology way" using CBR, or "model scale" way. Subsections for each.

Epi way:

Define the "initially represented fraction" as $\frac{\text{initially represented population size}}{\text{total GBD population size}}$. Note that in the case where our initialized population has no age or sex restrictions, this value is 1.

$\text{CBR} = \frac{\text{Live births}}{\text{Population}} \times 1000$

Then the `simulants_to_add_per_year` is $\frac{\text{initial simulated population size}}{\text{initially represented fraction}} \times \frac{\text{CBR}}{1000}$. Note: when initially represented fraction is 1, this is just $\text{initial simulated population size} \times \frac{\text{CBR}}{1000}$.

Model scale way:

Introduce the idea of "model scale", like a model-train set, which equals $\frac{\text{initial simulated population size}}{\text{initially represented population size}}$.

`simulants_to_add_per_year` is $\text{model scale} \times \text{Live births}$

In a side note: as relating to the previous subsection, model scale equals $\frac{\text{initial simulated population size}}{\text{total GBD population size} \times \text{initially represented fraction}}$. Show algebraic equivalence with previous method.

zmbc: Add TODO to move this into the "Vivarium Model Design" section

docs/source/models/other_models/fertility/index.rst
```diff
@@ -0,0 +1,138 @@
...
+
+age_start = 0, age_end = 1 --- simulated_fraction = fraction of population born in last year
+
+.. list-table:: Parameters
+  :header-rows: 1
+
+  * - Parameter
+    - Value
+    - Note
+  * -
```
zmbc, line 113: `simulants_to_add_per_timestep`

Value:

$\frac{\text{initial simulated population size}}{\int_{a=\text{age start}}^{\text{age end}} \text{GBD total population}(a) \, da} \times \text{Live births} \times \text{time step}/365$

docs/source/models/other_models/fertility/index.rst
zmbc: Maybe we could even add a note that the "model scale" is useful in _other_ ways: it can also be used to scale other things, e.g. DALY results from the sim back up to the "real world."

docs/source/model_design/designing_vivarium_model/fertility/index.rst
```diff
@@ -0,0 +1,151 @@
...
+  '''''''''''''''
+
+  The depth of each section level is determined by the order in which each
+  decorator is encountered below. If you need an even deeper section level, just
+  choose a new decorator symbol from the list here:
+  https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#sections
+  And then add it to the list of decorators above.
+
+===================================================
+Fertility Component that uses Crude Birth Rate Data
```
zmbc, line 32: It feels weird to me to say this "uses Crude Birth Rate Data," when, as described in "GBD modeling strategy," the data we get for this is _not_ a CBR, it is a count of "live births by sex."

Perhaps this component, both here and in [`vivarium_public_health`](https://github.com/ihmeuw/vivarium_public_health/blob/fcb9f4e3ec318a5b6d8f12fdde4308b5af9c0878/src/vivarium_public_health/population/add_new_birth_cohorts.py#L76), is misnamed?

Would be interested to hear others' thoughts.

docs/source/model_design/designing_vivarium_model/fertility/index.rst
```diff
@@ -0,0 +1,151 @@
...
+
+Overview
+-------------
+
+This document is intended to represent the fertility model for a simulation that adds new simulants based on a CBR.
+
+GBD Modeling Strategy
+----------------------
+
+Pregnancy and births are not explicit outcomes in the GBD study. However, there are location- and year-specfic GBD covariates related to fertility and births, and we use the "Live births by sex" covariate together with the GBD population estimates to calculate the CBR.  :math:`\text{CBR} = \frac{\text{Live births}}{\text{Population}} \times 1000`.
```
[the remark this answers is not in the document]
zmbc: The GBD population estimates are the midpoint population of the year, so after ~half of the live births in the numerator. We could add a note about this.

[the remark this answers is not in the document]
rmudambi: Our population scale is based on the initial population size as defined in the model spec, which is at the beginning of the time-step (i.e. before any fertility or mortality). Therefore, I think the population we care about here is actually the population at the beginning of the year (so maybe it's `pop = pop_GBD - (fertility + mortality) / 2` for total population. For age-group specific populations, it'd a bit more complicated to compute, but it'd basically be half the people who age into the age group and half of those who age out. 

Let me know if I'm thinking about this incorrectly.

tylerdy:
Making this a draft since #1735 makes some of these changes - still planning to come back to a more detailed explanation later

tylerdy:
> I'm a bit confused what happened here? Were the details removed from this PR?

@zmbc I'm actually not sure, thanks for pointing this out. At first I thought I must have clicked merge on this PR instead of the new minimal one, but that's merged too. I'm confused, will look into it tmrw morning.

PR #1643, opened 2025-05-07
Update effective coverage of probiotics 

pletale:
This PR updates our neonatal probiotics documentation so that only preterm early neonates are eligible for receiving probiotics. I made a very simple diagram of who's eligible. Please let me know if I should document this differently!


aflaxman (approved):
Here is a link to the rendered page: https://vivarium-research--1643.org.readthedocs.build/en/1643/models/intervention_models/neonatal/probiotics_intervention.html

And here is a screenshot of the new figure:
image

This seems pretty clear to me, but if the devs have any questions that could help us understand how to communicate complex details like this better in the future.

alibow (approved):

PR #1644, opened 2025-05-09
Correct and update note about COMO restrictions

zmbc:
Small update to this, which I ran up against/got curious about as I worked on the iron-responsiveness adjustment in the LSFF model.

I've shared the ticket with everyone tagged here.

### DIFF
--- a/docs/source/model_design/vivarium_features/other_causes_ylds/index.rst
+++ b/docs/source/model_design/vivarium_features/other_causes_ylds/index.rst
@@ -92,13 +92,15 @@ sequela prevalence.
 .. note::
 
   The COMO adjustment is run and maintained by IHME central computation. `The 
-  repository can be found here (as of May, 2023) <https://stash.ihme.washington.edu/projects/CCGMAC/repos/como/browse>`_.
+  repository can be found here (as of May, 2025) <https://stash.ihme.washington.edu/projects/CCGMAC/repos/como/browse>`_.
 
-  According to correspondence with a central computation representative: the COMO
-  adjustment ensures that no simulant may simultaneously possess mutually exclusive
-  sequelae (mild AND moderate diarrhea, for instance), nor may they simultaneously
-  possess mutually exclusive impairments (such as mild AND moderate anemia). A 
-  simulant may, however, simultaneously possess more than one sequelae that correspond
+  According to `correspondence with a central computation representative <https://help.ihme.washington.edu/servicedesk/customer/portal/5/CCMHD-22237>`_: the COMO
+  adjustment makes no restriction about the combinations of sequelae a simulant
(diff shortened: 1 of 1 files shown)


alibow (approved):
Interesting! Thanks Zeb, I was wondering what came of this.

There's another point in Catherine's response that is interesting to me too -- "All sequelae within a cause are mutually exclusive. For example, you can't have both mild AND moderate diarrhea. **It's up to teams to ensure the sequelae are by definition mutually exclusive**." So I am taking this as "sequelae for a given cause SHOULD be mutually exclusive," but there may be erroneous exceptions, like we see with maternal hemorrhage

zmbc:
@alibow My interpretation is that (1) someone wrote down at some point that sequelae of a cause should be mutually exclusive and (2) there is literally nowhere in GBD where that assumption is used or enforced. Given (2), it doesn't surprise me at all that some teams have strayed from (1).

zmbc:
I guess the part of (2) I'm not so sure about is how cause-level disability weights are calculated -- is that done centrally?

alibow:
> I guess the part of (2) I'm not so sure about is how cause-level disability weights are calculated -- is that done centrally?

I'm not sure that cause-level DWs are actually calculated in GBD. I think they calculate sequela-level YLDs using the DW for each sequela, and then the cause-level YLDs are just a sum of YLDs across that cause's sequelae. Vivarium calculates cause-level DWs as a prevalence-weighted average of sequelae DWs, but I think that is a weird convenience thing that only we do.

zmbc:
> I think they calculate sequela-level YLDs using the DW for each sequela, and then the cause-level YLDs are just a sum of YLDs across that cause's sequelae.

Except that there's a COMO adjustment, right?

And, I guess it would be interesting to determine whether or not it is enforced that sequela-level _prevalence_ sums to cause prevalence, which would imply mutual exclusivity.

alibow:
> > I think they calculate sequela-level YLDs using the DW for each sequela, and then the cause-level YLDs are just a sum of YLDs across that cause's sequelae.
> 
> Except that there's a COMO adjustment, right?
> 
> And, I guess it would be interesting to determine whether or not it is enforced that sequela-level _prevalence_ sums to cause prevalence, which would imply mutual exclusivity.

Yeah, but I think the COMO adjustment is done at the sequela-level, so then cause-level YLDs would still be a sum across sequela YLDs.

I just did a quick check and the YLDs for maternal hemorrhage are equal to the sum across its sequelae.

aflaxman (approved):

zmbc:
I made https://jira.ihme.washington.edu/browse/SSCI-2271 to look more into the sequela mutual exclusivity question.

PR #1646, opened 2025-05-09
Clean up of MNCNH concept model

alibow:
As I was doing recent V&V tracking I realized it would be helpful to make these changes.

Summary of what I did in this PR:
1. Updated observers table to what matches current outputs
2. Added observer 7 to reflect information that looked like it was in the old observer table but that I did not see in the outputs
3. Moved repeated info to a "defaults" table (including locations, population size, and number of draws that were repeated for each model and the base directory for all model results)
4. Removed the "status" column from the model runs table because we can see which runs have gone through V&V already in the V&V table
5. Brought table format in line with concept model template

Questions for the team (don't necessarily need to be resolved in this PR, but just want to make a record):
1. Do we want to stratify by random seed? Currently all of the results I am seeing are.
2. Confirm that the randomness key columns used in the model spec file (and listed in this PR) are adequate 
3. Confirm accurate representation of the age start/end parameters listed in the default specifications
4. Confirm that the antibiotics/probiotics/cpap observers are recording counts of births eligible and covered by the intervention

Immediate action items:
1. Make a new model run request that creates the maternal population count observer and perform V&V (will do after this PR is approved/merged)

### DIFF
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
@@ -543,49 +543,54 @@ defined as a module input in a subsequent row.
 
 Specific observer outputs and their stratifications may vary by model run as needs change. Modifications to default will be noted in the model run requests tables. Note that the observers and outputs listed here are different from the module outputs above. The outputs of the module are intended to be intermediate values that may or may not be included as observed simulated outputs.
 
+Default stratifications to all observers should include scenario and input draw.
+ 
+.. todo:: 
+
+  Confirm whether we want to continue stratifying by random seed? Currently all of the results dataframes are.
+
 .. list-table:: Simulation observers
   :header-rows: 1
 
-  * - Outcome
+  * - Observer
     - Default stratifications
-  * - ANC attendance counts
-    - 
-  * - Ultrasound coverage counts (AI-assisted, standard, none)
-    - 
-  * - Birth counts
-    - Delivery facility type, pregnancy outcome, child sex, scenario, and input draw
-  * - Pregnancy term counts (full or partial)
-    - 
-  * - Maternal GBD age group counts
-    - 
-  * - Delivery facility type counts (BeMONC, CeMONC, home)
-    - 
-  * - Azithromycin coverage counts
-    - 
-  * - Corticosteroid coverage counts
-    - 
-  * - Misoprostol coverage counts
-    - 
-  * - Maternal obstructed labor outcomes (deaths, YLLs, YLDs, incident cases)
-    - 
-  * - Maternal hemorrhage outcomes (deaths, YLLs, YLDs, incident cases)
-    - 
-  * - Maternal sepsis outcomes (deaths, YLLs, YLDs, incident cases)
-    - 
-  * - Count of neonatal deaths and YLLs by cause
-    - 
-  * - Neonatal antibioitics availability/coverage counts
-    - 
-  * - Neonatal probiotics availability/coverage counts
-    - 
-  * - CPAP availability/coverage counts
-    - 
+    - Note
+  * - Maternal disorders burden: cause-specific cases, deaths, YLLs, and YLDs
+    - * Maternal age group
+      * Pregnancy outcome
+      * Azithromycin coverage
+      * Misoprostol coverage
+    - Included with maternal age group stratification. Additional stratifiers to be added
+  * - Births (this observer includes ALL pregnancy outcomes, including partial term pregnancies that may not typically be considered "births")
+    - * Pregnancy outcome
+      * Child sex
+      * Delivery facility type
+    - Included
+  * - Neonatal deaths (cause-specific)
+    - * Child sex
+      * Child age group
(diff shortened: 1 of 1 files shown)


docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -543,49 +543,65 @@ defined as a module input in a subsequent row.
 
 Specific observer outputs and their stratifications may vary by model run as needs change. Modifications to default will be noted in the model run requests tables. Note that the observers and outputs listed here are different from the module outputs above. The outputs of the module are intended to be intermediate values that may or may not be included as observed simulated outputs.
 
+Default stratifications to all observers should include scenario and input draw.
+ 
+.. todo:: 
+
+  Confirm whether we want to continue stratifying by random seed? Currently all of the results dataframes are.
```
aflaxman, line 550: I have not thought about this before, but I don't think we ever want things stratified by random seed!

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -543,49 +543,65 @@ defined as a module input in a subsequent row.
...
+    - Included with maternal age group stratification. Additional stratifiers to be added
+  * - 2. Births (this observer includes ALL pregnancy outcomes, including partial term pregnancies that may not typically be considered "births")
+    - * Pregnancy outcome
+      * Child sex
+      * Delivery facility type
+      * Antibiotics availability
+      * Probiotics availability
+      * CPAP availability
+      * Corticosteroid coverage
+    - Included excpet for corticosteroid coverage, which has not yet been added
```
aflaxman, line 572:
```suggestion
    - Included except for corticosteroid coverage, which has not yet been added
```

aflaxman (approved):

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
alibow: We have before as an indirect way to assess the population size we should run with (see[ this notebook ](https://github.com/ihmeuw/vivarium_research_iv_iron/blob/main/validation/maternal/child%20seeds%20and%20draws%20analysis.ipynb)for an example of how we did it), but I definitely don't think we should do it by default. @pletale @NathanielBlairStahn do you know of any reason we have been using random seed stratification in our outputs?

NathanielBlairStahn: I don't know of any reason we'd need to stratify by random seed, but Alix has been the one actually dealing with the outputs.

SylLutze: This was a change introduced by the new results stratification system. I believe @stevebachmeier and I made this call. IIRC the conversation went something along the lines of "well occasionally we use this and it's easier for engineering to include them always then and it's not THAT hard for RT to deal with it so sure stick them in". But I agree from an RT perspective it's a tad annoying and I'm happy to have engineering remove this.

pletale: No I wasn't really using the random seed stratification when processing results, sounds good to drop it to me

pletale (approved):

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
alibow: Oh, interesting. That makes sense. In that case I think it's fine to leave in. Maybe we can touch base about this in RT/doc session, but I think I'll leave it as a todo for now in that case.

docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -603,251 +619,262 @@ Specific observer outputs and their stratifications may vary by model run as nee
...
+      * Pakistan (ID: 165)
+    -
+  * - Number of draws
+    - 10
+    - Based on calculations from the `Nutrition Optimization project <https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.html#production-run-specifications>`_: production run number divided in half for default V&V runs
+  * - Population size per draw
+    - 100,000
+    - Based on calculations from the `Nutrition Optimization project <https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_nutrition_optimization/kids/concept_model.html#production-run-specifications>`_: production run number divided in half for default V&V runs
+  * - Randomness key columns
+    - ['entrance_time','age']
```
alibow, line 640: do we need to make sure that the age value changes between timesteps because if it does not we will not refresh our randomness key???

alibow: Rajan says we get a new randomness key on every timestep even if age does not update... but is 99% sure and will check

PR #1651, opened 2025-05-13
MNCNH postpartum depression cause model

alibow:


pletale (approved):

docs/source/models/causes/maternal_disorders/gbd_2021_mncnh/postpartum_depression.rst
```diff
@@ -0,0 +1,275 @@
...
+
+- We are limited in that we do not consider mortality associated with PPD in our model.
+
+- We are limited in that we assume all PPD cases persist for the same average duration of a single MDD episode. This is limited in the sense that duration may vary by MDD severity (for example, a longer duration for more severe cases), which could cause our estimation of YLDs to be biased. Additionally, we assume that duration of PPD is equal to the duration of all MDD episodes.
+
+- We assume that the GBD MDD severity distribution, which is based on analysis of high-resource settings, generalizes to the severity of PPD in our simulation population of interest.
+
+- We assume that the onset of PPD occurs immediately following birth. However, the onset of PPD may peak around two or three months postpartum [Shorey-et-al-2018-mncnh]_.
+
+- We assume that the incidence of PPD does not vary by pregnancy outcome (incidence is constant across partial term pregnancy loss, stillbirth, and live births)
```
aflaxman, line 110: Is there any evidence about this in the Shorey et al paper?

aflaxman (approved):

docs/source/models/causes/maternal_disorders/gbd_2021_mncnh/postpartum_depression.rst
alibow: No, looking back at it now (this is the reference we used for the IV iron model back in the day) it's actually just among live full term births. There is some evidence that depression rates are higher among preterm infants, but the studies I have seen in my quick check are smaller among targeted populations. We could consider adding effect modification by preterm birth on depression incidence. I will add a note that we can come back to for this. I think the overall impact of adding this to the model will be small, so it may be a low priority to add.

PR #1653, opened 2025-05-14
Update NN antibiotics RR value and write up limitations/assumptions of our approach

pletale:
I also added a to-do to run the value past someone with some more expert knowledge on antibiotics!

### DIFF
--- a/docs/source/models/intervention_models/neonatal/antibiotics_intervention.rst
+++ b/docs/source/models/intervention_models/neonatal/antibiotics_intervention.rst
@@ -101,7 +101,12 @@ This reduces to the previous formula if there are no other interventions, and we
 
    \text{CSMR}^\text{sepsis}_{i, \text{original}} = \text{CSMR}^\text{sepsis}_{\text{BW}_i, \text{GA}_i}
 
-While we are searching the literature for an appropriate value for the relative risk, we will use a stand-in value with an origin I have failed to record.
+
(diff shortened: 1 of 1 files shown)


alibow:
Did/should we adjust for differences in LBWSG PAFs in this analysis?

docs/source/models/intervention_models/neonatal/antibiotics_intervention.rst
```diff
@@ -112,14 +117,19 @@ While we are searching the literature for an appropriate value for the relative
     - Distribution
     - Notes
   * - Relative Risk
-    - 1.39
+    - 38.0
     - :math:`\text{Normal}(1.39,0.08^2)`
```
zmbc, line 121: This needs to be updated too, I think!

alibow (changes requested):
Ok now that I'm back and reviewed the notebook you linked I am worried that LBWSG exposures are a significant confounder in this analysis! I think that if we should update the analysis used here to adjust for LBWSG. I think we can use the LBWSG PAFs as a covariate in a model that uses sepsis CSMR percentile to predict sepsis CSMR

aflaxman (approved):
Let's merge this to keep a history of our approaches, even though you are planning a cooler approach using a linear regression on Access to LRI Antibiotics, after adjusting NN Sepsis to control for LBWSG (and possibly switching from CSMR to EMR?)

PR #1658, opened 2025-05-29
Azithromycin model run request

alibow:
I wanted to add this info to make sure we will be all set for V&V! We need to add a new scenario to make sure we can do V&V as two of our locations do not have baseline coverage. Let me know if we want to structure this scenario differently!


pletale (approved):
Thanks Ali!

PR #1663, opened 2025-05-29
CPAP uncertainty interval docs and run request

alibow:
I added some documentation details around how the effect is applied because it was a little unclear to me as written and I think this improves it, but there shouldn't be any implementation changes! 

The only thing for @albrja to know about this PR is that we want to add uncertainty to the parameter in the artifact as described for the RR value in the parameter table (specifically look at line 127) for model run 7.0.3. Let me know if you have any questions.

And FYI @aflaxman I took a quick look at the cochrane review and it looks like the effect is actually for all cause mortality rather than RDS mortality. I added this as a limitation and noted that it may cause us to underestimate the impact of CPAP in our intervention.

### DIFF
--- a/docs/source/model_design/general_reference_material/bliss/race_ethnicity/index.rst
+++ b/docs/source/model_design/general_reference_material/bliss/race_ethnicity/index.rst
@@ -40,11 +40,13 @@ Inclusive Health Metrics: Race and Ethnicity
 Overview
 +++++++++++++++++++++
 
-On this page you will find ...
+On this page you will find our guiding principles,
+an introduction to race and ethnicity from a scientific perspective,
+a glossary of commonly used terms related
+to race and ethnicity, recommendations for how to address race- and ethnicity-related limitations in external data
+sources, discussion of the implications of imperfect proxy measures with regards to race and ethnicity,
+examples of how our team has incorporated BLISS into our work process, and more! 
 
-.. todo::
-
-    Fill out this section to describe what this guide will include.
 
 If you have any questions or comments about the content on this page - please reach out! You can
 contact us directly `on GitHub <https://github.com/ihmeuw/vivarium_research/issues?q=is%3Aopen+is%3Aissue+label%3Abliss>`_ or anonymously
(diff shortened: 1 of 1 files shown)


docs/source/models/intervention_models/neonatal/cpap_intervention.rst
```diff
@@ -196,7 +234,8 @@ Assumptions and Limitations
 
 - We assume that CPAP availability captures actual use, and not simply the machine being in the facility 
 - We assume that the delivery facility is the final facility in the care continum for deliveries that are transferred due to complications
-- We assume that the relative risk of RDS mortality with CPAP in practice is similar to that found in the Cochrane Review meta-analysis
+- We assume that the relative risk of RDS mortality with CPAP in practice is similar to that found in the Cochrane Review meta-analysis. Given that the review assessed overall mortality rather than RDS mortality, it is likely that we will underestimate the overall impact of CPAP on mortality in our simulation.
```
SylLutze, line 237: This is interesting. I guess we would need to back calculate an impact on RDS or at least pre-term that got the effect on overall mortality found here? 

I made a Jira ticket for this data analysis and update - feel free to adjust the description. I put it in model expansion high priority. If we want to improve this, I think it would be worthwhile. Ticket: https://jira.ihme.washington.edu/browse/SSCI-2286

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
```diff
@@ -87,7 +87,45 @@ We will then add an attribute to each simulant indicating whether the birth occu
...
+    - Value
+    - Note
+  * - :math:`\text{CSMRisk}^{\text{preterm with RDS}}_{\text{BW},\text{GA}, i}`
+    - Mortality risk due to preterm with RDS for a given simulant :math:`i` with a specific birth weight and gestational age exposure following modification from the CPAP intervention
+    - See equation above
+    - 
+  * - :math:`\text{CSMRisk}^{\text{preterm with RDS}}_{\text{BW},\text{GA}}`
+    - Mortality risk due to preterm with RDS for a simulant with a given birth weight and gestational age exposure before modification from the CPAP intervention
+    - Derived from instruction on the :ref:`neonatal preterm birth cause model document <2021_cause_preterm_birth_mncnh>`
+    -
```
NathanielBlairStahn, line 114: It could be confusing that the notation here is slightly different from what's on the preterm birth cause page, which I think uses $\text{CSMRisk}^{\text{preterm with RDS}}\_i$ and $\text{CSMRisk}^{\text{preterm with RDS}}\_{\text{BW},\text{GA}}$ for these, respectively.

alibow:
Thanks @NathanielBlairStahn -- I have updated the notation. Also, in case you haven't already seen it, here's a thread where we discuss some next steps for CPAP as an FYI.

NathanielBlairStahn (approved):

PR #1669, opened 2025-06-05
MNCNH model 7.1.1 vv

alibow:


pletale (approved):

albrja (approved):

aflaxman (approved):
Which figure in the validation notebook showed that CPAP intervention size was not computed as desired?

alibow:
> Which figure in the validation notebook showed that CPAP intervention size was not computed as desired?

There is not a figure, but at the very bottom of the notebook I evaluated distribution statistics of the artifact values and saw that they did not reflect the values in the documentation. image

This figure shows that the sim and artifact values for the effect of CPAP are reasonably close, though I should probably update to a boxplot or something

PR #1680, opened 2025-06-12
MNCNH wave II run requests (10.0 and 11.0)

alibow:


aflaxman (approved):

NathanielBlairStahn (approved):
Looks good. I did, however, notice a few anomalies on the postpartum depression page (sorry I didn't review that PR when it was open):

- The `cfr` row in the transition probability table [refers to sepsis instead of postpartum depression](https://vivarium-research.readthedocs.io/en/latest/models/causes/maternal_disorders/gbd_2021_mncnh/postpartum_depression.html#cause-postpartum-depression-mncnh:~:text=The%20probability%20that%20a%20simulant%20with%20sepsis%20or%20another%20maternal%20infection%20dies%20of%20that%20infection)
- The uncertainty for the incidence risk [specifies a normal distribution truncated at the 95% CI limits](https://vivarium-research.readthedocs.io/en/latest/models/causes/maternal_disorders/gbd_2021_mncnh/postpartum_depression.html#cause-postpartum-depression-mncnh:~:text=truncate%20at%2095%25%20CI%20limits), but I think we should truncate at 0 and 1 instead (the only point of truncating is to ensure that the number is a probability, and if we truncate to a narrower interval, it will reduce the uncertainty more than necessary)
- The documentation says the cause is nonfatal only, but the decision diagram still shows a case fatality rate -- this is confusing: Why bother including nodes and edges that will never be used? Also, if we *do* want to include this for some reason, the "parent alive" and "parent dead" nodes look like they were probably copied from the maternal sepsis page, but those nodes are mislabeled (I think there's a Jira ticket to fix them) -- they should be "parent died of PPD" and "parent did not die of PPD" instead, like on the other maternal disorders pages.
- It looks like PPD applies to all births (full-term or partial-term), whereas the other maternal disorders so far only applied to full-term births. This is fine, but I'd be interested in discussing all the implications of doing more with the partial term pregnancies in our sim, because up to this point, I think we had been sort of ignoring them because we weren't doing anything with them.

alibow:
Thanks @NathanielBlairStahn. I just pushed a commit that fixed the labeling and truncation issues you mentioned (and will let Jim know). I kept the CFR in the diagram because I figured it would make it more clear that we could reuse the existing maternal disorders code in the sim and just set the CFR to zero rather than make custom code changes and this way if we ever do get a CFR we could update it accordingly with just a trivial code change. I did apply this to all pregnancies but I have also wondered what we should do with the partial term pregnancies because they are kind of a pain in V&V and based on the interactive sim I think we are doing a decent amount of math on them that doesn't end up getting used so we're slowing things down by including them in the sim. I am tempted to just keep things as-is here but maybe bring this up for discussion in an upcoming MNCNH RT meeting?

PR #1692, opened 2025-06-24
MNCNH: Update data for GA errors; make facility choice language consistent

NathanielBlairStahn:
Jira tickets: [SSCI-2296](https://jira.ihme.washington.edu/browse/SSCI-2296), [SSCI-2297](https://jira.ihme.washington.edu/browse/SSCI-2297)

Makes the following changes:

- Update the standard deviations for error in estimated gestational age in AI ultrasound model
- Edit language on facility choice model page and all related module pages to be internally consistent. The affected pages are:
   
   - facility choice model page
   - facility choice module
   - initial attributes module
   - pregnancy module
   - pregnancy model page
   - ~~ANC module~~ (updates superseded #1690)
   - AI ultrasound module

I have **not** yet updated the language on the concept model page -- that will come in the next PR. Also, the diagram on the AI ultrasound page should probably get updated, but I haven't done that yet either.

PR #1694, opened 2025-06-25
MNCNH: Concept model abbreviations and language for facility choice

NathanielBlairStahn:
Jira ticket: [SSCI-2326](https://jira.ihme.washington.edu/browse/SSCI-2326)

Sequel to #1692

Update MNCNH concept model to make language consistent with facility choice docs. Specifically, I edited the module tables for the pregnancy component and intrapartum component.

When I changed "Delivery facility propensity" to  "IFD propensity" in the module tables, I thought I should define "IFD" somewhere. Then I realized that there were a ton of abbreviations throughout the page with no definitions, so I made an abbreviations table. **Please double check to make sure I have the correct definitions for all the abbreviations**, and let me know if I missed any!

### DIFF
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
@@ -229,10 +229,11 @@ defined as a module input in a subsequent row.
     - Outputs
     - Nested subcomponents
   * - :ref:`Facility choice <2024_vivarium_mncnh_portfolio_facility_choice_module>`
-    - * (Pregnancy term duration)
-      * Delivery facility propensity
-      * Believed gestational age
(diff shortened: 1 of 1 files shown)


alibow (approved):
Looks good to me! Thanks for adding this. It looks like it doesn't come up on this page, but I think we do use LBW and PTB on some subpages referenced on the concept model, so maybe it's worth adding those in too.

NathanielBlairStahn:
> Looks good to me! Thanks for adding this. It looks like it doesn't come up on this page, but I think we do use LBW and PTB on some subpages referenced on the concept model, so maybe it's worth adding those in too.

@alibow Sure, I can add these. Does PTB stand for "preterm birth"?

alibow:
Yep!

PR #1698, opened 2025-06-30
MNCNH anemia screening intervention

pletale:


alibow (commented):
I'm not sure if you're looking for review at this point or not with the draft status of this PR, but I have a couple of questions/thoughts in the meantime:

- I think the values from Annie's team are specific to coverage of blood draws at ANC, not necessarily whether or not hemoglobin concentration was assessed. For our purposes we will assume that hemoglobin was assessed if any blood was taken, but we should make it clear what the data represents and the assumptions we are making about it.
- Could you pull the values from Annie's data for the countries we are modeling for reference? I am curious to see what they are and to know if they are similar to the rough/low value you pulled for ferritin 
- How we've structured this I think it may be a bit complicated to have different values for baseline coverage of the hemoglobin and ferritin screenings... I think depending on what the blood draw baseline coverage values are (close to 4%) we could make an assumption that they are the same. Otherwise we may have to make a few edits to the hemoglobin page to make this work, but that can be a separate follow-up task

alibow (approved):

docs/source/models/intervention_models/mncnh_pregnancy/anemia_screening.rst
```diff
@@ -0,0 +1,152 @@
...
+is possible for non-iron-responsive anemias to also have low iron levels. See the :ref:`anemia impairment document <2019_anemia_impairment>` for a list of iron responsive and non iron responsive 
+causes of anemia in the GBD.
+
+Therefore, in our model we will use the severity-specific fraction of iron responsive anemia among all causes of anemia in GBD as a proxy measure for the fraction of anemia cases with low ferritin. 
+This approach is limited in that we may slightly underestimate total eligibility by not considering the proportion of the population who has low hemoglobin due to an iron-non-responsive cause and 
+also coincidentally has low ferritin.
+
+.. note::
+
+  Chris T. has suggested that we can use the fraction of iron deficiency anemia from the in-progress PRISMA study rather than GBD for this purpose. PRIMSA study results are expected in June or July of 2025.
```
aflaxman, line 75: It is now June or July of 2025, so we should check if these have become available

aflaxman (approved):

PR #1701, opened 2025-07-01
MNCNH model 11.2 vv summary

alibow:


albrja (approved):

alibow:
@aflaxman I am not exactly sure where your screenshot came from because I am not seeing those exact same values, but assuming this is some model version for the effect of misoprostol coverage: there are only about 10-20 maternal sepsis deaths in each draw/location/scenario we run, so I think this is just unlucky. Since it is not a consistent finding across draws (min RR: 0.26 and max RR: 5.8 in the version I am seeing) I think this is just random noise and I am not worried about an erroneous effect being applied.

PR #1705, opened 2025-07-03
 Second IHME python assignment

Osoti-Osoti:
Sharing my second python assignment for review


osotiihme_assignment2.ipynb
```diff
@@ -0,0 +1,1992 @@
+{
```
SylLutze, line 1: <p>This is cool doing the plot in 2 ways! There are aways so many ways to do things that it's neat to see other approaches, great work! </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1705/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

SylLutze, line 1: <p>This looks like it worked for the calculation, but it's best practice to include all the duplicate columns in the "on" section. First, then you don't get duplicates, which keeps things cleaner. For example, if you had <code>on=['location_name', 'age_group_id','sex_id']</code> then you would have one column for 'sex_id' in the combined dataset rather than sex_id_x and sex_id_y. But secondly, you can introduce errors by not matching on enough columns as data can be mismatched. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_research/pull/1705/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

SylLutze:
Overall this looks great! Thanks for completing and sending in the assignment!

osotiihme_assignment2.ipynb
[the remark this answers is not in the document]
Osoti-Osoti: <p>I will improve on this </p>

PR #1712, opened 2025-07-10
Update CPAP and ACS intervention pages with shared propensity value

pletale:


docs/source/models/intervention_models/intrapartum/acs_intervention.rst
```diff
@@ -176,52 +147,8 @@ of ACS for pregnant women and people at imminent risk of preterm delivery.
...
-
-.. math::
-
-   \mathbb{E}(\text{RR}) = p(\text{no ACS}) \cdot \text{RR}_\text{no ACS} + (1 - p(\text{no ACS})) \cdot 1.
-
-.. note:: 
-
-  Simulants should have the same propensity value for CPAP (see the :ref:`CPAP intervention model page <intervention_neonatal_cpap>`) and ACS, such that the same
-  simulants are exposed to both interventions (i.e., if coverage of both CPAP and ACS is 50%, then the same 50% of simulants will be getting each intevention).
+    - see the :ref:`CPAP intervention Calibration Strategy <intervention_neonatal_cpap>` for details on how to calculate PAF that is consistent with RR, risk exposure, and facility choice model.
```
alibow, line 150:
```suggestion
    - see the :ref:`CPAP intervention Calibration Strategy <intervention_neonatal_cpap>` for details on how to calculate joint PAF for ACS and CPAP on RDS that is consistent with RRs, risk exposures, and the facility choice model.
```

alibow (commented):
We should probably alter the equation in line 121 of the ACS doc to clarify how we are using the PAFs here. I suggest that we either:
- Make the "\text{CSMRisk}^\text{preterm with RDS}" term "\text{CSMRisk}^\text{preterm with RDS, modified by the CPAP intervention}" and then delete the (1 - PAF) term from this equation and keep the RR_no_ACS, or
- Add the same equation on both the ACS and CPAP pages that uses the joint PAF and the multiplies by the RRs for both interventions,
- or both! Let me know if you have questions about how to best do this

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
```diff
@@ -113,7 +123,7 @@ Where,
     - Derived from instruction on the :ref:`neonatal preterm birth cause model document <2021_cause_preterm_birth_mncnh>`
     - 
   * - :math:`\text{PAF}`
-    - Population attributable fraction of mortality due to preterm with RDS from access to CPAP intervention
+    - Population attributable fraction of mortality due to preterm with RDS from access to CPAP and ACS interventions
```
alibow, line 126:
```suggestion
    - Joint population attributable fraction of mortality due to preterm with RDS from access to CPAP and ACS interventions
```

docs/source/models/intervention_models/intrapartum/acs_intervention.rst
```diff
@@ -176,52 +147,8 @@ of ACS for pregnant women and people at imminent risk of preterm delivery.
...
+      No adjustments are needed to the PAF calculation for ACS, as it is the same as the CPAP intervention.
```
zmbc, line 151: I think along with Ali's suggestion above, this line should be removed; we are literally using one PAF for both, not using the same calculation method for each.

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
```diff
@@ -126,13 +136,20 @@ Where,
     - Relative risk of CPAP intervention on RDS mortality 
     - 0.53 (95% CI 0.34-0.83). Uncertaintly interval implemented as parameter uncertainty following a lognormal distribution
     - `2020 Cochrane review <https://pmc.ncbi.nlm.nih.gov/articles/PMC8094155/>`_. Note that this effect was measured for all cause mortality.
+  * - :math:`\text{RR}_\text{ACS}`
+    - Relative risk of ACS intervention on RDS mortality 
+    - Refer to :ref:`ACS intervention page <acs_intervention>` for this effect size.
+    - Only to be included in PAF calculation if simulant is within the gestational age range that is eligible for ACS (26-33 weeks).
```
zmbc, line 142: To me, this implies that it is _not_ true that "we assume the same simulants have access to (i.e., are exposed to) CPAP and ACS", since it sounds like there are simulants outside the gestational age range who could receive CPAP but not ACS. Am I missing something?

docs/source/models/intervention_models/neonatal/cpap_intervention.rst
```diff
@@ -126,13 +136,20 @@ Where,
...
 .. _cpap_calibration:
 
 Calibration Strategy
 --------------------
 
-The following decision tree shows all of the paths from delivery facility choice to CPAP availability.  Distinct paths in the tree correspond to disjoint events, which we can sum over to find the population probability of RDS mortality.  The goal here is to find internally consistent probabilities of RDS mortality for the subpopulations with and without access to CPAP, so that the baseline scenario can track who has access to CPAP and still match the baseline RDS mortality rate.
+The following decision tree shows all of the paths from delivery facility choice to CPAP availability.  Distinct paths in the tree correspond to 
+disjoint events, which we can sum over to find the population probability of RDS mortality.  The goal here is to find internally consistent probabilities 
+of RDS mortality for the subpopulations with and without access to CPAP and ACS, so that the baseline scenario can track who has access to CPAP and still match 
+the baseline RDS mortality rate.
```
zmbc, line 152: Just adding line breaks, right? I usually slightly prefer to add them between sentences but maybe that's just me

zmbc (approved):
Mostly seems good, I had a few questions

PR #1719, opened 2025-07-18
Add model run request for preterm prevalence bugfix

zmbc:
I have a pretty strong hunch that one reason the CSMR of preterm birth looks too high is because [this less-than sign in the calculation of preterm prevalence](https://github.com/ihmeuw/vivarium_gates_mncnh/blob/2efd1d73af83e282351de115da7030083389cedb/src/vivarium_gates_mncnh/data/loader.py#L644) should be less-than-or-equal-to, since the interval is open on the right.

### DIFF
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
@@ -1186,6 +1186,13 @@ Default stratifications to all observers should include scenario and input draw.
     - Added preterm status and believed preterm status to maternal
       population observer (#7)
     - Default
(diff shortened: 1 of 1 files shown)


zmbc:
I expect this will increase $p_\text{preterm}$ in [this equation](https://vivarium-research.readthedocs.io/en/latest/models/causes/neonatal/preterm_birth.html#modeling-strategy), thereby decreasing $\text{CSMRisk}$ for both preterm sub-causes, which are currently looking too high in V&V. It may also improve the underestimation of other neonatal causes' CSMRisks, since it should make the negative-other-causes-CSMRisk situation rarer or possibly eliminate it.

PR #1722, opened 2025-07-21
Add model 12.1 V&V to table

zmbc:
This is kind of half-finished, because I haven't yet addressed the V&V criteria of "Confirm LBWSG risk effects are working as expected."

But this was the main point of interest for this model, and 12.1 seems to have improved things substantially!

We think the remaining discrepancy may be an issue in our approach to calculating late neonatal PAFs.


alibow (approved):

zmbc:
Thanks Ali, I just noticed I should have updated that while I was reviewing your PR!

PR #1726, opened 2025-07-25
Update IFA documentation based on new Cochrane review

pletale:
We found a new Cochrane review for the IFA intervention, so this PR:
- Updates the reference for the IFA effect size on birthweight to be the new Cochrane review (which reported the same value as the previous review)
- Updates the effect size on maternal hemoglobin concentration (Note: both the 2015 and new 2024 Cochrane reviews focus on mainly high-income countries, with no representation of Sub-Saharan Africa, there is an RCT from Pakistan included in both though. I could note the limitation that this might overestimate effect size on Hb given SSA countries have higher rates of non-iron deficiency anemias, but I haven't yet!)


alibow (approved):
Thanks Alix!

zmbc (approved):

PR #1731, opened 2025-07-29
Burn in docs

tylerdy:
Document burn-in periods including examples of how we use them in our models.


pletale (approved):

zmbc (approved):
Overall this looks great to me! Some minor notes and included a replacement plot with less confusing years 😄

PR #1732, opened 2025-07-30
Model 13.3: Update to end-of-ENN prevalence for preterm in LNN CSMRisk equation

zmbc:
Using the prevalence at the beginning of the age group rather than at birth should _decrease_ $p_\text{preterm}$ (since preterm babies have higher mortality), thereby increasing CSMRisks and improving our current underestimation of mortality due to preterm (subcauses) in the LNN age group.

### DIFF
--- a/docs/source/models/causes/neonatal/preterm_birth.rst
+++ b/docs/source/models/causes/neonatal/preterm_birth.rst
@@ -238,7 +238,7 @@ age less than 37 weeks:
 
 .. math::
 
-  p_{\text{preterm},\text{ENN}} = \sum_{\{\text{cat}: \text{GA}<37\}} \text{lbwsg_birth_prevalence}_\text{cat}
+  p_{\text{preterm},\text{ENN}} = \sum_{\{\text{cat}: \text{GA}<37\}} \text{lbwsg_birth_prevalence}_\text{cat},
 
 where :math:`\text{lbwsg_birth_prevalence}` can be pulled from GBD with minor transformations,
 as detailed in the table below.
@@ -255,8 +255,13 @@ Similarly to the LNN PAF, *after* the early neonatal calculations are complete, 
 preterm at the end of the ENN age group should be calculated.
 This value should be used as :math:`p_{\text{preterm},\text{LNN}}` for the purposes
 of the CSMRisk equation.
-Concretely, in the PAF calculation pipeline, in which LBWSG categories have not been assigned using the birth prevalence (currently exactly equally between categories),
-the prevalence is calculated as follows:
+
+Determining the prevalence of preterm is a bit more complex than it sounds, because in the PAF calculation pipeline,
+the same number of simulants are assigned to each LBWSG category, rather than assigning each simulant
+to a random category with probability equal to that category's prevalence at birth.
+Due to this initialization strategy, all quantities calculated in the pipeline must use *weights*
+to account for the fact that the simulants in the categories with higher birth prevalence represent more people.
+Therefore, :math:`p_{\text{preterm},\text{LNN}}` is calculated as follows:
 
 .. math::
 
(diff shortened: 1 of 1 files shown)


docs/source/models/causes/neonatal/index.rst
```diff
@@ -208,7 +208,11 @@ unit time.
 Modeling Strategy
 +++++++++++++++++
 
-The neonatal death model requires only the probability of death (aka "mortality risk") for the early and late neonatal time periods. Rather than using GBD mortality rates and converting them into probability of deaths, we will use mortality risk as direct input data into our model. We will calculate mortality risk input data as age-specific death counts divided by live birth counts from GBD.
+The neonatal death model requires only the probability of death (aka "mortality risk") for the early and late neonatal time periods.
+These risks are age-group-, sex-, and location-specific.
```
aflaxman, line 212:
```suggestion
These mortality risks are age-group-, sex-, and location-specific.
```

docs/source/models/causes/neonatal/preterm_birth.rst
```diff
@@ -202,7 +209,64 @@ where :math:`k` is the subcause of interest (preterm birth with or without RDS),
...
+age less than 37 weeks:
+
+.. math::
+
+  p_{\text{preterm},\text{ENN}} = \sum_{\{\text{cat}: \text{GA}<37\}} \text{lbwsg_birth_prevalence}_\text{cat}
+
+where :math:`\text{lbwsg_birth_prevalence}` can be pulled from GBD with minor transformations,
+as detailed in the table below.
+
+For LNN the situation is more complicated, because we need to account
```
aflaxman, line 244:
```suggestion
For LNN, the situation is more complicated, because we need to account
```

aflaxman (approved):

docs/source/models/causes/neonatal/preterm_birth.rst
```diff
@@ -202,7 +209,64 @@ where :math:`k` is the subcause of interest (preterm birth with or without RDS),
...
+preterm at the end of the ENN age group should be calculated.
+This value should be used as :math:`p_{\text{preterm},\text{LNN}}` for the purposes
+of the CSMRisk equation.
+Concretely, in the PAF calculation pipeline, in which LBWSG categories have not been assigned using the birth prevalence (currently exactly equally between categories),
+the prevalence is calculated as follows:
+
+.. math::
+
+  p_{\text{preterm},\text{LNN}} = \frac{
+    \sum_{\{\text{cat}: \text{GA}<37\}} \text{lbwsg_birth_prevalence}_\text{cat} \times \frac{n_\text{cat} - n^\text{deaths}_\text{cat}}{n_\text{cat}}
```
hussain-jafari, line 262: Should this (and other instances) be less than or equal to instead of less than?

alibow (approved):

docs/source/models/causes/neonatal/preterm_birth.rst
```diff
@@ -202,7 +209,64 @@ where :math:`k` is the subcause of interest (preterm birth with or without RDS),
...
+where the :math:`\text{enn_death_count}` and :math:`\text{live_birth_count}` are
+quantities pulled from GBD, as detailed in the table below.
+
+The LNN CSMRisk is:
+
+.. math::
+
+  \text{CSMRisk}_\text{LNN} = \frac{\text{lnn_death_count}}{\text{live_birth_count} - \text{enn_all_cause_death_count}}
+
+:math:`p_{\text{preterm}}`, as mentioned above, represents the prevalence/exposure
```
NathanielBlairStahn, line 231:
```suggestion
  \text{CSMRisk}_\text{LNN} = \frac{\text{lnn_death_count}}{\text{live_birth_count} - \text{enn_all_cause_death_count}},

where, again, all quantities are pulled from GBD as detailed in the table below.

:math:`p_{\text{preterm}}`, as mentioned above, represents the prevalence/exposure
```

This was a bit confusing when I first read it because there was no period after the equation, but I think $p\_{\text{preterm}}$ is the start of a new sentence (and maybe a new paragraph?). Also, I think it will make things clearer to specify that these quantities also come from GBD like in the previous equation.

docs/source/models/causes/neonatal/preterm_birth.rst
```diff
@@ -202,7 +209,64 @@ where :math:`k` is the subcause of interest (preterm birth with or without RDS),
...
+prevalence from the same LBWSG PAF calculation pipeline used for :math:`Z`
+above.
+As detailed at :ref:`details_of_the_lbwsg_paf_calculation` on the neonatal all-cause
+mortality page, there are two iterative steps using microsimulation, with the late neonatal calculations
+using the result of the early neonatal calculations.
+Similarly to the LNN PAF, *after* the early neonatal calculations are complete, the prevalence of
+preterm at the end of the ENN age group should be calculated.
+This value should be used as :math:`p_{\text{preterm},\text{LNN}}` for the purposes
+of the CSMRisk equation.
+Concretely, in the PAF calculation pipeline, in which LBWSG categories have not been assigned using the birth prevalence (currently exactly equally between categories),
```
NathanielBlairStahn, line 256: I find this sentence confusing... Is this referring to the fact that in the PAF calculation sim, we initialize each category with the same number of simulants, rather than initializing each category with a number of simulants proportional to its birth prevalence? Is that documented somewhere? There must be a clearer way to get this across, maybe some way that explicitly mentions the "grid points" referred to parenthetically after the equation...

docs/source/models/causes/neonatal/preterm_birth.rst
```diff
@@ -202,7 +209,64 @@ where :math:`k` is the subcause of interest (preterm birth with or without RDS),
...
+
+.. math::
+
+  p_{\text{preterm},\text{LNN}} = \frac{
+    \sum_{\{\text{cat}: \text{GA}<37\}} \text{lbwsg_birth_prevalence}_\text{cat} \times \frac{n_\text{cat} - n^\text{deaths}_\text{cat}}{n_\text{cat}}
+  }{
+    \sum_{\text{cat}} \text{lbwsg_birth_prevalence}_\text{cat} \times \frac{n_\text{cat} - n^\text{deaths}_\text{cat}}{n_\text{cat}}
+  }
+
+where :math:`n_\text{cat}` is the number of simulants initialized into each category before mortality was applied (the number of grid points in each category) and :math:`n^\text{deaths}_\text{cat}` is the number of deaths in each category when ENN mortality was applied in the PAF calculation pipeline. Note that :math:`n_\text{cat}` will not vary by LBWSG exposure category under the current approach of assigning LBWSG exactly equally between categories.
```
NathanielBlairStahn, line 267: The phrase "assigning LBWSG exactly equally between categories" seems imprecise/unclear to me. Maybe:

```suggestion
where :math:`n_\text{cat}` is the number of simulants initialized into each category before mortality was applied (the number of grid points in each category) and :math:`n^\text{deaths}_\text{cat}` is the number of deaths in each category when ENN mortality was applied in the PAF calculation pipeline. Note that :math:`n_\text{cat}` will not vary by LBWSG exposure category under the current approach of calculating a weighted average after assigning the same number of simulants to each LBWSG category.
```

NathanielBlairStahn (approved):
Looks good overall! I left a few comments about improving the clarity in a couple places.

docs/source/models/causes/neonatal/preterm_birth.rst
zmbc: Yes it is, I will revise to clarify!

docs/source/models/causes/neonatal/preterm_birth.rst
zmbc: Thanks! I'm going to propagate this idea a bit more into the preceding paragraph.

PR #1735, opened 2025-08-05
Live births fertility component - minimal explanation

tylerdy:
Minimal version/re-write of #1642 that leaves out Crude Birth Rate and epidemiology way of calculating simulants to add. Incorporates some feedback from discussions with and PR feedback from Zeb, Rajan and Nathaniel that were relevant to this part of the explanation. The goal here is just to get a doc up that explains the simulants to add per timestep equation well enough to be used by engineering.

In the future I plan to make a more detailed PR that may address some or all the following depending on what people feel is useful/important:
* calculating the same value using CBR (alternate equivalent math for "epidemiology way" of thinking about it)
* add back any other text from original PR that may be worth including
* more explanation into why we specifically use the Live births by sex covariate vs other births covariates (maybe everyone understands this already but not clear to me)
* how to calculate this for age_start > 0 eg in Alzheimers
* explain units better?
* other documentation about fertility modeling in general, interpolation, or other related topics we don't currently have
* show model scale visually
* on new fertility page, discuss different types such as adding simulants, parent demography (see "types of vivarium components")
* discuss how randomness is involved
* extract general parts of Alzheimers methods and document them


docs/source/model_design/vivarium_model_components/fertility/index.rst
```diff
@@ -0,0 +1,127 @@
...
+:math:`\text{model scale} = \frac{\text{initial simulated population size}}{\int_{a=\text{age start}}^{\text{age end}} \text{GBD total population}(a) \, da}`
+
+:math:`\text{Initial simulated population size}` is the number of simulants we will initialize. 
+The integral limits indicate the age restrictions of the GBD population represented by our initialized simulants - e.g. 5 and under. 
+This can be generalized to models of populations restricted for other variables, such as sex. 
+
+.. note::
+  To use live births in the current year to determine the rate at which we should add new simulants, we must have :math:`\text{age start} = 0`.
+
+  Also note that with no age restrictions, our model scale is 1 -- we can simply use the Live births data directly.
```
aflaxman, line 98:
```suggestion
```

docs/source/model_design/vivarium_model_components/fertility/index.rst
```diff
@@ -0,0 +1,127 @@
...
+We multiply our model scale by the number of live births in the total GBD population and the fraction of a year which passes in each simulated time step 
+to get the total simulants to add per timestep, as shown in the below table.
+
+.. list-table:: Parameters
+  :header-rows: 1
+
+  * - Parameter
+    - Value
+  * - simulants_to_add_per_timestep
+    - :math:`\frac{\text{initial simulated population size}}{\int_{a=\text{age start}}^{\text{age end}} \text{GBD total population}(a) \, da} \times \text{Live births} \times \text{time step}/365`
```
aflaxman, line 109:
```suggestion
    - :math:`\frac{\text{initial simulated population size}}{\int_{a=\text{age start}}^{\text{age end}} \text{GBD total population}(a) \, da} \times \text{Live births} \times \text{time step in days}/365.25`
```

aflaxman (approved):

docs/source/model_design/vivarium_model_components/fertility/index.rst
tylerdy: https://github.com/ihmeuw/vivarium_research/pull/1735#discussion_r2255020646 thanks, fixed this - see my reply to Zeb

docs/source/model_design/vivarium_model_components/fertility/index.rst
```diff
@@ -0,0 +1,134 @@
...
+  Therefore, our model scale must reflect the total GBD population even if we are only initializing a subset of that population in our model. 
+  
+  In other words, the initialized population is different from the population that exists within the simulation scenario. 
+  The ratio of the (age-restricted) initialized population to the age-restricted GBD population gives us our model scale.
+  The ratio of the simulation scenario population to the total GBD population is also equal to the model scale 
+  (in this example the ratio would be 200 to 1000). 
+
+To represent model scale in general we can define the below equation:
+
+:math:`\text{model scale} = \frac{\text{Initialized population size}}{\int_{a=\text{age start}}^{\text{age end}} \text{GBD total population}(a) \, da}`
```
NathanielBlairStahn, line 94: I think the units of this integral don't make sense: population $\cdot$ age = person-time, not population. I.e., the integrand should be population per year or something like that instead of just population, since you're multiplying by $da$. I think it would probably be better to write it as a sum over age groups, which is how we'd actually compute it.

docs/source/model_design/vivarium_model_components/fertility/index.rst
```diff
@@ -0,0 +1,134 @@
...
+
+In a simulation where we do not model births by identifying which simulants give birth, we must instead use live births data
+to determine how many simulants to add in each time step.
+We will scale down the live births of the total GBD population by the same "model scale"  we use in our simulation.
+
+For example, imagine that country A has a GBD population of 1000, and we are modeling this total population, but only need to initialize children age 5 and under 
+(e.g. because they are the only ones susceptible to the disease we are modeling).
+Imagine there are 100 such children living in A, and we plan to initialize 20 children in our model. 
+
+In this example, our initialized population is :math:`\frac{1}{5}` the size of the age-restricted GBD population it corresponds to.
```
NathanielBlairStahn, line 74: I think it would be good to explicitly explain where you got 1/5:
```suggestion
In this example, our initialized population is :math:`\frac{1}{5}` the size of the age-restricted GBD population it corresponds to, i.e., 20 simulated children under 5 vs. 100 actual children under 5.
```
(Or switch to the bigger numbers I suggested above.)

docs/source/model_design/vivarium_model_components/fertility/index.rst
```diff
@@ -0,0 +1,134 @@
...
+
+Vivarium Modeling Strategy
+----------------------------
+
+In a simulation where we do not model births by identifying which simulants give birth, we must instead use live births data
+to determine how many simulants to add in each time step.
+We will scale down the live births of the total GBD population by the same "model scale"  we use in our simulation.
+
+For example, imagine that country A has a GBD population of 1000, and we are modeling this total population, but only need to initialize children age 5 and under 
+(e.g. because they are the only ones susceptible to the disease we are modeling).
```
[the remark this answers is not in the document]
tylerdy: I suppose thats fair, maybe I could say something like "imagine country A has a GBD population of 1000. Suppose we are modeling a disease that only affects young children, so we only need to initialize and track children age 5 years and under. That being said, this whole population will need to be represented (at a smaller scale) in the scenario we are simulating, since people still need to give birth to new stimulants."

Very open to other ideas on how I can concisely distinguish between tracked simulants and people who exist in the simulation scenario and affect it but are not tracked or "modeled".

docs/source/model_design/vivarium_model_components/fertility/index.rst
```diff
@@ -0,0 +1,134 @@
...
+Imagine there are 100 such children living in A, and we plan to initialize 20 children in our model. 
+
+In this example, our initialized population is :math:`\frac{1}{5}` the size of the age-restricted GBD population it corresponds to. 
+This means that the simulated population implied by our model scenario (which includes people over the age of 5 even though they are not initialized) is also :math:`\frac{1}{5}` 
+the size of the total unrestricted GBD population.
```
NathanielBlairStahn, line 76: Ok, I think I see what you're getting at here with considering the entire population instead of just the population under 5, but this framing is a bit confusing to me. Because we're _not_ simulating the rest of the population -- we basically never think about them when we're designing the simulation. The only reason we care about the total population here is because the "live births" covariate corresponds to the total population, so we need to know what it is in order to scale the births correctly.

So maybe just some kind of wording change would be helpful. What I find confusing is referring to the unsimulated people as the "simulated population," because they're literally not in our simulation. It's like some kind of "shadow population" or something, corresponding to the under-5 population we're actually simulating.

tylerdy: I ended up going with "model-implied" population for now since I thought that sounded OK and it matched Alzheimers docs, but I also like "total model population" and clarifying the difference between the model population and the simulation population. Didn't do that one partly bc in your other comment you said it seemed wrong to say we were modeling a population if the entire population isnt simulated, so it seemed like it might be confusing. But either way, should be easy to swap in a different term now that I have more careful definitions/descriptions.

docs/source/model_design/vivarium_model_components/fertility/index.rst
```diff
@@ -0,0 +1,149 @@
...
+----------------------------
+
+In a simulation where we do not model births by identifying which simulants give birth, we must instead use live births data
+to determine how many simulants to add in each time step.
+We will scale down the live births of the total GBD population by the same "model scale"  we use in our simulation.
+
+While we always scale down total GBD population live births when **adding** simulants,
+the number of simulants we **initialize** may be less than the scaled down total GBD population due to 
+restrictions on variables such as age.
+For example, for a child nutrition simulation, we may only simulate children under 5.
```
zmbc, line 73:
```suggestion
For example, for a child nutrition simulation, we may only initialize children under 5.
This is equivalent to initializing simulants of all ages and then immediately un-tracking simulants
older than 5, but without the unnecessary step; in other words, we can avoid initializing simulants
that are guaranteed never to be of interest to observe.
```

docs/source/model_design/vivarium_model_components/fertility/index.rst
```diff
@@ -0,0 +1,149 @@
...
+which will be useful in this discussion:
+
+  * Restricted GBD (full-scale)  
+  * Unrestricted/total GBD (full-scale) 
+  * Restricted "simulated" (model-scale)
+  * Unrestricted/total "model-implied" (model-scale)
+
+Note that we differentiate model-scale populations as either "simulated" or "model-implied" 
+depending on whether they are restricted or not. 
+"Simulated" reflects that the restiction defines the population that we create simulants for.
```
zmbc, line 85: I'm concerned that this term doesn't adequately capture that we could e.g. initialize simulants under 5 but then continue tracking them into adulthood. Clearly people over 5 in that case would be "simulated," just not _initialized_.

Though, one way to look at it then is that our population restriction is actually "cohorts born after 2018" or similar. We simulate those cohorts, but not older cohorts, so "simulated" makes sense. But I think this framing might require some notes elsewhere in the page.

docs/source/model_design/vivarium_model_components/fertility/index.rst
```diff
@@ -0,0 +1,149 @@
...
+"Model-implied" means that the lack of restriction includes people who exist in the context of the model but 
+are not simulated. For example, even if we are only simulating children under 5, people of other ages 
+are still implied to exist, otherwise there would be no one to give birth to new simulants!
+
+The model scale will be equal to the ratio between the restricted simulated 
+population and the restricted GBD population, as well as to the ratio between the 
+unrestricted model-implied population and the unrestricted GBD population. 
+
+For example, imagine that country A has a GBD population of one million ("unrestricted/total GBD population),
```
zmbc, line 94:
```suggestion
For example, imagine that country A has a GBD population of one million ("unrestricted/total" GBD population), 
```

docs/source/model_design/vivarium_model_components/fertility/index.rst
```diff
@@ -0,0 +1,149 @@
...
+"Initialized simulated population size" is the age-restricted, simulated population size -- the number of simulants we will initialize. 
+The summation limits indicate the age restrictions of the GBD population represented by our initialized simulants - e.g. under 5. 
+This can be generalized to models of populations restricted for other variables, such as sex. 
+
+.. note::
+  To use live births in the current year to determine the rate at which we should add new simulants, we must have :math:`\text{age start} = 0`.
+
+  Also note that with no age restrictions, the simulated population size in the numerator 
+  will be the same as the model-implied population size, 
+  and the integral value will simply be the total GBD population.
```
zmbc, line 119:
```suggestion
  and the sum will simply be the total GBD population. 
```

zmbc (approved):

docs/source/model_design/vivarium_model_components/fertility/index.rst
zmbc, line 85:
```suggestion
"Simulated" reflects that the restriction defines the population that we create simulants for.
```

PR #1744, opened 2025-08-11
Add model 13.3 V&V results to table

zmbc:


aflaxman (approved):

alibow:
This is looking much improved to me! Are you thinking that the late neonatal rates are still suspect because they have more variation? I think it makes sense to keep looking for bugs, but in the meantime I wonder if it would be worthwhile to launch another run with a larger population size -- the LNN group may just be noisier than ENN given that the mortality rate is so much lower so there are smaller counts.

zmbc:
The fact that all six sex-locations are low for LNN preterm birth risk seems unlikely to me to occur by chance. "Other causes" being high is a bit harder to tell, though.

Overall I'd be in favor of running with a larger population, since I think that is cheap to do.

PR #1760, opened 2025-08-25
Update to neonatal mortality model to account for causes affected vs. unaffected by LBWSG

alibow:
Ticket: https://jira.ihme.washington.edu/browse/SSCI-2363

Draft PR where we discussed this update previously: https://github.com/ihmeuw/vivarium_research/pull/1594

Please let me know if you have any questions or suggestions! I would like to wait until Zeb feels that he's finished up neonatal mortality de-bugging before we merge and make a run request, but I think we can still go through review in the meantime.


zmbc (approved):
Thanks Ali!

aflaxman (approved):

zmbc:
@alibow just curious what the status of this PR is?

docs/source/models/causes/neonatal/index.rst
```diff
@@ -121,13 +121,17 @@ Assumptions and Limitations
...
 
-- Mortality risk values very close to 1 for the individuals with the highest risk LBWSG exposures
+- We "cap" LBWSG RR values at a certain value in an attempt to eliminate the occurrence of individual all-cause mortality risk values greater than 1 on in our simulation (and therefore avoiding an associated underestimation of neonatal mortality) while also maintaining:
 
-However, in this implementation, we do not consider how modeled interventions may further increase individual-level mortality risk beyond the modifications from the LBWSG risk factor, so it is possible that we continue to have some mortality risk values greater that 1 in our simulation. 
+  - The relative difference in mortality risk values between LBWSG exposures that are not capped, and
+
+  - Mortality risk values very close to 1 for the individuals with the highest risk LBWSG exposures
+
+  However, in this implementation, we do not consider how modeled interventions may further increase individual-level mortality risk beyond the modifications from the LBWSG risk factor, so it is possible that we continue to have some mortality risk values greater that 1 in our simulation.
```
zmbc: > modeled interventions may further increase individual-level mortality risk

I don't understand this piece, shouldn't interventions be _decreasing_ risk?

docs/source/models/causes/neonatal/index.rst
```diff
@@ -121,13 +121,17 @@ Assumptions and Limitations
 
 - Preterm must be split into with and without respiratory distress syndrome (RDS) outside of GBD, since GBD does not distinguish preterms deaths with and without RDS.
 
-We "cap" LBWSG RR values at a certain value in an attempt to eliminate the occurance of individual all-cause mortality risk values greater than 1 on in our simulation (and therefore avoiding an associated underestimation of neonatal mortality) while also maintaining:
+- We assume the LBWSG RR values represent both the correlation between LBWSG exposure and all-cause neonatal mortality at baseline as well as the causal impact of LBWSG on cause-specific mortality for causes affected by LBWSG
 
-- The relative difference in mortality risk values between LBWSG exposures that are not capped, and
+- We assume that the LBWSG exposure distribution prior to applying intervention effects is the same as the LBWSG exposure distribution at baseline. These distributions will slightly diverge due to the baseline calibration of the :ref:`oral iron intervention <oral_iron_antenatal>` (and any other future modeled interventions that are present at baseline and affect LBWSG exposure), but should be roughly similar.
```
zmbc: I think this limitation does not spell out how/where we actually assume this. It is in calculating the LBWSG PAFs, correct?

docs/source/models/causes/neonatal/index.rst
```diff
@@ -235,28 +239,49 @@ and for a given cause of death:
...
+
+  - apply the change in mortality risk to the causes affected by LBWSG associated with intervention impact on LBWSG exposures that is in line with GBD assumptions,
+  - and also reflect the expected correlation between LBWSG exposures and mortality at baseline (improving the accuracy of background mortality rates across the LBWSG exposure distribution, which may influence our estimates of impact for interventions that are targeted by LBWSG exposure).
+
+To avoid the need to write out data between simulated scenarios for ease of implementation, we will assume that pre-intervention modified LBWSG exposure represents baseline LBWSG exposure.
```
zmbc: Let's mention IFA here, like we did above, because it took me a bit to understand why this wasn't exactly correct.

docs/source/models/causes/neonatal/index.rst
```diff
@@ -235,28 +239,49 @@ and for a given cause of death:
...
+- :math:`\text{PAF}_{\text{LBWSG}}` is the population attributable fraction for LBWSG, 
+- :math:`\text{RR}_{\text{BW},\text{GA}}` is the relative mortality risk for a specific birth weight :math:`\text{BW}` and gestational age :math:`\text{GA}` according to the exposure values in the current scenario, and 
+- :math:`\text{RR}_{\text{BW}^0,\text{GA}^0}` is the relative mortality risk for a specific birth weight :math:`\text{BW}` and gestational age :math:`\text{GA}` according to the exposure value prior to the application of any intervention effects on LBWSG exposure.
+
+To obtain the ACMRisk for a specific simulant (:math:`\text{ACMRisk}_i`), we subtract off the *population* CSMRisks for each modeled subcause :math:`k` for the birth weight and gestational age of the simulant, and then add back in the (potentially pipeline-modified) *individual* CSMRisks for the specific simulant, which might differ from baseline due to intervention coverage:
```
zmbc: The "population" in the first part of this sentence is hard for me to follow. It's already specific to the birth weight and gestational age, so this is "the population of potential people at exactly this simulant's birth weight and age"? That feels like kind of a stretch.

docs/source/models/causes/neonatal/index.rst
zmbc: On a bit more thought, this is a confusing limitation that will likely trip us up again in the future, and I would like to chat with @hussain-jafari about how much more difficult it would be to implement this correctly vs with this limitation (I believe both are a bit awkward in Vivarium, but maybe the "baseline" way isn't significantly more awkward than the "pre-intervention" way?)

docs/source/models/causes/neonatal/index.rst
alibow: No, this should remain as written. The GBD risk curves for hemoglobin are non-monotonic for BW and GA outcomes, so there are instances where receiving a hemoglobin-increasing interventions could lead to decreases in BW or GA at the individual level even though that is not expected to be the case on average. (see this notebook that has some plots demonstrating this relationship: https://app.reviewnb.com/ihmeuw/vivarium_gates_mncnh/blob/main/src/vivarium_gates_mncnh/data/hemoglobin_effects/function_tester.ipynb/ ). This is definitely contrary to the typical intervention mental model though so it would probably be an improvement to note this in that sentence.

docs/source/models/causes/neonatal/index.rst
alibow: I think there are a couple of issues here that are downstream of this:
- We have slightly different LBWSG exposure distributions between our PAF calculation and baseline simulated population given that we consider the impact of interventions on LBWSG in the baseline simulated population but not PAF calculation
- Our baseline simulated population will slightly diverge from the GBD exposure distribution

docs/source/models/causes/neonatal/index.rst
alibow: Yeah, agreed. Maybe just delete "*population*" and then add * emphasis on the "for the birth weight and gestational age of the simulant" part instead?

docs/source/models/causes/neonatal/index.rst
```diff
@@ -235,28 +239,49 @@ and for a given cause of death:
...
 
 .. math::
     \begin{aligned}
-    \text{ACMRisk}_i &= \text{ACMRisk}_{\text{BW}_i,\text{GA}_i} - \sum_k \text{CSMRisk}_{\text{BW}_i,\text{GA}_i}^{k}
-    + \sum_k \text{CSMRisk}_{i}^{k},
+    \text{ACMRisk}_i &= \text{ACMRisk}_{\text{BW}_i,\text{GA}_i}^{\text{BW}_i^0,\text{GA}_i^0} - \sum_{k \in \text{affected}} \text{CSMRisk}_{\text{BW}_i,\text{GA}_i}^{k} \\
+    & - \sum_{k \in \text{unaffected}} \text{CSMRisk}_{\text{BW}_i^0,\text{GA}_i^0}{k}
```
zmbc: This doesn't seem right to me:

- We still need to add in the individual's CSMRisks like in the line that was deleted
- There are no causes which are both modeled and unaffected, which I think was the intended group to be summed over here

PR #1761, opened 2025-08-26
vnv model 3

aflaxman:


docs/source/models/concept_models/vivarium_alzheimers/concept_model.rst
```diff
@@ -532,10 +539,11 @@ scenario, and input draw.
       aggregated results.
 
       * Everything from 2.0, except use FHS values for ACMR in the total mortality calculation
+      * Verify that ACMR decreases slightly from 2025 to 2050 and then levels off
```
NathanielBlairStahn, line 542: Do we actually mean ACMR here, or is this ACMR - CSMR + EMR? Or could that quantity also be called "ACMR" in the context of this sim where the entire population consists of people with AD?

Do you know why the mortality decreases and then levels off? Is this what we expect from the FHS data?

docs/source/models/concept_models/vivarium_alzheimers/concept_model.rst
```diff
@@ -532,10 +539,11 @@ scenario, and input draw.
...
       * Since there are so many (age groups, years, locations, sex) combinations that might be tested, it will be good enough to confirm that new simulant counts and total mortality rates line up for 2030, 2060, and 2090, and for two locations.
-    -
-    -
-
+    - * The number of new simulants entering the sim matches the target number, which leads to a prevalence counts higher than estimated by GBD/FHS, but closer than in Model 2.
```
NathanielBlairStahn, line 544: Does FHS estimate AD prevalence? Should we be using that data?

docs/source/models/concept_models/vivarium_alzheimers/concept_model.rst
aflaxman: I will change to ACMR - CSMR + EMR, which is more precise.  Mortality decreases (as I expected) because people are getting a little healthier over time, I suspect it is mostly because of less tobacco use.

docs/source/models/concept_models/vivarium_alzheimers/concept_model.rst
```diff
@@ -531,11 +532,12 @@ scenario, and input draw.
       group and sex, but it may be more prudent to start by looking at
       aggregated results.
 
-      * Everything from 2.0, except use FHS values for ACMR in the total mortality calculation
+      * Everything from 2.0, except use FHS values for ACMR in the overall mortality (ACMR - CSMR + EMR) vs artifact comparison
+      * Verify that ACMR decreases slightly from 2025 to 2050 and then levels off
```
aflaxman, line 536:
```suggestion
      * Verify that (ACMR - CSMR + EMR) decreases slightly from 2025 to 2050 and then levels off
```

docs/source/models/concept_models/vivarium_alzheimers/concept_model.rst
zmbc: > I am explaining it to myself as inconsistency between incidence, prevalence, and mortality

Or, violation of our assumption that other-causes mortality is uncorrelated with AD prevalence!

PR #1762, opened 2025-08-26
Alzheimer's: Add data tables to cause model with BBBM and MCI

NathanielBlairStahn:
Jira ticket: [SSCI-2389](https://jira.ihme.washington.edu/browse/SSCI-2389)

Fill out Alzheimer's cause model for Model 4 runs as follows:

- Add state and transition data tables, including:

  - Define relative initial prevalences of cause states in terms of durations of each cause state
  - Specify how to use time-dependent hazard function to transition from BBBM-AD to MCI
  - Calculate cause state prevalences in total population when including susceptible simulants

- Add data values and sources table, including:

  - Define hazard function for transitioning from BBBM-AD to MCI
  - Define disability weight of MCI
  - Calculate duration of AD-dementia cause state
  - Specify using forecasted data for population and mortality rate

- Add a to-do to further explain the MCI disability weight calculation

### DIFF
--- a/docs/source/models/causes/neonatal/preterm_birth.rst
+++ b/docs/source/models/causes/neonatal/preterm_birth.rst
@@ -314,7 +314,7 @@ Data Tables
       - 
     * - lbwsg_birth_prevalence
       - Birth prevalence of low birthweight and short gestation risk factor
-      - GBD with post-processing: rei_id = 339, then remove the extraneous category and rescale prevalence :ref:`as described here <rescaling_lbwsg_exposure_data_pulled_from_gbd_2019>`.
+      - GBD with post-processing: rei_id = 339, then remove the extraneous category and rescale prevalence :ref:`as described here <rescaling_lbwsg_exposure_data_pulled_from_gbd_2021>`.
       - 
     * - :math:`f_\text{preterm w RDS}`
       - fraction of preterm deaths with RDS
--- a/docs/source/models/concept_models/vivarium_alzheimers/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_alzheimers/concept_model.rst
@@ -384,6 +384,7 @@ scenario, and input draw.
     - Default
     - Default
 
+
 5.2 V & V Tracking
 ------------------------
 
@@ -531,11 +532,12 @@ scenario, and input draw.
       group and sex, but it may be more prudent to start by looking at
       aggregated results.
 
-      * Everything from 2.0, except use FHS values for ACMR in the total mortality calculation
+      * Everything from 2.0, except use FHS values for ACMR in the overall mortality (ACMR - CSMR + EMR) vs artifact comparison
+      * Verify that (ACMR - CSMR + EMR) decreases slightly from 2025 to 2050 and then levels off
       * Since there are so many (age groups, years, locations, sex) combinations that might be tested, it will be good enough to confirm that new simulant counts and total mortality rates line up for 2030, 2060, and 2090, and for two locations.
-    -
-    -
-
+    - * The number of new simulants entering the sim matches the target number, which leads to a prevalence counts higher than estimated by GBD/FHS, but closer than in Model 2.
+    - https://github.com/ihmeuw/vivarium_research_alzheimers/blob/32e7d3d44f540a9b9620b21b5a137f626631475c/verification_and_validation/2025_08_25b_model3.0_vv.ipynb
+  
 .. list-table:: Outstanding model verification and validation issues
   :header-rows: 1
 
@@ -547,7 +549,7 @@ scenario, and input draw.
     - Thought to be due to incorrect disability weight aggregation
     - Will be updated when we add severity levels, recheck then
     - Model 5
-  * - Total simulation population increasing in model 2
+  * - Total simulation population increasing in model 3
     - Thought to be due to GBD mismatch in mortality and incidence
-    - Population and ACMR's change in model 3 to use FHS data, reassess then
-    - Model 3
+    - Review again when we split AD and OD 
+    - Model 6
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
@@ -299,6 +299,7 @@ Pregnancy component
       - * :ref:`Hemoglobin risk exposure <2023_hemoglobin_exposure>`
         * :ref:`Oral iron supplementation intervention (IFA/MMS) <oral_iron_antenatal>`
         * :ref:`IV iron intervention <intervention_iv_iron_antenatal_mncnh>`
+        * :ref:`Anemia screening intervention <anemia_screening>`
       - New wave II module
     * - :ref:`Pregnancy II <2024_vivarium_mncnh_portfolio_pregnancy_module>`
       - * LBWSG category propensity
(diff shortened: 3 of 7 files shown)


docs/source/models/causes/alzheimers/presymptomatic_and_mci_gbd_2021/index.rst
```diff
@@ -178,10 +184,251 @@ Cause Model Diagram
...
+  * - :math:`\text{DW}_\text{c543}`
+    - Average disability weight of AD-dementia
+    - :math:`\sum_\limits{s\in \text{sequelae_c543}}
+      \text{DW}_s \cdot \text{prevalence}_s`
+    - Prevalence-weighted average disability weight over sequelae,
+      computed automatically by Vivarium Inputs. Used to calculate
+      YLDs.
+  * - :math:`\text{DW}_\text{MCI}`
+    - Disability weight of mild cognitive impairment
+    - ?
```
aflaxman, line 398: I have a notebook that calculates this, and the mean value is in it

docs/source/models/causes/alzheimers/presymptomatic_and_mci_gbd_2021/index.rst
```diff
@@ -178,10 +184,251 @@ Cause Model Diagram
...
+    - The time at which the simulant enters the BBBM-AD state
+    - Random variable for each simulant
+    -
+  * - :math:`\alpha`, :math:`\lambda`
+    - Shape and rate parameters, respectively, of gamma distribution for
+      waiting time :math:`T` in BBBM-AD
+    - * :math:`\alpha = 468.75`
+      * :math:`\lambda = 125`
+    - Chosen so that :math:`P(3.5 < T < 4) \approx 0.9` because client
+      said, "The BBBM+ state lasts about 3.5--4 years before
```
aflaxman, line 414: I wonder if they actually want something less concentrated than this; the parameters they have selected will imply a small effect for their treatment, and will make it relatively insensitive to *when* in the BBBM-AD period the treatment begins; perhaps they did not realize this.  A sensitivity analysis on this distribution could be something they value!

aflaxman (approved):

docs/source/models/causes/alzheimers/presymptomatic_and_mci_gbd_2021/index.rst
```diff
@@ -178,10 +184,251 @@ Cause Model Diagram
...
+      transitioning to MCI."
+  * - :math:`\Delta_\text{BBBM}`
+    - Average duration of BBBM-presymptomatic AD
+    - :math:`\alpha / \lambda`
+    - Mean of gamma distribution for waiting time :math:`T` in BBBM-AD
+  * - :math:`\Delta_\text{MCI}`
+    - Average duration of MCI due to AD
+    - 3.25 years
+    - Value from [[cite source]], assuming a constant hazard rate.
```
aflaxman, line 423: https://doi.org/10.1007/s40120-021-00272-1

docs/source/models/causes/alzheimers/presymptomatic_and_mci_gbd_2021/index.rst
NathanielBlairStahn: Yeah, I have the value in an email, but I haven't gotten around to looking it up yet. Do you want to share your notebook so we can post it on GitHub and I can link to it?

tylerdy: Can you link it? Will we have draws for this?

docs/source/models/causes/alzheimers/presymptomatic_and_mci_gbd_2021/index.rst
```diff
@@ -178,10 +184,251 @@ Cause Model Diagram
...
+      Corresponds to an annual probability of 0.735 of staying in MCI-AD
+      (or returning to asymptomatic), since :math:`\exp(-1 / 3.25)
```
tylerdy, line 425: > (or returning to asymptomatic)

We're assuming this is zero right? Does that require any adjustment of the value?

docs/source/models/causes/alzheimers/presymptomatic_and_mci_gbd_2021/index.rst
```diff
@@ -178,10 +184,251 @@ Cause Model Diagram
       long a simulant has been in the BBBM-Presymptomatic state, not a
       constant hazard like we usually use
   * - i_AD
-    - Alzheimer's incidence hazard
-    - Incidence hazard of Alzheimer's disease
+    - AD dementia incidence hazard
+    - Incidence hazard of Alzheimer's disease dementia
     - We will define this as a constant hazard rate for simulants in
       MCI-AD
```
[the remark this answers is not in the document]
NathanielBlairStahn: Yes

docs/source/models/causes/alzheimers/presymptomatic_and_mci_gbd_2021/index.rst
aflaxman: Yes, this notebook shows how to calculate them: https://github.com/ihmeuw/vivarium_research_alzheimers/blob/4d5dde0b74eb09ea997af7c2de88b81670ba7d61/2025_08_03a_alz_dw_explore.ipynb

```
mean     0.021
2.5%     0.013
97.5%    0.032
```

NathanielBlairStahn: I think we'll need to generate draws for this. And we should make sure that for each draw, the MCI DW is not greater than the mild dementia DW. Based on Abie's email, the means and 95% UIs are:

* MCI: 0.02 (0.01, 0.03)
* mild dementia: 0.07 (0.04, 0.10)

So if we sample independently, it's probably unlikely that we'd get a draw of MCI that's higher than mild dementia, but maybe not impossible...

NathanielBlairStahn: Oh, it looks like Abie's notebook already computes things at the draw level, so I think that means we shouldn't have to worry about overlaps since the DW draws are already properly correlated to avoid this.

docs/source/models/causes/alzheimers/presymptomatic_and_mci_gbd_2021/index.rst
```diff
@@ -168,20 +176,308 @@ Cause Model Diagram
...
+    - Entrance prevalence
+    - Excess mortality rate
+    - Disability weight
+  * - S
+    - 0
+    - 0
+    - 0
+    - 0
+  * - BBBM-AD
+    - :math:`\Delta_\text{BBBM} / \Delta_\text{(all AD states)}`
```
tylerdy, line 225: Thought of another question- why are the initial prevalences calculated in terms of durations?

NathanielBlairStahn: Because we don't have a better way to calculate them. Unless you know of some data that can inform this...? But if our durations are correct (they should all be close, but BBBM will be a little off, at least, because we're not accounting for deaths, as Rajan pointed out) and the model reaches equilibrium, then the relative durations should equal the relative prevalences. Basically because person-time in state X = (number of people in X) $\cdot$ (duration of state X), and average prevalence of X is calculated as the ratio of person-time in X to total person time. I think.

PR #1770, opened 2025-09-05
Model 3.1

tylerdy:
Add model 3.1 runs docs for forecasted population draws.


aflaxman (approved):
Can you think of a V&V strategy that will prove this has worked?

PR #1783, opened 2025-09-15
Request model run 16.5

pletale:
Per this discussion - https://ihme.slack.com/archives/C02944X688L/p1757956156625949 

Model run 16.5 will add low hemoglobin RR on depressive disorders so that we can include postpartum depression in our custom PAF calculations.

### DIFF
--- a/docs/source/models/causes/alzheimers/presymptomatic_and_mci_gbd_2021/index.rst
+++ b/docs/source/models/causes/alzheimers/presymptomatic_and_mci_gbd_2021/index.rst
@@ -240,15 +240,9 @@ in the :ref:`Alzheimer's population model
     - emr_c543
     - :math:`\text{DW}_\text{c543}`
 
-**Note:** The variable :math:`\Delta_\textsf{X}` denotes the average
-duration in cause state X, as defined in the :ref:`data values and
-sources table below
-<2021_cause_alzheimers_presymptomatic_mci_data_sources_table>`. If
-necessary, for age groups in which :math:`\Delta_\text{(all AD states)}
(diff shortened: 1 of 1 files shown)


docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -1458,6 +1458,13 @@ Default stratifications to all observers should include scenario and input draw.
     - Default
     - Default, but with preterm birth status stratification of neonatal deaths observers
     - Default
+  * - 16.5
+    - Artifact update to include low hemoglobin RR for depressive disorders. No need to actually re-run the model, we just need this RR value for our custom PAF calculations.
```
alibow, line 1462:
```suggestion
    - Inclusion of low hemoglobin RR for depressive disorders. No need to actually re-run the model, we just need this RR value active in the interactive context for our custom PAF calculations.
```

alibow (approved):
Let's plan to do some V&V on the RRs before using them in the PAF calculation. Here is an existing notebook that ran V&V on the hemoglobin RRs for sepsis and hemorrhage that you can use for the PPD RRs: https://github.com/ihmeuw/vivarium_research_mncnh_portfolio/blob/main/verification_and_validation/model_13.1_interactive_simulation_hemoglobin.ipynb

We'll only need to go as far as cell 23 for this though (the rest of the notebook checks that the RRs are being applied appropriately, and the PPD RRs won't be applied to anything yet)

aflaxman (approved):

PR #1790, opened 2025-09-16
Model 16.4 neonatal mortality V&V updates and 16.6 run request

zmbc:
Proposed a large run for neonatal mortality V&V as model 16.6, with observers for the continuous risk values, including statistics about how often (and by how much) other causes mortality goes negative.


docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -918,6 +918,61 @@ Default stratifications to all observers should include scenario and input draw.
...
+    - * Child sex
+      * Child age group
+      * Delivery facility type
+      * CPAP availability
+      * Antibiotics availability
+      * Probiotics availability
+      * Corticosteroid coverage
+      * ACS eligibility (dichotomous, 'eligible' if believed gestational age is between 26-33 weeks, 'not eligible' if gestational age is outside of this range)
+      * Preterm status (dichotomous at 37 weeks)
+    - Observe the following statistics about the CSMRisk pipelines just *before* they are normalized
```
alibow, line 948: I think a little more expansion on exactly what quantity this is would be helpful for interpretation - maybe linking to the relevant quantity on the [neonatal mortality document](https://vivarium-research.readthedocs.io/en/latest/models/causes/neonatal/index.html) and/or the quantity in the sim implementation?

I think this also applies to the parameter above and below as well!

alibow (approved):

PR #1794, opened 2025-09-18
Add conceptual diagram and description to residual maternal disorders cause model doc

alibow:
Updates as discussed in today's MNCNH RT/SE check-in. I decided not to make the equivalent updates to the [partial term pregnancy cause model doc](https://vivarium-research.readthedocs.io/en/latest/models/causes/maternal_disorders/gbd_2021_mncnh/partial_term_causes.html) because there is actually underlying incidence parameters for the included subcauses that is definitionally 1 for that cause model as written. Let me know if anyone feels strongly about making similar changes to that page though.


docs/source/models/causes/maternal_disorders/gbd_2021_mncnh/residual_maternal_disorders.rst
```diff
@@ -202,6 +199,58 @@ Assumptions and Limitations
...
+        may be a live birth or a still birth, to be determined in the
+        next step of the :ref:`pregnancy model
+        <other_models_pregnancy_closed_cohort_mncnh>`)
+    * - fr (fatality rate)
+      - The rate of death due to residual maternal disorders among all pregnancies with full term outcomes (live or still births) 
+
+While the above diagram represents the conceptual aims of the residual maternal disorders cause model, there are some convenient adjustments we can make to this diagram so that it continues to achieve the aims of our cause model while also achieving compatibility with the existing implementation of the `MaternalDisorders component <http://github.com/ihmeuw/vivarium_gates_mncnh/blob/main/src/vivarium_gates_mncnh/components/maternal_disorders.py>`_ used in the MNCNH simulation that has been used to model other maternal disorders subcauses such as :ref:`maternal sepsis <2021_cause_maternal_sepsis_mncnh>`. Specifically, although our modeling strategy for maternal disorders does not involve modeling incident cases, we can implement :code:`MaternalDisorders` component model as represented in the diagram below with an incidence risk (ir) value equal to 1.
```
zmbc, line 250:
```suggestion
While the above diagram represents the conceptual aims of the residual maternal disorders cause model, there are some convenient adjustments we can make to this diagram so that it continues to achieve the aims of our cause model while also achieving compatibility with the existing implementation of the `MaternalDisorder component <http://github.com/ihmeuw/vivarium_gates_mncnh/blob/main/src/vivarium_gates_mncnh/components/maternal_disorders.py>`_ used in the MNCNH simulation that has been used to model other maternal disorders subcauses such as :ref:`maternal sepsis <2021_cause_maternal_sepsis_mncnh>`. Specifically, although our modeling strategy for maternal disorders does not involve modeling incident cases, we can implement a :code:`MaternalDisorder` component model as represented in the diagram below with an incidence risk (ir) value equal to 1.
```

zmbc (approved):
Looks good!

I'm a bit conflicted on whether we should do the same on the partial term page. I see your logic (we _are_ actually tracking incident cases) but we aren't really doing it _within_ the cause model, so the IR is 1, which could be confusing. Maybe we don't add all this there but include a more prominent note about it

PR #1795, opened 2025-09-18
Iv iron notes

alibow:


docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -2096,6 +2117,7 @@ Default stratifications to all observers should include scenario and input draw.
         * Birth outcome rates
 
       * Confirm expected effects of IV iron on birth weight, gestational age, and birth outcome rates using the interactive simulation
+      * Confirm list of draws has been updated (will need to look in sim repo rather than results as we don't routinely run all 20 draws)
```
zmbc, line 2120: This draw is only listed in a comment for this reason:

https://github.com/ihmeuw/vivarium_gates_mncnh/blob/d3ae2392ec9631b3d74fe0ef802d61752c25c05b/src/vivarium_gates_mncnh/model_specifications/branches/scenarios.yaml#L7

But still worth calling out since it would be bad if that comment got out of sync. Perhaps even better would be to update the comment to link to these docs :)

zmbc (approved):

PR #1804, opened 2025-09-30
Model 7.0 (treatment) docs

tylerdy:


docs/source/models/intervention_models/alzheimers/hypothetical_treatment.rst
```diff
@@ -30,4 +30,232 @@
...
+        wane [label="Waning treatment effect LONG (update each step) [9 y]"]
+
+        treat_short [label = "Full treatment effect SHORT [6 mo]"]
+        wane_short [label = "Waning treatment effect SHORT (update each step) [2.5 y]"]
+
+
+        el -> test [label = "tested"]
+        el -> el [label = "not tested"]
+
+        test -> pos  [label = "(90%)", style=dashed]
```
stevebachmeier, line 105: So everyone who starts treatment either does it for 6 months OR the full 5 years? Nothing in between?

docs/source/models/intervention_models/alzheimers/hypothetical_treatment.rst
```diff
@@ -30,4 +30,232 @@
...
+        wane -> no_treat
+    }
+
+The diagram above illustrates how a simulant should progress through the various testing and treatment related 
+states defined by the client. Each simulant may transition to a new state on each time step. 
+
+Most states have a fixed duration (a multiple of the 
+time step length) where simulants will transition after :math:`\text{duration} / \text{time step}` time steps. 
+The duration is marked in the state node in brackets eg [6 mo]. As desribed in the :ref:`testing intervention <alzheimers_testing_intervention_bbbm>`, 
+some simulants in the `BBBM test eligible` state may transition to tested immediately (low propensity value), some may always self-transition
```
stevebachmeier, line 129: by "self-transition" you just mean don't get tested? (i.e. eligible -> eligible)

docs/source/models/intervention_models/alzheimers/hypothetical_treatment.rst
```diff
@@ -30,4 +30,232 @@
...
+transition (high propensity value), and some may self-transition for some number of time steps but eventually transition to tested
+as a result of the time-specific testing rate increasing.
+
+Some states have zero duration, illustrated with a dashed box (rather than the solid ovals for states with nonzero durations). 
+Transitions from a state with zero duration are illustrated with a dashed line. If a simulant transitions to a zero-duration state 
+on a time step, they should also immediately continue to the next state during that same time step, as a part of the same transition.
+For example, a simulant in `BBBM test eligible` who is tested and moves to `BBBM test received` would then immediately move to one of 
+that state's two sinks, and might even move to the `In treatment` state during the same transition/ time step.
```
stevebachmeier, line 137: to clarify, a simulant who lands in the test positive sink would _definitely_ immediately move into either in treatment or no treatment, right?

But simulants who are in the test negative sink would just hang out there for 3 years until they're eligible again?

aflaxman:
Correct, and it would be valuable to count the number of people who go just 6 months. I think there is either a request for an observer of this or there will be one coming soon.

docs/source/models/intervention_models/alzheimers/hypothetical_treatment.rst
tylerdy: Yes, exactly

docs/source/models/intervention_models/alzheimers/hypothetical_treatment.rst
```diff
@@ -30,4 +30,232 @@
...
+.. list-table:: Data values and sources
+  :widths: 15 15 30 15
+  :header-rows: 1
+
+  * - Variable
+    - Definition
+    - Source or value
+    - Notes
+  * - :math:`\text{prop}_I`
+    - Simulant lifetime testing "initiation propensity"
```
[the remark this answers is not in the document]
tylerdy: Sorry, this should say treatment initiation propensity, not testing. This should be a different propensity from the testing initiation propensity from the testing model. Will update this.

stevebachmeier (approved):

docs/source/models/intervention_models/alzheimers/hypothetical_treatment.rst
```diff
@@ -30,4 +30,234 @@
...
+.. list-table:: Affected Outcomes
+  :widths: 15 15 15 15
+  :header-rows: 1
+
+  * - Outcome
+    - Effect
+    - Modeled?
+    - Note (ex: is this relationship direct or mediated?)
+  * - Hypothetical treatment effect
+    - Slows progression to MCI (time-dependent effect size)
```
NathanielBlairStahn, line 73: The affected outcome should be the hazard rate for the transition from BBBM to MCI, and the effect should be multiplying the hazard rate by the hazard ratio. See these [docs for treatment of neonatal sepsis](https://vivarium-research.readthedocs.io/en/latest/models/intervention_models/neonatal/antibiotics_intervention.html) for an example.

docs/source/models/intervention_models/alzheimers/hypothetical_treatment.rst
```diff
@@ -30,4 +30,234 @@
...
+
+
+.. list-table:: Affected Outcomes
+  :widths: 15 15 15 15
+  :header-rows: 1
+
+  * - Outcome
+    - Effect
+    - Modeled?
+    - Note (ex: is this relationship direct or mediated?)
```
NathanielBlairStahn, line 71:
```suggestion
    - Note
```
I think we can get rid of the parenthetical remark since it's just an example for the template.

docs/source/models/causes/alzheimers/presymptomatic_and_mci_gbd_2021/index.rst
```diff
@@ -295,9 +297,10 @@ in cause state X, as defined in the :ref:`data values and sources table below
   * - i_MCI
     - BBBM-AD
     - MCI-AD
-    - :math:`h_\text{MCI}(t - T_\text{BBBM})`, where :math:`t` is the
-      current time in the simulation, and :math:`T_\text{BBBM}` is the
-      time the simulant entered the BBBM-AD state
+    - :math:`h_\text{MCI}(t - T_\text{BBBM}) * R_h`, where :math:`t` is the
+      current time in the simulation, :math:`T_\text{BBBM}` is the
+      time the simulant entered the BBBM-AD state, and :math:`R_h` is the 
+      :ref:`effect hazard ratio <alzheimers_intervention_treatment_data_table>`
```
NathanielBlairStahn, line 303: I think this formula including $R\_h$ should go on the intervention page rather than the cause model page. My main reason for thinking this is "[separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns)." When we're writing a cause model, we don't want to have to think about every possible intervention that can affect it, so my philosophy is generally that the cause model page should describe how the cause model works in the _baseline_ scenario, and each intervention page will describe how the intervention modifies the cause model in the appropriate intervention scenario(s). In more complicated simulations, there could be multiple interventions affecting the same cause, and it would get quite unwieldy trying to describe all the different effects on the cause model page.

I don't know if we've really discussed this explicitly as a team, so I'm open to other ways of thinking about it if @alibow @zmbc @aflaxman or @SylLutze have anything to say.

docs/source/models/intervention_models/alzheimers/hypothetical_treatment.rst
```diff
@@ -30,4 +30,234 @@
...
+  * - Variable
+    - Definition
+    - Source or value
+    - Notes
+  * - :math:`\text{prop}_I`
+    - Simulant lifetime treatment "initiation propensity"
+    - Drawn uniformly from :math:`[0,1)`
+    - Lower value means more likely to initiate testing. Independent from testing propensities.
+  * - :math:`I`
+    - Time- and location-specific testing initiation rate
```
NathanielBlairStahn, line 158:
```suggestion
    - Time- and location-specific treatment initiation rate
```
Is this what you mean?

docs/source/models/intervention_models/alzheimers/hypothetical_treatment.rst
```diff
@@ -30,4 +30,234 @@
...
+
+Some states have zero duration, illustrated with a dashed box (rather than the solid ovals for states with nonzero durations). 
+Transitions from a state with zero duration are illustrated with a dashed line. If a simulant transitions to a zero-duration state 
+on a time step, they should also immediately continue to the next state during that same time step, as a part of the same transition.
+For example, a simulant in `BBBM test eligible` who is tested and moves to `BBBM test received` would then immediately move to one of 
+that state's two sinks, and would even move directly to another state during the same transition/ time step on a positive test.
+
+Below are tables with details on how to model these states and transitions, and necessary data values. 
+The value of :math:`i_{MCI}` in the :ref:`cause model <2021_cause_alzheimers_presymptomatic_mci_transition_data_table>` is now updated
+to be equal to :math:`h_{adj} = h_{MCI} * R_h`.
```
NathanielBlairStahn, line 141:
```suggestion
to be equal to :math:`h_{adj} = h_{MCI} \cdot R_h`.
```
If we're going to use $\LaTeX$, we might as well use a standard multiplication symbol: either $\cdot$ (`\cdot`) or $\times$ (`\times`), or juxtaposition with no symbol in between.

docs/source/models/causes/alzheimers/presymptomatic_and_mci_gbd_2021/index.rst
tylerdy: Makes sense to me, I didn't think about the cause model being implicitly about the baseline but that seems like a good way to do it. I can leave it how it was and add a note that it changes in the intervention scenarios.

docs/source/models/intervention_models/alzheimers/hypothetical_treatment.rst
```diff
@@ -30,4 +30,236 @@
...
+    - Notes
+  * - :math:`\text{prop}_I`
+    - Simulant lifetime treatment "initiation propensity"
+    - Drawn uniformly from :math:`[0,1)`
+    - Lower value means more likely to initiate testing. Independent from testing propensities.
+  * - :math:`I`
+    - Time- and location-specific treatment initiation rate
+    - Lilly: "The percent of patients with a positive BBBM test who initiate treatment will vary by location and over time – but will not vary by age or sex. In the US: 30% of eligible patients initiate (constant 2030-2100); Japan: 80% of eligible patients initiate (constant 2030-2100); all other countries: 40% of eligible patients initiate in 2030, increasing linearly to 70% by 2035, remaining constant at 70% until 2100.""
+    - 
+  * - :math:`h_{adj}`
```
aflaxman, line 163: Request from engineers that h_adj and h_MCI not be in the table --- they are perhaps worth describing in a paragraph above or below the table, but they are not central enough to the treatment data values and sources to be included in the table.  (But this can be for future documentation, and does not need to block merging these docs).

docs/source/models/intervention_models/alzheimers/hypothetical_treatment.rst
```diff
@@ -30,4 +30,236 @@
...
+    - Duration depends on if simulant completes or discontinues treatment
+
+.. _alzheimers_intervention_treatment_assumptions:
+
+Assumptions and Limitations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Those who do not initiate treatment following their first positive BBBM test result, or those who discontinue, 
+  will never take the intervention, so propensity can be assigned for simulant lifetime
+- Treatment occurs instantaneously, with a six-month waiting period after BBBM testing. This interprets the
```
NathanielBlairStahn, line 260: I think I am confused by the "In treatment/ Waiting for treatment" state. The sentence, "Treatment occurs instantaneously, with a six-month waiting period after BBBM testing," seems self-contradictory: Does treatment start instantaneously on the same time-step the simulant was tested, or does it start on the following time-step 6 months later?

The Lilly requirement that “There is an average of 6 months between a positive BBBM test result and initiating treatment” suggests to me that (if we're interpreting the average as a fixed duration) no one should start treatment on the same time step they're tested (Time step 1). Everyone should wait 6 months and start treatment on the next time step (Time step 2).

Then, the requirement that “The treatment takes immediate full effect in the first 6-month time step” kicks in, and we assume people receive the full effect of treatment on Time step 2. Then people people immediately (still on Time step 2) either go to the "Full treatment effect LONG" or "Full treatment effect SHORT" state depending on whether they complete treatment or discontinue.

Is that description equivalent to what you're specifying, or not? I think the difference between this and your description is that I'm saying _everyone_ should wait 6 months for treatment, whereas it looks like you're saying we should conceptualize at least some people as starting treatment on the same time step as testing, but perhaps the two interpretations are functionally the same.

_If_ they're the functionally same, then I think it would be clearer if we changed the name of the "In treatment/ Waiting for treatment" state to just "Waiting for treatment," and add a transient box called "Start treatment" that simulants enter on Time step 2 after they've been in the "Waiting for treatment" state for 6 months. Upon entering the transient "Start treatment" box, simulants would then immediately travel down the "complete" or "discontinue" arrows to the appropriate "Full treatment effect LONG" or "Full treatment effect SHORT" state (still on Time step 2), and then continue from there.

Does that make sense, or am I way off base?

tylerdy: > Does treatment start instantaneously on the same time-step the simulant was tested, or does it start on the following time-step 6 months later?
Instantaneous refers to the duration of the treatment period being zero, not the time between BBBM test and treatment period beginning being zero. I'll make this more obvious.

> Everyone should wait 6 months and start treatment on the next time step (Time step 2).
Yes, exactly

> Then, the requirement that “The treatment takes immediate full effect in the first 6-month time step” kicks in, and we assume people receive the full effect of treatment on Time step 2.
Yes, that is the idea. 6 month wait, but treatment takes effect instantly on starting it, so it still takes effect after exactly 6 months 

>Then people people immediately (still on Time step 2) either go to the "Full treatment effect LONG" or "Full treatment effect SHORT" state 
Yes, 6 months after testing simulants will transition to a Full treatment effect state

> whereas it looks like you're saying we should conceptualize at least some people as starting treatment on the same time step as testing, 
No, I think the misunderstanding here was around "instantaneous"

>if we changed the name of the "In treatment/ Waiting for treatment" state to just "Waiting for treatment," and add a transient box called "Start treatment"
This is a good idea. You can imagine why I didn't want an extra box floating around, but I agree two different things shouldn't be happening in the same box.

PR #1806, opened 2025-10-02
MNCNH model run updates for GBD 2023

alibow:


SylLutze (approved):
This looks good. I just want to call out that depending on how long GBD 2023 updates take, models 21, 22 and 23 could be implemented with GBD 2021 data prior to models 19 and 20. Not an issue, we should just be careful with labeling things to make sure no one gets confused.

alibow:
> This looks good. I just want to call out that depending on how long GBD 2023 updates take, models 21, 22 and 23 could be implemented with GBD 2021 data prior to models 19 and 20. Not an issue, we should just be careful with labeling things to make sure no one gets confused.

Thanks Syl! I added a note about this.

PR #1807, opened 2025-10-03
Hemoglobin paf calculation description

alibow:
Updating docs to reflect hemoglobin calculation code in this PR https://github.com/ihmeuw/vivarium_gates_mncnh/pull/141

PAF calculation code is limited in a few ways as described here. Some of these limitations could be more easily resolved than others and if we wanted to do things totally correct, we may need to continue updating these calculations over time as additional components get added to the simulation. My general approach is to do the simple thing first and update if our model is not calibrating, but if others feel that we should make these calculations more comprehensive from the get-go let me know and we can make a plan.

### DIFF
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
@@ -1580,6 +1580,7 @@ Default stratifications to all observers should include scenario and input draw.
     - Default
   * - 18.1
     - Model 17.0 bugfixes:
+
       * Update so that only those who attend ANC are eligible for IFA/MMS
 
         * Include corresponding update in the baseline IFA calibration `laid out in this PR <https://github.com/ihmeuw/vivarium_research/pull/1803>`__
@@ -1595,13 +1596,15 @@ Default stratifications to all observers should include scenario and input draw.
 
       * Updated artifact key for excess shift of IFA on birthweight
       * Update hemoglobin risk effects to use intervention-affected hemoglobin exposure rather than raw hemoglobin exposure
+      * Update baseline calibration to use coverage among total pop rather than at ANC
+      * Fix common random numbers between scenarios with regard to ANC attendance
     - No run necessary, all V&V done in the interactive simulation
     - N/A
     - N
     - N/A
     - N/A
   * - 18.3
-    - Model 18.0 with updated observer requests to avoid >15 stratifications 
+    - Run with updated observer requests to avoid >15 stratifications and multiple bugfixes laid out in the outstanding V&V issues table `in this PR <https://github.com/ihmeuw/vivarium_research/pull/1809>`__ 
     - Baseline, MMS scale-up, and anemia screening scale-up scenarios
     - ``model18.3``
     - Default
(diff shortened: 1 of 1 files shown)


docs/source/models/risk_effects/hemoglobin/index.rst
```diff
@@ -196,17 +189,21 @@ Relative risk values to calculate custom PAFs for maternal disorders can be acce
                location_id=1, # NOTE: you do not need to specify a location_id. Specifying any location ID will return results specific to location_id=1
                ) 
 
-Use the custom-calculated PAF values such that the maternal disorder incidence rate for an individual :math:`i` for a given affected maternal disorder subcause is as follows:
+Before being applied in the simulation, these relative risk values are first re-scaled to be relative to the TMREL exposure value of 120 g/L. Linear interpolation is used to assign RR values for exposures between points provided by the GBD risk curve. For exposure values that fall outside of the range provided by the GBD risk curve, we assign the same RR value as the nearest point on the curve. `See the custom hemoglobin risk effect implementation used in the MNCNH simulation for specific details <https://github.com/ihmeuw/vivarium_gates_mncnh/blob/main/src/vivarium_gates_mncnh/components/hemoglobin.py>`__
```
aflaxman, line 192:
```suggestion
Before being applied in the simulation, these relative risk values are first re-scaled to be relative to the TMREL exposure value of 120 g/L. We used linear interpolation to assign RR values for exposures between points provided by the GBD risk curve. For exposure values that fall outside of the range provided by the GBD risk curve, we assign the same RR value as the nearest point on the curve. `See the custom hemoglobin risk effect implementation used in the MNCNH simulation for specific details <https://github.com/ihmeuw/vivarium_gates_mncnh/blob/main/src/vivarium_gates_mncnh/components/hemoglobin.py>`__
```

aflaxman (approved):

docs/source/models/risk_effects/hemoglobin/index.rst
```diff
@@ -196,17 +189,21 @@ Relative risk values to calculate custom PAFs for maternal disorders can be acce
...
+Before being applied in the simulation, these relative risk values are first re-scaled to be relative to the TMREL exposure value of 120 g/L. We used linear interpolation to assign RR values for exposures between points provided by the GBD risk curve. For exposure values that fall outside of the range provided by the GBD risk curve, we assign the same RR value as the nearest point on the curve. `See the custom hemoglobin risk effect implementation used in the MNCNH simulation for specific details <https://github.com/ihmeuw/vivarium_gates_mncnh/blob/main/src/vivarium_gates_mncnh/components/hemoglobin.py>`__
 
-.. math::
+We calculate custom PAFs for hemoglobin on maternal disorders outcome for use in the MNCNH simulation. Code to generate these PAFs can be found here ``TODO: POST LINK``. We utilize the interactive context to initialize a population with assigned hemoglobin exposure values, assign relative risk values in accordance with observed exposure, and calculate PAFs among the population as (mean_rr - 1) / mean_rr among the age-stratified population. This PAF calculation process is limited in the following ways:
```
zmbc, line 194: Is this a TODO we need to address before merging? If not, can we specify the timeline, e.g. add a TODO box saying to fill in this link once the GBD 2023 update is complete?

zmbc, line 194:
```suggestion
We calculate custom PAFs for hemoglobin on maternal disorders outcome for use in the MNCNH simulation. Code to generate these PAFs can be found here ``TODO: POST LINK``. We utilize the interactive context to initialize a population with assigned hemoglobin exposure values, assign relative risk values in accordance with observed exposure, and calculate PAFs for each age group and sex as (mean_rr - 1) / mean_rr among the simulants with that age group and sex. This PAF calculation process is limited in the following ways:
```

We do stratify by sex, right?

docs/source/models/risk_effects/hemoglobin/index.rst
```diff
@@ -196,17 +189,21 @@ Relative risk values to calculate custom PAFs for maternal disorders can be acce
...
 
-  ir_i = ir * (1 - PAF) * RR_i
+- The simulated population does not have a uniform distribution of maternal ages. The PAF values for age groups with less person time have less statistical precision than age groups with more person time represented in our simulation.
```
zmbc, line 196: Is this effectively saying that we _didn't_ implement the EvenlyDistributedPopulation optimization we did for LBWSG? If so, I feel like this doesn't quite rise to the level of a limitation and maybe should just be a note

docs/source/models/risk_effects/hemoglobin/index.rst
```diff
@@ -196,17 +189,21 @@ Relative risk values to calculate custom PAFs for maternal disorders can be acce
...
+- The PAF calculation does not take detailed correlation with other factors that affect maternal hemorrhage and maternal sepsis risk into account. For instance, the correlation between ANC attendance and in-facility delivery in our simulation will induce a correlation between hemoglobin (as a result of baseline IFA coverage distributed at ANC) and misoprostol coverage (available at home deliveries only), both of which affect maternal hemorrhage risk. We hypothesize that the impact of ignoring this sort of correlation will be small enough for our model to calibrate, but we can revisit this assumption if the model calibrates poorly.
+- We use the hemoglobin exposure distribution during pregnancy to calculate the PAF on postpartum depression rather than the hemoglobin exposure distribution amon simulants who survive to the postpartum period. Given that maternal mortality is a rare event, we assume this limitation will be small, but it can be revisited if the model calibrates poorly.
```
zmbc, line 198:
```suggestion
- We use the hemoglobin exposure distribution during pregnancy to calculate the PAF on postpartum depression rather than the hemoglobin exposure distribution among simulants who survive to the postpartum period. Given that maternal mortality is a rare event, we assume this limitation will be small, but it can be revisited if the model calibrates poorly.
```

docs/source/models/risk_effects/hemoglobin/index.rst
alibow: Well the hemoglobin risk factor and maternal disorders outcomes are "female only" in GBD by definition, so we don't have to and then for neonatal sepsis we use the female-only exposure among the pregnant population and use the RR data that exists at the neonatal sex and age specific level to calculate neonatal sex and age-group specific PAFs. I'll add some explanation of this to the docs.

alibow: Yeah I'll add a note that we will calculate and add links once the 2023 update is complete!

zmbc: Oh makes sense!

PR #1903, opened 2026-03-02
Update maternal hemorrhage docs to GBD 2023

zmbc:
The major changes here are:
* Track severity within the cause model and make mortality only an outcome of severe maternal hemorrhage
* Update maternal hemorrhage -> hemoglobin effects to use new GBD shift modeling, which says there is an impact for 9 months after the end of pregnancy
* As a result of the previous, introduce a new timestep after our old postpartum timestep to extend to 9 months after the end of pregnancy, update anemia YLDs calculation accordingly

This does _not_ fully prepare us for the E-MOTIVE bundle, which is going to introduce additional nuance to our cause model here due to its 300 mL trigger threshold. That will be addressed in a separate PR.


docs/source/models/concept_models/vivarium_mncnh_portfolio/anemia_component/module_document.rst
```diff
@@ -253,13 +273,17 @@ Note that simulants who died during labor should not experience any YLDs due to
...
 - We assume that interventions affect anemia YLDs at the time of administration at ANC (as according to the timed assumptions in the two prior bullets) with no additional delay 
-- We only track anemia YLDs for six weeks postpartum
+- We do not model any deaths in the 9 months after pregnancy, so we assume that all simulants who survive labor also survive 9 months afterward and therefore experience anemia YLDs during this period. This may lead to a slight overestimation of anemia YLDs.
+- We do not model simulants changing age groups during the 9 months after pregnancy, and thereby slightly underestimate the ages
+  of people in this period.
+  The impact of this on anemia YLDs depends on how hemoglobin changes with age: if hemoglobin increases with age, then we will underestimate anemia YLDs during this period, and if hemoglobin decreases with age, then we will overestimate anemia YLDs during this period.
```
pletale: So is it uncertain how hemoglobin changes with age? Or does it just depend on too many factors for us to make a more specific statement regarding this limitation's effect?

docs/source/models/risk_effects/maternal_hemorrhage/gbd2023/index.rst
```diff
@@ -0,0 +1,120 @@
...
+  The continuous curves could result in simulants moving between several anemia categories over the course of the postpartum period, which is not captured here.
+  This also means that the (relatively arbitrary) choices of durations of these periods are impactful.
+- We currently apply the same hemoglobin shift for all incident maternal hemorrhage cases, but in reality the hemoglobin shift likely differs by severity of maternal hemorrhage (moderate vs severe), which is not captured here.
+  We may address this in the future based on our approach in the :ref:`previous maternal hemorrhage effects model <2019_2021_risk_effect_maternal_hemorrhage>`.
+- The GBD shift is derived from data about *postpartum* hemorrhage only, and we are applying it to all maternal hemorrhage cases (including antepartum hemorrhage), which may not be appropriate.
```
pletale: This will also be a relevant limitation for the E-MOTIVE bundle intervention, right? My understanding was that E-MOTIVE is only for PPH not antepartum hemorrhage. Do you think that will impact our decision here about whether to split AP vs. PP hemorrhage?

pletale (approved):
This looks good to me! I had a couple questions, some of which I've left as comments on this PR, and others I thought would be helpful to talk through in real time for my understanding! Lmk if my questions here don't make sense

tylerdy (approved):
I'm catching up on this module but this all made pretty good sense to me I think. I am a bit confused on some GBD / hemoglobin background - it looks like hemoglobin affects several outcomes including hemorrhage, sepsis and postpartum depression, and hemhorrage also further affects hemoglobin. And also anemia YLDs are calculated based on hemoglobin levels, but anemia is not part of the hemoglobin risk effects doc? And we only talk about anemia but not any of the above risk effects in the post partum hemoglobin module docs? Is this about how we model fatal vs nonfatal outcomes? I'm sure its some GBD basics but I haven't thought much about this before

zmbc:
@tylerdy happy to chat on Zoom if it's helpful, but I think the key piece is that hemoglobin and anemia are essentially the same thing. Hemoglobin is the continuous measure (in g/dL) and anemia is the categorical variable of where your hemoglobin falls in between for a list of thresholds.

PR #1912, opened 2026-03-24
Update What is Vivarium Good for? page with more details on residual confounding

pletale:
Related JIRA ticket: https://jira.ihme.washington.edu/browse/SSCI-1921 

I'm not sure if the first sentence about compartmental or multiplication models is accurately written. Also, generally I'm leaving this pretty devoid of detail because I didn't want to be duplicative with the info on the Residual Confounding page that is linked, but I'm not sure it is very clear to me how residual confounding is specific to microsimulation and not other forms of population modeling!


zmbc (commented):
Looking forward to a conversation about this next week, but I'll quickly chime in here that I am also uncertain about this! My first thought is that this might be a limitation that is _surfaced_ by, not _introduced_ by, microsimulation.

zmbc:
> forces us to deal with it

Hasn't worked so far 😄

I agree that the data required to avoid residual confounding seems like it would be the same whether modeling at an individual level or not.

zmbc (approved):

aflaxman (approved):

PR #1924, opened 2026-04-15
Alzheimer's: Clarifications to rate calibration docs

NathanielBlairStahn:
Make several clarifications and improvements to the rate calibration documentation:

- Use consistent notation and definitions between different tables
- Update notation on treatment documentation page to reflect new notation in cause model
- Try to clarify how input data is being used
- Add missing artifact keys to list of calibration outputs, and remove references to nonexistent key
- Specify age range for the calibration procedure
- Reorganize subsections of rate calibration section and add additional explanation to help clarify things
- Note limitation of EMR/CSMR derived from DisMod
- Update labels on cause model page to reflect updated title and GBD year
- Rename dementia cause state to align with CSU tech report
- Add Wikipedia link to "multiple shootings" ODE solution technique used in calibration
- Fix some typos in equations
- Remove now-unused variables from data sources table

It would be great to get some more researchers and/or engineers to read through this to see whether everything makes sense.


aflaxman (approved):

tylerdy (approved):
I did a time-boxed review of this and while I did not follow all the math this is a detailed and thorough explanation with all the right pieces in place that I could understand better if I spent some more time with it. Eg the variable, parameter and constraint definitions as well as inputs, outputs, data sources etc all made sense to me.

NathanielBlairStahn:
> I did a time-boxed review of this and while I did not follow all the math this is a detailed and thorough explanation with all the right pieces in place that I could understand better if I spent some more time with it.

Thanks @tylerdy, I don't totally understand the math either, and I've therefore intentionally left some things vague...

> Eg the variable, parameter and constraint definitions as well as inputs, outputs, data sources etc all made sense to me.

Ok great, this was the main point of the PR -- to make sure everything was in the right place and that there's enough detail to implement the Vivarium simulation given Abie's consistent rates outputs, and to understand what data the simulation is using where.

PR #1939, opened 2026-05-12
MNCNH sepsis effect on Hb

pletale:
As noted in [this slack thread](https://ihme.slack.com/archives/C02944X688L/p1778622880946979), I'll update this page to include the new GBD 2023 anemia impairment page once it's ready. 

One side note - Here I use 'puerperal sepsis' and 'maternal sepsis' interchangeably, but maybe that's confusing or inaccurate. I could update everything to 'puerperal' because that is more in line with our BLISS guidelines anyways, but I didn't want to make a change like that when all of our other documentation about the cause uses 'maternal'. Lmk if others have thoughts! 

Also, I noticed there's a cutoff at 0 for the Hb-sepsis curve, but not for the Hb-hemorrhage curve. Does anyone know why the modelers chose to do this? Maybe I will understand better when I write up the anemia impairment page, but I'm curious!


docs/source/models/risk_effects/maternal_sepsis/gbd2023/index.rst
```diff
@@ -0,0 +1,119 @@
...
+     - Modelable entity
+     - 27596
+     - Hemoglobin concentration
+     - During the period from 6 weeks to 6 months after the end of pregnancy only; this is the "non-pregnant" hemoglobin distribution
+
+Hemoglobin effects after the end of pregnancy
++++++++++++++++++++++++++++++++++++++++++++++
+
+For simulants who experience an incident case of puerperal sepsis as determined in the :ref:`maternal sepsis <2021_cause_maternal_sepsis_mncnh>` page,
+we will decrease the simulant's hemoglobin concentration during two distinct time periods: the first six weeks after the end of pregnancy and the period from 6 weeks to 6 months after the end of pregnancy.
```
zmbc: I would recommend making this 9 months to prevent having to add a new timestep (we already have the 6 weeks to 9 months timestep). I also did this for APH.

docs/source/models/risk_effects/maternal_sepsis/gbd2023/index.rst
```diff
@@ -0,0 +1,119 @@
...
+- This modeling strategy applies an average hemoglobin shift for all incident maternal sepsis cases during the first six weeks after the end of pregnancy and an average hemoglobin shift for all incident maternal sepsis cases during the period from 6 weeks to 6 months after the end of pregnancy, which does not capture the continuous nature of hemoglobin changes over time.
+  The continuous curves could result in simulants moving between several anemia categories over the course of the postpartum period, which is not captured here.
+  This also means that the (relatively arbitrary) choices of durations of these periods are impactful.
+- The GBD shift is derived from USA MarketScan data, which may not be generalizable to other locations.
+  In particular, treatment in the USA may be more effective at mitigating the hemoglobin shift associated with maternal sepsis than treatment in other locations, which could lead to an underestimation of the hemoglobin shift in locations with less effective treatment.
+  Additionally, the USA has more access to ultrasound than our modeled locations which could lead to better prevention.
```
zmbc: I'm curious whether ultrasound is actually relevant here

zmbc (approved):

docs/source/models/risk_effects/maternal_sepsis/gbd2023/index.rst
zmbc: Note also that I plan to change this from "9 months" to a number of weeks. Will follow up with another PR though, good to have this be consistent with PPH.

docs/source/models/risk_effects/maternal_sepsis/gbd2023/index.rst
pletale: True - I could change this to antibiotic treatment or just remove this line

docs/source/models/risk_effects/maternal_sepsis/gbd2023/index.rst
pletale: Okay, that makes sense, you're changing it to 38 weeks?

zmbc: Not sure yet!

docs/source/models/risk_effects/maternal_sepsis/gbd2023/index.rst
```diff
@@ -0,0 +1,119 @@
...
+:ref:`maternal sepsis <2021_cause_maternal_sepsis_mncnh>` page.
+
+GBD 2023 Modeling Strategy
+-------------------------------
+
+.. todo:: 
+
+    Add link to 2023 anemia impairment page once it is created and describes the new anemia causal attribution process for GBD 2023. 
+
+GBD does not explicitly model maternal sepsis as a risk factor. However, as of GBD 2023, it models a hemoglobin shift associated with puerperal sepsis in the anemia causal attribution process.
```
tylerdy:
```suggestion
GBD does not explicitly model maternal (puerperal) sepsis as a risk factor. However, as of GBD 2023, it models a hemoglobin shift associated with puerperal sepsis in the anemia causal attribution process.
```

I do think switching bt the terms in these two sentences makes it seem like they might mean different things but I think totally fine to use puerperal on this page if you introduce it at the beginning by doing something like this

tylerdy (approved):
I found these to be quite clear and approachable docs!! I approve

aflaxman (approved):

PR #1943, opened 2026-05-18
Inpatient SAM limitations

alibow:


docs/source/models/concept_models/vivarium_nutrition_optimization_extension/concept_model.rst
```diff
@@ -55,7 +55,22 @@ Components with substantial updates (links/updates to come)
...
+
+* We apply the GBD wasting risk effects that are based on WHZ exposure alone to our modeled moderate acute malnutrition state. Given that there are cases that may be classified as moderate wasting by WHZ alone and classified as SAM when their MUAC and/or oedema exposures are also considered, this may cause us to overestimate the effect of MAM on morbidity and mortality in our model as some higher risk cases by the full acute malnutrition classification criteria will be (mis)classified as MAM using the WHZ-only GBD exposure assessment approach. This may contribute to an overestimation of overall child growth failure attributable mortality in our model.
+
+* Our wasting exposure model is informed by WHZ and oedema exposures only and does not consider MUAC exposures. This may cause us to underestimate the prevalence of acute malnutrition by excluding the cases that only meet the MUAC criteria and not the WHZ or oedema criteria.
```
zmbc, line 65: I don't fully understand where the MUAC exposures _are_ used in this model. If we don't use them anywhere, then that is internally consistent. Is this limitation relevant only to the family MUAC screening scenarios, maybe?

docs/source/models/concept_models/vivarium_nutrition_optimization_extension/concept_model.rst
```diff
@@ -55,7 +55,22 @@ Components with substantial updates (links/updates to come)
...
+
+* Our approach to modeling SAM mortality separate from the GBD CGF effects in this model causes us to underestimate the expected impact of SQLNS on mortality risk experienced in the SAM states as we will not capture any mortality risk reductions within the SAM states due to improvements in child stunting exposures (and additional associated downstream improvements in child underweight exposure).
```
zmbc, line 67: > we will not capture any mortality risk reductions within the SAM states due to improvements in child stunting exposures

Won't we capture some of this because uncomplicated SAM cases were less likely to be stunted in the studies from which we got mortality risks? Perhaps we are assuming this is approximately none because wasting is acute while stunting is cumulative?

zmbc (approved):

docs/source/models/concept_models/vivarium_nutrition_optimization_extension/concept_model.rst
alibow: MUAC exposures show up in our model due to us using SAM treatment literature to inform mortality rates of the SAM states as the admission criteria for the studies is based on the full rather than WHZ-alone classification. Does that clarify? I will add it to the docs if so. I think there is a separate limitation related to the MUAC-based screening I should probably also articulate here.

docs/source/models/concept_models/vivarium_nutrition_optimization_extension/concept_model.rst
alibow: Hmm, I am not really following your first point. My thought here is that for a given simulant:

- Scenario without SQLNS: simulant becomes a complicated SAM case with severe stunting. Mortality rate in our simulation will be the complicated SAM mortality rate. But in GBD: mortality rate would be a product of the severe wasting RR * severe stunting RR * severe underweight RR
- Scenario with SQLNS: this same simulant still progresses to complicated SAM, but SQLNS succeeds in preventing severe stunting, so now this simulant just has moderate stunting. Mortality rate in our simulation will be the complicated SAM mortality rate. But in GBD: mortality rate would be a product of the severe wasting RR * MODERATE stunting RR * MODERATE underweight RR

So in our simulation, SQLNS would not reduce the mortality risk of this simulant according to our model design. But if we had used GBD CGF risk effects to inform mortality in the SAM states as we did with the prior version of the NO model, there WOULD be mortality reductions due to SQLNS for this simulant (as reflected by the lower stunting and underweight RRs despite the unchanged wasting RR).

Does this clarify?

docs/source/models/concept_models/vivarium_nutrition_optimization_extension/concept_model.rst
zmbc: That makes sense!

docs/source/models/concept_models/vivarium_nutrition_optimization_extension/concept_model.rst
zmbc: Ahh I see, I did not realize that SQ-LNS could prevent stunting _without_ preventing wasting.

docs/source/models/concept_models/vivarium_nutrition_optimization_extension/concept_model.rst
```diff
@@ -55,7 +55,22 @@ Components with substantial updates (links/updates to come)
...
+
+* We do not consider dynamics related to wasting seasonality and/or treatment capacity.
+
+* We do not model morbidity associated with our PEM model.
```
SylLutze: If you did a BEC saying this is okay, should we add it here?

SylLutze (approved):

tylerdy (approved):

PR #1946, opened 2026-05-29
Adjust temporal ordering of maternal disorders

zmbc:
* Fixes a longstanding discrepancy between "component" descriptions and information actually used, by moving abortion/miscarriage/ectopic pregnancy maternal disorders out of the intrapartum "component"
* Limits antepartum hemorrhage to pregnancies that do not end in abortion/miscarriage/ectopic pregnancy
* Adjusts intrapartum maternal disorders cause models to account for mortality due to antepartum maternal disorders


tylerdy (approved):

docs/source/models/causes/maternal_disorders/gbd_2021_mncnh/antepartum_hemorrhage.rst
```diff
@@ -377,6 +368,8 @@ Limitations
   which is likely not true.
 * Splitting out maternal hemorrhage (modeled as one cause in the GBD) into antepartum and postpartum hemorrhage (modeled as two separate causes in our model, with a vicious cycle between them through hemoglobin)
   will lead us to overestimate the total burden of maternal hemorrhage relative to GBD due to cases that have both antepartum and postpartum hemorrhage and have double-shifted hemoglobin.
+* We assume that abortion/miscarriage/ectopic pregnancies are too short to experience in antepartum hemorrhage.
+  In reality, some abortions may occur late enough in pregnancy to experience antepartum hemorrhage.
```
pletale:
```suggestion
  In reality, some abortions or miscarriages may occur late enough in pregnancy to experience antepartum hemorrhage.
```
Not just abortions, right? (probably a nit sorry)

pletale (approved):

docs/source/models/causes/maternal_disorders/gbd_2021_mncnh/antepartum_hemorrhage.rst
zmbc: Not a nit and I had thought about this a bit, but I think I may have come to the wrong conclusion!

APH is defined as happening after 20 weeks gestation. Miscarriages are sometimes defined as happening before 20 weeks gestation (with the same event after that point categorized as a stillbirth). That would make APH during a pregnancy ending in miscarriage definitionally impossible.

But in our sim, the dividing line between miscarriage and stillbirth is 24 weeks! So I think your edit is correct

zmbc: Hmm no the reason I didn't include this is because it seems there is also some confusion about the timing of APH -- Wikipedia says after 24 weeks. https://en.wikipedia.org/wiki/Antepartum_bleeding

zmbc: @alibow do you happen to have any background knowledge to share here? I'm finding sources that disagree on these definitions without much explanation of why

hussain-jafari (approved):

docs/source/models/causes/maternal_disorders/gbd_2021_mncnh/antepartum_hemorrhage.rst
alibow: Hmm, nothing declarative, sorry. I don't have any insight into the 20 vs 24 week thresholds except for that I have also seen these inconsistencies across sources. I think it could be useful to take a look at the GBD ICD codes by maternal disorders subcause... like I think that miscarriage deaths typically happen due to blood loss or sepsis. I have been operating under the assumption that GBD would consider these as miscarriage deaths because they occurred prior to the gestational age threshold, but I am realizing that's probably something we should specifically confirm!

PR #1951, opened 2026-06-23
NO Doc Updates

SylLutze:
In getting the NO model to run in Kenya and 2023 for JKUAT, I identified some small needed updates. These have already been made in the code in this PR: https://github.com/ihmeuw/vivarium_gates_nutrition_optimization/pull/148

The two main changes are: 
- Updating the infant male sex section. I just removed the hard coded values and instead the notebook is linked and the GBD ID. This seemed easiest for maintaining and is more consistent with what we did for MNCNH. 
- Updating the code block for pulling hemoglobin values to use get_draws instead of get_outputs, which has draws and is preferred.

### DIFF
--- a/docs/source/models/concept_models/vivarium_mncnh_portfolio/anemia_component/module_document.rst
+++ b/docs/source/models/concept_models/vivarium_mncnh_portfolio/anemia_component/module_document.rst
@@ -237,23 +237,23 @@ Anemia YLDs are estimated according to the following steps:
 
 .. code-block:: python
 
-   get_outputs(
-       location_id=[165,179,214],
-       topic='rei',
-       rei_id=[206,206,207], # We also have rei_id=192 for all anemia and rei_id=432 for moderate and severe combined
-       population_group_id=16,
-       sex_id=2,
-       year_id=2023,
-       release_id=16, # release_id=33 also works
-       compare_version_id=8306,
-       measure_id=[3,5],
-       age_group_id=[7, 8, 9, 10, 11, 12, 13, 14, 15, 24, 169]
-   )
+   get_draws(
+        gbd_id_type='rei_id',
+        gbd_id=[205, 206, 207], # Mild, moderate, and severe anemia, respectively. We also have rei_id=192 for all anemia and rei_id=432 for moderate and severe combined
+        location_id=list(location_ids.location_id),
+        population_group_id=16, # pregnant population
+        sex_id=2, # female
+        year_id=2023,
+        release_id=33, # GBD 2023, special publications
+        source='como',
+        measure_id=5, # prevalence
+        metric_id=3, # rate
+        age_group_id=[7, 8, 9, 10, 11, 12, 13, 14, 15], # constituent age groups of 10-54 years (169)
+    )
 
 .. note::
 
    Make sure you have the latest version of ``db_queries`` to be able to use the ``population_group_id`` argument. To get pregnancy-specific results, the population group and the age groups need to be specified, because the default is all ages.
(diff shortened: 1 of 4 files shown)


docs/source/models/other_models/pregnancy/gbd_2021_closed_cohort/index.rst
```diff
@@ -301,25 +301,14 @@ Sex of infant
...
-        - 214
-        - 0.511785 
-    *   - Ethiopia
-        - 179
-        - 0.514271  
+For pregnancies that result in live birth or stillbirth outcomes,
+infant sex should be determined and recorded acording to the probability
+of male sex calculated using the live births by sex GBD covariate
+(ID 1106) for the appropriate GBD cycle (2021 or 2023 depending on the run).
+Values were calculated in this `sex of infant notebook <https://github.com/ihmeuw/vivarium_research_nutrition_optimization/blob/data_prep/data_prep/Live%20births%20by%20sex.ipynb>`_
```
zmbc, line 308: If this is done on the artifact building side now, perhaps:


```suggestion
The numeric values were previously provided here in the docs based on calculations in this `sex of infant notebook <https://github.com/ihmeuw/vivarium_research_nutrition_optimization/blob/data_prep/data_prep/Live%20births%20by%20sex.ipynb>`_
```

zmbc (approved):

docs/source/models/concept_models/vivarium_mncnh_portfolio/anemia_component/module_document.rst
```diff
@@ -237,23 +237,23 @@ Anemia YLDs are estimated according to the following steps:
...
+        release_id=33, # GBD 2023, special publications
+        source='como',
+        measure_id=5, # prevalence
+        metric_id=3, # rate
+        age_group_id=[7, 8, 9, 10, 11, 12, 13, 14, 15], # constituent age groups of 10-54 years (169)
+    )
 
 .. note::
 
    Make sure you have the latest version of ``db_queries`` to be able to use the ``population_group_id`` argument. To get pregnancy-specific results, the population group and the age groups need to be specified, because the default is all ages.
```
alibow: Should this note also get deleted now? I think the age group thing doesn't apply to get draws which returns most detailed age groups by default? But maybe we want to keep this here as a note in case someone wants to use get outputs? in which case it would be nice to note that we also need a compare version ID argument (at least I think that's been required?) which looks to be 8352 based on this page https://hub.ihme.washington.edu/spaces/GBDdirectory/pages/151718325/GBD+Home

docs/source/models/other_models/pregnancy/gbd_2021_closed_cohort/index.rst
alibow: Looks like it hasn't been updated to be done on the artifact building side. it would be nice to add a todo note about that here

alibow (approved):

docs/source/models/other_models/pregnancy/gbd_2021_closed_cohort/index.rst
zmbc: Gotcha, OK. Pretty sure this has moved to the artifact building side in MNCNH.

SylLutze: Yes, this is updated to pull directly in MNCNH, but not in NO. Since it wasn't an artifact key, I did not feel confident in updating it myself so I just updated the values.

I can add a to-do note here stating that we would like to update it, but I will reject this suggestion.

docs/source/models/concept_models/vivarium_mncnh_portfolio/anemia_component/module_document.rst
SylLutze: Thanks for the callout! I updated the note to specify that if you wanted to do a get_outputs call, you would need the age groups and added that note about a compare ID.

PR #1957, opened 2026-07-08
Create stub pages for MNCNH Portfolio Costing documentation

pletale:
As discussed in 7/7 doc session, we want to make sure we're documenting our process for the MNCNH portfolio costing analysis. Currently, this documentation lives in a few different sharepoint files, but eventually we will want to put everything in one place here!


aflaxman (approved):
I'm not sure I see cost-effectiveness estimates as a distinct bucket from unit costs and offset costs, but that is something we can discuss further as we go---we do not need to sort it out before merging these stub pages.

pletale:
> I'm not sure I see cost-effectiveness estimates as a distinct bucket from unit costs and offset costs, but that is something we can discuss further as we go---we do not need to sort it out before merging these stub pages.

Agreed, this was some draft language that I decided to just leave in but we can definitely reword. I meant this in terms of how we are using them (as in we are only using unit costs in our meta-regression, and offset costs and cost-effectiveness estimates will be used in validation, sensitivity analyses, or some other post-processing), but it's confusingly worded!

docs/source/models/other_models/mncnh_costing/costing_lit_review/index.rst
```diff
@@ -0,0 +1,72 @@
...
+-------------
+
+.. todo:: 
+
+    Add a description of the cost taxonomy used in the MNCNH costing model here, along with
+    the diagram that we created
+
+    Add list of interventions and which costs are associated with each intervention here
+
+There are three 'buckets' which different costs fall into in the MNCNH portfolio project:
```
NathanielBlairStahn: Here's a suggestion for a simple wording change to help address @aflaxman's [comment](https://github.com/ihmeuw/vivarium_research/pull/1957#pullrequestreview-4656076671): 

```suggestion
There are three 'buckets' which different extracted cost estimates fall into in the MNCNH portfolio project: 
```

NathanielBlairStahn (approved):
Looks good to me. I made one small suggestion to address Abie's comment.

PR #1964, opened 2026-08-11
Remove antepartum hemorrhage from the model

zmbc:


docs/source/models/concept_models/vivarium_mncnh_portfolio/concept_model.rst
```diff
@@ -1973,6 +1964,12 @@ Default stratifications to all observers should include scenario and input draw.
     - All
     - Default
     -
+  * -
+    - Remove APH
+    - Remove antepartum hemorrhage from the model; see `this PR <https://github.com/ihmeuw/vivarium_research/pull/1964>`__ for detailed changes
+    - Standard V&V scenarios
+    - Default
+    -
```
zmbc, line 1972: Note: more updates need to be made to this table to reflect that we have run and V&Ved more models, but I was hoping @tylerdy could do those

docs/source/models/risk_effects/hemoglobin/index.rst
```diff
@@ -173,6 +167,10 @@ Note that we will not be modeling direct effects of hemoglobin on the affected o
 
   Update this page with hypertension cause model links when ready
 
+Note that *antepartum* (as opposed to postpartum) hemorrhage is not impacted by hemoglobin in our model,
+and as a result we exclude the antepartum hemorrhage cause model from our simulation entirely.
+The Burden of Proof paper referenced above reports that "Three studies examining the association between maternal hemoglobin and antepartum hemorrhage found no statistically significant relationship across conventional anemia severity categories"
```
zmbc, line 172: This was the best place I could find to note this, but feels a bit awkward

docs/source/models/causes/maternal_disorders/gbd_2021_mncnh/obstructed_labor.rst
```diff
@@ -257,12 +257,12 @@ calculations as well as for the calculation of YLDs in the next section.
       - Note
     * - ir
       - obstructed labor and uterine rupture incidence risk per birth
-      - incidence_c370 / (birth_rate - antepartum_hemorrhage_csmr)
+      - incidence_c370 / birth_rate
       - The value of ir is a probabiity in [0,1]. Denominator includes
```
aflaxman:
```suggestion
      - The value of ir is a probability in [0,1]. Denominator includes
```

aflaxman (approved):

docs/source/models/causes/maternal_disorders/gbd_2021_mncnh/residual_maternal_disorders.rst
```diff
@@ -206,6 +206,12 @@ available for each cause).
     - False
     - False
     - 
+  * - Antepartum hemorrhage
```
tylerdy, line 209: Ok so we will still ultimately have YLLs YLDs and incidence for APH but it will be part of the residual bucket? I think that's helpful for understanding why it makes sense not to model it since none of our interventions affect it - we're not modeling it individually but not just ignoring that happens.

tylerdy (approved):

docs/source/models/causes/maternal_disorders/gbd_2021_mncnh/residual_maternal_disorders.rst
zmbc: Yes -- we will still have YLLs and YLDs for APH represented, though we won't say specifically which ones in the residual bucket are APH. We won't be tracking incidence, either.

PR #1965, opened 2026-08-13
Design document on automating modeling workflow

zmbc:
A few notes on what didn't make it into this PR:
- I partly mocked up what the MNCNH project's more complex pipeline might look like in this syntax, but it's unfinished and not updated with my latest thinking; not sure if it is worth my time to keep working on that until we've reviewed this
- I had an initial vision in which the workflow was more expansive, replacing the entirety of `loader.py` and moving away from using the artifact for most purposes; Patrick mostly talked me out of that as being too much to bundle in here, but I still do like that version better and am not sure if/where/how to share my thinking on that
- Data versioning was included in what I presented at all-team, but I have since separated that out


docs/source/model_design/vivarium_features/modeling_workflow_automation/index.rst
```diff
@@ -0,0 +1,250 @@
...
+We'll do this using a Vivarium simulation.
+We make a separate model specification, and add a custom component to observe the PAF.
+
+.. code::
+    :emphasize-lines: 4, 7-8
+
+    src/tutorial/
+        components/
+            sqlns_intervention.py
+            paf_observer.py
```
aflaxman, line 105: maybe show just the additions difference between the file listing above and here ; this one and the one on like 109, right?

docs/source/model_design/vivarium_features/modeling_workflow_automation/index.rst
```diff
@@ -0,0 +1,250 @@
...
+   We will pass an output directory for the PAF simulation results, also in the team directory, something like
+   ``/mnt/team/simulation_science/pub/models/tutorial/results/second_run/paf_simulation/``.
+3. Copy the results from the PAF simulation into a standard location for the artifact to pull from,
+   such as within the repo, or update a constant in the artifact generation code to point to the new PAF simulation output location.
+
+The rest of the process continues unchanged.
+
+Next, we will add one more custom process for demonstration purposes, one that does not use a Vivarium model.
+We've decided that we do not want to use the remission rate data from GBD for diarrheal diseases, but instead
+want to calculate a custom remission rate based on a compartmental model calibrated to a literature source,
```
aflaxman, line 149: This also comes up for us when the GBD analysis has not stored a remission rate that is consistent with the incidence, prevalence, and excess mortality rates

docs/source/model_design/vivarium_features/modeling_workflow_automation/index.rst
```diff
@@ -0,0 +1,250 @@
...
+            sqlns_intervention.py
+            paf_observer.py
+        constants/
+        data/
+            custom_paf/
+                paf_model_spec.yaml
+            custom_remission/
+                custom_remission.py
+                results/
+                    ethiopia.csv
```
aflaxman, line 167: are you thinking of this as the output of the custom_remission.py script when it runs?

docs/source/model_design/vivarium_features/modeling_workflow_automation/index.rst
```diff
@@ -0,0 +1,250 @@
...
+            interactive_vnv.ipynb
+        results/
+            results_vnv.ipynb
+    .gitignore
+    Makefile
+    environment.sh
+    pyproject.toml
+    ...
+
+We also update the ``loader.py`` to take the remission from the saved CSV file rather than from GBD.
```
aflaxman, line 188: loader.py is complicated enough already, maybe we should not have it show up more than once in the workflow

aflaxman (approved):
Do you understand what Rajan wants that is not in here?

tylerdy (approved):

docs/source/model_design/vivarium_features/modeling_workflow_automation/index.rst
zmbc: The difference is highlighted in the rendered version.

docs/source/model_design/vivarium_features/modeling_workflow_automation/index.rst
zmbc: This doc doesn't speak to _why_ we don't want to use the GBD remission rate.

docs/source/model_design/vivarium_features/modeling_workflow_automation/index.rst
zmbc: Yes

docs/source/model_design/vivarium_features/modeling_workflow_automation/index.rst
zmbc: I don't understand this comment

aflaxman: To say it another way, I would prefer a solution where we do not need to update loader.py iteratively to run this workflow. This line made me think the solution proposed here requires updating loader.py, which strikes me as overly complicated.

zmbc: This is in the "current practice" section of the document, describing what we do _now_. It represents one-time setup, not something that would need to be updated every time the model is run.

PR #1967, opened 2026-08-20
AI-assisted narrowing of wide graphviz diagrams

aflaxman:


zmbc:
@aflaxman did you review this? Links to the ReadTheDocs visual diffs of the relevant diagrams would help. Also from a very cursory look at the code it looks like some of these diagrams had some thought put into their layout already, so I worry this change may be too bold

aflaxman:
I did review it carefully by putting the old and new ones side by side in read the docs, and I think it is an improvement.  I can add some screenshots if it would be helpful to others.  Also, this is not urgent!

aflaxman:
Here is an artifact that can help reviewers compare the before and after of these changes: https://claude.ai/code/artifact/28ecb0f3-9d73-49ab-b140-5efaa386940f

zmbc:
I think this generally seems fine, except that the azithromycin changes introduced line breaks in the middle of (underscore-connected) words

aflaxman:
> I think this generally seems fine, except that the azithromycin changes introduced line breaks in the middle of (underscore-connected) words

Do you think it would be clear if there was an underscore before the line break?, e.g.

image

zmbc:
I guess that would be a bit better? But unless others have ideas I'd probably prefer simply reverting those two diagrams (I had missed that it also occurs in the probiotics one)

